PR #12, opened 2015-05-09
Rest

fabaff:
Add two simple examples on how to interact with the API.

### DIFF
--- a/source/_includes/custom/navigation.html
+++ b/source/_includes/custom/navigation.html
@@ -19,7 +19,7 @@
       <li><a href="{{ root_url}}/developers/add_new_platform.html">
         Adding platform support
       </a></li>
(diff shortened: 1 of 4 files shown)


balloob:
I think these two examples should be next to the line in this documentation about the `homeassistant.remote` package.

Also, we might want to add other python examples:

```python
import homeassistant.remote as remote

api = remote.API("host", "password")
living_room = remote.get_state(api, 'group.living_room')
```

```python
import homeassistant.remote as remote

api = remote.API("host", "password")
hass = remote.HomeAssistant(api)
hass.start()
living_room = hass.states.get('group.living_room')
```

fabaff:
I think that a page per API would make it easier in the future.

balloob:
I agree, thanks!

PR #18, opened 2015-06-02
Transmission switch page

fabaff:
This pull request contains only the page for the Transmission switch. 

The component page entry is just an idea to keep the entries together.


balloob:
I like it. 

Honestly, those two platforms should probably be integrated in a new Downloader component. But that's for the future

fabaff:
Would make sense.

PR #31, opened 2015-08-09
Update layout of components page

fabaff:
The tables are getting longer and longer...I was thinking about a layout change to a column layout instead of a large table. This could make it possible to group entities/services without breaking the responsiveness and "compress" the content. 

This is a mockup with the code from the front page. 

Mockup


balloob:
I am down to merge this but it looks like there are merge conflicts

fabaff:
Don't merge it, please. So far it's only done for the 'Entities' as an example to discuss the change.

fabaff:
I will finish the whole page and update this PR. I'm still hoping to get some feedback of a web designer about the layout. That person could maybe tell if the layout is compliant with the latest user experience.

PR #33, opened 2015-08-19
Update of verisure description

persandstrom:
I did some update to the Verisure description to make it a bit more clear what's supported for a Versirure user.


balloob:
Thanks for updating this.

PR #37, opened 2015-08-30
New layout of components index

fabaff:
This is the new PR and successor of #31 for the layout changes of the components index page. The indent level of the entries was not changed to keep the option of creating a big table of all entries.

### DIFF
--- a/source/components/index.markdown
+++ b/source/components/index.markdown
@@ -10,208 +10,264 @@ footer: true
 is_homepage: true
 ---
 
-Components add support for devices, automation and much much more to Home Assistant. The following things are supported out-of-the-box.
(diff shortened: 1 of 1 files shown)


balloob:
The screenshot looks good. Since your'e using the grid system the layout will also work on small devices.

I say: let's merge! :+1:

balloob:
Because I didn't want to create merge conflicts for your PR, the following links still need to be added to component page:
 - switch.mqtt
 - sensor.mqtt
 - device_tracker.aruba

fabaff:
Thanks. I will take care about the links.

balloob:
I added Aruba, didn't add MQTT because not sure if it fitted really, they are already linked underneath MQTT.

PR #38, opened 2015-09-02
Refactor Getting Started page.

rmkraus:
This commit is a large refactoring of the getting started page. This
version contains only the information necessary to get off the ground
and get your bearings. There is also one consistent flow on information
rather than a spaghetti flow. Advanced installation details have been
moved to their own page (Docker and daemons). Details about opening
firewall ports have been expanded a bit and moved to the
troubleshooting page. The install instructions contain details about
all three install types (production, beta, and dev).

Let me know what you think. I think this is a bit easier to follow.

### DIFF
--- a/source/getting-started/advanced.markdown
+++ b/source/getting-started/advanced.markdown
@@ -12,34 +12,46 @@ footer: true
 Here are some general tutorials on how to setup some of the more advanced deployments that are frequently requested.
 
 <div class='advanced-installs-container'>
-<input name='advanced-installs' type='radio' id='debian-install' checked>
-<input name='advanced-installs' type='radio' id='fedora-install'>
-<label class='menu-selector debian' for='debian-install'>Debian (Ubuntu, Raspbian) Daemon</label>
-<label class='menu-selector fedora' for='fedora-install'>Fedora (RHEL, CentOS) Daemon</label>
+<input name='advanced-installs' type='radio' id='upstart-install' checked>
+<input name='advanced-installs' type='radio' id='systemd-install'>
+<label class='menu-selector upstart' for='upstart-install'>Upstart Daemon</label>
+<label class='menu-selector systemd' for='systemd-install'>Systemd Daemon</label>
 
-<div class='advanced-installs debian'>
-**Debian Deamon**
-<p>Debian based systems, including Ubuntu and Raspbian for the Raspberry Pi use an application called Init to manage daemon services. Init will launch init scripts that are located in the directory <code>/etc/init.d/</code>. A sample init script for Debian based systems is <a href="https://raw.githubusercontent.com/balloob/home-assistant/dev/scripts/hass-daemon">maintained in this project</a>.</p>
 
-<p>To install this script, download it, tweak it to you liking, and install it by following the directions in the header. This script will setup Home Assistant to run when the system boots. To start/stop Home Assistant manually, issue the following commands:
+<div class='advanced-installs upstart'>
+Many linux distributions use the Upstart system (or similar) for managing daemons. Typically, systems based on Debian 7 or previous use Upstart. This includes Ubuntu releases before 15.04 and all current Raspian releases. If you are unsure if your system is using Upstart, you may check with the following command:
+
+```bash
+ps -p 1 -o comm=
+```
+
+If the preceding command returns the string `init`, you are likely using Upstart.
+
+Upstart will launch init scripts that are located in the directory <code>/etc/init.d/</code>. A sample init script for systems using Upstart is <a href="https://raw.githubusercontent.com/balloob/home-assistant/dev/scripts/hass-daemon">maintained by this project</a>.
+
+To install this script, download it, tweak it to you liking, and install it by following the directions in the header. This script will setup Home Assistant to run when the system boots. To start/stop Home Assistant manually, issue the following commands:
 ```bash
 sudo service hass-daemon start
 sudo service hass-daemon stop
 ```
-</p>
 
-<p>When running Home Assistant with this script, the configuration directory will be located at <code>/var/opt/homeassistant</code>. This directory will contain a verbose log rather than simply an error log.</p>
+When running Home Assistant with this script, the configuration directory will be located at <code>/var/opt/homeassistant</code>. This directory will contain a verbose log rather than simply an error log.
 
-<p>When running daemons, it is good practice to have the daemon run under its own user name rather than the default user's name. Instructions for setting this up are outside the scope of this document.</p>
-</div> <!-- DEBIAN -->
+When running daemons, it is good practice to have the daemon run under its own user name rather than the default user's name. Instructions for setting this up are outside the scope of this document.
+</div> <!-- UPSTART -->
 
-<div class='advanced-installs fedora'>
-**Fedora Daemon**
-<p>If you want that Home Assistant is lauched automatically, an extra step is needed to setup <code>systemd</code>. You need a service file to control Home Assistant with <code>systemd</code>. <!-- The <code>WorkingDirectory</code> and the <code>PYTHONPATH</code> must point to your clone git repository. --></p>
 
-<!-- WorkingDirectory=/home/fab/home-assistant/
-Environment="PYTHONPATH=/home/fab/home-assistant/" -->
 
+<div class='advanced-installs systemd'>
+Newer linux distributions are trending towards using systemd for managing daemons. Typically, systems based on Fedora or Debian 8 or later use systemd. This includes Ubuntu releases including and after 15.04, CentOS, and Red Hat. If you are unsure if your system is using systemd, you may check with the following command:
+
+```bash
+ps -p 1 -o comm=
+```
+
+If the preceding command returns the string `systemd`, you are likely using systemd.
+
+If you want Home Assistant to be launched automatically, an extra step is needed to setup systemd. You need a service file to control Home Assistant with systemd.
 
 ```bash
 su -c 'cat <<EOF >> /lib/systemd/system/home-assistant.service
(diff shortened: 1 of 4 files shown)


balloob:
I agree with streamlining it. There should be 1 page per installation platform but I think this approach is too confusing.

For one I don't like that the requirements section takes up so much height. This will scare people away with: oh shit, it seems like a lot of work to set up while it is 1 line. It's all psychological. I think we should stick with the getting started that we have now but should not be afraid to duplicate content. So let's get rid of the subtabs under local machine (Generic/Fedora etc) and just have these options on the top:
 - General
 - Raspberry Pi
 - Docker
 - Fedora
 - CentOS 

I also am not sure if we have to mention beta channel and development channel right on the getting started page. Development is covered already in [the development section](https://home-assistant.io/developers/#starting-development). Beta channel we could cover by adding a note to the bottom: run `pip3 install --pre --upgrade homeassistant` if you want to upgrade to the latest beta.

balloob:
I think that Babel has a [great example](http://babeljs.io/docs/setup/) of how to support multiple platforms in 1 setup page

rmkraus:
I did have the same thoughts about the requirements section. This definitely needs to be trimmed. 

For everything else, there were two major goals I was trying to achieve. 
1) Keep a consistent flow of information despite install path.
2) Eliminate extraneous information. By this I mean, detailed firewall management, Python installation instructions, and service daemon setups. These are important topics so I wanted to address them, but not in any detail as all of that is so specific and really outside our scope. It is simply too difficult to maintain all of these instructions for all potential platforms and configurations.

With that in mind, we are only left with two installation paths: Local Machine and Docker. I figured two wasn't really enough to justify the complexity of dividing the page up at the top level. I figured it couldn't hurt to propose something a bit different for discussion.

I also think moving the dev channel install to the development section would be perfectly fine. Moving the beta channel to the bottom would be ok to, but might make the page run a bit long. Just food for thought.

rmkraus:
Another thought that I had when I was rewriting some of this is that we probably shouldn't put too much effort into detailed instructions for enterprise Linux distributions like CentOS/RHEL. Realistically, if a user isn't comfortable figuring out how to install Python 3.4 or modify their firewall, they shouldn't be using an enterprise Linux distro.

balloob:
I don't mind having setup instructions per platform. We can still organize
that list to prioritize common platforms (common, RPi, Docker).

On Wed, Sep 2, 2015 at 9:41 AM, Ryan Kraus <notifications@github.com> wrote:

> Another thought that I had when I was rewriting some of this is that we
> probably shouldn't put too much effort into detailed instructions for
> enterprise Linux distributions like CentOS/RHEL. Realistically, if a user
> isn't comfortable figuring out how to install Python 3.4 or modify their
> firewall, they shouldn't be using an enterprise Linux distro.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/balloob/home-assistant.io/pull/38#issuecomment-137164076>
> .
>



-- 
PaulusSchoutsen.nl
It's nice to be important but it's more important to be nice.

rmkraus:
Now we are about halfway back to the original and still making good progress. I think this incorporates all of your comments. Let me know what you think. It still needs a page about contributing that should cover how to setup the dev environment as well as the fact that we use a fork and pull model. One thing at a time though.

balloob:
About the last part. That is covered in the developers section.

Will look at it tomorrow

rmkraus:
Cool cool.

source/getting-started/advanced.markdown
```diff
@@ -0,0 +1,87 @@
...
+```
+</p>
+
+<p>When running Home Assistant with this script, the configuration directory will be located at <code>/var/opt/homeassistant</code>. This directory will contain a verbose log rather than simply an error log.</p>
+
+<p>When running daemons, it is good practice to have the daemon run under its own user name rather than the default user's name. Instructions for setting this up are outside the scope of this document.</p>
+</div> <!-- DEBIAN -->
+
+<div class='advanced-installs fedora'>
+**Fedora Daemon**
```
fabaff: The systemd section is not only covering Fedora, Debian, Archlinux and a couple of derived distributions.

fabaff:
> Another thought that I had when I was rewriting some of this is that we probably shouldn't put too much effort into detailed instructions for enterprise Linux distributions like CentOS/RHEL. Realistically, if a user isn't comfortable figuring out how to install Python 3.4 or modify their firewall, they shouldn't be using an enterprise Linux distro.

Of course, where all those information publicly available but I think that if we can make the life of an user easier, we should. Copy-&-pasting a command is way faster than using Google first :-)

source/getting-started/advanced.markdown
rmkraus: Yes, it potentially could apply to all flavors of Linux. Debian based systems ship with Init now rather than systemd. It can be installed, but it is not the preferred method of managing services in Debian based systems.

The link at the top includes a few Fedora derivatives in the name to grab the reader's attention.

My thoughts on these advanced installation techniques is that they are just that. Advanced. There is no one size fits all solution unless we have one installer for every system that enforces setup procedures. I don't think we should pretend this is a copy and paste job because it simply isn't.

rmkraus:
I guess I have three contentions here:

1) Adding too much information makes the page cluttered and overwhelming. This page, the front page, and the demo are the three most important pages and they must not turn away users. Currently, when users start on this page, there is about a 45% drop off rate. When users see this page after their first interaction, there is a drop off rate of about 30%. After that, drop off rates settle at 20% (usually, users have seen the demo by now).

2) Adding too much information requires too many caveats (leading to confusion) and maintenance. If we provide specific instructions for a distro, then we need to provide all sorts of caveats about exactly what version it was tested against and under what conditions it worked. If we instruct a CentOS user that happens to be using IPTables that the proper way to open their firewall is with firewall-cmd, then they will get frustrated that our instructions don't work. Additionally, specific instructions require a great deal on maintenance to remain current and also give the impression that we are equipped/ready to answer any and all Linux configuration questions for them. This is outside our scope and we shouldn't pretend to pick it up.

3) We shouldn't try to make the life of a few users easier at the detriment of ourselves, the integrity of the information hosted, or the majority of other users. We all only have limited time to devote to this project and that time should not be wasted communicating with users how they should configure their Linux distribution. I'm not very apologetic at saying if you don't know how to use and don't want to research how to use Linux, then you shouldn't use Linux. Sorry, not sorry on this one. When we take time instructing users on how to use Linux, it takes away from time that could be better spent on more productive tasks like bug squashing, feature addition, or answering questions about important questions. Linux is, simply put, not for everyone. That's why there is Windows and Mac. Providing a reference to some solid Raspberry Pi instructions, I feel, can be useful because a large number of users need this information. The number of users coming with questions about CentOS, for example, is 0. However, providing full instructions on the page results in the flow of information being seriously and irrevocably changed. There is some very useful feedback for us on [the MySensors page regarding this](http://www.mysensors.org/controller/homeassistant). Sure we can make sure all the information appears in all views and still provide full install instructions, but then users will have a vastly different experience based on the technology their using as a server. This should not be the case.

Sorry, that turned into quite a rant there :scream:! I just think it is too much for us to pretend to be a one stop shop for all information and I find little to no utility in repeating data that already exists many times over on the internet. I'll also point out, that very little information was ACTUALLY removed by this PR. Firewall information was actually made more detailed and moved to the troubleshooting section. Daemon setups were moved to the Advanced Configuration section. Raspberry Pi Python install was reduced to an external link and the Python install for CentOS was removed.

source/getting-started/advanced.markdown
fabaff: I'm not following the discussions about the init systems in distributions but I think that Debian 8 is shipping systemd as default.

rmkraus: So they did, you are correct sir! I'm still on Ubuntu 14.04 and hadn't gotten the news that Debian was switching. So we should reword this somehow. Maybe separate the two sections by daemon manager and have each section include a note of a few popular distros that use those systems by default?

fabaff:
I agree with you, we will never be able to cover all kind of wired combinations but at the end of the day I guess that 90 % of the systems are running with the distribution's default and the users of the other 10 % are able to fix/setup everything by themselves.

I think that we are down to the removed instructions about Python 3 and RHEL/CentOS 7. After a little thinking, those information should never ended on this page. A one-liner on the troubleshooting page with a link to the details would have been enough. 

Anyway, thanks for your detailed answer. You have some valid points but we have different positions and that's good because this leads to an evolution of the getting-started section. At the end we will probably meet somewhere in the middle :-)

source/getting-started/advanced.markdown
fabaff: Sounds good to me.

rmkraus:
Definitely. My evil plan was to make a radical change and create
controversy. 😉 we needed to do something so now we have something to
iterate on. Other than changing some titles on the advanced installation
page, is there anything specific you'd like to see? I'm I hearing more
Python I station instructions? Possibly on a different page?

On Friday, September 4, 2015, Fabian Affolter <notifications@github.com>
wrote:

> I agree with you, we will never be able to cover all kind of wired
> combinations but at the end of the day I guess that 90 % of the systems are
> running with the distribution's default and the users of the other 10 % are
> able to fix/setup everything by themselves.
>
> I think that we are down to the removed instructions about Python 3 and
> RHEL/CentOS 7. After a little thinking, those information should never
> ended on this page. A one-liner on the troubleshooting page with a link to
> the details would have been enough.
>
> Anyway, thanks for your detailed answer. You have some valid points but we
> have different positions and that's good because this leads to an evolution
> of the getting-started section. At the end we will probably meet somewhere
> in the middle :-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/balloob/home-assistant.io/pull/38#issuecomment-137662230>
> .
>


-- 
Thanks,

Ryan Kraus

fabaff:
Just let's go with the current state and adjust if there is a need.

rmkraus:
That advanced install page has been updated per our discussions. Thanks for the feedback!

PR #52, opened 2015-09-12
Expand sonos text a little

rhooper:
Adjusted the Sonos component text to indicate it performs autodiscovery.


fabaff:
Thanks

balloob:
It would be cool if we add support for detection Sonos to netdisco. That way HA will auto load the platform too!

PR #54, opened 2015-09-13
add some notes about setting up for UI development

rhooper:
May not be fully accurate, but thought the extra content might help.


balloob:
I made the exact same change while we were talking about it, just didn't push it out until now.

PR #57, opened 2015-09-14
Examples for Python API

### LINKED ISSUE #50: Better (Python) API documentation with examples
Using the Python API wrapper, I'd like to turn a switch on. How do I do that?
I read the REST API documentation, I guess I have to call a service, I tried in the developer view, I can only turn on all switches. Looking at the source code did not really help me, so I think the API examples page should be extended.

### LINKED ISSUE #56: Better automation component documentation
The automation component is now the preferred way to do automations in Home Assistant. It has a lot of options. The current documentation should be put more in front of users (explicit part of the getting started?) and it should be cleaned up.

 - [ ] Add examples for event trigger
 - [ ] Add examples for if conditions
 - [ ] More usage examples
 - [ ] Mention logbook shows when automation is being triggered
 - [ ] More prominently integrated into getting started
 - [ ] Remove scheduler component documentation (it will be removed)

fabaff:
I will do this as PR because maybe other people have better samples to include.

This is an attempt to fix #50 .

### DIFF
--- a/source/developers/python_api.markdown
+++ b/source/developers/python_api.markdown
@@ -11,21 +11,99 @@ footer: true
 
 In the package [`homeassistant.remote`](https://github.com/balloob/home-assistant/blob/master/homeassistant/remote.py) a Python API on top of the [HTTP API](/developers/api.html) can be found.
 
-The two snippets below shows how to use the `homeassistant.remote` package:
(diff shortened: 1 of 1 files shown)


balloob:
Looks bueno :+1:

PR #58, opened 2015-09-15
Add Docs on AirPlay

maddox:
This adds some description to the iTunes component to explain that your can control your AirPlay endpoints.


fabaff:
Thanks

PR #59, opened 2015-09-15
Added multiple scene example

CCOSTAN:
Just added 2 more generic scenes to make it obvious how to add additional scenes.


fabaff:
Thanks

PR #70, opened 2015-09-22
Light Service required attributes change

CCOSTAN:
Changed the entries in the table to yes for the optional values like brightness and flash


fabaff:
Thanks

PR #71, opened 2015-09-23
Updated automation example

CCOSTAN:
Updated to reflect the new single automation block using alias to define each rule within.


fabaff:
Thanks

PR #74, opened 2015-09-23
New development steps

badele:


balloob:
Can we remove the reference to root folder? Just 
Git clone, cd, setup

badele:
No problem, you can remove the reference to root folder :)

PR #77, opened 2015-09-24
Added line about sensor, sensor 2, 3, etc..

CCOSTAN:
Since we don't have an actual sensor page, I think this is the best place for it to go so new users will get the gist of the format.

### DIFF
--- a/source/getting-started/devices.markdown
+++ b/source/getting-started/devices.markdown
@@ -15,6 +15,9 @@ routers, Belkin WeMo switches, Philips Hue bridges and Sonos speakers in your ne
 
 See the [components overview page](/components/) to find installation instructions for your devices
 and services.  Many components use the `sensor` platform.  Sensors need to be added into the `configuration.yaml` as `sensor:`, `sensor 2:`, `sensor 3:`, etc.  There should not be gaps in the sequence or your sensors might not load correctly.
+<p class='note'>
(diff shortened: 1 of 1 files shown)


balloob:
I think we should write this in a more generalized way. As it works the same for all components. Probably using the note format.

CCOSTAN:
I added a more generic note line.  Still hoping to see the dash format used in automation spread to these other components.

source/getting-started/devices.markdown
```diff
@@ -14,7 +14,10 @@ routers, Belkin WeMo switches, Philips Hue bridges and Sonos speakers in your ne
 [the discovery component]({{site_root}}/components/discovery.html) enabled (which is by default).
 
 See the [components overview page](/components/) to find installation instructions for your devices
-and services.
+and services.  Many components use the `sensor` platform.  Sensors need to be added into the `configuration.yaml` as `sensor:`, `sensor 2:`, `sensor 3:`, etc.  There should not be gaps in the sequence or your sensors might not load correctly.
+<p class='note'>
+Most components that support multiple entries within the `configuration.yaml` follow the `component:`, `component 2:` format.
```
balloob: Backticks do not work inside a note because Markdown parses this as HTML. Please use `<code>` and `</code>`

CCOSTAN:
Updated.

PR #79, opened 2015-09-25
Minor edits to bring up to date

CCOSTAN:
Removed some legacy labels.


balloob:
This is wrong. The script configuration syntax has not been updated.

CCOSTAN:
Ugh.. I guess this should be logged as an issue then?

balloob:
Yes, Bringing them in line would make sense

PR #88, opened 2015-10-05
Added example for using Offset

CCOSTAN:
Added example of using offset for Sunrise/sunset automations.  // @SEJeff


balloob:
I think it is confusing by naming the events since they are not real events fired in the system. Instead, we should just add a line linking to the [existing automation documentation for the sun trigger](https://home-assistant.io/components/automation.html#sun-trigger)

CCOSTAN:
Made a few changes.  Take a peek.

balloob:
I've copy pasted the line into the working branch I have now. Thanks!

PR #90, opened 2015-10-07
Sensor sequence note

CCOSTAN:
Added additional note about adding sensors in sequence.

### DIFF
--- a/source/components/sensor.forecast.markdown
+++ b/source/components/sensor.forecast.markdown
@@ -54,3 +54,5 @@ Configuration variables:
   - **ozone**: The columnar density of total atmospheric ozone in Dobson.
 
 Details about the API are available in the [Forecast.io documentation](https://developer.forecast.io/docs/v2).
+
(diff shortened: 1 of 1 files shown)


fabaff:
This circumstance is mentioned clearly at the beginning of the [devices](https://home-assistant.io/getting-started/devices.html) page and affects most if not all elements in the configuration file. I think that we should not add this to selected pages because from my point of view this leads to confusion.

CCOSTAN:
I agree except I keep fielding the same question in Chat. For whatever reason, people aren't finding it.  But Totally up to you guys.  I added it to a bunch of places but I also added it to the main device page thinking it would have resolved most queries.

Maybe all sensor pages should have a link back to the main device page instead?

fabaff:
Sorry, I moved to gitter and you to github...

The sample configuration.yaml file shows an example, too. Well, on the [devices](https://home-assistant.io/getting-started/devices.html) it looks a bit like that this format is needed for sensors only. Perhaps we should update the order.

PR #96, opened 2015-10-10
Correct configuration description

alanbowman:
The blinkstick configuration depends upon the serial number of the stick, not a port connection.  This corrects the description.


fabaff:
Thanks for the correction.

Issue #43, opened 2015-09-09
Refactor component page


balloob:
The new layout for the component page has helped a lot in making it more clear how the components work. I am however afraid that in the long run even this will turn into chaos. I think we should consider:
 - Splitting component page into three pages:
    - Devices (ie. Hue, MQTT, PushBullet)
    - Automation & Organization (ie. Automation, Scheduler, Group)
    - Functionality (ie. Logbook, Discovery)
 - Allow search/filter for device page ([example we can use](http://vestride.github.io/Shuffle/#demo), [example how it should look](http://www.google.com/cast/apps/#/gallery))

fabaff:
To be honest, at the end I was disappointed with the new column-based layout because it didn't turn out to make the page shorter or more appealing. With new stuff coming on a daily base we are on the same track as before only with column, unfortunately.

CCOSTAN:
That google cast example is really nice!

balloob:
Closing issue as this is now fixed in master.

PR #101, opened 2015-10-16
Components overview

fabaff:
In #43 we were talking about a new way to handle the components overview. This is a prototype of how it could be done. 

components

All elements are stored in an array which would make it easy to add a new one. A little Javascript and a little CSS, done :-) Well not really....
It would be nice to get some feedback before we go on. The last major rework of the components page was not very sustainable. 

It would also be nice to get some support from experienced web designers because my JS/jquery/CSS knowledge is limited and Octopress is "unknown territory" for me. Thanks.

### DIFF
--- a/sass/custom/_paulus.scss
+++ b/sass/custom/_paulus.scss
@@ -49,6 +49,7 @@
 
   &.overview {
     max-width: 100px;
+    vertical-align: middle;
   }
 }
 
--- a/source/components/index.markdown
+++ b/source/components/index.markdown
@@ -58,7 +58,7 @@ $(window).load(function(){
         }
     });
  
-    $('.componentFilter a').click(function(){
+    $('.filter-button-group').on('click', 'button', function() {
         $('.componentFilter .current').removeClass('current');
         $(this).addClass('current');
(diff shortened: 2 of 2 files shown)


fabaff:
Forgot to mention: This needs [Isotope](http://isotope.metafizzy.co/) to work.
components1

balloob:
This is awesome! 

I can fix some of the styling when I'm home again.

CCOSTAN:
Wow!  That's really great looking! :+1:

fabaff:
You are kidding right? It looks like sh....The images are horizontally not in the middle, the text link is placed somewhere depending on the image size, and there a for sure 10 other issue with the design/CSS.

I just tried to get the stone rolling...but I hope that we can create a solution that looks professional and is easy to maintain.

maddox:
maybe he can just see your vision :)

balloob:
So I took this PR, cleaned it up a bit and merged it into master. It still needs some more clean up but it's getting really pretty. Great work @fabaff !

screen shot 2015-10-24 at 21 56 31

PR #113, opened 2015-11-23
Multiple entries in config file

fabaff:
This is my idea to fix [Fix documentation about sensor, sensor 2.](https://www.pivotaltracker.com/n/projects/1250084/stories/108734488). I thought that it would be better to open a PR to gather some feedback about the comprehensibility because this topic is coming up again and again.

### DIFF
--- a/source/getting-started/devices.markdown
+++ b/source/getting-started/devices.markdown
@@ -11,13 +11,47 @@ footer: true
 
 Home Assistant will be able to automatically discover and configure any Google Chromecasts, Netgear routers, Belkin WeMo switches, Philips Hue bridges and Sonos speakers in your network if you have [the discovery component]({{site_root}}/components/discovery/) enabled (which is by default).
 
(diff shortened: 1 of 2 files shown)


source/getting-started/devices.markdown
```diff
@@ -11,13 +11,47 @@ footer: true
...
+  - platform: rest
+    resource: http://IP_ADDRESS/ENDPOINT
+
+switch:
+  - platform: vera
+```
+
+#### {% linkable_title Style 2 %}
+
+If you prefer to place your entries somewhere in the `configuration.yaml` file then you just choose this style. There should not be gaps in the sequence when you use numbers or your entities might not load correctly. Instead of numbers you are free to use strings (like `media_player livingroom:` or `media_player kitchen:`) to differentiate the entries as long as the appended strings are unique.
```
balloob: I got rid of the sequential requirement. Anything after it is fine.

balloob:
Looks good. You can merge it after fixing the one comment I had.

Also, you know you can start stories in Pivotal right to indicate there is a PR or that you are working on it (I do that now for you) :+1:

PR #120, opened 2015-12-08
Add explanation of using two device trackers

philipbl:
I have asked about this and I have seen a few other people ask about this (https://github.com/balloob/home-assistant/issues/696), so I thought I would put it in the documentation. Probably won't stop people from asking, but at least it is in there.


fabaff:
Thanks

PR #122, opened 2015-12-13
Add Let's Encrypt blogpost

balloob:
This adds a blog post how to set up encryption for Home Assistant using Let's Encrypt.

Before I put it out, can someone run the instructions and see if they also work for them?

Note: I will update the `date:` key before publishing to the correct date. This was needed to see the post in my dev environment.


fabaff:
Two alternatives to create the certificates:

Beside the quick way with Docker you could also generate the keys manually. First clone the git respository.

```bash
$ git clone https://github.com/letsencrypt/letsencrypt
$ cd letsencrypt
```

We only want a certificate thus we are using the `certonly` subcommand. The script is excepting some details like your e-mail address and the domain.  

```bash
$ ./letsencrypt-auto certonly --email $ ./letsencrypt-auto certonly --email your@email.address -d hass.duckdns.org
Updating letsencrypt and virtual environment dependencies.......
Running with virtualenv: sudo /path/letsencrypt/bin/letsencrypt certonly --email your@e-mail.address -d hass.duckdns.org

IMPORTANT NOTES:
 - Congratulations! Your certificate and chain have been saved at
   /etc/letsencrypt/live/hass.duckdns.org/fullchain.pem. Your cert
   will expire on 2016-03-12. To obtain a new version of the
   certificate in the future, simply run Let's Encrypt again.
 - If like Let's Encrypt, please consider supporting our work by:

   Donating to ISRG / Let's Encrypt:   https://letsencrypt.org/donate
   Donating to EFF:                    https://eff.org/donate-le
```

Keep an eye on the permissions. If you are running Home-Assistant as user then you need to allow this user read access for  `/etc/letsencrypt/live/hass.duckdns.org/`

The third way would be the [Get HTTPS for free!](https://gethttpsforfree.com/) webservice that is using [Let's Encrypt] but is for people who are familiar with `openssl` and don't want to use docker or the script.

fabaff:
A screenshot as proof ;-)

ha-ssl

balloob:
Thanks for the valuable feedback Fabian!

PR #124, opened 2015-12-15
fixed example typo and added Billion Router

Neko22:
Apologies if I've done this wrong


balloob:
Nop, perfect. It will go out the next release of the site.

Landrash:
Would we like to expand this list further or just have the current OID's as examples?`
If so I'll make a commit with DD-WRT.

PR #134, opened 2015-12-23
Rename geofancy to locative

philipbl:
Update documentation to refer to Locative instead of Geofancy (from https://github.com/balloob/home-assistant/pull/786).


GreenTurtwig:
Seems like they've rebranded :-)

PR #139, opened 2016-01-02
Evohome thermostat doc change

sander76:
added optional away_temperature and removed "Round" spec as last commit made all honeywell connected thermostats possible.


balloob:
Can you close this PR and create a new one that points at the `next` branch since your code has not been released yet.

PR #147, opened 2016-01-12
free_mobile platform documentation

HydrelioxGitHub:
I do not know if it's the right way to do this kind of stuff. But I try. 
There is also a picture and a logo


fabaff:
It's the right way :-) Thanks.

PR #150, opened 2016-01-13
Added documentation for the Universal Media Player

rmkraus:
Title says it all with this one.


balloob:
Can you point your PR at the `next` branch?

source/_components/media_player.universal.markdown
```diff
@@ -0,0 +1,100 @@
...
+title: "Universal"
+description: "Instructions how to create a universal media player in Home Assistant."
+date: 2016-01-12 22:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo:
+ha_category: Media Player
+featured: true
```
balloob: Please don't feature it as we can only feature a component if we un-feature something else (UI looks weird if there are not 15 featured components). I consider this an advanced component and prefer not to feature it right now.

balloob:
Two minor comments but looks good :dolphin: Merge once comments fixed.

source/_components/media_player.universal.markdown
rmkraus: My bad, I didn't mean to feature this one and totally agree with you.

PR #169, opened 2016-01-28
ZigBee component documentation

flyte:
This PR provides the documentation for the ZigBee integration and components.

I haven't been able to test this on HA itself because I wasn't able to get rake to install all of its components properly. It looks fine on Github though.


balloob:
This PR has to be pointed at the `next` branch as it is for the next release.

balloob:
Make sure you use the right Ruby version when trying to install gems. I use `rbenv` to use a different ruby version inside the HA.io repo.

flyte:
Which version of Ruby is the right version?

balloob:
cat .ruby-version

On Thu, Jan 28, 2016 at 9:10 AM, Ellis Percival <notifications@github.com>
wrote:

> Which version of Ruby is the right version?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/balloob/home-assistant.io/pull/169#issuecomment-176286992>
> .
>



-- 
PaulusSchoutsen.nl
It's nice to be important but it's more important to be nice.

MartinHjelmare:
RVM and Ruby 2.2.3 failed to install some of the gems. Then I tried Ruby 2.2.1, which I already had setup for Octopress, and that worked.

PR #173, opened 2016-01-29
Added Google Voice doc

w1ll1am23:
Documentation for https://github.com/balloob/home-assistant/pull/1044


balloob:
It's targeted at the master branch and also includes the html file that is not needed. I'll just copy paste your page into the next branch that I'm working on right now! Thanks.

PR #181, opened 2016-01-31
Add documentation about SCSGate

flavio:
This adds the documentation about the SCSGate code that has been merged with [this](https://github.com/balloob/home-assistant/pull/1056) pull request.

There's one thing I don't understand, only the main page about SCSGate and the one about the lights module show up. The others are not translated to html. Do you have any idea about why this is happening?

### DIFF
--- a/source/_components/light.scsgate.markdown
+++ b/source/_components/light.scsgate.markdown
@@ -1,0 +1,29 @@
+---
+layout: component
+title: "SCSGate light"
+description: "Instructions how to integrate SCSGate lights into Home Assistant."
+date: 2016-01-31 19:30
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: Light
+---
+
+The SCSGate device can control lights of the BTicino MyHome system.
+
+To enable SCSGate lights in your installation, add the following to your `configuration.yaml` file:
+
+```yaml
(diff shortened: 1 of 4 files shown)


balloob:
Can you point this PR at the `next` branch? Also, no need to create the HTML files. Those are only there for backwards compatibility/old URLs generated by the old version of our site generator.

balloob:
Jekyll is 'smart' and does not recompile every page on change if you're running it with `--incremental`. Restarting local compilation should have the pages be shown.

flavio:
> Jekyll is 'smart' and does not recompile every page on change if you're running it with --incremental. Restarting local compilation should show the pages.

That's strange, I never had this problem with Jekyll before (I also tried a `rake clean && rake generate` but it didn't help.

However, that fixed the issue.

I've just one little problem: when I go to the SCSGate component the lights are the only suggested module being shown. Neither the switch nor the roller shutter are there. Any idea?

flavio:
> Can you point this PR at the next branch? 

Sorry, my fault. I'll do that, but unfortunately I'll have to create a new PR.

flavio:
> Also, no need to create the HTML files. Those are only there for backwards compatibility/old URLs generated by the old version of our site generator.

I'll get rid of them with my next PR

PR #182, opened 2016-02-01
Add documentation about SCSGate

flavio:
This is a new version of https://github.com/balloob/home-assistant.io/pull/181

I've just one little problem: when I go to the SCSGate component the lights are the only suggested module being shown. Neither the switch nor the roller shutter are there. Any idea?


fabaff:
Thanks

PR #184, opened 2016-02-02
Added docker port and updated docs

jacobtomlinson:
The port exposure was missing from the docker command.

Also boot2docker has been replaced with Docker Machine, updated instructions to reflect that.


balloob:
We do not have to add the port because we're using `--net=host`. This will give Home Assistant full access to the host network. The reason we want this is because our discovery component does uPNP and mDNS broadcasts which will not work if isolated by docker.

jacobtomlinson:
Ah that makes sense.

PR #191, opened 2016-02-06
HTTP binary sensor and sensor docs

fabaff:
This PR contains some more details about the usage of the HA REST API. More information for sensors and binary sensors are provided.

Till now the only place to find information about this feature was embedded in the API docs. I hope that this could make the API more visible and stop me for doing stuff that is already there.

### DIFF
--- a/source/_components/binary_sensor.http.markdown
+++ b/source/_components/binary_sensor.http.markdown
@@ -11,18 +11,17 @@ logo: http.png
 ha_category: Binary Sensor
 ---
 
-
-The `http` binary sensor platform is not a real platform within the meaning of the terminology used around Home Assistant. Home Assistant's [REST API](/developers/rest_api/) is consuming and proceeding messages recieved over HTTP. 
-
-To use those kind of sensors in your installation no configuration in Home Assistant is needed. All configuration is done on the devices themself. This means that you must be able to edit the target URL or endpoint and the payload. The entity will be created after the first message has arrived.
-
-All [requests](/developers/rest_api/#post-apistatesltentity_id) needs to be sent to the endpoint of the device and must be **POST**. The URL looks like the example below:
+The URL for a binary sensor looks like the example below:
 
 ```bash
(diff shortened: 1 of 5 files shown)


fabaff:
I'm open for suggestions and feedback :-D

fabaff:
I would be nice if someone could verify the stuff I wrote. Just be sure that it's complete rubbish. Thanks.

balloob:
It looks good.

I think we should somehow be able to flag each page that we put out as a platform/component that is actually just a tutorial to use some other part of Home Assistant. There is `sensor.http`, `binary_sensor.http`, `notify.mqtt` and `ifttt.manything`.

Please be aware that this works for the `sensor` and `binary_sensor` platforms as the components do not provide any services to interact with the entities. This will not work for a light because services as `light.turn_on` etc will not work.

source/_components/binary_sensor.http.markdown
```diff
@@ -0,0 +1,78 @@
...
+---
+
+The URL for a binary sensor looks like the example below:
+
+```bash
+http://IP_ADDRESS:8123/api/states/binary_sensor.DEVICE_NAME
+```
+
+<p class='note'>
+It's suggested that you choose an unique device name (DEVICE_NAME) to avoid clashes with other devices.
```
balloob: It's not only suggested, it's necessary. "You should choose a unique…"

balloob:
You can merge it after fixing that 1 sentence. :dolphin:

fabaff:
> I think we should somehow be able to flag each page that we put out as a platform/component that is actually just a tutorial to use some other part of Home Assistant. There is sensor.http, binary_sensor.http, notify.mqtt and ifttt.manything.

Agreed. I started to realize while creating the MQTT notification page that this kind of documentation doesn't fit in the existing categories. Due to the lack of better ideas I just placed them in the category which fits best from my point of view.

balloob:
I think that is fine. That is the place people will look for it.

PR #193, opened 2016-02-06
Add more complex sensor.template example.

pavoni:
I thought it would be useful to show a more complex example.

Happy to drop if it's too complex!


balloob:
Can it be a bit more complex, maybe compare a number?

PR #194, opened 2016-02-06
Add switch.template.

pavoni:
First cut - please check!

### DIFF
--- a/source/_components/switch.template.markdown
+++ b/source/_components/switch.template.markdown
@@ -1,0 +1,102 @@
+---
+layout: component
+title: "Template switch"
+description: "Instructions how to integrate Template switches into Home Assistant."
+date: 2016-02-06 07:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: Switch
+---
+
+The `template` platform creates switches that combine other components.
+
+To enable Template switches in your installation, add the following to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+switch:
+    platform: template
+    switches:
+        skylight:
+            friendly_name: 'Skylight'
+            value_template: {% raw %}'{{ states.sensor.skylight.state }}'{% endraw %}
+            turn_on:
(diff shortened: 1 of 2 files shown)


source/_components/switch.template.markdown
```diff
@@ -0,0 +1,102 @@
...
+
+- **switches** array (*Required*): List of your switches.
+  - **friendly_name** (*Optional*): Name to use in the Frontend.
+  - **value_template** (*Required*): Defines a [template](/getting-started/templating/) to set the state of the switch.
+  - **turn_on** (*Required*): Defines an [action](/getting-started/automation/) to run when the switch is turned on.
+  - **turn_off** (*Required*): Defines an [action](/getting-started/automation/) to run when the switch is turned off.
+
+## {% linkable_title Examples %}
+
+In this section you find some real life examples of how to use this sensor.
```
balloob: use this *switch*

source/components/switch.template.html
```diff
@@ -0,0 +1 @@
+<script>document.location = '/components/switch.template/';</script>
```
balloob: No need to add this file, it's only for backwards compatibility for components that existed before we migrated jekyll version

source/_components/switch.template.markdown
```diff
@@ -0,0 +1,102 @@
...
+description: "Instructions how to integrate Template switches into Home Assistant."
+date: 2016-02-06 07:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: Switch
+---
+
+The `template` platform creates switches that combine other components.
```
balloob: We could have some more explanation here how other components are actually combined.

fabaff:
Sorry, I didn't check the PRs and added already something. Would be best to get rid of my commit, I assume.

balloob:
I went ahead and copy pasted the content into the docs page. Fastest solution.

pavoni:
Thanks guys.

PR #204, opened 2016-02-10
Fix git URLs and order of execution in developer guide bash commands.

flyte:
You'll want to use the SSH URL for your own repo and the HTTPS one for the upstream. Also, `cd` into the repo before trying to add upstream.

I don't know whether this should be pointed at `master` or `next` since it's a bugfix. I used the workflow of "edit this page on github" which defaulted to `master`.


fabaff:
`master` is fine. Thanks for the fix.

balloob:
Why would you want to use the SSH url over the HTTPs url ?

flyte:
@balloob mainly because the README was already using one ssh and one https url and it made more sense for them to be the either way around. However, I've just looked it up and it is no longer the case that you can't clone a repo you don't have push access to over ssh. It might in fact be the most compatible for them to both be https links since not all OSes contain an ssh client.

balloob:
I agree. I'll change them to be both https.

PR #206, opened 2016-02-11
Update honeywell docs to indicate region instead of thermostat id

kk7ds:
In #1196, we replaced the need for specifying the thermostat id
with just the region to indicate which type of interface we should
use.


.ruby-version
```diff
@@ -1 +1 @@
-2.2.3
+2.3.0
```
balloob: I'm down to upgrade but we should update gems and run bundle too

kk7ds: Sorry, didn't mean to commit this. Lemme remove.

PR #210, opened 2016-02-12
Speedtest

nkgilley:


source/_components/sensor.speedtest.markdown
```diff
@@ -0,0 +1,72 @@
...
+    - download
+    - upload
+```
+
+Configuration variables:
+
+- **monitored_conditions** array (*Required*): Conditions to display in the frontend.
+  - **ping**: A human-readable text summary.
+  - **download**: The type of precipitation occurring.
+  - **upload**: The average expected intensity of precipitation occurring.
```
philipbl: Ping, download and upload descriptions do not seem to be accurate. Unless bandwidth is somehow related to precipitation. :wink:

nkgilley: Ha, good catch!  Copied this from forecast.io... :grimacing:

PR #214, opened 2016-02-13
Update weblink.markdown

Gerto:
Added undocument icon option in example + configuration variables like most other components have


balloob:
Thanks!

PR #216, opened 2016-02-14
Update developer docs for using tox

kk7ds:


.ruby-version
```diff
@@ -1 +1 @@
-2.2.3
+2.3.0
```
balloob: …

kk7ds:
I really should rebuild my rbenv so I stop doing that. Apologies.

PR #220, opened 2016-02-15
Update mysensors main and light pages

MartinHjelmare:


MartinHjelmare:
I changed the date and time to today, without specifiiying timezone and now the updated pages don't exist, cause they are in the future :watch: . I'm going to try and add a timezone to the date in the front matter.

Found this as ref:
https://github.com/jekyll/jekyll/issues/4010

PR #223, opened 2016-02-15
Try to fix pages not showing by adding timezone

MartinHjelmare:
* Add timezone.
* Remove types from switches page.


balloob:
This is a good example as to why you should test your changes locally :+1:

MartinHjelmare:
Yeah, I did that by ```rake preview```, and it looked fine. Haven't really figured out why it didn't work after the build on the server, when the preview was good.

balloob:
Weird. It should use UTC locally to preview

PR #226, opened 2016-02-16
neurio documentation

infamy:
Added neurio documentation


infamy:
closing, needs to be next to next, not master to next.

PR #233, opened 2016-02-19
Add nx584 sensor platform documentation

kk7ds:


source/_components/binary_sensor.nx584.markdown
```diff
@@ -0,0 +1,50 @@
...
+layout: page
+title: NX584 zones
+description: "Instructions on how to set up nx584 zones as sensors"
+date: 2016-02-18 20:47
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: networx.png
+ha_category: Binary Sensor
```
balloob: Could you add:
```yaml
ha_iot_class: "Local Push"
```

kk7ds:
Note this needs to wait until #1308 merges because it describes the zone_types config element which is added in there.

PR #235, opened 2016-02-19
Updated google voice doc with automation notification example

w1ll1am23:
There have been a few people confused on how to setup the notification targets in a notification call. I think this should help with the confusion.

### DIFF
--- a/source/_components/notify.google_voice.markdown
+++ b/source/_components/notify.google_voice.markdown
@@ -37,14 +37,15 @@ Google Voice is a notify platform and thus can be controlled by calling the noti
 
 ```yaml
 # Example automation notification entry
-  - alias: Send a notification
+automation:
+  - alias: The sun has set
     trigger:
-      platform: event
-      event_type: 'some event'
+      platform: sun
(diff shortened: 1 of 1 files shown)


source/_components/notify.google_voice.markdown
```diff
@@ -34,3 +34,18 @@ Configuration variables:
 ### Usage
 
 Google Voice is a notify platform and thus can be controlled by calling the notify service [as described here](/components/notify/). It will send a notification to all devices listed in the notification **target**.
+
+```yaml
+# Example automation notification entry
+  - alias: Send a notification
```
balloob: Could you add the automation header above this line so people know it's for that component?

source/_components/notify.google_voice.markdown
```diff
@@ -34,3 +34,18 @@ Configuration variables:
...
+    trigger:
+      platform: event
```
balloob: Can we make a trigger that actually works, ie have it trigger at a specific time. That way people can copy paste and go.

PR #237, opened 2016-02-20
Allow preview with a different bind address

kk7ds:
This makes the Rakefile preview task honor a listen address instead of
always defaulting to 127.0.0.1. This makes it easier to run the preview
on a machine other than where you have a web browser running.

To use:

  rake preview[0.0.0.0]

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -77,13 +77,14 @@ task :watch do
 end
 
 desc "preview the site in a web browser"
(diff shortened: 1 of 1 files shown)


kk7ds:
This is how google tells me to do this...

balloob:
Must make sense to someone ;-)

PR #238, opened 2016-02-20
Add unifi device tracker documentation

kk7ds:


source/_components/device_tracker.unifi.markdown
```diff
@@ -0,0 +1,36 @@
...
+layout: page
+title: "Ubiquiti Unifi WAP"
+description: "Instructions how to use a Unifi WAP controller as a device tracker module."
+date: 2016-02-19 20:59
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: ubiquiti.png
+ha_category: Presence Detection
```
balloob: Could you add:
```yaml
ha_iot_class: "Local Poll"
```

balloob:
You can merge it after the comment is addressed.

source/_components/device_tracker.unifi.markdown
kk7ds: Should this be "Local Polling"?

https://github.com/balloob/home-assistant.io/blob/master/source/_components/light.hue.markdown

I guess I missed the start of the tagging effort, sorry about that. Excellent idea though.

balloob: Oops you're right. 

The tags are not being used yet but I plan to use it to highlight the more useful components on the site

happyleavesaoc: "Useful" is pretty subjective. "Accurate" might be a better descriptor.

PR #240, opened 2016-02-20
Development tutorial

fabaff:
This Pull Request is for tracking the creation of a simple tutorial about the development of a new Home Assistant component including frontend and tests. The tutorial will hopefully become something like the expended `hello_world` example for Home Assistant. The target audience for this tutorial are people who are looking for an overview of the involved steps or just wants to add something new but are only occasionally writing code.

Every kind of input is welcome. Especially for the missing sections (frontend & tests).

Pivotal story: https://www.pivotaltracker.com/story/show/113103013


source/developers/tutorial.markdown
```diff
@@ -0,0 +1,140 @@
...
+The setup of a development environment is described in the [Developers section](/developers/#starting-development) of the documentation.
+
+## {% linkable_title Component %}
+
+
+As a start, create a file `information.py` in your Git checkout of Home Assistant in the folder `homeassistant/component/`.
+
+```python
+"""
+homeassistant.components.information
```
balloob: I've been trying to adhere more files to the PEP257 documentation standard. Part of this is that we should update our headers to have a description of the module in the first line. As people reading this tutorial might become contributors it would be great if we give the right example here.

source/developers/tutorial.markdown
```diff
@@ -0,0 +1,140 @@
...
+information:
+```
+
+If you replace `_LOGGER.info("The 'information' component is ready!")` with or add
+
+```python
+hass.states.set('information.Text', 'Info component')
+```
+
+then the component will not be different to the sample included in the `config/custom_components` folder or shown as an [example](/cookbook/python_component_basic_state/). After a start or a restart of Home Assistant the component will be visible in the frontend.
```
balloob: I don't think people realize the config/custom_components folder exists because they are no longer cloning the repo. Let's make sure we link to it.

source/developers/tutorial.markdown
```diff
@@ -0,0 +1,140 @@
+---
+layout: page
+title: "Development tutorial"
```
balloob: The name is too generic. It should mention what the end result is of the tutorial.

balloob:
I wonder if a tutorial should be posted as a page under the developer section, a blog entry or as a cookbook entry. My preference would be one of the latter two because those do not occupy an entry in the menu ("ve been trying to keep the menu small because it becomes 1 big list on mobiles).

source/developers/tutorial.markdown
fabaff: Sure. Sorry, at the moment I'm a bit behind the current changes.

source/developers/tutorial.markdown
fabaff: Will be added.

source/developers/tutorial.markdown
balloob: I haven't announced it yet. I have just been sneaking it into my PRs when I fix other things

fabaff:
Now, I tend to place it in the cookbook section instead of the development section and  merge the new content with the existing ` Custom Python Component Examples` especially the `Basic State Setting Example`.

I think that `hello_state` can serve perfectly as an expended example. So there is no need to introduce an additional page.

balloob:
I didn't realize that this was done :disappointed: sorry! Merging it.

fabaff:
Part 1 is done. Part 2 (Frontend) is work in progress.

Doesn't make to sense to waiting any longer here. Perhaps I will get some more feedback if the stuff is online.

PR #246, opened 2016-02-22
Added SCS bus icon

t30:


balloob:
Can you clean up your commits?

PR #256, opened 2016-02-25
Binary template docs and an extra example

kk7ds:
This adds docs for the template binary_sensor.

Also adds an automation example using a sensor and a notification, which seemed to be uncovered so far (took me a little fiddling to figure it out myself).

### DIFF
--- a/source/_components/binary_sensor.template.markdown
+++ b/source/_components/binary_sensor.template.markdown
@@ -1,0 +1,51 @@
+---
+layout: page
+title: "Template Binary Sensor"
+description: "Instructions how to integrate Template binary sensors into Home Assistant."
+date: 2016-02-25 15:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: Binary Sensor
+---
+
+The `template` platform supports sensors which breaks out the state and `state_attributes` from other entities.
+
+To enable Template sensors in your installation, add the following to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+binary_sensor:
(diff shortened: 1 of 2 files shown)


source/_cookbook/notify_if_over_threshold.markdown
```diff
@@ -0,0 +1,50 @@
...
+      entity_id: sensor.furnace
+      below: 2
+    action:
+      service: notify.mypushbullet
+      data:
+        title: "Furnace fan is stopped"
+        message: "Fan stopped because current is {% raw %}{{ states.sensor.furnace.state }}{% endraw %} amps"
+```
+
+Note that in this case, passing the state value through `round()` is required because state is a string and we need it to be converted to an integer before we compare it to another one. If you were comparing against an attribute that is already a number, then `round()` is not required.
```
balloob: I don't see round being used in the examples

source/_components/binary_sensor.template.markdown
```diff
@@ -0,0 +1,51 @@
...
+description: "Instructions how to integrate Template binary sensors into Home Assistant."
+date: 2016-02-25 15:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: Binary Sensor
+---
+
+The `template` platform supports sensors which breaks out `state_attributes` from other entities.
```
balloob: Not just attributes, can be state too

kk7ds: I just copied this from sensor.template :)

fabaff:
Thanks

PR #258, opened 2016-02-27
Add ES6800 to supported Samsung TVs

timstanley1985:


timstanley1985:
Will add to master instead

PR #261, opened 2016-02-27
Add SendGrid notify platform

justincmoy:


source/components/notify.sendgrid.html
```diff
@@ -0,0 +1 @@
+<script>document.location = '/components/notify.sendgrid/';</script>
```
fabaff: This file is not needed for new parts.

balloob:
Can you point it at the `next` branch?

justincmoy:
https://github.com/balloob/home-assistant.io/pull/262

PR #263, opened 2016-02-28
Fix mysensors binary sensor and some fixes

MartinHjelmare:
* Fix binary sensor page. Move door, motion, smoke, and only part of
  sprinkler, leak, sound, vibration and moisture sensors to binary
  sensor from sensor page.
* Move back S_LIGHT to switch page from light page.


balloob:
I had on my to do list to ping you about this. Thanks for fixing it :+1: .

PR #269, opened 2016-03-01
Remove unwanted backslashes from filename

larsblumberg:
Back slashes should not be rendered in https://home-assistant.io/developers/add_new_platform/


fabaff:
Thanks

PR #273, opened 2016-03-02
Update mysensors light and switch pages

MartinHjelmare:
Clarify some of the V_TYPES used for these actuators.


balloob:
You can merge this without me approving.

PR #274, opened 2016-03-02
Missing disarmed state

v3rm0n:
No armed state in constants, but disarmed is missing here
https://github.com/balloob/home-assistant/blob/dev/homeassistant/const.py#L56


balloob:
Thanks

PR #276, opened 2016-03-04
Add another reported/confirmed Mikrotik SNMP OID

kk7ds:
This adds a new OID confirmed by a RouterOS 6.x user (the existing one
provides no results on that platform). Also correct misspelling in the
existing Mikrotik example.


balloob:
Is this a change only relevant to next version or also current?

kk7ds:
Unrelated to version, so this should be pushed to the current site. I was going to merge this to next and then ping you to push to master.

PR #285, opened 2016-03-08
Update binary_sensor.command_line.markdown

YanK-fr:


balloob:
No description?

YanK-fr:
Sorry,
On Docker version ping reply "1 packets received" and not "1 received".
I don't know if is a best solution but is one.

balloob:
Ping has different output on different platforms. Updating it for one over
the other won't help much. Just confuse other people.

Maybe we should update the example for something non ping related / cross
platform

YanK-fr:
Anyway the command line work in docker console but doesn't work in home-assistant.
You are right another exemple whitout ping and runing in docker is best.

fabaff:
My idea of the examples in the platform/component docs is/was to provide a starting point for that entity. I guess that one will figure out that `ss` is the successor of `netstat` if there are hints that `netstat` is not available on the used operating system.

The Binary sensor docs already show four different samples. Two are covering `ping` because checking the online status is something that is coming up over and over again. And the other two use a different approach. None will work without any modification (except the first one which will likely work on a all Linux machines).

YanK-fr:
My mistake was to look for a solution to do this on my Docker version of home assistant.
The last solution work on docker console and it's true this solution is a best alternative.
But i don't know why home-assistant command line doesn't work with this exemple on docker.

fabaff:
What do you mean by "home-assistant command line"?

YanK-fr:
If i use command_line platform or nmap_tracker both doesn't work but if i use nmap or ping directly in shell (in docker container) it's work.
I make this post on forum : https://automic.us/forum/viewtopic.php?p=1025&sid=4e55ee1d8c861c746a83ca01f06f58c7#p1025
I think there are something doesn't work with ping or nmap with python and docker.

balloob:
I agree with @fabaff here that examples are merely starting points. Closing this PR as the existing example will do.

PR #289, opened 2016-03-10
Update switch.command_line.markdown with example including statecmd

joemcmonagle:
Posted this example to Gitter: https://gitter.im/balloob/home-assistant?at=56e10e516fde057c26856411

None of the current examples for command_line switches show support for the statecmd along with value_template, so I think this might be a useful addition to the page. I'm happy to update the example with any other explanation


fabaff:
Thanks

PR #290, opened 2016-03-10
Update camera.foscam.markdown to include Pan/Tilt/Zoom Examples

joemcmonagle:
Posted this example to Gitter: https://gitter.im/balloob/home-assistant?at=56e10e516fde057c26856411

Foscam supports commands including Motion Detection Activation and Pan/Tilt/Zoom to be send via cURL. Added an example of this, but these could be used in any number of ways. I'm open to making the example less verbose, if there's any thought that something could be taken out, but I think this would be helpful for other Foscam users, as HA can control almost everything about the camera via the cURL commands.


fabaff:
Wouldn't the [Cookbook](https://home-assistant.io/cookbook/) be the better place for this addition? Sure it for a Foscam but involves various other parts.

joemcmonagle:
Fabaff, yes, that's a good point. Makes way more sense from an organization point of view.

I still think it would be helpful to make some note of the fact that HA can control more of Foscam with additional setup. Would it make sense just to leave a link to the Foscam CGI PDF along with a link to the to-be-created Cookbook page?

fabaff:
> I still think it would be helpful to make some note of the fact that HA can control more of Foscam with additional setup. Would it make sense just to leave a link to the Foscam CGI PDF along with a link to the to-be-created Cookbook page?

Sure. It makes perfect sense to add links between different parts of the documentation. We do this at a couple of places and so far nobody is complaining.

joemcmonagle:
Updated page to refer to cookbook entry. This was already pointing to `balloob:next` branch, and PR298 is pointing to `balloob:next`, so the new cookbook page should be created prior to this page being updated.

In hindsight, I probably should just added the cookbook entry to this PR I guess?

fabaff:
Thanks

PR #292, opened 2016-03-10
Update switch.markdown to include toggle

joemcmonagle:
Turn_on, turn_off, and *toggle* are all services for switches


fabaff:
Thanks

PR #297, opened 2016-03-11
Added scene as a platform

sander76:
Changed home assistant scene docs to a scene platform doc
Added powerview scene documentation.


fabaff:
Thanks

PR #298, opened 2016-03-12
Add Foscam automation to Cookbook

joemcmonagle:
Suggested by @fabaff that this would fit better as a cookbook. Reference in [PR29](https://github.com/balloob/home-assistant.io/pull/290)

> Foscam supports commands including Motion Detection Activation and Pan/Tilt/Zoom to be send via cURL. Added an example of this, but these could be used in any number of ways. I'm open to making the example less verbose, if there's any thought that something could be taken out, but I think this would be helpful for other Foscam users, as HA can control almost everything about the camera via the cURL commands.


fabaff:
Thanks.

PR #300, opened 2016-03-13
Make updating instructions more obvious

lindsaymarkward:
Make updating instructions more obvious.
Could I also suggest that the "Update Available" option that I see on my States page links directly to this fragment? Currently it just links to the main page and I have to scroll through to figure out how to update.
Thanks

### DIFF
--- a/source/getting-started/index.markdown
+++ b/source/getting-started/index.markdown
@@ -285,7 +285,8 @@ For additional help, in addition to this site, there are four sources:
  - [GitHub Page](https://github.com/balloob/home-assistant/issues) for issue reporting.
 
 ### Updating
-To update Home Assistant to the latest release run: `pip3 install --upgrade homeassistant`
+To update Home Assistant to the latest release run: `pip3 install --upgrade homeassistant`  
+You then have to restart Home Assistant for the changes to take effect. If you have installed it as an autostarting daemon (see below), then run: `sudo service hass-daemon restart` 
 
 ### What's next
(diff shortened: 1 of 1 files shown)


lindsaymarkward:
It looks like I had to restart Home Assistant, so I added that to the instructions.

source/getting-started/index.markdown
```diff
@@ -284,11 +284,13 @@ For additional help, in addition to this site, there are four sources:
  - [Gitter Chatroom](https://gitter.im/balloob/home-assistant) for general Home Assistant discussions and questions.
  - [GitHub Page](https://github.com/balloob/home-assistant/issues) for issue reporting.
 
+### Updating
+To update Home Assistant to the latest release run: `pip3 install --upgrade homeassistant`  
+You then have to restart Home Assistant for the changes to take effect. If you have installed it as an autostarting daemon (see below), then run: `sudo service hass-daemon restart`
```
fabaff: I think that we should not add a daemon-specific command here. From my point if a user is using some kind of autostarting mechanism she/he is aware of the commands to start/stop/restart the used daemon or init implementation.

fabaff:
I like the idea of a new section Makes it indeed more obvious.

lindsaymarkward:
Sure. Adjust to whatever is best. I expected it to work when I updated, but it didn't do anything new until I restarted it, so that's something newbies like me would need to know.

PR #311, opened 2016-03-18
Adds component entry for MoldIndicator sensor

xifle:
The Mold Indicator sensor component consumes information of two temperature sensors and a humidity sensor to give an indication for possible mold growth in your home. Also see the corresponding pull-request to the home-assistant project


bernamorillo:
I have synology 215j and I don`t know if it supports home assistant.

I have tried but it is not working.

Also I installed synology comunity package but it was launched in June 2015 (too old)

Any help?

I will use a rasberry only for this instead of synology NAS

Thanks and best regards

balloob:
Merged manually into next.

PR #312, opened 2016-03-19
add wake on lan page

joopert:
Add a wake on lan page, related to https://github.com/balloob/home-assistant/pull/1574


balloob:
Thanks

PR #313, opened 2016-03-19
Clarified and added detail for devices, customizations and groups

lindsaymarkward:
added example of group views that refer to other groups.

I've been going through the docs and using the Gitter chat to understand them, so I'm updating the docs to be clearer. 
(I think more examples of configuration files with images to match would be *really* helpful.)

### DIFF
--- a/source/_components/group.markdown
+++ b/source/_components/group.markdown
@@ -15,10 +15,11 @@ Groups allow the user to combine multiple entities into one. A group can be prom
 
 Check the **Set State** page from the **Developer Tools** and browse the **Current entities:** listing for all available entities.
 
+By default, every group appears in the HOME tab. If you name a group `default_view` it will REPLACE the contents of the HOME tab so you can customize it as you wish.
+
 ```yaml
 # Example configuration.yaml entry
 group:
-  # If you name an entry default_view it will REPLACE the contents of the "Home" tab
   default_view:
     view: yes
     entities:
@@ -42,9 +43,9 @@ group:
 
 Configuration variables:
 
+- **view** (*Optional*): If yes then the entry will be shown as a view (tab).
 - **name** (*Optional*): Name of the group.
-- **icon** (*Optional*): An optional icon to show in the Frontend.
-- **view** (*Optional*): If yes then the entry will be shown as a view.
+- **icon** (*Optional*): If you specify an icon it will show in the frontend instead of the name. This only applies to views; regular groups can't have icons.
 - **entities** array or comma delimited string (*Required*): List of entites to group.
 
 <p class='img'>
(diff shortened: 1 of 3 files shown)


source/_components/group.markdown
```diff
@@ -52,14 +53,24 @@ Configuration variables:
...
 
-If all entities are switches or lights they can be controlled as one with a switch at the top of the card. Grouped states should share the same type of states (ON/OFF or HOME/NOT_HOME).
+If all entities in a group are switches or lights then Home Assistant adds a switch at the top of the card that turns them all on/off at once.
+
+You can create views (tabs) that contain other groups.  
+Notice in the example below that in order to refer to the group "Living Room", you use `group.living_room` (lowercase and spaces replaced with underscores).
 
 ```yaml
-# Example configuration.yaml entry
-group:
```
balloob: These configuration examples are here to show that there are three different ways to specify a group config.

source/getting-started/devices.markdown
```diff
@@ -36,21 +36,22 @@ switch:
 
 #### {% linkable_title Style 2 %}
 
-If you prefer to place your entries somewhere in the `configuration.yaml` file then you just choose this style. You need to append numbers (see example below) or strings (like `media_player livingroom:` or `media_player kitchen:`) to differentiate the entries. The appended number or string must be unique.
+List each device separately.
```
balloob: We could add this line to the title: `Style 2: List each device separately`

lindsaymarkward:
New commit added that addresses these issues. @balloob I didn't realised I'd changed it so it didn't show the comma separated list. That's back now. I only see two ways though. What's the third?
I'm not sure if I need to squash this new commit into the last one. Do I?

source/_components/group.markdown
```diff
@@ -42,24 +43,32 @@ group:
...
+- **icon** (*Optional*): If the group is a view, this icon will show at the top in the frontend instead of the name. If it's not a view, then the icon shows when this group is used in another group.
+- **entities** (*Required*): array or comma delimited string, list of entities to group.
 
 <p class='img'>
 <img src='/images/blog/2016-01-release-12/views.png'>
 Example of groups shown as views in the frontend.
 </p>
 
-If all entities are switches or lights they can be controlled as one with a switch at the top of the card. Grouped states should share the same type of states (ON/OFF or HOME/NOT_HOME).
+If all entities in a group are switches or lights then Home Assistant adds a switch at the top of the card that turns them all on/off at once.
```
Landrash: I have groups with a mix of devices and Home Assistant still adds a switch on top of the card. 
Groups with devices that have an **Assumed State** will have this inherited from these devices and will show the corresponding switch for the group.

PR #316, opened 2016-03-21
Update zwave.markdown

devanl:
Adjust formatting on optional package creation


fabaff:
Did you test that locally? I think that this will not work because it's inside a `note` block.

devanl:
@fabaff Your comment on the markdown inside of a note is spot on but, the original was too confusing.  Maybe this isn't the best way, any ideas?

fabaff:
I think that the fastest way for now would be to just remove `bash`. ``` is rendered into <code> which is working in `note`.

balloob:
@fabaff that brings us back to the original state. So I'm closing this PR.

PR #318, opened 2016-03-21
remove the slash from the topic's preface

chilicheech:
The topic's preface doesn't actually start with a slash.


jer:
:+1: The latest bridge is defaulting to not having a slash.

stjohnjohnson:
:+1: 
yesss

PR #319, opened 2016-03-23
Created a documentation page for the pulseaudio_loopback switch.

Cinntax:
The instructions for submitting these were a little less clear to me than in the other repository, so forgive me if I've made mistakes here- I was able to preview them, so my markdown should be in good shape.
BTW- great project- very impressive stuff!


balloob:
Thanks!

PR #325, opened 2016-03-25
Added hidden and local picture examples

Landrash:
- Moved hidden value to a new entry since example of setting a picture and hiding it seems counter intuitive.
 - Added example of using local picture as icon.


fabaff:
Thanks

PR #328, opened 2016-03-25
Update zwave.markdown

bassclarinetl2:
Removing $ prompt to allow for better cut-paste.


balloob:
We're doing this consistently throughout the rest of the website. It gives a good indication of what is a command versus what is expected to be the output from running that command.

I am down to use an approach that uses CSS `:before` selectors to add the `$` (like done on the front page). The proposed solution in this PR is however not a good idea.

bassclarinetl2:
fair enough

PR #336, opened 2016-03-26
Added Yamaha Receiver Markdown Page

aoakeson:


balloob:
Should have gone to the next branch. Since I'm playing with that right now I'll take care of it and get it added, thanks!

PR #342, opened 2016-03-28
Lots of typo corrections, added some useful examples. 

lindsaymarkward:
In automation / numeric state, I removed the config lines that used value_template for battery when the condition was for a temp sensor. I would have replaced it with a more valid template, but I haven't used that before, so I'm not sure what it should be. I did add a comment about it though. Add a relevant template back in to the example if you like.


robbiet480:
Thanks for all the fixes!!

PR #343, opened 2016-03-28
Added known issue to MJPEG cameras, moved customize polling interval to its own category, minor fixes

lindsaymarkward:
Added known issue to MJPEG cameras, moved customize polling interval to its own category, minor fixes

### DIFF
--- a/source/_cookbook/customize_polling_interval.markdown
+++ b/source/_cookbook/customize_polling_interval.markdown
@@ -7,8 +7,8 @@ sidebar: true
 comments: false
 sharing: true
 footer: true
-ha_category: Example configuration.yaml
----
+ha_category: Customize Defaults
+-------------------------------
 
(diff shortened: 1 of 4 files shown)


source/_cookbook/customize_polling_interval.markdown
```diff
@@ -7,8 +7,8 @@ sidebar: true
 comments: false
 sharing: true
 footer: true
-ha_category: Example configuration.yaml
----
+ha_category: Customize Defaults
+-------------------------------
```
balloob: Why did you add these extra dashes? Are you sure that jekyll picks that up correctly?

lindsaymarkward: Those were automatically added by PHPStorm and the MultiMarkdown plugin... I didn't notice it wasn't supposed to be like that. I'll fix it.

balloob:
Looks good :dolphin: You can merge it when comment addressed.

PR #345, opened 2016-03-28
Fix OwnTracks URL

glynhudson:


fabaff:
Just a `/` in front would be enough to fix the link.

glynhudson:
No problems, happy to help. Awesome project :+1:

PR #347, opened 2016-03-28
Update Sun documentation (`elevation` parameter)

MinchinWeb:
`elevation` parameter provides physical elevation. Solar elevation changes throughout the day and is provided by the `Sun` entity, not a parameter to it.

Clarify parameters: that times given are in UTC (not local time), and the `elevation` provided here is different from the `elevation` parameter above.

Also update provided elevation to that at provided location.


balloob:
Thanks!

PR #348, opened 2016-03-28
fix value_template string  Markdown issue

glynhudson:
value_template string was getting lost in markdown resulting in web page just showing `value_template: ''`. This should fix the issue...blatent copy from  https://github.com/balloob/home-assistant.io/blob/master/source/_cookbook/track_battery_level.markdown :-)


balloob:
Thanks

PR #351, opened 2016-03-29
Update zwave documentation

srcLurker:
Add more information about the zwave services and give an example of using soft_reset and heal_network services.


balloob:
Thanks!

PR #357, opened 2016-03-30
Alexa documentation improvements

robbiet480:
Improve the Alexa documentation to explain differences between the built-in component and @auchter's Haaska.

Also, I made this page featured because Alexa is goddamn awesome and would be a big draw for new users IMHO.

### DIFF
--- a/source/_components/alexa.markdown
+++ b/source/_components/alexa.markdown
@@ -9,7 +9,7 @@ sharing: true
 footer: true
 logo: amazon-echo.png
 ha_category: Voice
-featured: true
+featured: false
 ---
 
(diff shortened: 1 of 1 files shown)


source/_components/alexa.markdown
```diff
@@ -9,17 +9,50 @@ sharing: true
 footer: true
 logo: amazon-echo.png
 ha_category: Voice
-featured: false
+featured: true
```
balloob: Look at this page: https://home-assistant.io/components/

One can only be added if we either remove 1 or add 4 more. You can remove forecast.io being featured.

balloob:
Looks good, can merge after you addressed the feature comment.

PR #359, opened 2016-03-30
Update config of rfxtrx switch

turbokongen:
Describes the config option 
```yaml
 fire_event: True
```


Danielhiversen:
Nice, you should add it to the light component too

turbokongen:
Forgot about that. Done. :+1:

robbiet480:
@turbokongen Thanks!

PR #362, opened 2016-03-30
Add documentation for zwave's test_network

srcLurker:
Add documentation for test_network which is a call through to openzwave's TestNetwork()


balloob:
The `next` branch contains the website for the 0.17 release. Could you point it at that branch?

PR #370, opened 2016-04-03
Update autostart.markdown

emartec-ca:
Per the link systemd is used by default in Jessie instead of upstart (not specifically for the Pi but still): https://wiki.debian.org/FAQsFromDebianUser#systemd

Running ps -p 1 -o comm= on my pi gives systemd as well.


fabaff:
Thanks

PR #373, opened 2016-04-06
Add blog post about a static website

fabaff:
A small article about using Home Assistant to serve a web site. Of course with a dark theme :smile:


robbiet480:
Hey @fabaff this post looks good. I would suggest that instead of instructing users to setup HBMQTT, they can instead use the `/api/stream` endpoint which uses [Server-Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) to push data to a browser/client. It is unidirectional unlike WebSockets and is easier to implement on the front end in my experience.

balloob:
It might be easier if it wasn't for the case that this already works 👍

balloob:
I'm down with either. The blog post looks good. @fabaff you can release when you feel like. Do note that our website creates urls based on the UTC date. So it's now April 7 in UTC time…

robbiet480:
Sorry, didn't notice the https://github.com/fabaff/home-assistant-display link in the post. Carry on :)

Landrash:
This looks really cool!

fabaff:
Thanks guys.

fabaff:
Going to us the "Confirm squash and merge" option for the first time :baby_chick:

PR #385, opened 2016-04-10
Add note about python3-dev requirement

synth3tk:


robbiet480:
@synth3tk Thanks for this!!! :cookie:

PR #390, opened 2016-04-13
Remove warning from media_player.yamaha page

Kernald:
The front-end loading issue (home-assistant/home-assistant#1686) has been solved, hence removing the warning.


balloob:
Thanks!

PR #391, opened 2016-04-14
Add example sketches

MartinHjelmare:
* Add binary sensor sketch.
* Add light sketch.
* Add sensor sketch.
* Add switch sketch.
* Change example yaml to default value for optimistic setting.


balloob:
Looks good! :dolphin:

PR #393, opened 2016-04-14
Update Install Instructions for DSM 6.0

MinchinWeb:
Changes from DSM 5.2:

- Synology Python package is now Python 3.5 (was 3.4)
- it appears the install location for the Python binaries has changed
- does not assume `pip` is installed with the Python package (it wasn't on my machine)
- `-r` wasn't an option for `chown` or `chmod`, but `-R` is (applies the commands recursively. I'm not sure whether this was originally a typo, or I have a different version of these commands
- `chmod` flags set to `660` for the `/volume1/homeassistant` directory made it so that the admin users can't enter it. `664` seems to work well.

The instructions for the [autostart](https://home-assistant.io/getting-started/autostart/) should be updated too, but I haven't got that part working yet.

### DIFF
--- a/source/getting-started/index.markdown
+++ b/source/getting-started/index.markdown
@@ -82,7 +82,7 @@ There is also a [video tutorial](https://www.youtube.com/watch?v=GjzOXkPb7XE) cr
 
 <div class='install-instructions synology' markdown='1'>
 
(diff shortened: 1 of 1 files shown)


fabaff:
Thanks

PR #397, opened 2016-04-16
Update input_select.markdown

hartmms:
Added a sentance about how changes of value generate a state event that can be used as trigger for automation.


balloob:
Thanks! 🐬

PR #403, opened 2016-04-19
Update switch.command_line.markdown

joemcmonagle:
Looks like this is missing the trailing ``` to close the final yaml configuration, which causes it to be improperly displayed on the page


balloob:
Thanks!

PR #404, opened 2016-04-19
added webostv

TheRealLink:


balloob:
Bueno 🐬

source/_components/media_player.webostv.markdown
```diff
@@ -0,0 +1,35 @@
...
+description: "Instructions how to integrate a LG WebOS TV within Home Assistant."
+date: 2016-04-18 23:24
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: webostv.png
+ha_category: Media Player
+ha_iot_class: "Local Poll"
+--------------------------
```
balloob: Not sure where you get your example of, but if you don't have exactly three dashes the page won't render.

PR #408, opened 2016-04-20
Suggest the flake8 git commit hook

JshWright:
Given the number of PRs that require style cleanups, it seems like it might be helpful to suggest users add the flake8 commit hook, to catch issues before they get committed to the repo.


balloob:
Love it.

PR #414, opened 2016-04-22
Make sure the 'hass' user is in the right group

JshWright:
In order to access the USB Z-Wave controller, the newly created user needs to be in the 'dialout' group (on Debian)


Landrash:
This would make assumptions that the user is using Debian so the instructions should probably reflect this.

JshWright:
The introduction includes the following:

    We’ll be using Debian in this example (as many HA users are running Raspbian on a Raspberry Pi)

PR #420, opened 2016-04-23
Update garage_door.mqtt.markdown

PombeirP:
The `.x` doesn't bring any value (no idea where that comes from). If I publish "STATE_OPEN" on my MQTT instance, then we want HA to compare with the raw value.

### DIFF
--- a/source/_components/garage_door.mqtt.markdown
+++ b/source/_components/garage_door.mqtt.markdown
@@ -36,7 +36,7 @@ garage_door:
   state_closed: "STATE_CLOSED"
   service_open: "SERVICE_OPEN"
   service_close: "SERVICE_CLOSE"
-  value_template: '{% raw %}{{ value.x }}{% endraw %}'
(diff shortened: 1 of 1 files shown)


fabaff:
The given configuration is an example which shown all available configuration options. There is no need to use `value_template:` if the payload matches the configuration thus marked as optional.  `value_template:` is useful if you can't change the MQTT settings of your device.

Well, `{{ value_json.x }}` would maybe be the better sample for the payload but `{{ value.x }}` it's only a placeholder and if one follows the link the templating docs it should become clear.

PombeirP:
> There is no need to use value_template: if the payload matches the configuration thus marked as optional.
That's strange, because I did try removing that and it still didn't work. It only started working when I changed it to `value_template: '{{ value }}'`.

fabaff:
#597 will update the docs for the new cover component which combines garage door and roller shutter.

PR #433, opened 2016-04-30
Update device_tracker.netgear.markdown with optional port number setting

agjendem:


agjendem:
R7500 did not expose port 5000, scanned the router with nmap and tried the various ports. Turns out that port 80 works fine on this model with latest firmware.

fabaff:
Can you please the conflict? Thanks.

PR #435, opened 2016-04-30
Add version of HA supported

captainnapalm:
Might be useful to know which version of HA components are supported on for those who don't upgrade regularly, or have to downgrade due to broken components.


fabaff:
Good point.

fabaff:
The detail about the release will show up in the sidebar. This makes this PR obsolete because I decided to implement it a bit different. It's just an entry in the file header: `ha_release: 0.17`.

Anyway, thanks for bringing this up.

PR #436, opened 2016-04-30
Display HA release

fabaff:
I agree with @captainnapalm to display the Home Assistant release in which the platform/component was introduced. This could create a better user experience for people who don't update regularly.

Related: #435


balloob:
:+1: like it

Landrash:
A bit late to the game but wouldn't "Minimum version" or similar be a more logical addition?

Assuming someone would use has on an appliance level it may be feasible that it wont be updated for a long time (lets hope note) and potential users would then be limited to the version the appliance is using.

PR #441, opened 2016-05-04
Update sensor.tellstick.markdown

Danielhiversen:


Landrash:
Isn't there a risk of confusion when calling the Tellstick devices Tellstick Hub when there's a Telldus Live?

It might just be semantics but may cause confusion for new users.

Danielhiversen:
@Landrash I agree. Do you have a suggestion for a better name than Tellstick Hub?

Landrash:
@Danielhiversen No really good answer.
Telldus mentions the networked devices as tellstick also but they are in general mentioned as gateways.
Maybe **Tellstick USB** and in the future specify that the Telldus component is for use with the **Tellstick Gatemway** devices.

PR #444, opened 2016-05-04
Update binary_sensor.template.markdown

Danielhiversen:


fabaff:
I think that it would be easier and less work in the future if we just link to the sensor class overview as the MQTT Binary sensor is doing it.

Danielhiversen:
@fabaff 
I agree

PR #449, opened 2016-05-07
Updating the example docker config for the new docker image location

jer:
In the SmartThings MQTT Bridge post. Just updating the Home Assistant image path to stay current.


balloob:
Thanks!

PR #452, opened 2016-05-07
Update device_tracker.icloud.markdown

aaronpace:
Adding a note regarding that iCloud in hass is not compatible with 2FA at this time. Took me a while to figure this out and then found the bug report on github (#1789)


balloob:
Thanks

PR #453, opened 2016-05-08
Improve the example light platform code

basilfx:
The code was incomplete and invalid (e.g. missing properties and def keyword). This PR extends the example code with turn_on/turn_off, brightness control and a display name.

This is the piece of code I was looking for when I experimented with adding a platform yesterday.


fabaff:
Thanks

PR #457, opened 2016-05-10
Added subreddit to help page

Landrash:
Added subreddit advertised in Gitter chatroom.


balloob:
The subreddit is not an official maintained part of the community. So I'm not sure if we should promote it as such

balloob:
Never mind, gave it some more thought and it is fine 👍

PR #458, opened 2016-05-10
Update zwave.markdown

fignuts:
Make it more clear that devices are not polled by default.


Landrash:
Wouldn't it be better to specify that the default is 0, no pollin?

fignuts:
In zwave.py there is no default value. Right now undefined and 0 behave the same, but I didn't want to say 0 was the default when it really is not.

PR #462, opened 2016-05-11
ecobee notify component

nkgilley:


balloob:
Please create a new PR that points at the `next` release (which is for 0.20)

PR #465, opened 2016-05-12
Update development_environment.markdown

kellerza:
- Extended the rebase section
- Windows setup
- Removed Squash instructions, since I understand this is easy to do from Github Interface now?
  Squashing also confuses `farcy` a bit at the moment.
- Running `hass` does not require `tox`, so moved the statement
- Added some links

### DIFF
--- a/source/developers/development_environment.markdown
+++ b/source/developers/development_environment.markdown
@@ -22,7 +22,7 @@ Home Assistant is open-source and MIT licensed. The source can be found here:
 You will need to set up a development environment if you want to start developing a new feature or component for Home Assistant. Please follow these steps to get setup.
 Visit the [the Home Assistant repository](https://github.com/home-assistant/home-assistant) first and click fork in the top right.
 
-We suggest that you setup a [virtual environment](https://docs.python.org/3.4/library/venv.html) aka `venv` before running the setup script.
+We suggest that you setup a [virtual environment](https://home-assistant.io/getting-started/installation-virtualenv/) aka [`venv`](https://docs.python.org/3.4/library/venv.html) before running the setup script.
 
 ```bash
 $ git clone https://github.com/YOUR_GIT_USERNAME/home-assistant.git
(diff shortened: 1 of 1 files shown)


source/developers/development_environment.markdown
```diff
@@ -22,14 +22,17 @@ Home Assistant is open-source and MIT licensed. The source can be found here:
 You will need to set up a development environment if you want to start developing a new feature or component for Home Assistant. Please follow these steps to get setup.
 Visit the [the Home Assistant repository](https://github.com/home-assistant/home-assistant) first and click fork in the top right.
 
-We suggest that you setup a [virtual environment](https://docs.python.org/3.4/library/venv.html) aka `venv` before running the setup script.
+We suggest that you setup a [virtual environment](https://home-assistant.io/getting-started/installation-virtualenv/) aka [`venv`](https://docs.python.org/3.4/library/venv.html) before running the setup script.
```
fabaff: I wouldn't link to the [virtual environment](https://home-assistant.io/getting-started/installation-virtualenv/) docs as with the required version of Python for Home Assistant the `venv` module provides a simple way to setup a virtual environment without additional packages.

source/developers/development_environment.markdown
```diff
@@ -22,14 +22,17 @@ Home Assistant is open-source and MIT licensed. The source can be found here:
...
 
 ```bash
 $ git clone https://github.com/YOUR_GIT_USERNAME/home-assistant.git
 $ cd home-assistant
 $ git remote add upstream https://github.com/home-assistant/home-assistant.git
 $ script/setup
 ```
+On Windows you can use `python setup.py develop` instead of the setup script
```
fabaff: Period is missing.

source/developers/development_environment.markdown
kellerza: Ok, not sure I understand 100%, I used those instructions, but will take this out

source/developers/development_environment.markdown
kellerza: Thanks Mr farcy :wink:

source/developers/development_environment.markdown
balloob: There are two versions: virtualenv is the original cool and venv is the official one that now comes bundled with Python. So it's not an "aka"

source/developers/development_environment.markdown
balloob: Should it be `python3` ?

kellerza: Not on windows (it was a clean install)

PR #466, opened 2016-05-12
Generic platform discovery intro

kellerza:
Introduce generic platform discovery with an example.

Unsure on how to add this to the menu.

This should probably not replace the other netdisco/discovery page, so I create a new one.


kellerza:
Ok, something went wrong here..... there is 54 files in the PR? :sweat: 

Should only be one new

balloob:
You branched off master but are merging into next. Master is currently ahead of next. It's fine since we need to update next with master anyway.

Your page is fine but there is currently nothing linking to it.

PR #467, opened 2016-05-12
Actually document the Google Maps Travel Time sensor

robbiet480:
See also home-assistant/home-assistant#2047


robbiet480:
Wrong branch :(

PR #468, opened 2016-05-13
Actually document the Google Maps Travel Time sensor

robbiet480:
See also home-assistant/home-assistant#2047


Danielhiversen:
Nice work.
But you should also document the different options: mode, language, avoid, units....

Do we support multiple origins and destinations?
I think the state only will show the travel time of the first origin and destination?

An example is also always nice: 

```yaml
sensor:
  platform: google_travel_time
  api_key: API_KEY
  origin: 37.8156380,-122.2628050
  destination: 37.7968520,-122.4055500
  options:
    mode: transit
    transit_mode: rail
    transit_routing_preference: less_walking
```

PR #472, opened 2016-05-13
Update zwave.markdown

seedzero:
Changed command substitution to use $() instead of backticks. This is clearer for newer users, who sometimes confuse backticks with quotes. Also corrected a typo.


balloob:
Thanks

PR #477, opened 2016-05-14
Update sensor.rest.markdown

devdelay:


robbiet480:
@devdelay Can you change it to `http://IP_ADDRESS:8123/api/states/sensor.weather_temperature?api_password=PASSWORD`? Thanks

fabaff:
Isn't the current `resource` perfectly fine for a setup without password?

PR #483, opened 2016-05-16
Changing input_boolean from IFTTT

arsaboo:
Added instruction to changing `input_boolean` from IFTTT


balloob:
I think that one example with a screenshot for calling a service in HA is enough.

PR #488, opened 2016-05-19
iBeacon blog part 2

pavoni:
Add the second part of the iBeacon blog (slightly tidied since the version you read @balloob ).

Proof reading welcome!


Landrash:
You mention 
> ...and a iBeacon on my drive so that a location update is triggered when I arrive.
Is that supposed to be driveway or is it placed so you pass it on your drive home?

PR #502, opened 2016-05-22
Added documentation for new LIRC component

partofthething:
These are the docs for a new LIRC component.


source/_components/lirc.markdown
```diff
@@ -0,0 +1,96 @@
...
+[LIRC](http://www.lirc.org/) integration for Home Assistant allows you to receive signals from an infrared remote control and control actions based on the buttons you press. You can use them to set scenes or trigger any other [automation](https://home-assistant.io/components/automation/). 
+
+Sending IR commands is not supported in this component (yet), but can be accomplished using the [shell_command component](https://home-assistant.io/components/shell_command/) in conjunction with the `irsend` command. 
+
+### {% linkable_title Installation %}
+
+To allow Home Assistant to talk to your IR receiver, you need to first make sure you have the correct dependencies installed:
+
+```bash
+$ sudo apt-get install lirc python-lirc
```
balloob: python-lirc will automatically be installed by Home Assistant.

partofthething:
Oh really? I had to disable its auto-installation because it depends on lirc headers that are not available on the CI server (they come with the `lirc` package) and the build was failing, so I was under the impression that the user had to install it manually kind of like `python-openzwave`.

partofthething:
@balloob considering https://github.com/home-assistant/home-assistant/issues/2266 and the fact that `requirements_all.txt` has:
```
# homeassistant.components.lirc
# python-lirc==1.2.1
```
because the build of `python-lirc` fails unless the user has the `lirc` package installed (which caused CI failures in https://github.com/home-assistant/home-assistant/pull/2139)

What's the best way to ensure people have `1.2.1`? Can I mention it here in the documentation?

balloob:
Home Assistant will install it because it doesn't use this file but instead
uses the REQUIREMENTS constant

PR #507, opened 2016-05-23
Formatting changes

arsaboo:
so that upgrade instructions use bullets


balloob:
Bullets are not done by adding a space but by adding new lines before and after the list. This already has been fixed.

PR #517, opened 2016-05-28
Add Vagrant installation docs

shaftoe:
Full credits to @spuder who's started the whole Vagrant idea and first version of this document


spuder:
Need to clarify that the user first needs to cd to home-assistant/virtualization/vagrant

shaftoe:
@spuder can you elaborate on that? The user is reminded to cd into the right folder in every code snippet, at lines 39, 50, 63, 72 and 81. Do you mean perhaps we should clarify why is needed?

spuder:
The instructions assume that the user is cloning into their home directory. 

Maybe instead of using an absolute path, the instructions should use relative paths? 

```
git clone https://github.com/home-assistant/home-assistant.git
cd ~/home-assistant/virtualization/vagrant
```

```
git clone https://github.com/home-assistant/home-assistant.git
cd ./home-assistant/virtualization/vagrant
```

shaftoe:
@spuder got it. Yeah why not, the clearer the better. I also try to put in some note before the snippets to explain a bit why is needed

PR #557, opened 2016-06-12
Added information about voltage sensor

arsaboo:
Added information for the BloomSky voltage sensor (corresponding PR - home-assistant/home-assistant#2285)


balloob:
Thanks.

PR #560, opened 2016-06-15
Added more detail around getting the API key

rpr69:
Since I had a hard time getting the Ecobee thermostat working based on the instructions, I added more detail to help the next person.


balloob:
👍 Looks good 🐬

PR #564, opened 2016-06-16
Update letsencrypt blog

MartinHjelmare:
* Use certbot with --standalone option instead of letsencrypt-auto.
* Update port instructions. Only port 80 is needed during challenge.

### DIFF
--- a/source/_posts/2015-12-13-setup-encryption-using-lets-encrypt.markdown
+++ b/source/_posts/2015-12-13-setup-encryption-using-lets-encrypt.markdown
@@ -13,7 +13,7 @@ og_image: /images/blog/2015-12-lets-encrypt/letsencrypt-secured-fb.png
 
 Exposing your Home Assistant instance outside of your network always has been tricky. You have to set up port forwarding on your router and most likely add a dynamic DNS service to work around your ISP changing your IP. After this you would be able to use Home Assistant from anywhere but there is one big red flag: no encryption.
 
-This tutorial will take you through the steps to setup a dynamic DNS for your IP and allow trusted encrypted connection to it - for free using [DuckDNS] and [Let's Encrypt]. It will use the Raspberry Pi 2 with raspbian Jessie as example platform.
+This tutorial will take you through the steps to setup a dynamic DNS for your IP and allow trusted encrypted connection to it - for free using [DuckDNS] and [Let's Encrypt].
 
 <p class='img'>
 <img src='/images/blog/2015-12-lets-encrypt/letsencrypt-secured.png' />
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
I didn't update the date. Should we write somewhere that the page was updated?

Landrash:
@MartinHjelmare As someone who follows the blog I would appreciate there being at least a reference in next release post mentioning that the blog has been updated. 
Also writing something that the post has been updated gives confidence for those who read it for instructions that it's not outdated.

MartinHjelmare:
Yes I agree. If I change the date in the front matter, I guess the post will be sorted to the front of the list of blog posts? Or is this dependent on the file name? Would that be a good solution?

balloob:
No, we shouldn't change the url that this post is being served on. Instead just add a note after the introduction when and why it was updated

source/_posts/2015-12-13-setup-encryption-using-lets-encrypt.markdown
```diff
@@ -46,48 +46,45 @@ First step is to acquire and set up our domain name. For this, go to [DuckDNS],
...
-IMPORTANT NOTES:
- - Congratulations! Your certificate and chain have been saved at
-   /etc/letsencrypt/live/hass-example.duckdns.org/fullchain.pem. Your cert
-   will expire on 2016-03-12. To obtain a new version of the
-   certificate in the future, simply run Let's Encrypt again.
- - If like Let's Encrypt, please consider supporting our work by:
-
-   Donating to ISRG / Let's Encrypt:   https://letsencrypt.org/donate
-   Donating to EFF:                    https://eff.org/donate-le
+$ echo 'deb http://ftp.debian.org/debian jessie-backports main' | sudo tee --append /etc/apt/sources.list.d/backports.list
```
balloob: I would love to keep it platform independent like it was before. What is wrong with git cloning the repo? (or can we pip install it?)

MartinHjelmare: I'm following the recommended install procedure. Also personally I would never suggest to someone that they should git clone and install, as the first resort, if there is a maintained apt or similar package for their platform. Installing a package from the OS package manager is usually the simplest and most error less way, and updates will be seamless.

I can add a link to the docs where there are instructions for all platforms.

You're call though.

MartinHjelmare: I can test if pip install is working without additional configuration. That could be an option in that case.

MartinHjelmare: What the heck, I'll download and run the platform independent script instead. It's not so much more effort. I'll link to the docs and let users follow whatever install method they like.

Thanks for input.

balloob:
Looks good! :dolphin:

PR #576, opened 2016-06-21
Example of using apache as a proxy

mmaret:
This help to deal with subdomains, https redirection...
If you already have a server hosting other services, this can be
convenient to use such a configuration

This is the configuration I use on a ArchLinux system. Path may be different on ubuntu.
Moreover, feel free to improve my English :)

This PR was inspired by the discussion from https://github.com/home-assistant/home-assistant/pull/2341


fabaff:
Thanks

PR #577, opened 2016-06-22
Fixed misleading / incorrect documentation

amahlaka:
The component name to call is command_switch not command_line


fabaff:
The documentation is right. The platform is called [`command_line`](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/switch/command_line.py) not `command_switch` starting with release [0.14](https://home-assistant.io/blog/2016/02/27/steam-d-link-smart-plugs-and-neurio-energy-sensors/).

PR #579, opened 2016-06-23
Added configuration for card order to customize docs

keatontaylor:


balloob:
This behavior should not be encouraged. The order attribute is automatically assigned by the order that groups are defined in the configuration.

PR #594, opened 2016-06-28
WinSCP SFTP instructions to avoid permission errors

arsaboo:


Landrash:
This really shouldn't be recommended usage since it's not good practice from a security standpoint.
The better option would be to have the HA user and the pi user share a group and set owner ship of the files and folders that may need to be edited to that group.

balloob:
Agreeing with @Landrash here.

PR #595, opened 2016-06-29
Add template sample

fabaff:
Add a sample for a possible template sensor.


Landrash:
I really like including a template example in the component documentation.
It really should be considered best practice where applicable. 😄

fabaff:
The template sensor sample was included because most details are available as attributes in this case.

Perhaps adding lists of the attributes would show the full potential of the sensors and a link to the template sensor. I think that with adding template examples in a lot of places we are just going to repeat ourselves.

PR #597, opened 2016-06-29
Combine Garage and Rollershutter Components -> Cover (home-assistant#1949)

n8henrie:
Documentation changes relevant to new cover component (see: home-assistant/home-assistant#1949)

PR #600, opened 2016-06-30
Fix default up/down values in rollershutter.mqtt

nvella:
The default `payload_up` and `payload_down` values as found in the Home Assistant MQTT Rollershutter module are `UP` and `DOWN` respectively (not `OPEN` and `CLOSE`). These values can be found on line 28 of homeassistant/components/rollershutter/mqtt.py [here](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/rollershutter/mqtt.py#L28). This pull request updates the documentation to match the component.


fabaff:
Rollershutter and garage door will be merged soon.

#597 is relevant.

balloob:
The configuration is showing how these values can be customized from the default value.

PR #619, opened 2016-07-05
Update persistent_notification.markdown

andrew-curtis:
included a little 'gotcha' that caught me out for a bit. =)


balloob:
Oh! Good point

PR #627, opened 2016-07-08
(resubmitting) New cookbook info for setting up Tor

n8fr8:
This is an example about how you can configure Tor to provide secure remote access to your home assitance instance as an Onion site, through Tor’s Hidden Service feature. With this enabled, you do not need to open your firewall ports or setup HTTPS to enable secure remote access.

(Now for the master branch)
(Looks like I accidentally included another fix from "next" but since it is a spelling change, why not include it!)

### DIFF
--- a/source/_components/persistent_notification.markdown
+++ b/source/_components/persistent_notification.markdown
@@ -1,7 +1,7 @@
 ---
 layout: page
 title: "Persistent notification"
-description: "Instructions on how to integrate persistant notifications into Home Assistant."
+description: "Instructions on how to integrate persistent notifications into Home Assistant."
 date: 2016-06-25 10:00
 sidebar: true
 comments: false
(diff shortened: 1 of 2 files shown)


heytcass:
@n8fr8 I tried setting up based on your page on the HA site, but only get to the loading image, the frontend never loads. I tried enabling the API_PASSWORD in http to see if that would get anything else to load, but no dice. Any advice?

screenshot_20160709-073217-2

n8fr8:
I think you need to enable Javascript in Orfox. See the menu->tools->NoScript option and turn that on. HA is very javascript dependent it seems?

balloob:
HA is 100% JavaScript dependent. We do not offer an HTML only version

Fledi:
Hi I had the same problem on my Android smartphone when opening it with the
standard browser. But it worked fine in Chrome.

heytcass:
Whitelisting did it. Thanks, @n8fr8! Also, sorry about the giant screen shot, thought GH would scale it.

n8fr8:
We are working on the "Security Slider" idea that Tor Browser has for use in Orfox, to make it more clear what kinds of capabilities you have granted or can grant to the current site. Sorry it is so buried at the moment.

PR #633, opened 2016-07-12
add joaoapps join component and platform.

nkgilley:
I wasn't sure what category to place the Join component in so it's a Hub for now.


fabaff:
Cool.

PR #641, opened 2016-07-14
Added dynamic entity tracking documentation

bah2830:


bah2830:
Not sure why it's now showing commit 2b4fb87

balloob:
Because master has changes that are not in next yet. It's fine however as we need to merge those anyway. This is bueno now 🐬

balloob:
Doh, I should not have squashed that 👎  Hope git will be able to figure it out.

PR #646, opened 2016-07-15
home assistant website update for telegram send location

keatontaylor:
update for the config options for telegram location sending.


fabaff:
I will do it manually as we now have conflicts.

PR #647, opened 2016-07-16
Make link to PushBullet Component Absolute

ClashTheBunny:
As of now, the link tries to go to a deep subdirectory of the documentation.  Making it an absolute link will make it work on the built documentation on the site.  Neither way currently works in github, but I doubt that's the goal.


fabaff:
Thanks

PR #653, opened 2016-07-17
Clarification on using place IDs

exxamalte:
Link added on how to find place ID, and clarified that a place ID requires to be prefixed in the configuration.


fabaff:
Thanks

PR #657, opened 2016-07-18
Basic MQTT light example (ESP8266)

mertenats:
Add a basic example to control a led with MQTT and the light component (ESP8266)


fabaff:
Can you please place the example outside of the warning? Thanks

PR #662, opened 2016-07-19
Updated BraviaTV docs to show example bravia.conf

laf:
If you have the details for pairing the TV already then it's possible to specify these manually. Details now in the documentation.


fabaff:
Thanks

PR #666, opened 2016-07-21
Line wrap was causing formatting shenaigans

m00dawg:
There were a ton of trailing spaces at the end of some of the YAML definitions which caused the resulting output to look rather ugly.


fabaff:
Thanks

PR #670, opened 2016-07-22
Update binary_sensor.http.markdown

tguerena:
The sample call was missing -H "Content-Type: application/json" for the POST section, causing it to fail.


robbiet480:
Thank you!!!! 🍪

robbiet480:
🎉

PR #672, opened 2016-07-24
Updated KNX documenation

open-homeautomation:
- Added documentation for KNX switch and KNX thermostat
- Removed some stuff from KNX platform component that was invalid (copy-paste error from Zigbee component).


fabaff:
Thanks

PR #677, opened 2016-07-26
Remove doc for google voice

w1ll1am23:
https://github.com/w1ll1am23/pygooglevoice-sms no longer functions so support for google voice sms is being removed.


balloob:
🐬

PR #678, opened 2016-07-26
Blog post update: switch laundry automation to nodemcu and update yaml

nkgilley:
I rewrote my laundry automation for NodeMCU which seems to be a more popular and cheaper solution than moteino.  Updated this blog post to reflect that as well as update the code examples.


fabaff:
How about to leave the old post in place and create a entirely new one? I think that the old one could still be useful and I fear that your work will not get the attention it deserves if don't update the date/file as well.

At the end it's your call.

nkgilley:
I couldn't decide whether to create a new post or not.  It's ok with me
either way.  I was considering taking a stab at rewriting the arduino code
in MicroPython and that might be worth a new post.

nkgilley:
I should also include a circuit or block diagram for it to be worthy of a new post. on my todo...

PR #681, opened 2016-07-27
Samsung TVs that don't work

CCOSTAN:
Tested but not working Samsung Models.  - Because it's annoying to set up stuff and then find out they don't work.


fabaff:
Thanks

PR #688, opened 2016-07-27
Manual Speedtest Service information

CCOSTAN:
Added new information I just learned about Speedtest from @nkgilley.


fabaff:
Thanks

PR #694, opened 2016-07-28
Quotation marks needed

PerMalmberg:
The quotation marks are needed for digit-only name/ids, as I disovered when adding my devices.
I assume the same goes for devices under the light: domain?

Perhaps this is actually a YAML-issue, but it doesn't hurt to explicitly state this to help the users.

### DIFF
--- a/source/_components/switch.rfxtrx.markdown
+++ b/source/_components/switch.rfxtrx.markdown
@@ -51,9 +51,11 @@ switch:
   signal_repetitions: 2
   devices:
     0b1100ce3213c7f210010f70:
-      name: Movment1
+      name: Movement1
     0b11000a02ef2gf210010f50:
-      name: Movment2
+      name: Movement2
+    "0710010041010160":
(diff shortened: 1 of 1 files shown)


balloob:
This should be fixed instead of adding a workaround to the documentation.

CC @Danielhiversen

ktpx:
I got a bunch of devices defined under rfxtrx , switches and lights.  I'm not using quotes around the id's, and it's working ok.  Seems this could be something else?

PerMalmberg:
Do you have any digit-only ids, Thomas?

Looking at the code, and practical test all points to Python interpreting the id in the example as a member.

ktpx:
Per,
Yeah..disregard my statement..looking closer, they all got at least one character in them ..

balloob:
It's a yaml thing

Danielhiversen:
I can make a fix to handle digit only when I  am back from holiday.

PR #695, opened 2016-07-28
Clarify Onkyo Name parameter

miniconfig:
The "name" parameter said it was "optional, required if host is specified," which is kind of confusing since, as far as I can tell, name is completely ignored if you don't specify the host. Unless I'm missing something, @danieljkemp ?


danieljkemp:
Yeah you're right, the docs were misleading/confusing. If no host is
specified the component will attempt to discover potentially multiple
receivers and there would be no way to map them.

PR #696, opened 2016-07-28
AIO also has zwave already included

PuckStar:
Found this out after running the scripts. Could've saved me some time :)


fabaff:
Thanks

PR #697, opened 2016-07-29
Adding an example for a mathematical addition.

jwl17330536:


source/_topics/templating.markdown
```diff
@@ -66,7 +66,7 @@ Home Assistant adds extensions to allow templates to access all of the current s
 - `is_state('device_tracker.paulus', 'home')` will test if the given entity is specified state.
 - `is_state_attr('device_tracker.paulus', 'battery', 40)` will test if the given entity is specified state.
 - Filter `multiply(x)` will convert the input to a number and multiply it with `x`.
-- Filter `round(x)` will convert the input to a number and round it to `x` decimals.
+- Filter `round(x)` will convert the input to a number and round it to `x` decimals. (Additional arithmetic can be done from here. See examples below.)
```
kellerza: Should this addition not be moved one line up?

jwl17330536: Yes, it should be.  Good catch.

kellerza:
What about mentioning `float()` as well, so we can push std template language even more?

balloob:
Instead of doing `multiply(1)`, the example should use `float`.

PR #698, opened 2016-07-30
Proximity List and Configurable Unit of Measure

Teagan42:
List of proximity component and valid unit of measurement configuration


source/_components/proximity.markdown
```diff
@@ -47,6 +47,7 @@ proximity:
     - device_tracker.eleanorsiphone
     - device_tracker.tsiphone
   tolerance: 50
+  unit_of_measurement: mi
```
balloob: Is this one still relevant?

Teagan42: It allows you to have the proximity in a different unit than the
application wide unit system. It should be marked optional though

PR #705, opened 2016-07-30
Fix copy-paste errors in the new iframe documentation.

corbanmailloux:
Looks like the text for `icon` and `url` were copied from elsewhere. I've added some text stolen from the [weblink docs](https://github.com/home-assistant/home-assistant.io/blob/master/source/_components/weblink.markdown).


fabaff:
Thanks

PR #707, opened 2016-07-31
Update core configuration to accept distance unit

Teagan42:
Application wide unit of measure for distance.


Teagan42:
Closing for now - refactoring to use unit system

PR #708, opened 2016-07-31
Add warning to the MPC-HC media_player component documentation.

abcminiuser:
I thought it prudent to add a warning to the new MPC-HC media player component, telling users to be cautious about exposing the web UI to actors outside of their trusted network (it allows for arbitrary filesystem browsing and media player control without authentication).

I've also added a link to a simple proxy script I wrote last night that can act as a bridge between MPC-HC running with the web UI restricted to `localhost` and a remote client, redacting and restricting the command set to a sane subset. I'm using this to allow HA to observe and have only basic control over my main PC's MPC-HC, while redacting the current playing information.


fabaff:
Thanks

PR #710, opened 2016-07-31
HotFix - Needs to pipe integer to delay

Teagan42:
Using the delay template as specified in the documentation worked when the delay template was being implemented, but input_slider is now floating point - update documentation to pipe the float into an integer.

@balloob This may be considered a hotfix so I labelled it as such. Someone already had issues in the gitter chat following the example I provided in the docs.


fabaff:
Thanks

PR #711, opened 2016-07-31
Document that the plex component does not support HTTPS.

Dutchy-:
My plex was not getting autodiscovered and I got an error after configuring it manually. Turns out that the component does not support HTTPS and I want to document this.


dale3h:
Looks good!

PR #714, opened 2016-08-01
Added note for windows users to getting started with vagrant

ttu:
Added note for windows users that git's core.autocrlf setting must be input, or scripts executed during provision won't work.


fabaff:
Thanks

PR #715, opened 2016-08-01
Locative now work on Android

boced66:
We have worked with the developper of locative and now it's work on
Android


fabaff:
Thanks

PR #722, opened 2016-08-02
Add node_event documentation and example.

mezz64:
Adds a node_event example and updates the scene_activated example to utilize the object_id rather than the entity_id.


balloob:
Awesome 🐬  thanks

PR #724, opened 2016-08-03
Update notify.telegram.markdown

ThinkPadNL:
Added small explanation on how to send a simple message as the page was lacking this explanation.
Swapped method for getting chat_id around, opening a URL is easier than creating a Python script.


fabaff:
Various ways to send a message are covered in the [notify overview](https://home-assistant.io/components/notify/) and the [automation](https://home-assistant.io/getting-started/automation-action/) docs itself.

While I agree that the first option is simple than using a script, should the script be placed before the result.

ThinkPadNL:
But that is a few pages away, why not post a simple explanation to get people started quickly ;) ?
And why place the script before the result? The result of the script will look different than the result of the url?

fabaff:
> But that is a few pages away, why not post a simple explanation to get people started quickly ;) ?

The parent documentation for the notification is only one click away. Today we have 26 notification platforms. It will become a nightmare to maintain the same examples for this amount of docs.

Yeah, I updated the script to spit out the client ID directly. Sorry, my bad. I was still thinking that the output was just JSON like the link provide.

ThinkPadNL:
Alrighty :)

PR #726, opened 2016-08-03
Add documentation with regards to cover art

deonspengler:
This adds support for cover art in mpd. A web server like nginx will need to have access to the mpd music directory and a image of the cover art needs exist in each album.


robbiet480:
Closed along with home-assistant/home-assistant#2721

PR #730, opened 2016-08-03
Added DEV command

CCOSTAN:
Added Dev upgrade command and warnings etc...


balloob:
I don't want this to be in the main instructions as I don't want people that should not use it, use it.

PR #734, opened 2016-08-04
Fix systemd service file's path

smurfix:
Local changes to systemd belong in /etc, not /lib.

Also, running journalctl doesn't require sudo.


fabaff:
> Also, running journalctl doesn't require sudo.

True, but this will limit the output to your user and not give you the details about the system itself as you probably use a different user to run HA.

smurfix:
On 05.08.2016 13:50, Fabian Affolter wrote:
>
>     Also, running journalctl doesn't require sudo.
>
> True, but this will limit the output to your user and not give you the
> details about the system itself as you probably use a different user
> to run HA.
>
True, but somebody who may run arbitrary sudo commands is probably in
group adm which can read the journal anyway.

-- 
-- Matthias Urlichs

PR #743, opened 2016-08-07
Documentation for the pilight component

DavidLP:
https://github.com/home-assistant/home-assistant/pull/2742


source/_components/pilight.markdown
```diff
@@ -0,0 +1,52 @@
+---
+layout: page
+title: "Pilight"
+description: "Instructions how to setup MQTT within Home Assistant."
```
fabaff: This line needs an update.

source/_components/pilight.markdown
```diff
@@ -0,0 +1,52 @@
...
+date: 2015-08-07 18:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: pilight.png
+ha_category: Hub
+featured: true
```
fabaff: Can you set this to `false` please?

source/_components/pilight.markdown
```diff
@@ -0,0 +1,52 @@
...
+ha_release: pre 0.26
```
fabaff: `0.25` is fine. We just used `pre` before 0.7 as we didn't have a proper versioning schema at that time.

fabaff:
I will merge that and fix the remaining issue.

PR #749, opened 2016-08-08
Add details about math, comparision, and logic

fabaff:
Related: #697, https://github.com/home-assistant/home-assistant/pull/2652


balloob:
That's a good fix 👍 

We should probably keep the multiply filter around in the code base as people depend on it, just not promote it any longer.

PR #774, opened 2016-08-12
Whoops, fixed a typo in the new addition

demonspork:
somehow added 2 extra characters to the "endraw" in my previous edit, which breaks some of the other elements further down the page. Second edit ever and I am already breaking things.


fabaff:
This was fixed as part of #771.

PR #775, opened 2016-08-12
Update installation-raspberry-pi-all-in-one.markdown

buut-vrij:
A clean Raspbian Jessie Lite installation on a Raspberry Pi 3 found not working as promised by the manual. Suggest to remove the reference or even discourage Jessie Lite while installer/instructions are not updated for the Lite version.


Landrash:
@buut-vrij Care to add some more information at what step you got stuck? 
My install is Debian Jessie Lite  but used the AIO install as a reference last time i reinstalled and can't remember any problems. I'm very familiar with Linux so it's highly possible I just fixed for myself what ever broke.

PR #795, opened 2016-08-17
Update input_slider.markdown

andrew-curtis:
Was missing practical automation code examples. hopefully someone else will find this useful.

Can someone please check my comments in the code examples about the use of 'data_template: '? Is 'input variables' the correct term here?

### DIFF
--- a/source/_components/input_slider.markdown
+++ b/source/_components/input_slider.markdown
@@ -34,3 +34,71 @@ Configuration variables:
 - **max** (*Optional*): Maximum value for the slider.
 - **step** (*Optional*): Step value for the slider.
 
+# Automation Examples
+
+Here's an example of `input_slider` being used as a trigger in an automation.
+```
+# Example configuration.yaml entry using 'input_slider' as a trigger in an automation
+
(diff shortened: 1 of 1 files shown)


source/_components/input_slider.markdown
```diff
@@ -34,3 +34,71 @@ Configuration variables:
...
+  - alias: Bedroom Light - Adjust Brightness
+    trigger:
+      platform: state
+      entity_id: input_slider.bedroom_brightness
+    action:
+      - service: light.turn_on
+# Note the use of 'data_template:' below rather than the normal 'data:' if you weren't using an input variable
+        data_template:
+          entity_id: light.bedroom
+          brightness: '{{ trigger.to_state.state | int }}'
```
fabaff: Templates needed to be marked with `{% raw %}...{% endraw %}` otherwise they will get rendered and appear blank.

andrew-curtis: ok no worries. so do I need to do that for the entirety of the code examples? or just the tempting sections of the code?

PR #799, opened 2016-08-18
Update automation-action.markdown

andrew-curtis:
tempting was showing up blank as it wasn't wrapped properly. fixed it.


source/getting-started/automation-action.markdown
```diff
@@ -51,6 +51,7 @@ automation 2:
 
 Conditions can also be part of an action:
 
+{% raw %}
 ```yaml
```
fabaff: `{% raw %}` needs to be inside the YAML block. I guess that's my fault because I told you to do it that way last time.

PR #805, opened 2016-08-20
Remove note about root access from dht sensor

roidayan:
This update is related to #2900

Signed-off-by: Roi Dayan <roi.dayan@gmail.com>


fabaff:
Thanks

PR #811, opened 2016-08-21
Add documentation on how to test requirements

technicalpickles:
It's possible to test different requirement versions in development by using `pip install` to the Home Assistant's config directory's dep, and then `hass --skip-pip`.

I discovered this while working on https://github.com/home-assistant/home-assistant/pull/2914 that needed fixes to an requirement to work.

### DIFF
--- a/source/developers/component_deps_and_reqs.markdown
+++ b/source/developers/component_deps_and_reqs.markdown
@@ -11,16 +11,33 @@ footer: true
 
 Home Assistant allows components and platforms to specify their dependencies and requirements using the variables `DEPENDENCIES` and `REQUIREMENTS`. Both are lists that contain strings.
 
+## Dependencies
+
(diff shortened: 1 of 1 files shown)


technicalpickles:
It might be easier to use [setuptool's development mode](http://setuptools.readthedocs.io/en/latest/setuptools.html#development-mode), but I haven't used it before and not well versed in python land to know how to use it in the context of homeassistant

balloob:
`python3 setup.py develop` will install a package in your current Python environment (either global or virtual) that points at your local directory. A similar thing should work for requirements.

Anyway, this is a good start and I'll merge it. I'll update it further when I have time 👍

technicalpickles:
Thanks!

Home Assistant has its own directory of packages, i.e. `deps`, which is what made me think it's tricky. If hass previously installed a requirement there, then you `python setup.py develop`... which version gets used?

technicalpickles:
I was doing more research on this, and read that `pip install -e path/to/checkout` is considered better than `python setup.py develop`.

PR #817, opened 2016-08-21
Adding AVR receivers to supported list

wittycoder:
Have tested on older AVR-2312CI and it is able to control volume/mute, power and basic media support while playing internet audio sources.


fabaff:
Thanks

PR #824, opened 2016-08-23
Update group.markdown

Chariyski:
It took me a while to understand where is "Set State page from the Developer Tools and browse the Current entities:", so that I'm adding a link to it.


balloob:
We can't link to localhost, people might not host Home Assistant there, making it more confusing.

Instead, we should show the icon for the states tool:

```
<img src='/images/screenshots/developer-tool-services-icon.png' class='no-shadow' height='38' />
```

balloob:
Thanks!

PR #832, opened 2016-08-25
Update z-wave.markdown

devdelay:
moar edits


Landrash:
Changing this to "also" could be interpreted as doing both is an option.

devdelay:
@Landrash good point

robbiet480:
@devdelay @Landrash Status of this?

Landrash:
@robbiet480 It's been fixed in #836 by @devdelay. This one can be closed.

PR #835, opened 2016-08-25
Ffmpeg

pvizeli:
please square on merge.


Landrash:
For consistency I would suggest documenatation parts that mention FFmpeg that are not variables or links use the proper capitalisation of FFmpeg. 😄

PR #839, opened 2016-08-25
Update the second configuration example

mertenats:
For the second example, the brightness functionality was mentioned, but it doesn't include the "brightness_state_topic" and the "brightness_command_topic" instructions.


fabaff:
Thanks

PR #848, opened 2016-08-27
Add info on how to install python, pip3 and python3-libyaml on a Debian system

llauren:
I got a bunch of errors when installing Home Assistant on Debian Jessie, running no a PINE64. So i added info on how to install python, pip3 and python3-libyaml on it.


fabaff:
Thanks

PR #854, opened 2016-08-28
Update sensor.imap.markdown

Danielhiversen:
After this PR https://github.com/home-assistant/home-assistant/pull/2903 'user' changed to 'username' for imap.
Should maybe be added as a breaking change to the release note?


robbiet480:
Dupe #852.

PR #855, opened 2016-08-28
Update sensor.mhz19.markdown

ThinkPadNL:
CO2 level outside isn't 250 - 350pmm, but assumed to be 400ppm. It has risen over the years, due to global warming :-(
Most sensors that do automatic calibration assume 400ppm as lowest.

See also https://www.co2.earth/ for actual CO2 level on the earth.


fabaff:
Thanks

PR #857, opened 2016-08-28
fix grammatical error for "climate"

captainnapalm:


fabaff:
Thanks

robbiet480:
See also #853.

PR #868, opened 2016-08-29
Added example for Flux Led to use effect:random

tchellomello:
Added on the documentation an example to define ```effect: random``` for Flex Led light bulbs

### DIFF
--- a/source/_components/light.flux_led.markdown
+++ b/source/_components/light.flux_led.markdown
@@ -54,7 +54,7 @@ light:
     automatic_add: True
 ```
 
-Will add two lights with given name and create an automation rule to randomically set color each 45 seconds:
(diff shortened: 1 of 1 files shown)


source/_components/light.flux_led.markdown
```diff
@@ -54,7 +54,7 @@ light:
     automatic_add: True
 ```
 
-Will add two lights with given name:
+Will add two lights with given name and create an automation rule to randomically set color each 45 seconds:
```
Landrash: Shouldn't it be `randomly ` instead of `randomically`?

tchellomello: Yes, make sense. I'll fix it. Thank you

tchellomello:
Fixed!

fabaff:
Thanks

PR #872, opened 2016-08-31
Remove insteon fan docs

Teagan42:
Remove docs associated with insteon fan as bug in library is preventing the changes.


fabaff:
Or we can just set the date to day in the future. This would not render the page anymore but it would stay around till the bug is fixed and the platform ready.

Teagan42:
That's an amazing idea!

PR #877, opened 2016-08-31
Added required dependency

Tomek985:
Since release 0.27.1 it became mandatory to define apcupsd and associated server and port. Missing configuration was not caught by validation script and it prevented all other sensors from loading. See more here:
https://community.home-assistant.io/t/0-27-1-broke-customize

### DIFF
--- a/source/_components/sensor.apcupsd.markdown
+++ b/source/_components/sensor.apcupsd.markdown
@@ -105,3 +105,5 @@ sensor:
       - loadpct
       - itemp
 ```
+
+**NOTE:** Before using above sensor, you must enable  [apcupsd](https://home-assistant.io/components/apcupsd/) in configuration file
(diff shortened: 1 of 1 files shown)


Landrash:
I would suggest adding the change to the top of the page in a similar manner as the [Tellstick Sensor component](https://home-assistant.io/components/sensor.tellstick/).

source/_components/sensor.apcupsd.markdown
```diff
@@ -105,3 +105,5 @@ sensor:
       - loadpct
       - itemp
 ```
+
+**NOTE:** Before using above sensor, you must enable  [apcupsd](https://home-assistant.io/components/apcupsd/) in configuration file
```
fabaff: Use this snippet would create a box.

```html
<p class='note warning'>
...
</p>
```

PR #881, opened 2016-09-01
Added information to turn off Disco Lights on Aeon Gen5

CCOSTAN:
Although not HA specific, this information wasn't easily found.  (Found it eventually in an AMAZON review!). So since ZWave Sticks are so popular with HA users, I thought I would contribute this information.

### DIFF
--- a/source/getting-started/z-wave-settings.markdown
+++ b/source/getting-started/z-wave-settings.markdown
@@ -13,7 +13,7 @@ You may wish to modify the zwave settings ozw*.xml file in your .homeassistant r
 
 The reasoning for using these tools is that your Zwave controller stores the values and data that are used to control the network. The XML file in the .homeassistant folder acts as a settings/values cache for the zwave network, so modifying it directly won't change the network values. The Open-Zwave Control Panel writes values directly to the network, and will provide you with an updated .xml file to overwrite in your .homeassistant folder. This is the most foolproof way to make modifications to your zwave devices.
 
-Although totally normal for your ZWave stick (Aeon Aeotec Z-Stick Gen5 for example) to cycle through LEDs (Yellow, Blue and Red) while plugged into your Pi, it's also super annoying.  If you want to turn off the cycling color, you can use the following commands from a putty session tot your Pi.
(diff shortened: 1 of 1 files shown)


source/getting-started/z-wave-settings.markdown
```diff
@@ -12,3 +12,13 @@ footer: true
...
 
 The reasoning for using these tools is that your Zwave controller stores the values and data that are used to control the network. The XML file in the .homeassistant folder acts as a settings/values cache for the zwave network, so modifying it directly won't change the network values. The Open-Zwave Control Panel writes values directly to the network, and will provide you with an updated .xml file to overwrite in your .homeassistant folder. This is the most foolproof way to make modifications to your zwave devices.
+
+Although totally normal for your ZWave stick (Aeon Aeotec Z-Stick Gen5 for example) to cycle through LEDs (Yellow, Blue and Red) while plugged into your Pi, it's also super annoying.  If you want to turn off the cycling color, you can use the following commands from a putty session tot your Pi.
```
fabaff: `from a putty session tot your Pi.` would be better to say `terminal session` as Putty is not available on all OS platform. Also, can this be done from any computer not just a Raspberry Pi.

CCOSTAN: I made the change to the putty (to terminal).  These commands are specific to the PI but those HEX values are universal to the ZWave stick.  If a user was using Windows or another platform, they would use another process to send those HEX commands but that's easier to research than the Hex codes.

fabaff: The commands looks like that they should work with small modification on other platforms.

CCOSTAN: Correct.  I found the info here : https://www.amazon.com/review/R2VZHZDYQKCVA8/ref=cm_cr_dp_title?ie=UTF8&ASIN=B00X0AWA6E&channel=detail-glance&nodeID=228013&store=hi

You'll see how it was adapted from Windows to Pi (and also has a Home Assistant Shout out!)

PR #873, opened 2016-08-31
Add user to gpio group

jettw:
This command helped me control my DHT11 & Binary Motion Sensor without running HA as root

PR #883, opened 2016-09-01
added sudo and some error fix language

bionutrient:
sudo su is needed on raspberry pi to create the systemd service file

if the file is initially created using the wrong directory for hass, then the file, at least when I did it, needed to be removed, and systemctl reloaded prior to fixing the service file.

### DIFF
--- a/source/getting-started/autostart-systemd.markdown
+++ b/source/getting-started/autostart-systemd.markdown
@@ -20,7 +20,7 @@ If the preceding command returns the string `systemd`, you are likely using `sys
 If you want Home Assistant to be launched automatically, an extra step is needed to setup `systemd`. You need a service file to control Home Assistant with `systemd`. If you are using a Raspberry Pi with Raspbian then replace the `[your user]` with `pi` otherwise use your user you want to run Home Assistant. `ExecStart` contains the path to `hass` and this may vary. Check with `whereis hass` for the location.
 
 ```bash
-$ su -c 'cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
+$ sudo su -c 'cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
 [Unit]
 Description=Home Assistant
 After=network.target
@@ -62,6 +62,13 @@ $ sudo systemctl status home-assistant@[your user] -l
            └─30426 /usr/bin/python3 /usr/bin/hass
 [...]
 ```
+
+IF, you get an error message, and it fails to start Home Assistant, double check that your ExecStart=/usr/bin/hass points to the correct location...it's likely ExecStart=/usr/local/bin/hass  If you didn't update it correctly, you may have to remove the home-assistant@pi.service file and then start over at the top of this instruction set using the correct directory.
+
+'''bash
+$ sudo rm /etc/systemd/system/home-assistant@pi.service
+$ sudo systemctl --system daemon-reload
+'''
 
 To get Home Assistant's logging output, simple use `journalctl`.
 


source/getting-started/autostart-systemd.markdown
```diff
@@ -20,7 +20,7 @@ If the preceding command returns the string `systemd`, you are likely using `sys
 If you want Home Assistant to be launched automatically, an extra step is needed to setup `systemd`. You need a service file to control Home Assistant with `systemd`. If you are using a Raspberry Pi with Raspbian then replace the `[your user]` with `pi` otherwise use your user you want to run Home Assistant. `ExecStart` contains the path to `hass` and this may vary. Check with `whereis hass` for the location.
 
 ```bash
-$ su -c 'cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
+$ sudo su -c 'cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
```
fabaff: `sudo su` looks always a bit wired. This will only work if the user is listed in the `/etc/sudoers` file which is no a lot of systems not the case. If this is Rapbian specific then we should mention that and keep the guide generic.

source/getting-started/autostart-systemd.markdown
```diff
@@ -63,6 +63,13 @@ $ sudo systemctl status home-assistant@[your user] -l
 [...]
 ```
 
+IF, you get an error message, and it fails to start Home Assistant, double check that your ExecStart=/usr/bin/hass points to the correct location...it's likely ExecStart=/usr/local/bin/hass  If you didn't update it correctly, you may have to remove the home-assistant@pi.service file and then start over at the top of this instruction set using the correct directory.
```
fabaff: Please follow the same style here and embed stuff like `ExecStart=` with \`. It would also be possible to simple move and edit the file. 

`...it's likely ExecStart=/usr/local/bin/hass  If you` looks a bit misplaced.

source/getting-started/autostart-systemd.markdown
```diff
@@ -63,6 +63,13 @@ $ sudo systemctl status home-assistant@[your user] -l
...
+
+'''bash
```
fabaff: You need to use three \`.

source/getting-started/autostart-systemd.markdown
bionutrient: Sudo is Raspbian specific (I think)... but it caused me much consternation... So I would propose we have something to indicate that you may need to run as sudo if using raspbian.

source/getting-started/autostart-systemd.markdown
bionutrient: Thanks. got it

source/getting-started/autostart-systemd.markdown
fabaff: In another PR (#873) we were talking about a Troubleshooting page for Rasbian. I think that this should be added there.

Landrash: @bionutrient `sudo` is defiantly not specific to Raspbian or Linux for that part. It's a command that let's you execute a command with escalated privileges. 

The `su` command  when ran without specifying a user will default to the root user and is often called the superuser command(a bit misleading). 

In this case your escalating privileges to root to run another command as root.
I think(haven't tested) that you could replace the command in question with
```bash
$ sudo cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
```

Landrash:
@bionutrient Any more work on this?

Landrash:
Closing this since it's gone stale. Feel free to reopen it when the comments have been fixed.

PR #884, opened 2016-09-01
Added an example of an Automation using IFTTT

CCOSTAN:
For me, the data entry was confusing so hopefully this example helps clear it up.


fabaff:
Thanks

PR #885, opened 2016-09-01
GoControl can be opened/closed

jwl17330536:
The text is misleading as GoControl integrated with Wink CAN be opened/closed.


robbiet480:
Thanks! 🐬 🍪 💯

PR #887, opened 2016-09-02
Update sensor.sabnzbd.markdown

xehn:
Added clarification about the name of the sensors created to help new users who might expect the sensor name and component name to match.

Also, it would seem that after 0.27, configuration of this component may have changed.  Using "http" or "https" as indicated in the documentation seems to raise an exception.  Using the port number on the host line also causes an error.  Omitting this (at least without SSL) seems to fix the issue.  I don't know enough about how this component work to troubleshoot, so I left that part as-is.  Maybe someone who knows this component better could take a look?


fabaff:
This commit https://github.com/home-assistant/home-assistant.github.io/commit/520b761f0dc291dabed62d5d3476fd7f6ae52db2 contains more fixes.

PR #897, opened 2016-09-04
Pine64 offical Z-wave module setup

shift:
Extended the docs for the Pine64 Z-wave module (which arrived for me on Friday from the Kickstarter), so there maybe an influx of people to try this due to the AMD64 Serial Port OpenHAB JAR not being AMD64 compatible.


Landrash:
Are you certain that is looked to that tty on any os/dist? 
In general the normal instructions are better to avoid confusion. 

I'm working on rPi hardware specific instructions and considering continuing with the pine64 after. If you wouldn't mind giving me a hand so reach out to me on gitter @landrash.

Landrash:
Integrated in #935.

PR #920, opened 2016-09-11
added sudo

sytone:
if you are using sudo for apt you will need it for pip

### DIFF
--- a/source/_components/recorder.markdown
+++ b/source/_components/recorder.markdown
@@ -49,6 +49,5 @@ For MySQL you may have to install a few dependencies:
 
 ```bash
 $ sudo apt-get install libmysqlclient-dev
-$ pip3 install mysqlclient
+$ sudo pip3 install mysqlclient
 ```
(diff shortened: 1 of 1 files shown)


Landrash:
You won't if your installing in a Virtualenv. 
Most users that have will installed with the AIO installer or followed the Raspbeerry pi guide and will have done this.

We could reformat to reflect this

balloob:
Using sudo is bad practice. If people are in a virtualenv using sudo will actually mess up the permissions, hence it should not be part of the docs.

sytone:
OK. I'll update it again to reflect virtual environments to make it clearer

Sent from Outlook for iPhone




On Sun, Sep 11, 2016 at 11:42 PM -0700, "Paulus Schoutsen" <notifications@github.com> wrote:





Using sudo is bad practice. If people are in a virtualenv using sudo will actually mess up the permissions, hence it should not be part of the docs.

PR #926, opened 2016-09-12
Simplepush docs

fabaff:
Initial docs for Simplepush notifications.

### DIFF
--- a/source/_components/notify.simplepush.markdown
+++ b/source/_components/notify.simplepush.markdown
@@ -21,7 +21,7 @@ To add Simplepush to your installation, add the following to your `configuration
 notify:
   - name: NOTIFIER_NAME
     platform: simplepush
-    device_key: ABCDEFGHJKLMNOPQRSTUVXYZ
+    device_key: ABCDE
 ```
 
 Configuration variables:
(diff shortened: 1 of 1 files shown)


source/_components/notify.simplepush.markdown
```diff
@@ -0,0 +1,38 @@
...
+The `simplepush` platform uses [Simplepush](https://simplepush.io/) to delivery notifications from Home Assistant to your Android device.
+
+To add Simplepush to your installation, add the following to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+notify:
+  - name: NOTIFIER_NAME
+    platform: simplepush
+    device_key: ABCDEFGHJKLMNOPQRSTUVXYZ
```
Landrash: Tested and keys I got are much shorter. Could be good to shorten it in example to avoid confusion.

fabaff: It's a 5 chars key in my case. Agreed I will change it.

Landrash: Same here on multiple devices.

Landrash:
Looks good :dolphin:

PR #931, opened 2016-09-13
Remove executable mode from non-executable files

persandstrom:
Noticed that some files were marked as executable when they should not be.


fabaff:
Thanks

PR #943, opened 2016-09-14
Create Component page for mysensors based HVAC

kaustubhphatak:


source/_components/climate.mysensors.markdown
```diff
@@ -0,0 +1,102 @@
...
+description: "Instructions how to integrate MySensors climate into Home Assistant."
+date: 2016-09-14 18:20 +0100
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: mysensors.png
+ha_category: Climate
+featured: false
+ha_release: 0.28+
```
fabaff: No `+` needed.

source/_components/climate.mysensors.markdown
```diff
@@ -0,0 +1,102 @@
...
+/*
+ * Documentation: http://www.mysensors.org
+ * Support Forum: http://forum.mysensors.org
+ *
+ */
+
+#include <MySensor.h>
+...
+...
+...
```
fabaff: Dots?

source/_components/climate.mysensors.markdown
MartinHjelmare: It should be 0.29.

source/_components/climate.mysensors.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+Integrates MySensors HVAC into Home Assistant. See the [main component] for configuration instructions.
+
+The following actuator types are supported:
+
+##### MySensors version 1.5 and higher
+
+S_TYPE      | V_TYPE
+------------|-------------
+S_HVAC      | [V_HVAC_SETPOINT_HEAT, V_HVAC_SETPOINT_COOL, V_HVAC_FLOW_STATE, V_HVAC_SPEED]
```
MartinHjelmare: Maybe we should note that `V_HVAC_FLOW_STATE` is required to add the device in home assistant?

kaustubhphatak: Right... How does this look now?

source/_components/climate.mysensors.markdown
kaustubhphatak: done...

source/_components/climate.mysensors.markdown
MartinHjelmare: I think you can remove the brackets. Previously I've only used brackets to signify options for the same function. That's partly the case here as well, but also not entirely. I've denoted required attributes with a star on the other platform pages. Add a star after `V_HVAC_FLOW_STATE` and an explanation what the star means. Looks good otherwise.

kaustubhphatak: done...

MartinHjelmare:
Looks good!

PR #946, opened 2016-09-16
Include units to description of elevation config

sgauche:
The elevation units have never been specified.  The example configuration file sent with Home Assistant should be updated to include the elevation units also.
https://pythonhosted.org/astral/module.html#astral.Location.elevation


fabaff:
Yes, should be added as there are people who will do this manually.

PR #950, opened 2016-09-16
Added docs for email sensor

sam-io:
Documentation for email sensor


Landrash:
This will result in no new files? Forgot to add the new file after renaming @sam-io ?

PR #956, opened 2016-09-18
Automation: Documentation for hiding automation entities

milaq:
Documentation for https://github.com/home-assistant/home-assistant/pull/3442

### DIFF
--- a/source/getting-started/automation-create-first.markdown
+++ b/source/getting-started/automation-create-first.markdown
@@ -99,6 +99,28 @@ automation:
     entity_id: group.living_room
 ```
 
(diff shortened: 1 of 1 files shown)


Landrash:
One small comment. Should it be mentioned that they are grouped in the Automation group by default?
With that exception, looks good!

milaq:
I could add a remark about that.
Should I also mention that the group itself can only be hidden as a whole when all automation entities are hidden explicitly?

robbiet480:
@Landrash

milaq:
fixex via 4adaa72d8c0627977b669eb92d21c6ab1bb7713d

PR #959, opened 2016-09-19
Added documentation for manually adding a Yamaha AVR

simonszu:
I have added a feature to manually specify a Yamaha Receiver's IP address, and this is the pull request for the according documentation

### DIFF
--- a/source/_components/media_player.yamaha.markdown
+++ b/source/_components/media_player.yamaha.markdown
@@ -35,12 +35,12 @@ media_player:
 Configuration variables:
 
 - **name** (*Optional*): Name of the device
-- **host** (*Optional*): IP address of the device
+- **host** (*Optional*): IP address or hostname of the device
 
 A few notes:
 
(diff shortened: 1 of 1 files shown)


source/_components/media_player.yamaha.markdown
```diff
@@ -29,14 +29,18 @@ To add a Yamaha Network Receiver to your installation, add the following to your
 # Example configuration.yaml entry
 media_player:
   platform: yamaha
+  host: 192.168.1.100
   name: 'Basement Receiver'
 ```
 Configuration variables:
 
 - **name** (*Optional*): Name of the device
+- **host** (*Optional*): IP address of the device
```
Landrash: Can this be also be a hostname or only an IP address? If so the extended text below should probably be updated to reflect this.

simonszu: It can. I'll update it right on the fly in Github's web frontend

Landrash: Looks good! :dolphin:

PR #964, opened 2016-09-19
Add Documentation for Netatmo binary sensor

jabesq:


source/_components/netatmo.markdown
```diff
@@ -24,6 +24,7 @@ netatmo:
   secret_key: YOUR_SECRET_KEY
   username: YOUR_USERNAME
   password: YOUR_PASSWORD
+  discovery: False
```
Landrash: Example code should always be minimized and since this in an optional variable it should not be in the example.
Feel free to have an advanced example also.

source/_components/binary_sensor.netatmo.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+
+The `netatmo` binary sensor platform is consuming the information provided by a [Netatmo Welcome](https://www.netatmo.com) camera. This component allows you to get the latest event seen by the camera.
+
+To enable the Netatmo binary sensor, you first have to set up [netatmo](/components/netatmo/), this will use discovery to add you binary sensor, nothing else to do.
+
+If you want to select a specific sensor, set discovery to False for [netatmo](/components/netatmo/) and add the following lines to your `configuration.yaml`:
+
+```yaml
+# Example configuration.yaml entry
```
Landrash: Minimize this example to it's bare limits. In general assume it's going to be cut and pasted.

source/_components/camera.netatmo.markdown
```diff
@@ -12,10 +12,15 @@ ha_category: Camera
 ha_release: 0.22
 ---
 
+### {% linkable_title Basic Configuration %}
 
 The `netatmo` camera platform is consuming the information provided by a [Netatmo Welcome](https://www.netatmo.com) camera. This component allows you to view the current photo created by the Camera.
 
-To enable the Netatmo camera, you first have to set up [netatmo](/components/netatmo/), and add the following lines to your `configuration.yaml`:
+To enable the Netatmo camera, you have to set up [netatmo](/components/netatmo/), this will use discovery to add your camera, **nothing else to do.**
```
Landrash: I would remove the "** nothing else to do.** ".

source/_components/binary_sensor.netatmo.markdown
jabesq: This component uses user's specific value (his camera's name) so can't be copy/paste, but I reduced it to just one camera

PR #965, opened 2016-09-20
Fix typos and pass payload that matches example

danielolson13:


Landrash:
Good find!
Thank you.

PR #966, opened 2016-09-20
Removed additional spaces in the yaml

arsaboo:
The yaml entry is incorrect. The additional spaces give an error:
`expected <block end>, but found '?' in "/root/.homeassistant/configuration.yaml"`


Landrash:
Most likely a remnant from that example being part of a larger set.
Looks good!

PR #971, opened 2016-09-21
Updating 'thermostat' to 'climate'

heytcass:


fabaff:
This is already fixed (#940).

PR #973, opened 2016-09-22
Link virtualenv to z-wave documentation

technicalpickles:
It looks like this documentation may have originated somewhere else,
since it refers to documentation on home-assistant.io. This is
home-assistant.io, so may as well just link to it.


Landrash:
Good find! Thank you :+1:

PR #977, opened 2016-09-23
GPMDP expects an IP address in "host" (not "address")

ssimms:
Here's a documentation fix -- the GPMDP component uses homeassistant.const's CONF_HOST to figure out which configuration variable ("host") to use for the IP address.

The documentation currently has "address" instead.  The component ignores this field, defaults to localhost, and (if GPMDP isn't running on the same computer as Home Assistant) gives an error saying that the connection was refused.


fabaff:
Thanks :bird:

PR #983, opened 2016-09-24
Update developer tool icon

byte-bender:
I was searching quite a while for the entities, till I understood that I had to look in the section labeled with the updated icon and the stated one.

### DIFF
--- a/source/_components/group.markdown
+++ b/source/_components/group.markdown
@@ -13,7 +13,7 @@ ha_category: Organization
 
 Groups allow the user to combine multiple entities into one. A group can be promoted to a **view** by setting the `view` option to `yes`. This will make the group available as a new tab in the frontend.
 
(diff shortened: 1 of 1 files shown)


source/_components/group.markdown
```diff
@@ -13,7 +13,7 @@ ha_category: Organization
 
 Groups allow the user to combine multiple entities into one. A group can be promoted to a **view** by setting the `view` option to `yes`. This will make the group available as a new tab in the frontend.
 
-Check the **Set State** page from the **Developer Tools** <img src='/images/screenshots/developer-tool-services-icon.png' class='no-shadow' height='38' /> and browse the **Current entities:** listing for all available entities.
+Check the **Set State** page from the **Developer Tools** ![<>](https://cdn.rawgit.com/Templarian/MaterialDesign/master/icons/svg/code-tags.svg) and browse the **Current entities:** listing for all available entities.
```
balloob: You can't just link to external images, this all wrong. You have to link to the other developer tool image icon that's in the same dir.

byte-bender:
Will make a clean pull request with the internal file, thanks for the note.
BTW: The usage of the external file was copied from another part of the documentation, which I will also update.

PR #984, opened 2016-09-24
Add information to allow AiO installs to work.

CCOSTAN:
Added sudo adduser hass gpio

### DIFF
--- a/source/_components/switch.rpi_rf.markdown
+++ b/source/_components/switch.rpi_rf.markdown
@@ -15,6 +15,8 @@ ha_release: 0.19
 
 The `rpi_rf` switch platform allows you to control devices over 433/315MHz LPD/SRD signals with generic low-cost GPIO RF modules on a [Raspberry Pi](https://www.raspberrypi.org/).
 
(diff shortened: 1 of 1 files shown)


source/_components/switch.rpi_rf.markdown
```diff
@@ -15,6 +15,8 @@ ha_release: 0.19
 
 The `rpi_rf` switch platform allows you to control devices over 433/315MHz LPD/SRD signals with generic low-cost GPIO RF modules on a [Raspberry Pi](https://www.raspberrypi.org/).
 
+Please note that if you are using the All In One installer for Home Assistant on the RPI, you will need to grant the default user HASS access to the GPIOs.  Run the following command `sudo adduser hass gpio` while in a terminal session on your Pi.  This is a one time configuration change to allow HA access to the GPIOs.
```
Landrash: user is not name `HASS ` but ``hass`.
OS specific settings was removed from the components and moved to the installation guide.
If you would like you could replicate this in the AiO installation instructions.

PR #997, opened 2016-09-27
Add `hass` user to sudoers

AdithyanI:
I was confused on why the simple switch command was not working. Then I realized I had to `hass` user to `sudo` group before using switch command.
Maybe, some other beginner also has a same question.

### DIFF
--- a/source/_components/switch.command_line.markdown
+++ b/source/_components/switch.command_line.markdown
@@ -41,6 +41,8 @@ Configuration variables:
     - **value_template** (*Optional*): If specified, `command_state` will ignore the result code of the command but the template evaluating to `true` will indicate the switch is on.
     - **friendly_name** (*Optional*): The name used to display the switch in the frontend.
 
+NOTE :In case, you used Raspberry Pi All-In-One Installer method to install, then by default installation makes use of a Python Virtualenv. The virtualenv is run under the username: `hass`. Again, by default this `hass` user does not have `sudo` access. Hence, in some cases if you would like to run a command with `sudo` rights, then it will not work. First, you need to add the `hass` user to sudoers group, to make it work.
+
 ## {% linkable_title Examples %}
 
 In this section you find some real life examples of how to use this switch.


Landrash:
`hass` user is a system user. This user account is not meant to have elevated privileges. 
To issue commands with elevated privileges use the `pi` user.

AdithyanI:
@Landrash 
For example, if I have a switch

``` yaml
switch:
  platform: command_line
  switches:
    arest_pin4:
      command_on: "sudo /home/pi/433Utils/RPi_utils/./codesend 1135923 1 183"
``` 

My log says this,

```log
16-09-27 19:23:24 homeassistant.components.switch.command_line: Command failed: sudo /home/pi/433Utils/RPi_utils/./codesend 1135923 1 183
```
I was able to resolve this problem by adding `hass` to `sudoers` is that wrong?


In my case, how can I elevate myself to `pi` user and then run the command?
I used raspberry-pi-all-in-one-installed to get HA.

Landrash:
@AdithyanI Your use case is a real corner case. It's possible to run as your doing now and by granting `sudo` privileges to your `hass` account.
Another workaround would be to use the [https://en.wikipedia.org/wiki/Setuid] flag on your executable. This on the other hand wont work if the `codesend` command is a shell script since Linux in general is configured to ignore this flag on shell scripts.
The proper way on the other hand would most likely be to figure out what permissions is needed for the `hass` account to run the `codesend` command and granting it only these.

Reasoning for not adding `sudo` privileges to the `hass` account in general is that we always should design with security in mind and do our best to keep permissions to a bare minimum.

AdithyanI:
Hi @Landrash 
Thanks for the detailed reply :+1: 
So, definitely my solution would be the wrong one. 
But, at-least we could include a note describing what you have written above.
Beginners like me at-least would know the cause of the error and then each can decide the solution that best fits his purpose.

Landrash:
I'm even a bit unsure about the note since it's preferable that the docs be simple and straight to the point.
If you can come up with a good one then please open up a pull request and I'll happily have a look at it :)

PR #1000, opened 2016-09-29
Fix/shorten title for component

tinglis1:
The title in the existing documentation was far to long and did not sort it alphabetically.


Landrash:
Looks good. In the future always try new documentation in jekyll locally first to see how it will look on home-assistant.io

PR #1001, opened 2016-09-29
Add docs for Weather component

fabaff:


source/_components/weather.markdown
```diff
@@ -0,0 +1,15 @@
...
+date: 2016-09-28 14:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+---
+
+Weather platforms are gathering meteorological information from web services and displays the conditions and other details about the weather at the given location. 
+
+Home Assistant currently supports a free web services and such which requires a registration. Please check the sidebar for a full list of supported sensor platforms.
```
Landrash: > ..a free web services..
Singular and plural. Unsure how the sentence should be structured since there's currently a platform supported but in time there will be more.
> Home Assistant currently supports free web services and such which requires a registration. Please check the sidebar for a full list of supported sensor platforms.
Could be an option.

PR #1005, opened 2016-09-30
fix multiple misspellings of 'means'

zebpalmer:


zebpalmer:
huge PR, I know....  but the multiple misspellings of the same word were glaring at me.

PR #1020, opened 2016-09-30
Minimizing of the configuration sample (Hub)

fabaff:


source/_components/mysensors.markdown
```diff
@@ -10,6 +10,7 @@ footer: true
 logo: mysensors.png
 ha_category: Hub
 featured: true
+ha_iot_class: "Local Polling"
```
MartinHjelmare: Local Push

fabaff: Thanks

PR #1027, opened 2016-09-30
add to existing current documentation about new client for presence

nemik:
add to existing current documentation about new client for presence detection


fabaff:
This PR looks the same as #1004.

nemik:
Yes, I'm sorry. I meant to put it in "current" because it is only a small
addition to documentation, not a brand new feature for the next release. I
think I misfiled the previous one. My apologies.

fabaff:
I have to fix the merge issues anyway when we deploy the docs for 0.30 but I guess that this will not cause a problem.

PR #1060, opened 2016-10-02
Add Volvo on Call docs

fabaff:


source/_components/device_tracker.volvooncall.markdown
```diff
@@ -0,0 +1,33 @@
...
+    username: username
+    password: password
+```
+
+Configuration variables:
+
+- **username** (*Required*): The username associated with your Volvo On Call account.
+- **password** (*Required*): The password for your given Volvo On Call account.
+
+See the [device tracker component page](/components/device_tracker/) for instructions how to configure the people to be tracked.
```
Landrash: Is it really people that are tracked in this case or cars?

fabaff: Just copied the content. I guess that it should be cars.

PR #1066, opened 2016-10-02
Reference to outdated "echo" attribute

ih8gates:
Reworded the "expose_by_default" information. It used the "echo" attribute, which if I'm not mistaken, is deprecated.


robbiet480:
Thanks! 🐬 🍪 💯

Issue #1071, opened 2016-10-03
Make sure all components and platforms have logos and they are up to date


robbiet480:
Sometimes someone is so excited to submit documentation for their new bit of code that they forget to add a logo. We also have some logos which may be older or incorrect. 

Sadly you can't upload images to GitHub in browser. You will need to [setup a local developer environment](https://home-assistant.io/developers/frontend/). Once you find a component/platform that needs a logo update, you should find the relevant Markdown file in your local source code (they all exist in `./source/_components`). Place the new image into `./source/images/supported_brands` and then update the `logo:` variable at the top of the Markdown file to reference the image filename.

balloob:
Example PR that adds a logo: https://github.com/home-assistant/home-assistant.github.io/pull/1072

requenym:
Thoughts on replacing existing logos with icons/symbols were possible? (when the logo is not a wordmark style logo)

See the sample attached...
![wordmark-symbol](https://cloud.githubusercontent.com/assets/174867/19025491/8236f170-88e7-11e6-8689-d2c7a8024f7e.png)

robbiet480:
@requenym I'd say that it's up to you on a case by case basis. If I really don't agree with it for some reason (I don't see why I would) then i'll bring it up in the PR.

requenym:
A related thought, what about defaulting to the home assistant logo for components that don't have a defined logo on the components listing page?  Or a variation of it....perhaps a grayscale home assistant logo?

Reviewing the Jekyll template it looks fairly straightforward to knock this out.

Landrash:
@requenym The Home Assistant logo is usually used for Home Assistants integrated components such as history, logger and similar. The grayscale option sounds like a good plan! Feel free to go ahead with that.

robbiet480:
@requenym Agree on the grayed out logo!

requenym:
Thanks for the feedback -- this is what I'm thinking:
![home-assistant-grayscale](https://cloud.githubusercontent.com/assets/174867/19090576/17bd03a4-8a4d-11e6-8cee-3b576e804dae.png)

Landrash:
Love it!

robbiet480:
@requenym Can you make this logo the default if none is provided?

requenym:
@robbiet480 yeah, work has had me busy these last few days.  Perhaps tomorrow I'll try to knock this out.

PR #1075, opened 2016-10-03
add logo for coinmarketcap

fuga2136:
relates to #1071


robbiet480:
Thanks! 🐬 🍪 💯

fabaff:
This logo looks like the first hit on Google. @fuga2136 can you please point me to a Coinmarketcap page that is using this a logo? Thanks.

fuga2136:
@fabaff 

While I don't have a link to one of their pages that is using that logo, the reason I felt it was valid is because Line 16 of their webpage source. This image is what is used when you link to their site via Facebook and Twitter (the most common Open Graph supporters)

Their favicon I didn't think was their actual logo otherwise I would have used that.

fabaff:
I guess that they are using the Bitcoin logo because it's public domain and the most recognizable one of all currencies they support.

Issue #1052, opened 2016-10-02
Check grammar and find typos in the documentation


balloob:
Our documentation is made up of contributions from people from all over the world. English is not always their native tongue (mine included) and so the site could use a check up on grammar and sentence structure.
## 2. Click on the `Edit this page on GitHub` link.

![screen shot 2016-10-02 at 11 00 49](https://cloud.githubusercontent.com/assets/1444314/19022547/94164376-888f-11e6-8bd8-8ad1aebfea23.png)
## 3. Click on the `Edit file` button

![screen shot 2016-10-02 at 11 05 52](https://cloud.githubusercontent.com/assets/1444314/19022600/9412dc30-8890-11e6-9ac1-6104559c0108.png)
## 4. Fix the text

Switch between `Edit File` and `Preview Changes` to ensure that the changes look correct. 
## 5. Commit your changes

![screen shot 2016-10-02 at 11 13 14](https://cloud.githubusercontent.com/assets/1444314/19022636/80f58926-8891-11e6-9eb9-26da45ea35b8.png)
## 6. Submit pull request

Make sure to fill in the comment section if the changes are more than just fixing typos/sentence structure.
![screen shot 2016-10-02 at 11 15 33](https://cloud.githubusercontent.com/assets/1444314/19022639/aeace526-8891-11e6-87b3-08ea00615cc5.png)

Make sure that your Pull Request points to **current**. This ensures that your changes goes online after merging.

For more details, check the [Website documentation](https://home-assistant.io/developers/website/).

rusrushal13:
@balloob I am up for it!!!

salman-bhai:
@balloob I'm ready to do help in contributing.

Kietzmann:
Fixed little grammar typo.
[#1086](https://github.com/home-assistant/home-assistant.github.io/pull/1086)

anchepiece:
There are a few areas where rewording could be useful. There may also be a few cases where grammar and spelling differ between British and North America such as favorite and favourite, organize and organise. Minor nitpick anyways.

Ativerc:
I noticed that sentences could use commas at places and sentences could be written differently to make it easier to read. This is a good project... I might use it in my home!

PR #1088, opened 2016-10-03
Update installation-synology.markdown ref #1052

anchepiece:


fabaff:
Can you point your Pull Requests to **current**? Seems not possible to change the base for this one.

justweb1:
@fabaff I got it to change branch but now there is a conflict that @w1ll1am23 will need to fix. That is probably why it didn't work the first time.

w1ll1am23:
@justweb1 not sure what I need to do here? It looks like maybe you merged your local branch with next? Shouldn't you only see the changes for installation-synology.markdown?

robbiet480:
This PR will need to be entirely recreated. Make the changes directly off the current branch and resubmit. Thanks!

PR #1094, opened 2016-10-03
Added include and exclude functionality to history component

wokar:
Updated configuration description to explain include/exclude functionality. This is added with PR
home-assistant/home-assistant#3674


source/_components/history.markdown
```diff
@@ -18,8 +18,33 @@ The `history` component will track everything that is going on within Home Assis
 To enable the history option in your installation, add the following to your `configuration.yaml` file:
 
 ```yaml
-# Example configuration.yaml entry
+# Example configuration.yaml entry without any filters
```
Landrash: This should only be a basic minimal config. Remove the "without any filters" part. This is for consistency in documentation. Explain more in the others examples instead.

wokar: Moved the examples and improved the description a bit.

PR #1095, opened 2016-10-03
Minimizing of the configuration sample (Z-Wave)

fabaff:


fabaff:
According to the code is the config path determined by Python OpenZWave itself if not given.

PR #1096, opened 2016-10-03
Update templating.markdown

fanaticDavid:
now and utcnow became functions in version 0.29 of Home Assistant


Landrash:
In the future try to make changes that are already ready to go to the `current` branch. That way we can publish them directly. Since this is no breaking change and the next release is this weekend I'll just merge it into `next`.

fanaticDavid:
Ah, thanks @Landrash! I'll keep that in mind 🙂

PR #1100, opened 2016-10-04
Add Pushbullet and Home Assistant version to update notification

philhawthorne:
Adds an example configuration for someone who wants to be alerted when a
new Home Assistant is released, and they would like the notification to
include the new Home Assistant version in the message.

### DIFF
--- a/source/_cookbook/notify_if__new_ha_release.markdown
+++ b/source/_cookbook/notify_if__new_ha_release.markdown
@@ -30,3 +30,24 @@ automation:
       data:
         message: 'There is a new Home Assistant release available.'
 ```
+
+You can use [templates](/topics/templating/) to include the release number of Home Assistant if you prefer. The following example sends a notification via [Pushbullet](/components/notify.pushbullet/) with the Home Assistant version in the message.
+
+```yaml
+notify:
+  platform: pushbullet
+  api_key: 'YOUR_KEY_HERE'
+  name: pushbullet
+
+automation:
+  - alias: Update notifications
+  trigger:
+    - platform: state
+      entity_id: updater.updater
+  action:
+    service: notify.pushbullet
+    data: 
+      title: 'New Home Assistant Release'
+      target: 'YOUR_TARGET_HERE' #See Pushbullet component for usage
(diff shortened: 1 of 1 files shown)


source/_cookbook/notify_if__new_ha_release.markdown
```diff
@@ -30,3 +30,24 @@ automation:
...
+  trigger:
+    - platform: state
+      entity_id: updater.updater
+  action:
+    service: notify.pushbullet
+    data: 
+      title: 'New Home Assistant Release'
+      target: 'YOUR_TARGET_HERE' #See Pushbullet component for usage
+      message: "Home Assistant {{ states.updater.updater.state }} is now available."
+```
```
Landrash: Needs an empty line after for formatting to behave properly. With that exception this looks good!

fabaff:
The [updater page](https://home-assistant.io/components/updater/) contains already a very similar sample. What do you guys of extending the updater example instead of an additional cookbook entry?

Landrash:
@fabaff Additional entries in the cookbook is a bad thing in my mind. Since many most likely use it as inspiration it might help as a example since this is a bit more complex.
A good compromise could be to retrofit the example on the updater [page](https://home-assistant.io/components/updater/) to show the release nr.
@philhawthorne How do you feel about it?

philhawthorne:
@Landrash extra line added.

So I have mixed feelings about that. In my search for finding out how to add the HA version to the notification, I had to look at examples from the Notify Platform. Pushbullet Platform, Templating, and this cookbook page. I didn't even know there was updater page with that example.

I knew from the Cookbook I wanted a notification when there was a new HA release. So in my opinion, it makes sense that this example goes into the cookbook. Otherwise, if someone was in my shoes and wants to include the HA version in the notification, they'll need to look outside of that page to get the template example. After all, both examples are achieving the same thing, notification of a new HA release.

Perhaps this example could also be added to the updater page. However that just means more duplicate examples that need to be maintained.

Landrash:
Good point @philhawthorne. Lets add this then =)

PR #1104, opened 2016-10-04
Add config path if following the virtual environment setup

nivekmai:


fabaff:
Thanks

superm1:
I don't think this is correct actually.  I was just doing the install entirely in the python virtual environment as referred in other documentation and mine ended up like this:
`/srv/hass/lib/python3.4/site-packages/libopenzwave-0.3.1-py3.4-linux-armv7l.egg/config/`

nivekmai:
That install directory is if you used pip to install python-openzwave. If you follow the instructions in the getting started guide, you create a python-openzwave directory in your /srv/hass directory, and then install there.

Actually, looking at the docs again, we're both wrong. You actually are supposed to install it in /srv/hass/src, not /srv/hass.

FYI, I'm using https://home-assistant.io/getting-started/installation-virtualenv/#installing-python-openzwave as the "installing with venv" guide.

superm1:
Ah the reason I got confused was that I ran through that guide last week and didn't do the zwave part yet so I instead came to https://home-assistant.io/getting-started/z-wave/ directly.

So I think it would be clearer to someone new (like me) if at https://home-assistant.io/getting-started/installation-virtualenv/#installing-python-openzwave there wasn't the zwave section directly but instead it linked to https://home-assistant.io/getting-started/z-wave/ and there was an entire subsection for virtualenv.  https://home-assistant.io/getting-started/z-wave/ didn't really talk about where to store anything when using virtualenv so I was working at ~ of the hass user.

PR #1106, opened 2016-10-04
Minor fix

DailyMatters:
In this particular case, i think using 'into' makes more sense than 'onto', as 'onto' normally refers to moving to a location on the surface of something or moving aboard with the intention of traveling in it.


fabaff:
Thanks

PR #1110, opened 2016-10-04
Added config location

arsaboo:


source/getting-started/installation-raspberry-pi-image.markdown
```diff
@@ -21,7 +21,7 @@ The easiest way to install Home Assistant on your Raspberry Pi is by using Hassb
 
 These instructions are also available as a [video](https://www.youtube.com/watch?v=iIz6XqDwHEk).
 
-Home Assistant will now be available by navigating with a browser to `http://ip-address-of-pi:8123`. The default username is `pi` and password is `raspberry` (please change this by running `passwd`)
+Home Assistant will now be available by navigating with a browser to `http://ip-address-of-pi:8123`. The default username is `pi` and password is `raspberry` (please change this by running `passwd`). The Home Assistant configuration is located at `/home/homeassistant/.homeassistant/configuration.yaml`.
```
Landrash: This would technically be incorrect.The configuration is the whole `.homeassistant` directory.
I would prefer 
>  The Home Assistant configuration directory can be found at `/home/homeassistant/.homeassistant/`.

On the other hand so is this a copy of the same information as in the getting started guide.

arsaboo: In light of https://github.com/home-assistant/home-assistant/pull/3660, this may not even be required, but there were few questions about the path and hence I thought of including it. We can leave this out, if not required.

Landrash:
I'm not sure this is the best of fixes but enough for now.

PR #1130, opened 2016-10-05
Update example to use metric unit_system instead of the legacy temperature unit

mweinelt:
Update the example in the getting started guide, to reflect the metric system instead of the old temperature based unit.


Landrash:
A bit of a miss. Thank you for finding this!

PR #1131, opened 2016-10-05
update installation info

pinksocks:
Added packages that must be installed to run firetv-server on hassbian/ubuntu 16.04.


Landrash:
Looks good!

PR #1135, opened 2016-10-05
Update z-wave-device-specific.markdown

turbokongen:
Add info and example of entry in dev-service to set configuration parameter for zwave.
Add more info on security devices.


source/getting-started/z-wave-device-specific.markdown
```diff
@@ -11,15 +11,32 @@ footer: true
...
+}
+```
 
 ##### {% linkable_title Locks and other secure devices %}
 
 These devices require a network key to be set for the Z-Wave network before they are paired. This key is set in OpenZwave's `options.xml` which is located in OpenZWave's directory. This should also be the same directory as `config_path:` in your `configuration.yaml`. If it's not, make sure you have the same values in all the files you are using.
 The option is commented out by default in `options.xml` and is a default key. Make your own unique key. The key is in Hexadecimals.
-It is best to pair these devices in OpenZWave Control Panel or other Z-wave tool that can show you logs while pairing. Test the device before you save the configuration.
+It is best to pair these devices in OpenZWave Control Panel or other Z-wave tool that can show you logs while pairing. Home Assistant show logs from zwave too and it is `OZW.log` in the Home Assistant config directory. 
+You should see communication from the node with lines starting with `info: NONCES` in `OZW.log` when the device is paired secure successfully. If you use Open Zwave Control Panel to Pair, test the device before you save the configuration.
```
Landrash: Pair with capital letter.

Landrash: No problem. Will merge this when I'm at a computer and tag it accordingly.

PR #1140, opened 2016-10-06
Passing variables to scripts

fanaticDavid:
Explain how to pass variables along to scripts in more detail, using actual examples instead of JSON code. Not alot of people are aware of this (previously) nearly undocumented feature.


Landrash:
Looks good @fanaticDavid. Good addition and it's very inline with the cleanup of the documentation.

PR #1142, opened 2016-10-06
Added DTE Energy logo

tchellomello:
Added DTE Energy logo 

Refers to issue #1071


fabaff:
BTW, changes for existing docs can go directly to `current`.

tchellomello:
@fabaff  sounds good I'll send the next ones on the current branch. 

Thanks

fabaff:
`next` will be deployed when 0.30 is released. `current` is published right after merging.

Issue #1141, opened 2016-10-06
Cookbook sorting and additions


Landrash:
Our cookbook of examples could use some sorting and additions.
Needed would be a system to sort the cookbook for a better overview and ease of use.
Also request for additional cookbook entries can be made here and we'll compile it into new issues to be resolved.

PR #1149, opened 2016-10-06
Details about the web server

fabaff:
A simple example about how to use a networks scanner to determine how the Home Assistant web server looks.


fabaff:
Related to #1141.

PR #1157, opened 2016-10-06
Add include configuration to logbook

wokar:
Add documentation on include configuration in the logbook component. This is added with PR
home-assistant/home-assistant#3739


Landrash:
Looks good and can be merged when the related pull request is merged.

PR #1162, opened 2016-10-07
Added logo to Alarm Control Panel Manual

tchellomello:
Added logo to Alarm Control Panel Manual

Refers to issue #1071


robbiet480:
Is this the default gray HA logo we discussed in #1071? I've asked to make that the default logo. I will leave this PR open for now until that gets resolved.

tchellomello:
@robbiet480 sounds good to me

Landrash:
@robbiet480 In this case this is a built-in component of HA so it should use the normal blue icon?

PR #1170, opened 2016-10-07
Added Syslog logo

tchellomello:
Added logo to syslog

Refers to issue #1071


robbiet480:
Syslog is not the same as syslog-ng. Maybe just provide a generic terminal logs icon?

tchellomello:
@robbiet480  Yeah, makes sense. I think a terminal icon will look cool for a syslog entry. 

Thanks

PR #1176, opened 2016-10-07
Add support for Monzo

lwis:
Ref home-assistant/home-assistant#3747


source/_components/sensor.monzo.markdown
```diff
@@ -0,0 +1,47 @@
...
+      - local_currency
+      - local_exchange_rate
+```
+
+Configuration variables:
+
+- **access_token** (*Required*): Your access token for http://monzo.com/.
+- **account_id** (_Optional_): An account id, if omitted the first account will be used.
+- **monitored_variables** array (*Required*): Variables to display in the frontend.
+  - **balance**: The account balance.
```
Landrash: Are all these required or optional.
Update them accordingly and try to minimize the configuration example.

Landrash: Are all these required or optional.
Update them accordingly and try to minimize the configuration example.

lwis: @Landrash I tried to follow https://home-assistant.io/components/sensor.darksky/ as an example, which lists all configurable parameters - I've opted not to include `account_id` in the example as it's optional.

Landrash: @lwis Your correct. It looks fine then :)

lwis: @Landrash cheers, bud.

PR #1184, opened 2016-10-08
Updated config example

arsaboo:
The example config did not show `name`


fabaff:
As mentioned in the Release notes/blog post are now all configuration samples we use at ha.io minimized, thus only the required variables are shown.

Issue #1055, opened 2016-10-02
Improve the Hassbian getting started page


balloob:
We have [just introduced](https://home-assistant.io/blog/2016/10/01/we-have-raspberry-image-now/) a new Raspberry Pi image with Home Assistant integrated.

However, our getting started tutorial is still very basic. Some things that would be good to get added are:
- Enable Mosquitto MQTT
- Configure Mosquitto MQTT
- Compile OpenZwave
- ~~How to login via shell (?) or at least mention how to work with a Pi~~
- ~~How to upgrade, install and remove OS packages.~~
- ~~How to restart Home Assistant and the Raspberry PI~~
- ~~How to manually launch Home Assistant~~
- ~~How to check config with validation script.~~
- ~~How to see the logs for config validation~~
- ~~How to update the config~~
- ~~Update the [Updating ](https://home-assistant.io/getting-started/updating/) page to include HASSbian~~

Landrash:
I'll adopt this one and merge together any changes. Edited the issue to include some more things.
Anyone who needs help just send me a message on Gitter or grab my email from the Github profile.
#1073 is a good start.

instantchow:
I was thinking about adding a suggestion for host name, like `HASSelhoff` but I'm keeping that for myself. I'm doing this Hasbian install now on a Pi 2 and I'll keep committing changes if I see anything.

Landrash:
How ever much I would love that name I don't think we dweeb use it. No one
should ever call him anything else than The Hoff  🚤.
Maybe K.I.T.T would be more appropriate. /s

Feel free to submit anything you feel is missing or if there's something
you feel is lacking you can just that and I'll try to write it up asap I
got time.

chuxxsss:
I was really looking forward to Hassbian . After finding the right location of the yaml file, which is now in home/hass/.homeassisant/homeassisant.yaml . I followed Ben's instruction on changing the yaml file. First I change the api_password, which straight away broken home assistants home page.  I don't understand why something so simple would break the system. Anyway I still want to use this great piece of software thank you to everyone who have made a the time to develop this automation system.

Landrash:
@chuxxsss Just changing that shouldn't stop break HA. I have a update to the documentation rolling out very soon.

chuxxsss:
Thanks.

PR #1189, opened 2016-10-09
contributes to #1055 updates related to Hassbian Image install

instantchow:
* Fixed a typo on Rasbian getting started page as well as adding how to change hostname
  * Adding git as a requirement, don't assume you have it, ex. Hassbian Image 1.0.


Landrash:
If you could separate this out it would be a good thing.

The changing of hostname really shouldn't be in that list since it not a must do.
We can add that further down with system management and updating.
I'll add it to my list of rewrites.

The change to the ozw docs should be good to go.

Landrash:
Similar changes where merged into the documentation with #1262.
If you feel something is missing well revisit this one :)

PR #1191, opened 2016-10-09
Document Climate Services

philhawthorne:
Adds documentation for all the services exposed by the new Climate
platform.

Based on the cover platform documentation.

PR #1194, opened 2016-10-09
Document Climate Services

philhawthorne:
Adds documentation for all the services exposed by the new Climate
platform.

Based on the cover platform documentation.

Related: #1191


Landrash:
Thank you :dolphin:

PR #1195, opened 2016-10-09
Display template tags in Pushbullet Home Assistant release example

philhawthorne:
The template example introduced in #1100 needs some tags around the
template tags so that Jekyll doesn't try to interpret the code.

Rebase of #1193


Landrash:
Thank you :dolphin:

PR #1210, opened 2016-10-11
Add example of tag usage inside yaml

T3m3z:
This helps users to see how extra information passed to notification should be placed under data->data. In this case it's data->data->tag.

### DIFF
--- a/source/_components/notify.html5.markdown
+++ b/source/_components/notify.html5.markdown
@@ -111,6 +111,22 @@ By default, every notification sent has a randomly generated UUID (v4) set as it
 }
 ```
 
+Example of adding a tag to your configuration. This won't create new notification if there already exists one with the same tag.
+
+```yaml
+  - alias: Push/update notification of sensor state with tag
+    trigger:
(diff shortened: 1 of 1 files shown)


source/_components/notify.html5.markdown
```diff
@@ -111,6 +111,22 @@ By default, every notification sent has a randomly generated UUID (v4) set as it
...
+
+```yaml
+  - alias: Push/update notification of sensor state with tag
+    trigger:
+      - platform: state
+        entity_id: sensor.sensor
+    action:
+      service: notify.html5
+      data_template:
+        message: "Last known sensor state is {{ states('sensor.sensor') }}."
```
fabaff: You need to embed the template with `{% raw %}{% endraw %}` otherwise it will be rendered.

Seems that this part was missing in the [website documentation](https://home-assistant.io/developers/website/).

T3m3z:
Oh, thanks @fabaff ! I checked examples of raw/endraw from other pages and added them to this PR.

PR #1212, opened 2016-10-11
Description for the new pilight sensor component

DavidLP:
home-assistant/home-assistant#3822

### DIFF
--- a/source/_components/sensor.pilight.markdown
+++ b/source/_components/sensor.pilight.markdown
@@ -14,9 +14,9 @@ ha_iot_class: depends
 ---
 
 
-This `pilight` sensor platform for 433 MHz devices uses a value in the message payload as the sensor value. Unique identifiers (e.g. _uuid_) can be set to distinguish between multile pilight devices. The pilight HA hub has to set up.
+This `pilight` sensor platform for 433 MHz devices uses a value in the message payload as the sensor value. Unique identifiers (e.g. _uuid_) can be set to distinguish between multile pilight devices. To use a pilight sensor the pilight home assistant hub has to be set up.
 
-To use your sensor via pilight, add the following to your `configuration.yaml` file:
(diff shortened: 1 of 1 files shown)


source/_components/sensor.pilight.markdown
```diff
@@ -0,0 +1,67 @@
...
+- **unit_of_measurement** (*Optional*): Defines the units of measurement of the sensor, if any.
+
+## {% linkable_title Example: Weather station %}
+
+This section shows a real life example how to use values of a weather station.
+
+
+```yaml
+# Example configuration.yml entry
+sensor 1:
```
Landrash: This could all be nested under the same `sensor:`

source/_components/sensor.pilight.markdown
```diff
@@ -0,0 +1,67 @@
...
+description: "Instructions how to integrate pilight sensors within Home Assistant."
+date: 2016-10-08 23:21
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: pilight.png
+ha_category: Sensor
+ha_release: 0.31
+ha_iot_class: depends
```
DavidLP: A 433 MHz sensor only reports its state. So is it assumed then?

source/_components/sensor.pilight.markdown
```diff
@@ -0,0 +1,67 @@
...
+---
+
+
+This `pilight` sensor platform for 433 MHz devices uses a value in the message payload as the sensor value. Unique identifiers (e.g. _uuid_) can be set to distinguish between multile pilight devices. The pilight HA hub has to set up.
```
DavidLP: Sure I can use ling forms. What I want to say is, that the pilight hub (a different component) has to be setup to make this pilight sensor work.

source/_components/sensor.pilight.markdown
DavidLP: Yes, can do.

source/_components/sensor.pilight.markdown
DavidLP: You maybe know how to link to other components?

PR #1219, opened 2016-10-12
sensor name vs platform name

xehn:
Added note about difference between platform name and sensor name.


fabaff:
Damn, seems that I wasn't consistently with the naming :-(

source/_components/sensor.darksky.markdown
```diff
@@ -86,4 +86,6 @@ Configuration variables:
 - **units** (*Optional*): Specify the unit system. Default to `si` or `us` based on the temperature preference in Home Assistant. Other options are `auto`, `us`, `si`, `ca`, and `uk2`.
 `auto` will let forecast.io decide the unit system based on location.
 
+Note: While the platform is called "darksky" the sensors will show up in Home Assistant as "dark_sky" (eg: sensor.dark_sky_summary).
```
fabaff: Why not put that in a box? 

```html
<p class='note warning'>
  While the platform is called....
</p>
```

fabaff:
Thanks :bird:

PR #1227, opened 2016-10-13
Update Yamaha markdown

T3m3z:
Functionality for source select has been around since home-assistant/home-assistant@f1d8667d7eddae1994007c72957cf7e520b5117c but documentation hasn't been updated.

This PR tells users that this component supports source select too.


T3m3z:
Woops, I forgot to rebase. Is that why there exists so many commits in this PR? Should I rebase and create new PR?

fabaff:
Yes, it is.

PR #1237, opened 2016-10-14
Add docs for Emby media_player component.

mezz64:
Documentation for new Emby component.

home-assistant/home-assistant#3862


source/_components/media_player.emby.markdown
```diff
@@ -0,0 +1,33 @@
...
+title: "Emby"
+description: "Instructions how to integrate Emby into Home Assistant."
+date: 2016-10-13 22:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: emby.png
+ha_category: Media Player
+ha_release: "0.31"
```
Landrash: Hopefully this will be in 0.32 so this would need to be updated.

PR #1242, opened 2016-10-14
Updates the influxdb documentation with minimal setup instructions

kirichkov:
Adds instructions on creating the default influxdb database, which were
previously located in the blog post.

Clarifies that the default influxdb configuration doesn’t require
authentication.

### DIFF
--- a/source/_components/influxdb.markdown
+++ b/source/_components/influxdb.markdown
@@ -21,28 +21,7 @@ The default InfluxDB configuration doesn't enforce authentication. If you have i
 influxdb:
 ```
 
-You will still need to create a database named `home_assistant` via InfluxDB's web interface or command line. 
-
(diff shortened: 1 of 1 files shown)


fabaff:
> Adds instructions on creating the default influxdb database, which were previously located in the blog post.

Usually full guides should go to the blog because with the current available resources it's nearly impossible to keep up with changes of the used services. If it's a bog then we can it's out-dated which is difficult when we ship those details as part of our documentation. A link to the [InfluxDB setup docs](https://docs.influxdata.com/influxdb/v1.0/introduction/getting_started/#creating-a-database) would help, though.

kirichkov:
OK, I can remove that and just add a link to the InfluxDB page and update the PR.

PR #1246, opened 2016-10-14
Added port configuration for Pioneer AVRs

ehagan:
home-assistant/home-assistant#3878

### DIFF
--- a/source/_components/media_player.pioneer.markdown
+++ b/source/_components/media_player.pioneer.markdown
@@ -23,12 +23,13 @@ To add a Pioneer receiver to your installation, add the following to your `confi
 media_player:
   - platform: pioneer
     host: 192.168.0.10
-    port: 23
 ```
 
 Configuration variables:
 
 - **host** (*Required*): The IP of the Pioneer device,  eg. `192.168.0.10`
 - **name** (*Optional*): The name you would like to give to the receiver.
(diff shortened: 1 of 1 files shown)


source/_components/media_player.pioneer.markdown
```diff
@@ -23,10 +23,12 @@ To add a Pioneer receiver to your installation, add the following to your `confi
 media_player:
   - platform: pioneer
     host: 192.168.0.10
+    port: 23
```
fabaff: `port` is optional. Our sample should only show the minimal needed part and port defaults to 23 according your PR.

source/_components/media_player.pioneer.markdown
```diff
@@ -23,10 +23,12 @@ To add a Pioneer receiver to your installation, add the following to your `confi
...
 ```
 
 Configuration variables:
 
 - **host** (*Required*): The IP of the Pioneer device,  eg. `192.168.0.10`
 - **name** (*Optional*): The name you would like to give to the receiver.
+- **port** (*Optional*): The port on which the Pioneer device listens, e.g. `23`(default) or `8102`
```
fabaff: Would be good to mention here that some devices are using 8102 to communicate instead of the [source code](https://github.com/home-assistant/home-assistant/pull/3878/files#diff-7240066205c596fa0760f415052ff249R22).

Seems like that after 23 is a space missing.

PR #1259, opened 2016-10-18
switch.anel_pwrctrl: correct ha_release type

mweinelt:
The release version is being treated as float and presented as `0.3` instead of `0.30`, converting it to string should prevent that.


Landrash:
This should actually be formatted as is. There's something else happening with the formatting.
I'll look into it.

Landrash:
I was wrong and at the previous releases that ended with a 0 it was solved in the same way :disappointed:

PR #1272, opened 2016-10-20
zwave exposes >10 services, fixing docs.

hcooper:
I count 12 services.


Landrash:
Good find! Funny things that happens when documentation expands so fast. There's a merge conflict in the 0.31 update with the Z-Wave documentation so this one will have to wait until after the weekend.

PR #1277, opened 2016-10-20
Introduce the new updater data collection

kellerza:
**Description:**
New updater component data & `opt_out` feature

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** https://github.com/home-assistant/home-assistant/pull/3781

### DIFF
--- a/source/_components/updater.markdown
+++ b/source/_components/updater.markdown
@@ -34,6 +34,6 @@ automation:
       message: 'Update for Home Assistant is available.'
 ```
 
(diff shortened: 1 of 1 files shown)


source/_components/updater.markdown
```diff
@@ -34,3 +33,7 @@ automation:
     data:
       message: 'Update for Home Assistant is available.'
 ```
+
+The updater component will collect basic information about Home Assistant and it's environment. The information includes the current Home Assistant version, the timezone, Python version and operating system infomation. No identifyable information (i.e. IP, GPS coordinates) will ever be collected. If you are concerned about your privacy, you are welcome to scrutinize the Python [source code](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/updater.py#L91).
+
+If you choose not to share any information when checking for updates, you can add `opt_out: True` to the updater section of your configuration.
```
fabaff: If you check your configuration in three month I will be hard to determine for most users what `opt_out: True` is. I suggest that we re-name that variable to something like `reporting: ` which reflect what the value about.

PR #1289, opened 2016-10-21
Clean-up notify.nma

Kernald:
**Description:**
Fix a small typo (#1052), and remove the optional Name variable from the example (somehow missed in #1034)

### DIFF
--- a/source/_components/notify.nma.markdown
+++ b/source/_components/notify.nma.markdown
@@ -15,15 +15,14 @@ ha_release: pre 0.7
 
 The `nma` platform uses [Notify My Android (NMA)](http://www.notifymyandroid.com/) to delivery notifications from Home Assistant to your Android device.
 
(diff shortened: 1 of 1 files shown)


source/_components/notify.nma.markdown
```diff
@@ -15,15 +15,14 @@ ha_release: pre 0.7
...
 
-Go to the [NMA website](https://www.notifymyandroid.com) and create a new API key. If you are using the trial offer then keep in mind that your limit is five (5) message per day.
+Go to the [NMA website](https://www.notifymyandroid.com) and create a new API key. If you are using the trial offer then keep in mind that your limit is five messages per day.
 
 To add NMA to your installation, add the following to your `configuration.yaml` file:
 
 ```yaml
 notify:
-  - name: NOTIFIER_NAME
```
fabaff: We decided to keep the `- name:` entry for all notification platforms. Sure, it's optional but the name is used for the service which very handy if you use multiple notification platforms.

Landrash:
@Kernald Make a new pull request with the first part :)

Kernald:
Sure, I'll do that :-)

PR #1293, opened 2016-10-22
Update installation-synology.markdown

lcersly:
**Description:**
Changed su to sudo as DSM 6.0.
Changed some of the commands to be more newbie friendly.


Landrash:
Looked over the changes in Synology 6.0.
Not a fan of using `sudo` in such a manner but it seems to be the only way.
Could you add that this applies from 6.0 and up @lcersly ?

Landrash:
Read over the docs and did some further research. This should be enough.
Thank you :cookie:

PR #1297, opened 2016-10-22
Add warning about MS Windows to Asuswrt

fabaff:
**Description:**
This platform doesn't seem to run on Microsoft Windows installations.

Feedback from windows users would be appreciated.

Issue: https://github.com/home-assistant/home-assistant/issues/3983


Landrash:
Merging this. Better to have this out there to prevent confusion and if it changes then we can remove it.

josephdouce:
it works as long as you use telnet, ssh is not supported

PR #1331, opened 2016-10-26
Adding links to help newbies

PollieKrismis:
**Description:**
Adding additional links and homework for people new to contributing open source and/or some of the tools/concepts used in the home assistant project. 

### DIFF
--- a/source/developers/development.markdown
+++ b/source/developers/development.markdown
@@ -11,9 +11,22 @@ footer: true
 
 Home Assistant is built from the ground up to be easily extensible using components. Home Assistant uses [Python 3](https://www.python.org/) for the backend and [Polymer (Web components)](https://www.polymer-project.org/) for the frontend.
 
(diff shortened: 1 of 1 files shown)


fabaff:
At lot of those links are already present in the documentation. If a newbie goes through the "Starting with Development" section she/he will get context around those links.

PollieKrismis:
Hi Fabian, I agree, but if you grew up on these types of projects, it is enough. Coming in completely blind, requires a little more background. The issue is that external guides you had to fall back on start clashing or teach bad/outdated habits you are trying to avoid here.

So my idea was having a separate section in the guides that expands a little more and highlights some common mistakes. Something separate so as not to bore the people for whom the fast track path is enough and that can be expanded on by newbies.

So essentially a way to play without needing any of your time and having to also get to grips with the Home Assistant architecture as well.

I just wasn't sure where to slot it in or link to it.

Regards
Paul

PR #1335, opened 2016-10-27
Updates to Alexa Flash Briefing documentation

jumpkick:
- Clarify configuration steps
- Remove reference to YOUR_API_PASSWORD from flash briefing since its not needed and there's no explanation what this value is for
- Add note that Amazon won't make calls to non-standard http ports for flash briefings
- Add some info on how to enable the briefing once it's been created

### DIFF
--- a/source/_components/alexa.markdown
+++ b/source/_components/alexa.markdown
@@ -337,19 +337,24 @@ Please refer to the [Amazon documentation][flash-briefing-api-docs] for more inf
 ### {% linkable_title Configuring your Flash Briefing skill %}
 
 - Log in to [Amazon developer console][amazon-dev-console]
-- Click the Alexa button at the top of the console
+- Click the Alexa navigation tab at the top of the console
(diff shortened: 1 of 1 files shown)


robbiet480:
99.999% sure that API password is required but not at a computer so could be wrong. 

The rest of this looks great.

robbiet480:
Confirmed that API password is required. 'requires_auth = False' isn't in code

jumpkick:
@robbiet480 in your flash briefing?  My flash briefing feed does not have a YOUR_API_PASSWORD it works for me.

jumpkick:
Okay @robbiet480 clarified that YOUR_API_PASSWORD is your console login password, if you have one, which I do not, which is why it wasn't required in my case... back to the drawing board.

robbiet480:
@jumpkick Went ahead and set URL back to `https://YOUR_HOST/api/alexa/flash_briefings/BRIEFING_ID?api_password=YOUR_API_PASSWORD`

PR #1340, opened 2016-10-27
Adding documentation for influxdb sensor

bestlibre:
**Description:**

Adding documentation for influxdb sensor

### DIFF
--- a/source/_components/sensor.influxdb.markdown
+++ b/source/_components/sensor.influxdb.markdown
@@ -18,7 +18,7 @@ To configure this sensor, you need to define the sensor connection variables and
 
 Configuration variables for the server:
 
-- **host** (*Optional*): IP address of your database host, eg. http://192.168.1.10. Defaults to `localhost`.
+- **host** (*Optional*): IP address of your database host, eg. 192.168.1.10. Defaults to `localhost`.
 - **port** (*Optional*): Port to use. Defaults to 8086.
 - **username** (*Optional*): The username of the database user.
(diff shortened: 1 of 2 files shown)


source/_components/sensor.influxdb.markdown
```diff
@@ -0,0 +1,71 @@
...
+footer: true
+logo: influxdb.png
+ha_category: Sensor
+ha_release: 0.32
+---
+
+The `InfluxDB` sensor allows you to use values from an InfluxDB database to populate a sensor state.
+
+To configure this sensor, you need to define the sensor connection variables and a list of queries. A sensor will be created for each query.
+
```
fabaff: Please add a minimal sample to get users started quickly.

```yaml
sensor:
  - platform: influxdb
    queries:
      - name: last value of foo
        unit_of_measurement: °C
        where: '"name" = ''foo'''
        measurement: '"°C"'
``

File: source/_components/sensor.influxdb.markdown [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+Configuration variables for the server:
+
+- **host** (*Optional*): IP address of your database host, eg. http://192.168.1.10. Defaults to `localhost`.
```
fabaff: Can you please do me a favor and check if that's right? We are talking about the IP address but show an URL. I guess that wrong in the InfluxDB component documentation as well.

bestlibre: I have corrected both. The influxdb client api wait for a hostname/ip address

PR #1345, opened 2016-10-27
Update the MQTT fan documentation

wingrunr21:
This is a documentation update for the MQTT Fan component. When the original component was merged (via home-assistant/home-assistant#3095) incomplete/incorrect documentation was added at the same time. This brings the docs inline with the component.

### DIFF
--- a/source/_components/fan.mqtt.markdown
+++ b/source/_components/fan.mqtt.markdown
@@ -17,31 +17,61 @@ The `mqtt` fan platform let you control your MQTT enabled fans.
 
 In an ideal scenario, the MQTT device will have a `state_topic` to publish state changes. If these messages are published with `RETAIN` flag, the MQTT fan will receive an instant state update after subscription and will start with correct state. Otherwise, the initial state of the fan will be `false`.
 
-When a `state_topic` is not available, the fan will work in optimistic mode. In this mode, the fan will immediately change state after every command. Otherwise, the fan will wait for state confirmation from device (message from `state_topic`).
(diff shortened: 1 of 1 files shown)


source/_components/fan.mqtt.markdown
```diff
@@ -17,31 +17,61 @@ The `mqtt` fan platform let you control your MQTT enabled fans.
...
+When a `state_topic` is not available, the fan will work in optimistic mode. In this mode, the fan will immediately change state after every command. Otherwise, the fan will wait for state confirmation from the device (message from `state_topic`).
 
-Optimistic mode can be forced, even if state topic is available. Try to enable it, if experiencing incorrect fan operation.
+Optimistic mode can be forced even if a `state_topic` is available. Try to enable it if you are experiencing incorrect fan operation.
 
 To enable MQTT fans in your installation, add the following to your `configuration.yaml` file:
 
 ```yaml
 # Example configuration.yml entry
 fan:
```
Landrash: This is a great addition *but* the example code should always try to be minimized.
Feel free to have a fully fleshed out example also but avoid having the **Optional** variables in the minimized one.

wingrunr21: So happy to make the change but the way the original author defined the component would mean the minimal version would ONLY be the command topic. Would you be ok with some kind of happy medium?

fabaff:
With https://github.com/home-assistant/home-assistant/issues/4134 was the documentation updated without taking that PR into account.

Duplicated work. I will fix my mess.

PR #1346, opened 2016-10-27
Add sensor types to Envisalink binary sensor docs

wingrunr21:
I recently set up this component and had to dig into the source to find these types. Figured dropping them into the docs directly might save some people some time in the future.


source/_components/binary_sensor.envisalink.markdown
```diff
@@ -14,5 +14,26 @@ ha_iot_class: "Local Push"
 ---
 
 The `envisalink` platform allows you to get data from your [Envisalink](http://www.eyezon.com/) binary sensors from within Home Assistant.
-  
+
 The requirement is that you have setup your [Envisalink hub](/components/envisalink/).
+
+#### Sensor Types
```
fabaff: It is already documented at https://home-assistant.io/components/binary_sensor/

fabaff: and https://home-assistant.io/getting-started/customizing-devices/#possible-values

wingrunr21: Ok, I missed that documentation somehow. What're your thoughts on removing the Envisalink binary sensor docs and directly linking there instead? There doesn't appear to be anything specialized about the Envisalink binary sensor that the general component docs don't cover. The `zones` area can include the link/documentation to the sensor type similar to how the CLI binary sensor does it.

fabaff: No, we can't remove the Envisalink binary sensor docs because we are linking from the configuration check to the docs.

PR #1347, opened 2016-10-27
New Details category

fabaff:
**Description:**
So far the "details" were shown as "Advanced Topics". It feels that they were misplaced there as this category is for advanced topics around the configuration.


Landrash:
Seems like a good plan! :)

PR #1372, opened 2016-10-31
Corrected second example, removing URL path content for pin

moellere:
**Description:**

Second example included full URL path.  Current implementation of aRest automatically adds /digital/X to URL, and specifying it in the configuration file causes the component not to load properly


robbiet480:
@moellere Thanks for this change but you need to make it against the `current` branch instead because it should go out ASAP instead of waiting for the next release. Please open a new PR with your change.

PR #1379, opened 2016-10-31
Added pydocstyle test

StaticCube:
**Description:**
Added the pydocstyle test to the bash syntax mainly because this is the one I always forget (and get a few messages from my new friend balloobbot) and to make the test sequence complete


robbiet480:
Thanks! 🐬 🍪 💯

PR #1416, opened 2016-11-07
Update z-wave.markdown

alanmimms:
**Description:**

The config path was missing "src" along the way. At least my installation required it.


Landrash:
@alanmimms This is actually correct for an install that's been created from the [Installation in a VirtualEnv](https://home-assistant.io/getting-started/installation-virtualenv/) guide. It differs from the All-in-One guide.

PR #1433, opened 2016-11-12
added extra config to use with AIO

godloth:
**Description:**

added the needed grant to be run as a non admin with AIO


Landrash:
Wasn't this discussed and and recommended not to be implemented since it's give the python instance privileges that are not fully tested and could potentially be exploited?

PR #1446, opened 2016-11-15
modified an error on the config syntax

Nicopython:
**Description:**

as stated here
https://home-assistant.io/getting-started/scripts-conditions/
under condition: can't set platform: time but condition: time


source/getting-started/automation-examples.markdown
```diff
@@ -30,7 +30,7 @@ automation:
     - platform: state
```
fabaff: I guess that this entry should be updated as well.

Landrash:
Addressed feedback from @fabaff and merging this to prevent problems for new users.

PR #1451, opened 2016-11-17
WAQI sensor

valentinalexeev:
**Description:**
Description for the WAQI sensor component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#4434

### DIFF
--- a/source/_components/sensor.waqi.markdown
+++ b/source/_components/sensor.waqi.markdown
@@ -1,0 +1,39 @@
+---
+layout: page
+title: "World Air Quality Index"
+description: "Instructions how to setup World Air Quality Index sensor in Home Assistant."
+date: 2016-11-17 06:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: smtp.png
+ha_category: Sensor
+ha_release: 0.32
+ha_iot_class: "Local Polling"
+---
+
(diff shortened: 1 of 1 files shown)


source/_components/sensor.waqi.markdown
```diff
@@ -0,0 +1,39 @@
...
+      - beijin
+```
+
+Configuration variables:
+
+- **locations** (*Required*): a list of location names to look for air quality data. In case a specific location has multiple registered stations all of them will be added to Home Assistant
+
+The value reported is an overall AQ index for the location. The values of the index can be interpreted as following:
+
+- from 0 to 50 - **Good**. Air quality is considered satisfactory, and air pollution poses little or no risk,
```
Landrash: Is the word `from` really needed in the list?

valentinalexeev: @Landrash I would rather have it as a table as on the original web site, however for some reason my markdown syntax wasn't good enough.

Landrash: Tables are a bit funny in markdown but [here's](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables) a guide if you would like to make one,

PR #1472, opened 2016-11-21
Added in helpful tip

CCOSTAN:
Excluding entries in logbook every 3 minutes.

### DIFF
--- a/source/_components/cover.garadget.markdown
+++ b/source/_components/cover.garadget.markdown
@@ -89,7 +89,7 @@ customize:
     icon: mdi:wifi
 ```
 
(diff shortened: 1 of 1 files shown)


source/_components/cover.garadget.markdown
```diff
@@ -89,3 +89,12 @@ customize:
     icon: mdi:wifi
 ```
 
+Some of the Garadget sensors can create a lot of clutter in the logbook.  Use this section of code in your `configuration.yaml` to exclude those pesky entries.
```
Landrash: I would prefer if you remove the word pesky to make this look a bit more professional.

Landrash:
Looks good. Thank you☃️

PR #1479, opened 2016-11-22
Quick time window update

CCOSTAN:
Added example of windo crossing midnight as per https://github.com/home-assistant/home-assistant/pull/2158


Landrash:
Good addition :)

PR #1480, opened 2016-11-22
Update sensor.pvoutput.markdown

Marcel030nl:
**Description:**
Added an example using the template component.

### DIFF
--- a/source/_components/sensor.pvoutput.markdown
+++ b/source/_components/sensor.pvoutput.markdown
@@ -35,3 +35,30 @@ Configuration variables:
 It's recommended to set `scan_interval:` according to a value greater than 60 seconds. The service only allows 60 requests per hour but the sensor's default is 30 seconds.
 </p>
 
+To format the PVoutput sensor you can you the [template component](/topics/templating/). For example:
+
+```yaml
+sensor:
+  - platform: pvoutput
+    system_id: YOUR_SYSTEM_ID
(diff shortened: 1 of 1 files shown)


source/_components/sensor.pvoutput.markdown
```diff
@@ -35,3 +35,30 @@ Configuration variables:
...
+```yaml
+sensor:
+  - platform: pvoutput
+    system_id: YOUR_SYSTEM_ID
+    api_key: YOUR_API_KEY
+    scan_interval: 150
+  - platform: template
+    sensors:
+      power_consumption:
+        value_template: '{% if is_state_attr("sensor.pvoutput", "power_consumption", "NaN") %}0{% else %}{{ states.sensor.pvoutput.attributes.power_consumption }}{% endif %}'
```
fabaff: All templates needs to be wrapped with `{% raw %}...{% endraw %}` otherwise they are rendered.

Landrash:
Wouldn't this be better suited in the [Example/Cookbook](https://home-assistant.io/cookbook/) section since it's a good generalized example also?

fabaff:
I though about that too but for me it looks like that the template stuff it's very sensor-specific and could help users of the PVOutput sensor to find all details at one place.

Landrash:
Good work. Merging this in but feel free to make a copy or similar in the Cookbook section.

PR #1486, opened 2016-11-23
Corrected the mapping for HA vs MySensors State

kaustubhphatak:
Correction to the mapping of states between HA and MySensors


Landrash:
Interesting mistake. Good find and thank you. :cookie:

PR #1488, opened 2016-11-23
Add docs for NUT sensor

mezz64:
**Description:**
Documentation for the NUT sensor platform.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#4551


source/_components/sensor.nut.markdown
```diff
@@ -0,0 +1,95 @@
...
+    host: 192.168.11.5
+    alias: ups_name
+    username: user
+    password: pass
+    resources:
+      - ups.load
+      - ups.realpower.nominal
+      - input.voltage
+      - battery.runtime
+```
```
fabaff: We try to keep the configuration sample minimal thus only show the required variables. This helps people to get started quickly by copy-&-paste the information without the hassle of thinking the optional stuff.

source/_components/sensor.nut.markdown
```diff
@@ -0,0 +1,94 @@
...
+date: 2016-11-23
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: nut.png
+ha_category: Sensor
+---
+
+The `nut` sensor platform allows you to monitor a UPS (battery backup) by using data from the NUT server.
```
Landrash: This should most likely be "a NUT server" since I assume there can potentionaly be more than one host.

source/_components/sensor.nut.markdown
```diff
@@ -0,0 +1,94 @@
...
+layout: page
+title: NUT Sensor
+description: "Instructions on how to set up NUT sensors within Home Assistant."
+date: 2016-11-23
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: nut.png
+ha_category: Sensor
```
Landrash: Don't forget  "ha_version" for when this is merged in.
Example if this is merged in before next release then this would be `ha_version: 0.34`.

source/_components/sensor.nut.markdown
```diff
@@ -0,0 +1,94 @@
...
+      - input.voltage
+      - battery.runtime
+```
+
+Configuration variables:
+
+- **name** (*Optional*): Name prefix for defined sensors. Defaults to 'NUT UPS'.
+- **host** (*Optional*): The host name or address of the device that is running NUT. Defaults to localhost.
+- **port** (*Optional*): The port number. Defaults to 3493.
+- **alias** (*Optional*): Name of the ups on the NUT server. Will default to the first ups name listed.
```
Landrash: Should be UPS instead of ups.

source/_components/sensor.nut.markdown
```diff
@@ -0,0 +1,94 @@
...
+'battery.charge': '100', 
+'input.sensitivity': 'medium', 
+'input.transfer.low': '88'
+'input.transfer.high': '147', 
+'input.voltage': '121.0', 
+'input.voltage.nominal': '120', 
+'input.transfer.reason': 'input voltage out of range', 
+```
+
+Use the (case insensitive) values from the left hand column.  Support is included for most values with ups, battery, and input prefixes.
```
Landrash: Should be UPS instead of ups.

mezz64: I left this lowercase but put it in '' since I'm directly referring to how it will be in the yaml.

PR #1492, opened 2016-11-24
Proper documentation for foursquare

simonszu:
Follow up for https://github.com/home-assistant/home-assistant.github.io/pull/1471

### DIFF
--- a/source/_components/foursquare.markdown
+++ b/source/_components/foursquare.markdown
@@ -28,6 +28,20 @@ Configuration variables:
 - **access_token** (*Required*): A Foursquare API access token.
 - **push_secret** (*Required*): The push secret that Foursquare provides to you in the app dashboard.
 
+#### Getting the access token ####
+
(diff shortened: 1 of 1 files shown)


Landrash:
Looks good but it's against `next` branch instead of current. I can merge into that if you would like but it won't get published until the weekend after this.

simonszu:
I'm okay with this. I had no idea that committing in the current branch was possible, since the developer's guide said one should add documentation for new components in the next branch, so i assumed that the build of the website occurs only on releasing a new version. Well, my attempt to merge master branch into my working copy's next branch should have tell me otherwise, but it's okay. 
I have figured it out how to obtain the access_token, and if other's would have had problems until now, there would be more questions in the forums - so if noone has these problems beside than me, the publication can wait some more days.

Landrash:
Great, thanks for the great addition. 🎂

PR #1501, opened 2016-11-26
Fix binary_sensor.command_line example

michaelarnauts:
The example to check if an IP is alive didn't work. It seems that the `&> /dev/null` redirector didn't work. `> /dev/null 2>&1` works fine however.

This can be verified with the following config:
```yaml
- platform: command_line
  name: "testtrue1"
  command: "true &> /dev/null && echo success || echo fail"
  scan_interval: 5
  payload_on: "success"
  payload_off: "fail"

- platform: command_line
  name: "testfalse1"
  command: "false &> /dev/null && echo success || echo fail"
  scan_interval: 5
  payload_on: "success"
  payload_off: "fail"

- platform: command_line
  name: "testtrue2"
  command: "true > /dev/null 2>&1 && echo success || echo fail"
  scan_interval: 5
  payload_on: "success"
  payload_off: "fail"

- platform: command_line
  name: "testfalse2"
  command: "false > /dev/null 2>&1 && echo success || echo fail"
  scan_interval: 5
  payload_on: "success"
  payload_off: "fail"
```

Also a `-W 1` lowers the timeout to one second.

### DIFF
--- a/source/_components/binary_sensor.command_line.markdown
+++ b/source/_components/binary_sensor.command_line.markdown
@@ -73,7 +73,7 @@ An alternative solution could look like this:
 binary_sensor:
   platform: command_line
   name: Printer
-  command: ping -c 1 192.168.1.10 &> /dev/null && echo success || echo fail
+  command: ping -W 1 -c 1 192.168.1.10 > /dev/null 2>&1 && echo success || echo fail
   sensor_class: connectivity
   payload_on: "success"
   payload_off: "fail"


fabaff:
`&> /dev/null` seems to work fine with `GNU bash, version 4.3.42(1)-release (x86_64-redhat-linux-gnu)`.

source/_components/binary_sensor.command_line.markdown
```diff
@@ -73,7 +73,7 @@ An alternative solution could look like this:
 binary_sensor:
   platform: command_line
   name: Printer
-  command: ping -c 1 192.168.1.10 &> /dev/null && echo success || echo fail
+  command: ping -W 1 -c 1 192.168.1.10 > /dev/null 2>&1 && echo success || echo fail
```
fabaff: The timeout is a nice addition.

michaelarnauts:
Hmm, I'm running in the default Docker container. When I execute in a shell, it works fine, maybe it's a limitation of the Python process library? Have you tried it from within a binary sensor?

fabaff:
Works for a binary sensor as well.

michaelarnauts:
Strange... Are your running hass inside a docker? I am, and maybe there is a different bash there?

fabaff:
No docker. As far as I remember is Ubuntu using `dash` instead of `bash`.

michaelarnauts:
My host is running Ubuntu 14.04 and it worked fine there, also from the cli inside the hass container. Could it be related to the Python library?

michaelarnauts:
I'm using the exact configuration from above. See the result of `testfalse1`.

screenshot from 2016-11-27 08-08-02

From inside the container, it is working fine. I guess this must be related to Python then.
```bash
root@9e40628f9282:/usr/src/app# false &> /dev/null && echo success || echo fail                                                             
fail
```

I've spend over an hour to troubleshoot this, the reason for this PR was to help others who also copy this code snippet so they don't lose this time. 
If you don't want to change the example, can we at least put a notice below the code to indicate that `> /dev/null 2>&1` can give the required results in case `&> /dev/null` doesn't?

michaelarnauts:
It has to be python related... What output do you get when you execute this?

```python
import subprocess

command = "false &> /dev/null && echo success || echo fail"
return_value = subprocess.check_output(command, shell=True, timeout=15)
value = return_value.strip().decode('utf-8')
print(value)

command = "false > /dev/null 2>&1 && echo success || echo fail"
return_value = subprocess.check_output(command, shell=True, timeout=15)
value = return_value.strip().decode('utf-8')
print(value)
```

root@9e40628f9282:/usr/src/app# python test.py 
success
fail

fabaff:
Output is:

```bash
...-assistant]$ python3 shell-test.py 
fail
fail
```

michaelarnauts:
Wtf... I get above output on HASS docker and on it's host. Host is Ubuntu 14.04.

Landrash:
This could be a question about what underlying shell is used by the user that executes this.
What shell are the both of you using? You can check with `echo $0`.

fabaff:
As mentioned before `GNU bash, version 4.3.42(1)-release (x86_64-redhat-linux-gnu)`.

lwis:
My vote is for `> /dev/null 2>&1` as it's the most consistent (albeit verbose) solution across shells.

Also, is it worth adding something like `nc -zw 1 google 80` as another example of testing a connection to a service?

michaelarnauts:
I honestly don't understand why it's a problem to improve an example that works in more shells. Especially since examples are mainly for people who don't know the exact working of hass, so in my opinion, it should not cause extra troubles to find out why the example doesn't work in their case like I had to do.

Also, docker is a supported way of installing hass, and the "official" hass docker image has this problem.

lwis:
@michaelarnauts I'm with you, bud. Docs like this should aim for majority compatibility.

michaelarnauts:
Anyway, it seems docker is using /bin/sh as shell in python.
```python
command = "echo $0"
return_value = subprocess.check_output(command, shell=True, timeout=15)
value = return_value.strip().decode('utf-8')
print(value)
```
returns `/bin/sh`

fabaff:
It's not a problem at all, to improve to a way that covers most of the ground. Nobody is against it.

Well, for an simple change like this we perhaps spent a bit too much time on it. The benefit is that we now have something in our hands if it become an issue again in the future.

PR #1520, opened 2016-11-30
Use adduser instead of useradd

brandonweeks:
`adduser` is the "preferred" way to add a new user in Debian and the command is present in Raspbian.

### DIFF
--- a/source/getting-started/installation-raspberry-pi.markdown
+++ b/source/getting-started/installation-raspberry-pi.markdown
@@ -15,7 +15,7 @@ There's currently three documented ways to install Home Assistant on a Raspberry
  - [Manual installation](/getting-started/installation-raspberry-pi/#Manual-Installation). Following this guide doing each step manually. This is highly recommended as a first installation since you get a good overview of the installation.
  - [Hassbian image](/getting-started/installation-raspberry-pi-image). Basic installation with the same settings as following the manual installation guide. Some additional software is preinstalled to make installation quicker and easier. Installation uses `homeassistant` user.
  - [All-in-One Installer](/getting-started/installation-raspberry-pi-all-in-one/). Fabric based installation script that installs and compiles many of the things an advanced Home Assistant install is likely to need. Installation uses `hass` user.
- 
+
  <p class='note note'>
   Since each installation type uses a different user for Home Assistant, be sure to note and use the correct username for the `adduser` commands listed below for camera and GPIO extensions.
 </p>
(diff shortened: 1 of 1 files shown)


Landrash:
`adduser` uses `useradd` in the background. Issuing your command would break the rest of this script since there wont be a home directory created. `useradd` is also the tool that's used to create the default pi user in the build script and therefore the way I choose to use in this guide to keep it consistent. 

If I where to rewrite this guide to work on other platforms like FreeBSD then I would consider your change to keep it consistent over platforms.
In reference to preferred I would love a source on that.

brandonweeks:
@Landrash my original PR did indeed break the script due to the lack of `--group`. `adduser` does however create a home directory by default.

Landrash:
@brandonweeks `adduser` creates a home directory by default but not for a `system account`.
Passing the `--system` variable then requires you to pass the `-m` or the friendlier `--create-home` variable. Based on this your command would have to be `sudo adduser --group --system --create-home  homeassistant` instead of `sudo useradd -rm homeassistant`.

In short changing this would have no real benefit to the end user in my eyes and the only potentially bad thing is introducing a new user to more command-line tools such as `useradd`.

brandonweeks:
I feel like we're referencing potentially different versions of `adduser`? This is what is present in the manpage of a recent version of Debian and is functionally accurate from my testing. I'll admit this change is trivial but it does work. :)

```
   Add a system user
       If  called  with  one  non-option  argument and the --system option, adduser will add a system
       user. If a user with the same name already exists in the system uid range (or, if the  uid  is
       specified,  if  a  user  with that uid already exists), adduser will exit with a warning. This
       warning can be suppressed by adding "--quiet".

       adduser will choose the first available UID from the range specified for system users  in  the
       configuration file (FIRST_SYSTEM_UID and LAST_SYSTEM_UID). If you want to have a specific UID,
       you can specify it using the --uid option.

       By default, system users are placed in the nogroup group.  To place the new system user in  an
       already existing group, use the --gid or --ingroup options.  To place the new system user in a
       new group with the same ID, use the --group option.

       A home directory is created by the same rules as for normal users.  The new system  user  will
       have  the  shell  /bin/false (unless overridden with the --shell option), and have logins dis‐
       abled.  Skeletal configuration files are not copied.
```

Landrash:
I'll consider it after I've verified it on a fresh install. Just realized I've used `adduser` in the components sections lower down so standardizing over this document might be good after all.

Most likely will have to wait until this weekend.

andrey-git:
@Landrash any updates on this?

Landrash:
@forgot about it and it will have to wait untilt he refactor of the Getting-started is done.

andrey-git:
This should either be fixed to use the new docs location or closed.

Landrash:
Closing since the docs are being reworked.

PR #1521, opened 2016-11-30
Update z-wave.markdown

Landrash:
**Description:**

Specify version of cython that installed.


Landrash:
Updated to include maximum version of cython to avoid having build fail if version 0.25 is used.

PR #1522, opened 2016-12-01
Example of Data_template with IFTTT

CCOSTAN:
This drove me nuts for a few hours trying to get the exact syntax down so I thought I would share.


Landrash:
Looks good and going to trust you that it works ;)

PR #1524, opened 2016-12-02
Add keyword 'condition' for completeness

naren8642:
**Description:**

Added the 'condition:' keyword to make it clear that for a basic condition definition, you will need that keyword twice.


Landrash:
@fabaff Looks good but I'm not aware if this was done on purpose from the beginning?

fabaff:
I really don't know. Let's merge and see how it works out.

PR #1548, opened 2016-12-05
show a template switch example for harmony

nkgilley:
**Description:**
I thought this ought to be included in these docs.  Describes how to use a template switch to view and control the state of a harmony activity.

### DIFF
--- a/source/_components/remote.harmony.markdown
+++ b/source/_components/remote.harmony.markdown
@@ -67,7 +67,7 @@ switch:
   - platform: template
     switches:
       tv:
-        value_template: "{% if is_state('remote.family_room', 'on') %}on{% else %}off{% endif %}"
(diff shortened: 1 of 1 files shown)


source/_components/remote.harmony.markdown
```diff
@@ -60,6 +60,22 @@ Supported services:
...
 ### {% linkable_title Examples %}
 
+A template switch can be used to display and control the state of an activity in the frontend.
+
+```yaml
+switch:
+  - platform: template
+    switches:
+      tv:
+        value_template: "{% if is_state('remote.family_room', 'on') %}on{% else %}off{% endif %}"
```
fabaff: Templates needs to be wrapped in ` {% raw %}...{% endraw %}` otherwise they will appear blank after the page is created.

PR #1551, opened 2016-12-06
Update light.lifx.markdown

mehuman:
**Description:**
Added tip to include broadcast if Lifx lights aren't immediately showing up on a restart


Landrash:
This would be better if there was an example of what the broadcast address can be.

mehuman:
Updated with an example broadcast address. Happy to add an explanation of what how to find the user's broadcast address if not outside the scope of this document.

Landrash:
I think this is enough and gives enough of a hint to find it yourself if needed.
Good job and thank you 🍸

PR #1562, opened 2016-12-08
example for optional values

keatontaylor:
**Description:**
Added where to place optional values relay_time and state_pull_mode.

https://github.com/home-assistant/home-assistant/issues/4818

### DIFF
--- a/source/_components/cover.rpi_gpio.markdown
+++ b/source/_components/cover.rpi_gpio.markdown
@@ -27,6 +27,8 @@ To enable Raspberry Pi Covers in your installation, add the following to your `c
 # Example configuration.yaml entry
 cover:
   platform: rpi_gpio
+  relay_time: 0.2
+  state_pull_mode: 'UP'
   covers:
     - relay_pin: 10
       state_pin: 11


source/_components/cover.rpi_gpio.markdown
```diff
@@ -27,6 +27,8 @@ To enable Raspberry Pi Covers in your installation, add the following to your `c
 # Example configuration.yaml entry
 cover:
   platform: rpi_gpio
+  relay_time: 0.2
+  state_pull_mode: 'UP'
```
fabaff: Those are optional entries. As we try to keep the configuration sample minimal, please add them only to the "Configuration variables".

keatontaylor: They're already added to the Configuration variables, but it wasn't clear where 'relay_time' and 'state_pull_mode' were supposed to go in the configuration. Essentially leading us on a multi hour long troubleshooting. 

We had been trying to add them like this:

```
cover:
  platform: rpi_gpio
  covers:
    - relay_pin: 10
      state_pin: 11
    - relay_pin: 12
      state_pin: 13
      name: 'Right door'
```

But the configuration actually calls for them to to be like this:
```
cover:
  platform: rpi_gpio
  relay_time: 0.2
  state_pull_mode: 'UP'
  covers:
    - relay_pin: 10
      state_pin: 11
    - relay_pin: 12
      state_pin: 13
      name: 'Right door'
```

How do you suggest we avoid that confusion for others?

fabaff: By adding an [example](https://home-assistant.io/components/cover.command_line/#examples) section. It happens a couple of times in the past that people copied the sample configuration without using their brain and then setup failed because the values didn't match their environment. 

I guess that the "Configuration variables:" listing should be updated as well to show the right position of the entries.

PR #1566, opened 2016-12-08
added example of changing input boolean.

CCOSTAN:
Setting it programmatically.


source/_components/input_boolean.markdown
```diff
@@ -50,3 +50,9 @@ automation:
       title: ""
       message: "Honey, I'm home!"
 ```
+
+You can also set or change the status of an `input_boolean` by using `input_boolean.turn_on` and `input_boolean.turn_off` in your automations.
+
+    - service: input_boolean.turn_on
```
fabaff: This should be formatted as YAML.

CCOSTAN:
Done.

PR #1597, opened 2016-12-15
Replaces 'sudo su' with 'sudo -s'

kirichkov:
**Description:**

There's no need to use `sudo su -s ...` as this invokes first sudo, becoming super-user and then again as super-user invoking `su`.

To replicate the functionality it's sufficient to invoke `sudo` with the follogin parameters
- `-u USERNAME` execute as the given username
- `-s` start a shell
- `-H` set HOME environment variable


Landrash:
@kirichkov This change could be made but would large parts of our documentation inconsistent.
`sudo` is a very powerful command but the part's I've written I've chosen to follow the format used in the current documentation and in the Raspbian documentation.

In general the Home Assistant user is supposed to be setup as a system account without a shell specified and that why we pass the shell variable manually. 
There's also an inconsistency in how you have chosen to lay out the options over the documentation.

kirichkov:
OK. This makes sense.

PR #1599, opened 2016-12-15
iCloud depency fail, manual install notice

BruurTycho:
**Description:**

This will explain the user when the auto install of the depency fails the user can manually install it using the command in the text.


Landrash:
This seem counter-intuitive to add to the documentation. 
It informs you in the logs about the error and adding this to each component will quickly get messy.
@fabaff I think we shouldn't add this to the documentation if it's not a temporary workaround since it an indication of and error either in HA or with the dependency. Your thoughts?

andrey-git:
Any updates on this?

andrey-git:
Closing, since there are no updates. Please reopen if still relevant.

PR #1609, opened 2016-12-16
Initial documentation for switch.tilt component

JerryWorkman:
**Description:**
Combines a tilt sensor and a relay switch to control a garage door opener or motorized gate.
When this switch is turned on it will toggle the relay to activate the garage door if it is
not already open. The reverse will happen when it is turned off.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) home-assistant/home-assistant#4921


fabaff:
Is that a duplicate of #1609?

JerryWorkman:
Must be. Still learning the ropes...

fabaff:
I would like to suggest that you create a new branch for your new docs and then you can push to that branch.

PR #1610, opened 2016-12-17
s/sensor/switch/ in mfi docs

pgenera:
**Description:**

This appears to be a copy & paste error; it does need to both switch and sensor stanzas to work in my installation.


Landrash:
Thank you. :cookie:

PR #1629, opened 2016-12-19
Removing references to URL and replacing with IP address

terrycarlin:
The homeassistant/components/media_player/itunes.py requires an IP address not a URL

**Description:**


Landrash:
I'm fairly certain that this can be either a `hostname` or IP-address @terrycarlin

PR #1631, opened 2016-12-20
Security Tip for changing AiO MQTT password.

CCOSTAN:
went back and forth on whether to put this here or in the All in One installer docs.  I think it is best suited here since users will come here once they want to learn more about MQTT and their particular installation which could be a significantly time after when they first did the install.

### DIFF
--- a/source/_components/mqtt.markdown
+++ b/source/_components/mqtt.markdown
@@ -99,6 +99,13 @@ mqtt:
     # Your HBMQTT config here. Example at:
     # http://hbmqtt.readthedocs.org/en/latest/references/broker.html#broker-configuration
 ```
+<p class='note'>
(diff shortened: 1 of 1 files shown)


CCOSTAN:
I moved this to the All In One installer section.

dale3h:
It wouldn't hurt to have something similar to this on the MQTT components page to show users how to add new passwords to mosquitto.

CCOSTAN:
I restored the branch and reopened in case @fabaff wants it in both places.

Landrash:
I think the addition of it only in the AiO docs and in the Fabric repository is enough.
Having this in the MQTT documentation for an external component will only cause confusion.
If anything it could be documented in the Ecosystem part of our documentation.

PR #1634, opened 2016-12-20
Update tts.markdown

hokagegano:
**Description:**
known problem with pulseaudio when hass running as root


Landrash:
Home Assistant should not be run as root.
In general permissions for specific function can be applied to a none root account.

PR #1638, opened 2016-12-21
Remove superfluous -

kylegordon:
Documentation fix.
Lifx broadcast doesn't need -

**Description:**


arsaboo:
Wondering if we should remove the `-` in front of `platform` as well?

mehuman:
Looking at other light components there isn't a standard in regards to a `-` being in front of the platform. I'm leaning towards having it there so when users add more platforms the copy and pasting just works.

PR #1640, opened 2016-12-22
Add note about new default Raspbian behavior

bassclarinetl2:
figured this note should be present to save some facedesking.

apologies for the bad markdown (assuming it fails to render correctly.)

**Description:**

Since I had to go digging (read: somewhat obscure blog post) to find out why I couldn't SSH in to a fresh image, and it's a reasonable assumption that users doing this will be running headless, it might be a good idea to include this.


Landrash:
Small change was made but looks good. Thank you :christmas_tree:

PR #1642, opened 2016-12-22
hdmi_cec enhancement

konikvranik:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** home-assistant/home-assistant#4781

### DIFF
--- a/source/_components/hdmi_cec.markdown
+++ b/source/_components/hdmi_cec.markdown
@@ -106,7 +106,7 @@ Choose just one schema. Mixing both approaches is not possible.
 
 ### {% linkable_title Select Device %}
 
-Call the `hdmi_cec/select_device` service with the name of the device from config or entity_id or prysical address"to select it, for example:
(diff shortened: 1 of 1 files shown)


Landrash:
Review of this one will wait until component is done.
Just a quick overview of this and it looks way more complicated than before to configure. If we update the config of the component we should always try to keep it compatible with the previous version to prevent breakage. If there's breakage don't forget to document it and how to update the previous configurations to match the new requirement.

konikvranik:
@Landrash 
Do you find it more complicated? I changed format of config because previous approach seems complicated to me.
But maybe there is posibility to retain both approaches. If number is on left side and string on right, we use original way and if string on left and pattern %x.%x.%x.%x we use new format.

Landrash:
@konikvranik Your right, either will work. If there's no reason to change I would recommend not doing it as it will break some deployments of Home Assistant.

konikvranik:
@Landrash 
OK, you're right, backwards compatibility is important. Reworked to offer both approaches.

robbiet480:
Thanks! 🐬 🍪 💯

PR #1656, opened 2016-12-26
"homeassistant" changed in "hass"

PuckStar:
Couldn't update Home Assistant. 
Found out the word homeassistant was changed in hass.
After changing that the update worked.

**Description:**

### DIFF
--- a/source/getting-started/installation-raspberry-pi-all-in-one.markdown
+++ b/source/getting-started/installation-raspberry-pi-all-in-one.markdown
@@ -57,8 +57,8 @@ The All-In-One Installer script will do the following automatically:
 To upgrade the All-In-One setup manually:
 
 *  Login to Raspberry Pi `ssh pi@your_raspberry_pi_ip`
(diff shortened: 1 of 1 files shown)


PuckStar:
Hi Fredrik,


I just here from @dale3h that perhaps I had such an old version of HASS that that's the reason I needed to use the username HASS instead of homeassistant.

So perhaps for new installations the username should be be: homeassistant.

In that case I think we need to revert the pull request I did. And perhaps mention for which version people need to use hass and for which version the need to use homeassistant.

What do you think?


Lukas


________________________________
Van: Fredrik Lindqvist <notifications@github.com>
Verzonden: maandag 26 december 2016 16:05
Aan: home-assistant/home-assistant.github.io
CC: PuckStar; Mention
Onderwerp: Re: [home-assistant/home-assistant.github.io] "homeassistant" changed in "hass" (#1656)


@Landrash approved this pull request.

Looks good. Thank you. ??

-

Landrash:
You are correct, reverting. Sorry about that. Apparently I should follow the AiO installer more carefully. :)

PR #1666, opened 2016-12-28
Fix example automation syntax in light.flux_led.markdown

ahnooie:
**Description:**
Tried to follow this example and the configuration syntax was invalid.  Fixed.


tchellomello:
Looks good!

PR #1671, opened 2016-12-29
media_player/vlc: More details about inner working

tbarbette:
- Say that only music is supported
- Explain that it launches a VLC in background and this does not allow to control a GUI instance.

@fabaff this maybe not the better phrasing, but I think this bit of documentation is needed.


fabaff:
The `vlc` doc was only a stub.

PR #1687, opened 2016-12-31
Update systemd template used in manual installs

tmcneal:
**Description:** The systemd template included on the "Getting Started" guide assumes that the user running Home Assistant is called 'hass', however the [Manual Installation guide](https://home-assistant.io/getting-started/installation-raspberry-pi/) uses the username 'homeassistant'.

Testing:
- Installed the new template and ran `sudo systemctl start home-assistant@homeassistant`.  Verified  that `hass` is running and that Home Assistant is accessible via port 8123.

### DIFF
--- a/source/getting-started/autostart-systemd.markdown
+++ b/source/getting-started/autostart-systemd.markdown
@@ -35,7 +35,7 @@ WantedBy=multi-user.target
 EOF'
 ```
 
-If you've setup Home Assistant in `virtualenv` following the guide the following template should work for you.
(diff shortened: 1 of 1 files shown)


attzonko:
It would be nice if somehow the directions can allow the person to fill in the username they used for their own installation. For example:
User={USERNAME_YOU_CREATED}

That is just my opinion. I am fine with the PR as it stands since it makes the example match the username from the Manual Installation guide.

ryqiem:
The name used in the template is the name used in the VirtuelEnv guide: https://home-assistant.io/getting-started/installation-virtualenv/

I guess we 'ought to fix that inconsistency as the lowest hanging fruit.

PR #1690, opened 2017-01-01
Fix formatting error in Samsung TV component docs

jimrollenhagen:
I assume this was meant to be in backticks, also open to just removing the backtick and carat. :)


GreenTurtwig:
Great :cat: 
Such a silly thing to be missed, thank you! :smile_cat:

PR #1700, opened 2017-01-02
Merge documentation update for PR 4981

Duoxilian:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#4981

### DIFF
--- a/source/_components/climate.markdown
+++ b/source/_components/climate.markdown
@@ -75,6 +75,28 @@ automation:
         away_mode: true
 ```
 
+### {% linkable_title Service `climate.set_home_mode` %}
+
+Turn home mode on/off for climate device
+
+| Service data attribute | Optional | Description |
+| ---------------------- | -------- | ----------- |
+| `entity_id` | yes | String or list of strings that point at `entity_id`'s of climate devices to control. Else targets all.
+| `home_mode` | no | New value of home mode.
+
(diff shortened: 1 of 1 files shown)


source/_components/climate.markdown
```diff
@@ -75,6 +75,28 @@ automation:
         away_mode: true
 ```
 
+### {% linkable_title Service `climate.set_home_mode` %}
+
+Turn home mode on/off for climate device
+
+| Service data attribute | Optional | Description |
+| ---------------------- | -------- | ----------- |
+| `entity_id` | yes | String or list of strings that point at `entity_id`'s of climate devices to control. Else targets all.
```
philhawthorne: I would change

> that point at `entity_id`'s of climate devices

to read

> that point at `entity_id`'s climate device(s)

Duoxilian: Note that the whole text and also the text for other devices reads "devices". While I agree with you that your suggestion is better stylistically, without changing this phrase everywhere, it would just be inconsistent usage.

philhawthorne: Ah yes good call

Duoxilian:
Based on discussions concerning the implementation, this feature will be implemented at the ecobee level. Therefore, this documentation change will be withdrawn.

PR #1703, opened 2017-01-03
Bom weather docs

Zac-HD:
I've been working on BOM weather information, now that there's a new weather component.  This is the corresponding docs for home-assistant/home-assistant#5153

### DIFF
--- a/source/_components/sensor.bom.markdown
+++ b/source/_components/sensor.bom.markdown
@@ -12,25 +12,11 @@ ha_category: Weather
 ha_release: 0.29
 ---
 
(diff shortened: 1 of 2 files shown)


Zac-HD:
Thanks :smile:

I've fixed the typos, but it seems that in markdown this is the only way to write a `note` block.  The same pattern is used in (eg) `sensor.octoprint`, the old `sensor.bom` page, `getting-started/z-wave`, etc.  If there is a better way I'd love to hear about it!

Zac-HD:
Rebased, with minor edits to stay current with the main pull request.

PR #1707, opened 2017-01-03
Update autostart-systemd.markdown

ptutt1:
I needed to change a few settings to get this to work.
1. Using a parameter (%1) for the username when using pi did not work.  I received a 217/user error when viewing the log after attempting to start the service.
2. I needed to prefix the su with sudo, as I understand there is no password for the su?

**Description:**

### DIFF
--- a/source/getting-started/autostart-systemd.markdown
+++ b/source/getting-started/autostart-systemd.markdown
@@ -17,10 +17,10 @@ $ ps -p 1 -o comm=
 
 If the preceding command returns the string `systemd`, you are likely using `systemd`.
 
-If you want Home Assistant to be launched automatically, an extra step is needed to setup `systemd`. You need a service file to control Home Assistant with `systemd`. If you are using a Raspberry Pi with Raspbian then replace the `[your user]` with `pi` otherwise use your user you want to run Home Assistant. `ExecStart` contains the path to `hass` and this may vary. Check with `whereis hass` for the location.
+If you want Home Assistant to be launched automatically, an extra step is needed to setup `systemd`. You need a service file to control Home Assistant with `systemd`. If you are using a Raspberry Pi with Raspbian then replace the `[your user]` and %1 with `pi` otherwise use your user you want to run Home Assistant. `ExecStart` contains the path to `hass` and this may vary. Check with `whereis hass` for the location.
 
 ```bash
-$ su -c 'cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
+$ sudo su -c 'cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
 [Unit]
 Description=Home Assistant
 After=network.target


source/getting-started/autostart-systemd.markdown
```diff
@@ -17,10 +17,10 @@ $ ps -p 1 -o comm=
 
 If the preceding command returns the string `systemd`, you are likely using `systemd`.
 
-If you want Home Assistant to be launched automatically, an extra step is needed to setup `systemd`. You need a service file to control Home Assistant with `systemd`. If you are using a Raspberry Pi with Raspbian then replace the `[your user]` with `pi` otherwise use your user you want to run Home Assistant. `ExecStart` contains the path to `hass` and this may vary. Check with `whereis hass` for the location.
+If you want Home Assistant to be launched automatically, an extra step is needed to setup `systemd`. You need a service file to control Home Assistant with `systemd`. If you are using a Raspberry Pi with Raspbian then replace the `[your user]` and %1 with `pi` otherwise use your user you want to run Home Assistant. `ExecStart` contains the path to `hass` and this may vary. Check with `whereis hass` for the location.
 
 ```bash
-$ su -c 'cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
+$ sudo su -c 'cat <<EOF >> /etc/systemd/system/home-assistant@[your user].service
```
smurfix: Why are you naming this file @youruser? that does not make sense. You should name it `home-assistant@.service`, which also would make the %i work.

smurfix: Also, the '>>' is wrong; what if the file exists and the user wants to replace it? Use '>'.

smurfix: … and finally, "sudo su" is not a good idea because /bin/su creates a new session with logind. "sudo sh" is much better.

ptutt1: Honestly, I would count myself a noob with this stuff.  I made the changes to get it to work.
You can see, I only made 2 changes...i) insert sudo.  This was required because I believe out of the box a raspberry pi does not have a su password...and if it did, I don't know what it is.  I was just following recommendations for other sites on the web.  Your suggestion could well be better than mine...but I can say that mine got it over the line.

ptutt1: ii) I replaced i% with the actual user name as it was not working.  If your suggestion to remove the [your user] from the filename works, then that would be a better change.  I don't know why the file was named with the [your user] suffix.  I guess we would need to confirm that your change works for both the pi user and the created user.  I have only confirmed that my change works for the pi user.

ptutt1: In general, from my limited experience I agree your suggested changes are better than mine...I got it to work, but I am pretty blind in just following other recommendations for the web.

ptutt1: Also agree with your >> change to >.  I don't think it makes sense to append.

fabaff: > I don't know why the file was named with the [your user] suffix.

Because this lead to one of the simplest service units and allows one to run HA as the user how installed it including access to the configuration file.

smurfix: @fabaff No, that's not how systemd wants to do this.

The correct way is to **name** the service file `foo@.service`.
You then **use** it like "`systemctl start foo@username.service`.
See the systemd manpages for details.

fabaff: Usually people are not running multiple instances of HA, thus there is not need to create a template which can be invoked multiple times. As far as I remember is systemd checking for `foo@username.service` first and if that template is not found for `foo@.service`. Yes, that's still the [case](https://www.freedesktop.org/software/systemd/man/systemd.unit.html).

For the sake of flexibility I agree that `foo@.service` would be the better choice.

andrey-git:
Is this PR still relevant?

ptutt1:
I dont believe so.  The other contributors had better suggestions than mine.

Landrash:
Closing since pr is no longer relevant.

PR #1712, opened 2017-01-04
Add documentation for the pilight echo option

janLo:
**Description:**

Add documentation for the `echo` option in the pilight switch config.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#5056


source/_components/switch.pilight.markdown
```diff
@@ -55,6 +55,12 @@ Variables for the different codes:
 - **'off'** (*Optional*): `1` or `0`
 - **'on'** (*Optional*): `1` or `0`
 
+Besides that the he `on_code_receive` and `off_code_receive` have the following special option:
+
+- **echo** (*Optional*) `true` if the on-/off-code should be sent if the given code was received.
+
+This is useful if you have paired your sender directly with the receiver to prevent sending the signal twice.
```
mnoorenberghe: It's not clear when looking at the output that this paragraph is only referring to the `echo` option. Perhaps the paragraph below should move before the additions and/or `echo` should be in a sub-section?

source/_components/switch.pilight.markdown
```diff
@@ -55,6 +55,12 @@ Variables for the different codes:
 - **'off'** (*Optional*): `1` or `0`
 - **'on'** (*Optional*): `1` or `0`
 
+Besides that the he `on_code_receive` and `off_code_receive` have the following special option:
```
mnoorenberghe: Please cleanup "Besides that the he". Also I think this would be less verbose/clearer as a linkable section heading e.g. "`on_code_receive`/`off_code_receive` options" with a link from "echo" documented under both on_code_receive/off_code_receive's variable section above (lines 45+46).

PR #1723, opened 2017-01-06
Add ping docs

fabaff:
**Description:**
Initial `ping` device tracker docs.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#5176

### DIFF
--- a/source/_components/device_tracker.ping.markdown
+++ b/source/_components/device_tracker.ping.markdown
@@ -27,7 +27,7 @@ device_tracker:
 
 Configuration variables:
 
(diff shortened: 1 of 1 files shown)


michaelarnauts:
Nice!
The closing parenthesis of `count` is missing though. 
It might also be better to word the `hosts` as *List of device names and their corresponding IP address or hostname.*
This makes it clear that it can do a DNS lookup and that the key is just a identifier.

fabaff:
Thanks for your feedback.

PR #1730, opened 2017-01-06
Heartbeats section added

brahmafear:
Added basic documentation of smartSleep functionality within HA based on receiving a heartbeat signal form MySensor node.

**Description:**

### DIFF
--- a/source/_components/mysensors.markdown
+++ b/source/_components/mysensors.markdown
@@ -178,5 +178,4 @@ Sending a heartbeat from the MySensors device to Home Assistant activates the Sm
 
 Visit the [library api] of MySensors for more information.
 
-[MySensors 1.x library api]: https://www.mysensors.org/download/sensor_api_15
-[MySensors 2.x library api]: https://www.mysensors.org/download/sensor_api_20
+[MySensors library api]: http://www.mysensors.org/download
--- a/source/_components/mysensors.markdown
+++ b/source/_components/mysensors.markdown
@@ -173,9 +173,7 @@ void receive(const MyMessage &message) {
 
 ### {% linkable_title Heartbeats %}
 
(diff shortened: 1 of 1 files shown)


source/_components/mysensors.markdown
```diff
@@ -171,6 +171,14 @@ void receive(const MyMessage &message) {
 }
 ```
 
+### {% linkable_title Heartbeats %}
+
+Sending a heartbeat from the MySensors sketch to home assistant activates the smartSleep functionality in home assistant. This means that messages are buffered and only sent to the device upon receiving a heartbeat from the device. State changes are stored so that only the last requested state change is sent to the device. Other types of messages are queued in a FIFO queue. SmartSleep is useful if you want to sleep a device, eg to be able to have it battery powered.  See the MySensors library API for information on how to send heartbeats and sleep device.
```
MartinHjelmare: * I think you should write: "Sending a heartbeat from the MySensors device to Home Assistant activates...".
* I think we write home assistant with capital H and A in the docs: Home Assistant. Please check that we're consistent with the rest of the docs.
* SmartSleep is specifically useful for actuator devices, that expects incoming messages. Sensors that only send messages do not really need to be "smartSleeping". Please clarify this.

source/_components/mysensors.markdown
```diff
@@ -171,6 +171,12 @@ void receive(const MyMessage &message) {
...
 
+### {% linkable_title Heartbeats %}
+
+Sending a heartbeat from the MySensors device to Home Assistant activates the SmartSleep functionality in Home Assistant. This means that messages are buffered and only sent to the device upon receiving a heartbeat from the device. State changes are stored so that only the last requested state change is sent to the device. Other types of messages are queued in a FIFO queue. SmartSleep is useful for battery powered actuators that are waiting for commands.  See the MySensors library API for information on how to send heartbeats and sleep device.  
+
+
 Visit the [library api] of MySensors for more information.
 
-[library api]: https://www.mysensors.org/download/sensor_api_15
+[MySensors 1.x library api]: https://www.mysensors.org/download/sensor_api_15
```
MartinHjelmare: You can remove this link cause the old page has been removed so link leads to a 404 page.

MartinHjelmare:
Thanks again!

PR #1747, opened 2017-01-10
Create packages

kellerza:
**Description:**
Documentation for configuration packages.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#5140


source/_topics/packages.markdown
```diff
@@ -0,0 +1,85 @@
...
+light:
+  - platform: rpi
+    ...
+```
+
+There are some rules for packages that will be merged:
+
+1. Component names may only use the basic form (e.g. `switch` and `switch 1` or `switch aa` is not accepted)
+2. Platform based components (`light`, `switch`, etc) can always be merged
+3. Components where entities are identified by a key that will represent the entity_id (`{key: config}`) need to have unique 'keys' between packages and the main configuration file (`input_boolean`, `group`)
```
MartinHjelmare: I think this needs an example of what not to do. It was hard to understand the first time I read this.

PR #1750, opened 2017-01-11
Add known issues for the pilight.send service

janLo:
**Description:**

As we don't implement the pilight api but only provide a transport to
the daemon, we should inform the user, that its their responsibility to
choose the correct data type.

As we convert all the users data to string if it specifies it in a
data_template without giving them an option to change it back, we
should put a warning, that templating will not work at all with the
send service of pilight.

I think, these warnings should be persisted in the documentation,
as we're not intended to fix the issues.

### DIFF
--- a/source/_components/pilight.markdown
+++ b/source/_components/pilight.markdown
@@ -53,6 +53,11 @@ pilight:
       - 42
 ```
 
+## {% linkable_title Known issues %}
+
+- You always have to use the correct types for service data for the `pilight.send` service, as all data is sent unchecked and unchanged to the daemon. If you choose the wrong type, your call gets rejected and all you get is a `insufficient number of arguments` in the pilight log. To retrieve the right argument type, read the [implementation of your protocol](https://github.com/pilight/pilight/tree/master/libs/pilight/protocols).
+- The `pilight.send` does **NOT WORK** if you specify the service data with `data_template` instead of `data`. It does not matter if you use an actual template or not. But choosing `data_template` converts all data to strings which gets rejected by the pilight daemon.
+
 ## {% linkable_title Troubleshooting %}
 
 - A list of tested RF transceiver hardware is available [here](https://wiki.pilight.org/doku.php/electronics). This might be useful before buying.


source/_components/pilight.markdown
```diff
@@ -53,6 +53,11 @@ pilight:
       - 42
 ```
 
+## {% linkable_title Known issues %}
+
+- You always have to use the correct types for service data for the `pilight.send` service, as all data is sent unchecked and unchanged to the daemon. If you choose the wrong type, your call gets rejected and all you get is a `insufficient number of arguments` in the pilight log. To retrieve the right argument type, read the [implementation of your protocol](https://github.com/pilight/pilight/tree/master/libs/pilight/protocols).
```
DavidLP: and all you get is --> remove _all_

DavidLP: is a `insufficient number of arguments` in the pilight log --> a `Pilight send failed` in the home assistant log

DavidLP: To retrieve the right argument type you can read the [implementation of your protocol](https://github.com/pilight/pilight/tree/master/libs/pilight/protocols) or use common sense: settings with numbers only are numbers and everything else are strings. For example `id: 123` and `uuid: '123-A'`

source/_components/pilight.markdown
```diff
@@ -53,6 +53,11 @@ pilight:
...
+- The `pilight.send` does **NOT WORK** if you specify the service data with `data_template` instead of `data`. It does not matter if you use an actual template or not. But choosing `data_template` converts all data to strings which gets rejected by the pilight daemon.
```
DavidLP: But choosing `data_template` converts all data to strings which gets rejected by the pilight daemon. --> Because `data_template` converts all data to strings which gets rejected by the pilight daemon.

source/_components/pilight.markdown
janLo: > is a insufficient number of arguments in the pilight log --> a Pilight send failed in the home assistant log
That's an generic error that does not point to the actual data type problem. The only hint that pilight got your data but rejected them is the log message in the pilight error log. It is necessary to point the user to the position where he can see that its not a connection problem but wrong argument types.

DavidLP: The message actually reads: `Pilight send failed for _message data_`. Check [here](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/pilight.py#L95).

janLo: It also says that if there was a died pilight daemon or any other problem. The only hint that point that there is something wrong with the actual _message data_ is the error in the pilight log. The log message in the hass log says only unspecific that _something_ went wrong with the service call.

So if the user encounters this error he can just check the pilight error log and knows that he probably hit that known issue. So it is the more specific way to check.

DavidLP:
> The log message in the hass log says only unspecific that something went wrong

No. It very specificly means that a certain command was not processed by the pilight c-library.

> the error in the pilight log

What and where is the _pilight log_?

andrey-git:
Are there any updates on this PR?

janLo:
The issues are still present. I'm indifferent about the wording. So if someone wants to change it, feel free. If you think that the users should not be bothered with this, then just close it.

Landrash:
@DavidLP Do you wan't this merged as is or do you want any changes?

Landrash:
Closing since PR has gone stale.
@DavidLP if you still want this merged then well reopen it.

janLo:
What exactly is the point in dismissing work as the maintainer does not respond? It really discuraged me to work on this project. Sorry.

Landrash:
Sorry, I was under the impression that DavidLP was the maintainer. 
Pr's are closed if they go stale. Since the last response was over a month ago I closed it but can just as easily open the PR again. 

Closing PR's when they go stale are done to make the project maintainable since fabaff and mostly me try to edit and review most of the Pr's for the documentation. When it comes to some component specific things we leave it to the folks with knowledge about the subject and in this case I assumed it was abandoned.

DavidLP:
I have no time right now for home assistant, due to heavy work load, sorry. Will come back at some point.

PR #1755, opened 2017-01-13
Use bots to get the chat_id

magnushacker:
**Description:**


michaelarnauts:
Shouldn't it make sense to also keep the old instructions to not depend on 3th party bots?

fabaff:
I would like to keep the different instructions. The way with a bot seems to be the easiest one but bots come and go...

magnushacker:
Makes sense, I can revert the change and only add the information about the
ID bot.

Maybe an even better way would be to let the component discover the chat ID
itself? Is there a specific reason for requiring the user to provide the
chat ID in the config file?

fabaff:
I picked the change.

PR #1769, opened 2017-01-14
Add instructions for installing node.js via nvm

armills:
**Description:**
Add instructions for how to install node.js. There currently isn't any mention of node in the frontend development docs, and `script/setup` fails without node installed.


armills:
Reopened against current branch #1770.

Issue #1768, opened 2017-01-14
Getting Started is unclear about configuring frontend


uSpike:
As a new HA user, I tried following instructions in the getting-started section.

There's a section titled "Password protecting the web interface", however nothing in the getting-started section mentions how to enable the web interface.  AFAIK, you must include `frontend:` in your `configuration.yaml` for the web interface to show up.  I think it would be helpful to include this on https://home-assistant.io/getting-started/basic/

Landrash:
Feel free to add what you feel is missing @uSpike. There's an edit button on top of each page that let's you submit a pull request.
Somethings you stop seeing after using Home Assistant for a while and those are easier to document for newcomers.

fabaff:
As far as I remember is the frontend  enabled by default. Anyway, we should address the observation by @uSpike with https://github.com/home-assistant/home-assistant.github.io/issues/1603.

uSpike:
@Landrash I created #1779 in case you don't want to handle it in #1603 

@fabaff if you run `hass --open-ui` as the minimal introduction, it creates `~/.homeassistant/configuration.yaml` which contains:

```yaml
# Enables the frontend
frontend:
```
I tested 0.35.3 and the dev branch, both do not expose a frontend without this.

uSpike:
I think there's also a lack of documentation for the `frontend` section.  It's a core component but there's not any documentation for it that I can see, and somewhat confusingly the `http` section controls some aspects of it.

balloob:
Frontend is enabled by default: https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/config.py#L52-L53

There is no need to further document it because there is no config for it - it represents the state of the devices and groups as configured throughout the rest of the config.

uSpike:
@balloob I'm sorry, but you're mistaken.  Maybe the frontend **should** be enabled by default, but it **is not** in 0.35.3 and the dev branch.

Simple test case:
* create directory `conf`
* create `conf/configuration.yaml` with contents:
```yaml
homeassistant:
  name: Home Test
  unit_system: imperial
  latitude: 32.87336
  longitude: 117.22743
  elevation: 100
  time_zone: America/Chicago

http:
  api_password: assword
```
* run `hass --open-ui -c conf` and the web browser opens to a page which says only `404: Not Found`

By adding `frontend:` to `conf/configuration.yaml`, the frontend now works as expected.  *This is not documented.*

[...]

PR #1779, opened 2017-01-15
Add "Enabling the web interface" section

uSpike:
See #1768

**Description:**
Add a very simple section in the getting started/basic configuration for how to enable the frontend, which apparently is not enabled by default in 0.35.3 and later.


balloob:
The frontend is enabled by default: https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/config.py#L52-L53

PR #1785, opened 2017-01-15
Update systemd unit file for virtualenv

asbach:
**Description:**
The current systemd unit file will provoke an error because the 'ExecPre' line calls a shell built-in function and all systemd Execs need to use an absolute path. The proposed change sets the python environment using the provided 'Environment' calls as used by systemd.

### DIFF
--- a/source/getting-started/autostart-systemd.markdown
+++ b/source/getting-started/autostart-systemd.markdown
@@ -45,7 +45,9 @@ After=network.target
 [Service]
 Type=simple
 User=homeassistant
-ExecStartPre=source /srv/homeassistant/homeassistant_venv/bin/activate
+#make sure the virtualenv python binary is used
+Environment=VIRTUAL_ENV="/srv/homeassistant/homeassistant_venv"
+Environment=PATH="$VIRTUAL_ENV/bin:$PATH"
 ExecStart=/srv/homeassistant/homeassistant_venv/bin/hass -c "/home/homeassistant/.homeassistant"
 
 [Install]


fabaff (approved):
Thanks :bird:

WeeBull:
These environment lines aren't necessary I believe. Any python executables installed inside a virtualenv have their python executable specifically defined.

For example, my hass executable looks like this:

```
#!/srv/homeassistant/homeassistant_venv/bin/python3

# -*- coding: utf-8 -*-
import re
import sys

from homeassistant.__main__ import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

Note the hashbang line.

andrey-git:
@asbach This doesn't work
Using `$` in `Environment` is not allowed.

The resulting path is `"$VIRTUAL_ENV/bin:$PATH"` (literally)

asbach:
@andrey-git I just reread the docs and as you already noticed, unit files do not support variable expansion in the `Environment=` directive.
Are there any objections to using the style as suggested by @WeeBull?

andrey-git:
I don't understand his proposal.

For me I just replaced "$VIRTUAL_ENV/bin:$PATH" by manually expanded values.

asbach:
Because the virtualenv already creates binaries with a hashbang pointing to the correct binary of python. Maybe one of the main Developers can jump in? Does home assistant depend on a correct `PATH` in relation to the virtual environment binaries?

PR #1804, opened 2017-01-17
Updating development environment setup guide

shortbloke:
**Description:**
Added more complete instructions for setting up a development environment for Home Assistant. Including details of dependencies, setting up venv. This is based on notes captured from setting up a dedicated hass dev environment on Debian 8.6.0

### DIFF
--- a/source/developers/development_environment.markdown
+++ b/source/developers/development_environment.markdown
@@ -62,7 +62,7 @@ $ script/setup
 
 * On Windows, you can use `python setup.py develop` instead of the setup script.
 
-* Run `hass` to invoke your local installation of the backend server or `hass --open-ui` to start the server and serve the web interface on http://localhost:8123
+* Run `hass` to invoke your local installation.
 
 ### {% linkable_title Logging %}
 
--- a/source/developers/development_environment.markdown
+++ b/source/developers/development_environment.markdown
@@ -33,12 +33,7 @@ $ sudo apt-get install python3-pip python3-dev libssl-dev
 Different distributions have different package installation mechanisms and sometimes packages names as well. For example Centos would use: `sudo yum install epel-release python34 python34-devel mysql-devel`
 </p>
 
-#### {% linkable_title Frontend dependencies (optional) %} 
-
-In order to run or develop the [Frontend](https://home-assistant.io/developers/frontend/) you will need node.js. The preferred method of installing node is [nvm](https://github.com/creationix/nvm). Install nvm using the instructions in the [README](https://github.com/creationix/nvm#install-script), and install node.js by running the following command:
-```bash
-$ nvm install node
-```
+Additional dependencies exist if you you plan to perform Frontend Development, please read the [Frontend](https://home-assistant.io/developers/frontend/) section to learn more.
 
 #### {% linkable_title Setting up virtual environment (optional) %} 
 
--- a/source/developers/development_environment.markdown
+++ b/source/developers/development_environment.markdown
@@ -29,6 +29,9 @@ Install the core dependencies.
 ```bash
 $ sudo apt-get install python3-pip python3-dev libssl-dev
 ```
+<p class='note'>
+Different distributions have different package installation mechanisms and sometimes packages names as well. For example Centos would use: `sudo yum install epel-release python34 python34-devel mysql-devel`
+</p>
 
 #### {% linkable_title Frontend dependencies (optional) %} 
 
@@ -51,7 +54,7 @@ $ source venv/bin/activate
 
 ### {% linkable_title Setup and Run %}
 
-* On Mac OS X and Linux 
+* On Mac OS X and Linux (remember to activate your virtual environment before running setup, if you are using one):
 ```bash
 $ cd home-assistant
 $ script/setup
--- a/source/developers/development_environment.markdown
+++ b/source/developers/development_environment.markdown
@@ -42,13 +42,18 @@ $ nvm install node
 
 #### {% linkable_title Setting up virtual environment (optional) %} 
 
-If you plan on providing isolation to your environment using [`venv`](https://docs.python.org/3.4/library/venv.html). You'll need to install the python3-venv package:
+If you plan on providing isolation to your environment using [`venv`](https://docs.python.org/3.4/library/venv.html). Within the `home-assistant` directory, create and activate your virtual environment.
+
+If using Python < 3.6 `pyvenv` is the recommended method of setting up virtual environments, you'll need to install the `python3-venv` package.
 ```bash
 $ sudo apt-get install python3-venv
-```
-Within the `home-assistant` directory, create and activate your virtual environment using the commands:
-```bash
 $ pyvenv venv
+$ source venv/bin/activate
+```
+
+If using Python 3.6 `pyvenv` has been deprecated:
+```bash
+$ python3 -m venv venv
 $ source venv/bin/activate
 ```
 
--- a/source/developers/development_environment.markdown
+++ b/source/developers/development_environment.markdown
@@ -30,7 +30,7 @@ Install the core dependencies.
 $ sudo apt-get install python3-pip python3-dev libssl-dev
 ```
 <p class='note'>
-Different distributions have different package installation mechanisms and sometimes packages names as well. For example Centos would use: `sudo yum install epel-release python34 python34-devel mysql-devel`
+Different distributions have different package installation mechanisms and sometimes packages names as well. For example Centos would use: `sudo yum install epel-release && sudo yum install python34 python34-devel mysql-devel`
 </p>
 
 Additional dependencies exist if you you plan to perform Frontend Development, please read the [Frontend](https://home-assistant.io/developers/frontend/) section to learn more.
--- a/source/_components/recorder.markdown
+++ b/source/_components/recorder.markdown
@@ -28,15 +28,64 @@ recorder:
 Configuration variables:
 
 - **purge_days** (*Optional*): Delete events and states older than x days.
+- **exclude** (*Optional*): Configure which components should be excluded from recordings.
+  - **entities** (*Optional*): The list of entity ids to be excluded from recordings.
+  - **domains** (*Optional*): The list of domains to be excluded from recordings.
+- **include** (*Optional*): Configure which components should be included in recordings. If set, all other entities will not be recorded.
+  - **entities** (*Optional*): The list of entity ids to be included from the history.
+  - **domains** (*Optional*): The list of domains to be included from the history.
 - **db_url** (*Optional*): The URL which point to your database. 
 
 
+Define domains and entities to `exclude` (aka. blacklist). This is convenient when you are basically happy with the information recorded, but just want to remove some entities or domains. Usually these are entities/domains which do not change (like `weblink`) or rarely change (`updater` or `automation`).
+
 ```yaml
-# Example configuration.yaml entry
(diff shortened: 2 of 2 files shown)


source/developers/development_environment.markdown
```diff
@@ -11,20 +11,55 @@ footer: true
...
+
+#### {% linkable_title Setting up virtual environment (optional) %} 
+
+If you plan on providing isolation to your environment using [`venv`](https://docs.python.org/3.4/library/venv.html). You'll need to install the python3-venv package:
+```bash
+$ sudo apt-get install python3-venv
+```
+Within the `home-assistant` directory, create and activate your virtual environment using the commands:
+```bash
+$ pyvenv venv
```
fabaff: At this point the repo is cloned. Why not just `pyvenv .`?

source/developers/development_environment.markdown
```diff
@@ -11,20 +11,55 @@ footer: true
...
+$ source venv/bin/activate
+```
+
+### {% linkable_title Setup and Run %}
+
+* On Mac OS X and Linux 
+```bash
+$ cd home-assistant
 $ script/setup
```
fabaff: `script/setup` should only be run if the venv is activated. I think that a little hint wouldn't harm.

source/developers/development_environment.markdown
shortbloke: By specifying venv, it places all the virtual environment in a separate directory. One which is configured in the .gitignore file to be excluded. Otherwise your home-assistant directory is populated/polluted by:
```bash
martinr@ha-dev:~/home-assistant-backup/venv$ ls -go
total 28
drwxr-xr-x 3 4096 Jan 18 11:50 bin
drwxr-xr-x 2 4096 Jan 18 11:50 include
drwxr-xr-x 5 4096 Jan 18 11:50 lib
lrwxrwxrwx 1   37 Jan 18 11:50 lib64 -> /home/martinr/home-assistant/venv/lib
-rw-r--r-- 1   60 Jan 18 11:50 pip-selfcheck.json
-rw-r--r-- 1   69 Jan 18 11:50 pyvenv.cfg
drwxr-xr-x 3 4096 Jan 18 11:50 selenium
drwxr-xr-x 4 4096 Jan 18 11:50 share
```
Does that make sense?

source/developers/development_environment.markdown
shortbloke: Setting up a venv is optional, so if someone isn't using venv this would still apply. Are you suggesting a reminder to activate the venv (which is covered earlier in the page), would be a helpful bit of extra reinforcement?

source/developers/development_environment.markdown
```diff
@@ -11,20 +11,55 @@ footer: true
...
+### {% linkable_title Prepare Your Environment %}
+
+#### {% linkable_title Core dependencies %} 
+
+Install the core dependencies.
+```bash
+$ sudo apt-get install python3-pip python3-dev libssl-dev
+```
+
+#### {% linkable_title Frontend dependencies (optional) %}
```
kellerza: Doesn't this belong on the dedicated frontend dev page?
https://home-assistant.io/developers/frontend/

source/developers/development_environment.markdown
```diff
@@ -11,20 +11,55 @@ footer: true
...
 $ git remote add upstream https://github.com/home-assistant/home-assistant.git
+```
+
+### {% linkable_title Prepare Your Environment %}
+
+#### {% linkable_title Core dependencies %} 
+
+Install the core dependencies.
+```bash
+$ sudo apt-get install python3-pip python3-dev libssl-dev
```
kellerza: And what about `yum install`? Not sure if we want to be too specific on package managers (and even lib names that can be different between distros)

source/developers/development_environment.markdown
shortbloke: The reason it's here, is that the page (before my changes) states you run script/setup which sets up both the server and the frontend. So whilst I may not be doing any front end development, it stops new developers from hitting errors on their initial environment setup. 

The alternative would be to make the setup script only setup the server unless you specify an option to also install the frontend. However that is a broader discussion that documenting what the current setup scripts do.

source/developers/development_environment.markdown
shortbloke: This is valid. However I was following existing documentation that only refers to apt-get. e.g:

- https://home-assistant.io/getting-started/installation-virtualenv/
- https://home-assistant.io/getting-started/installation-raspberry-pi/ - Though I accept most people would run raspbian, but they could run another ARM distro if they wanted.
- https://home-assistant.io/getting-started/troubleshooting/ 

In fact a google search for yum site:home-assisatnt.io returns pretty much nothing, wheras searching for apt-get finds a number of pages and community threads.

kellerza: 👍  lets keep `apt-get`

source/developers/development_environment.markdown
kellerza: You should be able to run `script/setup` at any point. Actually you should run `script/setup` after each update you pull in from the git repo as well

@home-assistant/frontend devs can confirm, but surely if you run `script/setup` without node and simply run it again when you add node it will result in the same

shortbloke: @kellerza If script/setup is needed/expected to run run whenever you do a pull, then it we could add a section on updating here, but perhaps that would better belong on the page: https://home-assistant.io/developers/development_catching_up/

You can run and re-run the setup script, I did ;-) Whilst I was working through different options for pulling Node into the venv. As part of another PR that @armills worked on.

The goal of this PR is to ideally set everything up for a new developer, if we can avoid them/me seeing errors when following the published instructions. A developer who knows enough and doesn't want to install node may be happy to see an error during setup and be able to understand the impact.

justweb1:
@kellerza 'script/build_frontend' updates the fronted without having to restart Hass.

source/developers/development_environment.markdown
armills: We can always turn this into a list of commands for different distros.

source/developers/development_environment.markdown
armills: It's true that `script/setup` calls `script/boostrap`, which calls `script/boostrap_frontend`, which requires node.

I think what we might want to do is modify `script/bootstrap` to first check `command -v npm`, and just skip boostrap frontend if it's not installed, while emitting a warning. Most people don't work on the frontend, and we shouldn't have them install frontend dependencies just so our script works.

source/developers/development_environment.markdown
kellerza: Might be a bit extreme, the other distro's are not that common :-)

But for reference, on Centos that would be:

```bash
sudo yum install epel-release 
sudo yum install python34 python34-devel mysql-devel
```

source/developers/development_environment.markdown
kellerza: Since we had an issue opened for this, I submitted a PR

source/developers/development_environment.markdown
kellerza: A hint would be good.

source/developers/development_environment.markdown
armills: Hey, you're talking to a Gentoo user, so watch out. 😄 

In any case if we wanted to start collecting the alternate distro commands, they should go in a secondary location so we're not cluttering up the guide.

shortbloke: Added a bit more info, with centos as an example in the latest commit.

source/developers/development_environment.markdown
shortbloke: Added hint in the latest submit.

source/developers/development_environment.markdown
balloob: Python will automatically pick up the venv if it is added to a directory. By adding it to a subdirectory you lose this functionality.

source/developers/development_environment.markdown
```diff
@@ -11,20 +11,58 @@ footer: true
...
+* On Mac OS X and Linux (remember to activate your virtual environment before running setup, if you are using one):
+```bash
+$ cd home-assistant
 $ script/setup
 ```
 
 * On Windows, you can use `python setup.py develop` instead of the setup script.
 
-* Run `hass` to invoke your local installation.
+* Run `hass` to invoke your local installation of the backend server or `hass --open-ui` to start the server and serve the web interface on http://localhost:8123
```
balloob: Please revert this change. The only thing that `--open-ui` does is that it opens a browser window. The web interface is being served depending on your config.

source/developers/development_environment.markdown
shortbloke: When you say automatically pick up the venv, are you suggesting you don't need to run the activate script? As just a quick test doesn't show this. regardless of if I setup the venv in a sub directory or at the top level, I still need to run activate in order for pip list or pip3 list to show only the modules installed in the virtual environment. 

Also it is recommended to install that the top level, then it may also need changes to .gitignore to help avoid parts of the venv being included in submissions.

I'm happy to make further changes, just need a definitive agreement, or I if you can point me at some public docs that talk about the reason for setting it up in the top level, perhaps we can include both options.

Note, in my latest submit I've also included the fact that pyvenv is being dreprecated and replaced by python3 -m venv.

source/developers/development_environment.markdown
```diff
@@ -11,14 +11,52 @@ footer: true
...
+If using Python < 3.6 `pyvenv` is the recommended method of setting up virtual environments, you'll need to install the `python3-venv` package.
+```bash
+$ sudo apt-get install python3-venv
+$ pyvenv venv
+$ source venv/bin/activate
+```
+
+If using Python 3.6 `pyvenv` has been deprecated:
+```bash
+$ python3 -m venv venv
```
kellerza: This syntax works perfectly fine in Python 3.4 on Linux as well and used on the 3.4 docs page as well (albeit only for Windows).

I guess `pyvenv` was a wrapper for `python -m venv`

source/developers/development_environment.markdown
```diff
@@ -11,14 +11,52 @@ footer: true
...
+### {% linkable_title Prepare Your Environment %}
+
+#### {% linkable_title Core dependencies %} 
+
+Install the core dependencies.
+```bash
+$ sudo apt-get install python3-pip python3-dev libssl-dev
+```
+<p class='note'>
+Different distributions have different package installation mechanisms and sometimes packages names as well. For example Centos would use: `sudo yum install epel-release python34 python34-devel mysql-devel`
```
shortbloke: @kellerza would: `sudo yum install epel-release && sudo yum install python34 python34-devel mysql-devel` work to keep it to a single line?

source/developers/development_environment.markdown
shortbloke: @kellerza I know they both work, I'm not positive it's as wrapper. In the 3.6 docs: https://docs.python.org/3/library/venv.html it sates

> Deprecated since version 3.6: pyvenv was the recommended tool for creating virtual environments for Python 3.3 and 3.4, and is deprecated in Python 3.6. Changed in version 3.5: The use of venv is now recommended for creating virtual environments.

As such I **assume** that using the new command is not recommended for < 3.6. So I'm not 100% sure that they are identical.

kellerza: It simply a wrapper in Centos (`cat $(command -v pyvenv)`), but no problem to keep as is

kellerza:
Thanks  🐬 !

PR #1805, opened 2017-01-17
Exaple code fixed

lee-js:
Example config caused hass to fail to start. Corrected and confirmed working.

**Description:**

### DIFF
--- a/source/_components/light.yeelight.markdown
+++ b/source/_components/light.yeelight.markdown
@@ -19,11 +19,11 @@ To enable those lights, add the following lines to your `configuration.yaml` fil
 ```yaml
 # Example configuration.yaml entry
 light:
-  platform: yeelight
+  - platform: yeelight
     devices:
       192.168.1.25:
         name: Front Door
-      192.168.1.13
+      192.168.1.13:
         name: Living Room
 ```
 


SilvrrGIT:
lee-js.  The existing entry as written will work.  I'm guessing you have multiple lights options (ie. hue and yeelight) which would require the hyphen.  @balloob maybe we need to come up with a standard for how these are presented.  I think if the hyphen is used with only a single component it will still work  (ill try to remember to test tonight), maybe all of the component pages should be updated to be consistent.

lee-js:
I only have yeelights, two in fact, which is why the code as included was
so fortuitous. With the code as is I got an error. I added a missing colon
and it still errored. It was only when I added both the dash and the colon
that the code worked and hass started.

SilvrrGIT:
Yeah, the missing colon is an issue, not sure how I missed that.  Your existing option you submitted should work so I don't see why it shouldn't get approved.   Im curious to know why after adding the colon it didn't work for you.  I don't have the hyphen and it works.  Try copy and pasting the example code into your yaml editor again.  Does it use tabs instead of spaces on the paste?  Some do and that will break the config file.  Other than that I have no idea.

lee-js:
Tried it without the - and I get the error "mapping values are not allowed" and a reference to the position of the colon after devices.

light:
  platform: yeelight
    devices:
      192.168.1.111:
        name: Office Lamp
      192.168.1.113:
        name: Living Room Lamp

lee-js:
Each of the lines is indented by 2 spaces in accordance with the sample code, not tabs. I checked them all. The code above deleted my spaces.

SilvrrGIT:
@lee-js  

Oye, I see another error in my config now.  devices shouldn't be indented.  When you added the hyphen it moved the platform over two spaces and provided the correct formatting.  Thanks for submitting the pull request to fix  my errors.

PR #1823, opened 2017-01-19
Update history.markdown

nodomain:
Elaborated the fact that the "recorder" component storage settings have an effect on the "history" component. It was unclear to me before. See https://community.home-assistant.io/t/history-with-mysql/10027 for details.

**Description:**

### DIFF
--- a/source/_components/history.markdown
+++ b/source/_components/history.markdown
@@ -13,7 +13,7 @@ ha_release: pre 0.7
 ---
 
 
-The `history` component will track everything that is going on within Home Assistant and allows the user to browse through it. It depends on the `recorder` component for storing the data. This means that if the `recorder` component is set up to use e.g. mySQL as data store, the `history` component does not use the default SQLite database to store data.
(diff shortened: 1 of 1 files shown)


source/_components/history.markdown
```diff
@@ -13,7 +13,7 @@ ha_release: pre 0.7
 ---
 
 
-The `history` component will track everything that is going on within Home Assistant and allows the user to browse through it.
+The `history` component will track everything that is going on within Home Assistant and allows the user to browse through it. It depends on the `recorder` component for storing the data. This means that if the `recorder` component is set up to use e.g. mySQL as data store, the `history` component does not use the default SQLite database to store data.
```
kellerza: Is it not better to simply state that history uses the same database setting as the recorder?

Also what is quite important is if you exclude any entities from being recorded, there will be no history from them

nodomain: True. You are right.

Landrash:
@nodomain Any update to address the changes suggested?

nodomain:
Done

kellerza:
Thanks!

PR #1848, opened 2017-01-22
Create oauth2 documenation

demonspork:
Adds documentation for modifying a pre-existing nginx configuration to allow oauth2 authentication rather than just password based authentication.

**Description:**

### DIFF
--- a/source/_ecosystem/oauth2.markdown
+++ b/source/_ecosystem/oauth2.markdown
@@ -1,0 +1,202 @@
+---
+layout: page
+title: "oauth2"
+description: "Documentation about setting up Home Assistant with an oauth2 Auth provider using nginx and bitly/oauth2_proxy."
+release_date: 2017-1-22 12:00:00 -0600
+sidebar: true
+comments: true
+sharing: true
+footer: true
+---
+
+Once you have [NGINX](https://home-assistant.io/ecosystem/nginx/) set up with Home Assistant, your own domain, and an SSL Certificate, you can set up authentication to require sign in with an oauth2 authentication provider suchas Google, Github, Linkedin, and others using [bitly\oauth2_proxy](https://github.com/bitly/oauth2_proxy).
+
+This also allows 2-factor authentication when using an oauth2 providers that support it.
+
+These instructions will walk you through modifying your nginx configuration to support one of these providers by setting up your own oauth2_proxy and using the auth_request directive.
+
+### {% linkable_title 1. Pick your oauth provider and acquire the authentication keys. %}
+
+This is taken directly from the oauth2_proxy [Github Page](https://github.com/bitly/oauth2_proxy). For other providers, visit that page as it has more detailed instructions.
+
+For Google, the registration steps are:
+
+1. Create a new project: https://console.developers.google.com/project
+2. Choose the new project from the top right project dropdown (only if another project is selected)
+3. In the project Dashboard center pane, choose **"Enable and manage APIs"**
+4. In the left Nav pane, choose **"Credentials"**
+5. In the center pane, choose **"OAuth consent screen"** tab. Fill in **"Product name shown to users"** and hit save.
+6. In the center pane, choose **"Credentials"** tab.
+   * Open the **"New credentials"** drop down
+   * Choose **"OAuth client ID"**
+   * Choose **"Web application"**
+   * Application name is freeform, choose something appropriate
+   * Authorized JavaScript origins is your domain ex: `https://internal.yourcompany.com`
+   * Authorized redirect URIs is the location of oath2/callback ex: `https://internal.yourcompany.com/oauth2/callback`
+   * Choose **"Create"**
+4. Take note of the **Client ID** and **Client Secret**
+
+### {% linkable_title 2. Install oauth2_proxy %}
+
+I am going to take the cheap route and send you directly to the [Github Page](https://github.com/bitly/oauth2_proxy) for oauth2_proxy for general installation of the product.
+
+If you are using docker, you can use this command to run it and reference the configuration files we create in the next step:
+
+```
+ docker run -d --name="oauth2_proxy" --net="bridge" -e -p 8095:8095/tcp -v "/path/to/appdata/oauth2_proxy":"/config":rw bitly/oauth2_proxy -config="/config/oauth2_proxy.cfg"
+```
+
+### {% linkable_title 3. Configure oauth2_proxy %}
+
+We are going to create 2 files, "oauth2_proxy.cfg" and "auth_emails", which will contain a list of email addresses that will be allowed in.
+
+In oauth2_proxy.cfg, the biggest non-standard thing we are going to do is to proxy it to file:///dev/null so that it doesn't attempt to return anything other than the 200 OK response to nginx because oauth2_proxy is only being used to auth, not to proxy.
+
+```
+http_address = "0.0.0.0:8095"
+upstreams = [
+    "file:///dev/null"
+ ]
+client_id = "Client_ID"
+client_secret = "Client_Secret"
+authenticated_emails_file = "/config/auth_emails"
+cookie_name = "_home_assistant"
+cookie_secret = "4C4F1B247E837B62"
+cookie_domain = "example.com"
+cookie_expire = "72h"
+cookie_refresh = "4h"
+cookie_secure = false
+```
+
+### {% linkable_title 4. Edit your nginx configuration %}
+
+We need to add several things to the nginx configuration. If you are currently using basic authentication with nginx, you probably want to disable them at this point or you are going to have several separate layers to sign in.
+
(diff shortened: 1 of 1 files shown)


source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
+---
+layout: page
+title: "Oauth2"
+description: "Documentation about setting up Home Assistant with an oauth2 Auth provider using nginx and bitly/oauth2_proxy."
+release_date: 2017-1-22 12:00:00 -0600
```
robbiet480: The month must have a trailing `0`

source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
...
+sidebar: true
+comments: true
+sharing: true
+footer: true
+---
+
+Once you have [NGINX](https://home-assistant.io/ecosystem/nginx/) set up with Home Assistant, your own domain, and an SSL Certificate, you can set up authentication to require sign in with an oauth2 authentication provider suchas Google, Github, Linkedin, and others using [bitly\oauth2_proxy](https://github.com/bitly/oauth2_proxy).
```
robbiet480: `bitly\oauth2_proxy` should be `bitly/oauth2_proxy`

source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
...
+
+These instructions will walk you through modifying your nginx configuration to support one of these providers by setting up your own oauth2_proxy and using the auth_request directive.
+
+### {% linkable_title 1. Pick your oauth provider and acquire the authentication keys. %}
+
+This is taken directly from the oauth2_proxy [Github Page](https://github.com/bitly/oauth2_proxy). For other providers, visit that page as it has more detailed instructions.
+
+For Google, the registration steps are:
+
+1. Create a new project: https://console.developers.google.com/project
```
robbiet480: Link this URL

source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
...
+4. Take note of the **Client ID** and **Client Secret**
+
+### {% linkable_title 2. Install oauth2_proxy %}
+
+I am going to take the cheap route and send you directly to the [Github Page](https://github.com/bitly/oauth2_proxy) for oauth2_proxy for general installation of the product.
+
+If you are using docker, you can use this command to run it and reference the configuration files we create in the next step:
+
+```
+ docker run -d --name="oauth2_proxy" --net="bridge" -e -p 8095:8095/tcp -v "/path/to/appdata/oauth2_proxy":"/config":rw bitly/oauth2_proxy -config="/config/oauth2_proxy.cfg"
```
robbiet480: Remove the extra space before `docker`.

source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
...
+```
+
+### {% linkable_title 3. Configure oauth2_proxy %}
+
+We are going to create 2 files, "oauth2_proxy.cfg" and "auth_emails", which will contain a list of email addresses that will be allowed in.
+
+In oauth2_proxy.cfg, the biggest non-standard thing we are going to do is to proxy it to file:///dev/null so that it doesn't attempt to return anything other than the 200 OK response to nginx because oauth2_proxy is only being used to auth, not to proxy.
```
robbiet480: For file names and paths we usually like to make it look like code. You can do this by surrounding a string with backticks (`) i.e. file:///dev/null should instead be `file:///dev/null`. May want to do the same for `oauth2_proxy.cfg` etc.

source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
...
+### {% linkable_title Oauth2 Locations %}
+
+We need to create locations to proxy the /oauth2 commands through nginx when they are requested.
+```
+  location = /oauth2/auth {
+      internal;
+      proxy_pass http://10.10.10.200:8095;
+      proxy_set_header Host $host;
+      proxy_pass_request_body off;
+      proxy_set_header Content-Length "";
```
robbiet480: Why are we emptying the `Content-Length` header?

source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  location = /oauth2/start {
+      internal;
+      proxy_pass http://10.10.10.200:8095;
+      proxy_set_header Host $host;
+      proxy_pass_request_body off;
+      proxy_set_header Content-Length "";
```
robbiet480: Same question as right above this, why are we emptying `Content-Length`?

source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
...
+  location = /oauth2/callback {
+      auth_request off;
+      proxy_pass http://10.10.10.200:8095;
+      proxy_set_header Host $host;
+  }
+```  
+
+### {% linkable_title Home Assistant API %}
+
+Most external uses of the Home Assistant API are not going to be able to authenticate through the oauth2 layer. To get around this, we need to configure nginx to recognize the Home Assistant api_password and skip the oauth layer if that happens.
```
robbiet480: Backticks around `api_password`

source/_ecosystem/oauth2.markdown
```diff
@@ -0,0 +1,202 @@
...
+      proxy_set_header Host $host;
+      proxy_http_version 1.1;
+      proxy_set_header Upgrade $http_upgrade;
+      proxy_set_header Connection "upgrade";
+  }
+```
+
+### {% linkable_title notify.html5 %}
+
+If you are using the notify.html5, you have to add 2 special locations before the /api location. The /api/notify.html5/callback uses its own authentication, and is safe to skip the oauth2_proxy step, and /manifest.json needs to be available with no authentication.
```
robbiet480: > the notify.html5 **platform**

(add "platform")

Backticks around `/api/notify.html5/callback` too please, as well as `/manifest.json`, `/api` `oauth2_proxy`.

source/_ecosystem/oauth2.markdown
robbiet480: This comment is still valid, GitHub just hid it because you pushed a change after I left it.

source/_ecosystem/oauth2.markdown
robbiet480: This comment is also still valid.

robbiet480:
You're a mindreader! You fixed all my backtick requests before I actually requested it! :cookie:

source/_ecosystem/oauth2.markdown
demonspork: Because the content will be empty. oauth2_proxy has an upstream of `file:///dev/null`

robbiet480: Ah, good point. Okay!

demonspork:
Ok, there is a small issue with this setup that I am hoping someone can help me solve.
Scenario:
Home Assistant is loaded in my browser, I successfully complete the oauth2 authentication, and I gain access. I do not have api_password set in the configuration.yaml, my only auth is the oauth2 layer described in this PR.
After 72 hours (as currently configured) the cookie expires for the oauth2 layer.
Home-Assistant Web interface detects that it can no longer access the server, and loads the Password dialog that you would usually use to sign in when using api_password in configuration.yaml rather than following the redirect that is being sent by nginx to sign in with oauth2.
To get past this, you have to use ctrl+refresh to get past the cache'd Home Assistant Web Interface that is only showing the password dialog, then you can complete the oauth2 authentication and regain access.

This is similar to an issue I had where I tried to add other locations such as `location /ozwcp` to my nginx configuration but would always get the Home Assistant login dialog without the ctrl+refresh because Home Assistant seemed to do some aggressive caching in the browser that would cause anything on the same domain try to load part of Home Assistant. I ended up having to move Home Assistant to its own subdomain.

What would need to be changed (I assume in Polymer?) to resolve this? Or do you know of a way to prevent this from occurring without having to modify anything in Home-Assistant itself, maybe a setting in nginx?

demonspork:
It appears there is already an issue in polymer repo about this. https://github.com/home-assistant/home-assistant-polymer/issues/110

Any thoughts?

source/_ecosystem/oauth2.markdown
demonspork: I was actually wrong. The /oauth2/auth request is special, and wouldn't return anything anyway. the `/dev/null` proxy is just housekeeping because it requires _something_ to be put in the upstreams.
Write value, wrong reason.

robbiet480:
Going to close this until we come to a resolution on the Polymer issues, and because this PR is blocking creation of a new PR on `current`.

PR #1850, opened 2017-01-22
Added supportability matrix link from the github page of python-amcrest project

tchellomello:
**Description:**

Added supportability matrix link from the github page of python-amcrest project .

Should be a quick reference to users to identify if their camera model as been tested on the backend python module `amcrest`.


colinodell:
:+1: looks good to me

PR #1851, opened 2017-01-22
Fixed menu to contain also iOS.

iBobik:
Because article it also about iOS.

**Description:**
Article is about adding to homescreen on both mobile OS, so also menu title should mention both OS.

How about also change URL? Currently it is /getting-started/android/ but more descriptive will be /getting-started/mobile-homescreen/. Could I change it? Also it will be good to add jekyll-redirect and use it there.


source/_includes/asides/getting_started_navigation.html
```diff
@@ -67,7 +67,7 @@ <h1 class="title delta">Getting Started Guide</h1>
       </li>
       <li>{% active_link {{site_root}}/topics/ Advanced Topics %}</li>
       <li>{% active_link {{site_root}}/details/ Details %}</li>
-      <li>{% active_link /getting-started/android/ Add to Android Homescreen %}</li>
+      <li>{% active_link /getting-started/android/ Add to Android/iOS Homescreen %}</li>
```
Landrash: This is a bit misleading since the page it's linking to is specifically called Android.

iBobik:
@Landrash [This page](https://home-assistant.io/getting-started/android/) describes the same feature on both iOS and Android. Scroll down :-)

Landrash:
Sorry, I was a bit unclear. 
Instructions are for both Android and IOS on the page but the page file is called Android.
A better solution would be to rename the page and check that there's no link in the documentation to the old link. Taking the shortest path like this patch would do only leads to more cleanup in the end for the maintainers.

iBobik:
Agree. And can I also add redirect from old URL?

Landrash:
I would actually just replace the `android` one with for example `mobile` or similar.
Redirects should be needed if we just check for old links. Worst case someone gets a 404 and have to go looking again.

iBobik:
I think somebody getting 404 is not acceptable. :-) Also search engines gives worse rankings if they see 404 errors.

I have very good experience with jekyll-redirect plugin on other sites and will be pleased to set it up.

Landrash:
I'm all for this after we add on the redirect.
Can you build that on this pull request?

Landrash:
@fabaff This adds a bit of complexity to the current .io page.
Are you ok with this since it's also adds a new feature that might be useful for when deprecating old components that are replaced?

PR #1853, opened 2017-01-22
Fix componentes/camera link

adrianlzt:


Landrash:
Thanks for the fix! :cake:

PR #1858, opened 2017-01-23
Clarify zwave refresh delay configurations

armills:
**Description:**
This helps clarify the description of the refresh delay configuration options. Currently it sounds like `refresh_value` would be an integer.


armills:
Ah, good catch. :)

PR #1873, opened 2017-01-25
Revert "correct value_template format for command_state"

fabaff:
Reverts home-assistant/home-assistant.github.io#1868


n8henrie:
@fabaff -- I was trying to figure out how to "review" this last night (haven't done a review on GH before). It didn't seem quite right, but I couldn't figure out why not -- is this `return_value` behavior documented anywhere? If not should I add something about it? I didn't see any mention in the templating docs.

PR #1883, opened 2017-01-26
Necessary quotation in Matrix room id or alias

janwh:
**Description:**

[YAML syntax](https://symfony.com/doc/current/components/yaml/yaml_format.html) suggests that strings containing `!` and `#` should be wrapped in single or double quotes. Otherwise it will be recognised as an inline comment.


mweinelt:
Maybe call it 

> room id or alias

Otherwise lgtm.

PR #1884, opened 2017-01-26
Update mqtt.markdown

JammyDodger231:
The main point I couldn't find any documentation to instruct how to use mosquitto, letsencrypt and home assistant and this was my main stumbling block, hopefully I can save someone else some time :smile: 

After having my own MQTT server running on a VPS that is hosted elsewhere from my home assistant server I found that after using lets encrypt it would not allow me to use the .pem files created with home assistant. They worked natively with mosquitto_sub and owntracks but this is the only method I found to work. The change should help anyone else who can't get a secure connection and save hours of searching. :grimacing: 

**Description:**

### DIFF
--- a/source/_components/mqtt.markdown
+++ b/source/_components/mqtt.markdown
@@ -119,6 +119,10 @@ mqtt:
 There is an issue with the Mosquitto package included in Ubuntu 14.04 LTS. Specify `protocol: 3.1` in your MQTT configuration to work around this issue.
 </p>
 
+<p class='note'>
+If you are running a mosquitto instance on a different server with proper SSL encryption using a service like letsencrypt you may have to set the certificate to the operating systems own `.crt` certificates file. In the instance of ubuntu this would be `certificate: /etc/ssl/certs/ca-certificates.crt`
+</p>
+
 ### {% linkable_title Public broker %}
 
 The Mosquitto project runs a [public broker](http://test.mosquitto.org). This is the easiest to set up, but there is no privacy as all messages are public. Use this only for testing purposes and not for real tracking of your devices.


source/_components/mqtt.markdown
```diff
@@ -119,6 +119,10 @@ mqtt:
 There is an issue with the Mosquitto package included in Ubuntu 14.04 LTS. Specify `protocol: 3.1` in your MQTT configuration to work around this issue.
 </p>
 
+<p class='note'>
+If you are running a mosquitto instance on a different server with proper SSL encryption using a service like letsencrypt you may have to set the certificate to the operating systems own `.crt` certificates file. In the instance of ubuntu this would be `certificate: /etc/ssl/certs/ca-certificates.crt`
```
JammyDodger231: Thanks for the feedback :slightly_smiling_face: the only thing related I can find is the this [superuser](http://superuser.com/questions/437330/how-do-you-add-a-certificate-authority-ca-to-ubuntu) link, I don't really see anyone else using a dedicated external server for their MQTT broker and a dedicated internal server at home that isn't a raspberry pi, so I am struggling to find anything :disappointed:

JammyDodger231:
I agree that it could use some more clarification in the form of references but I cannot seem to find anything regarding what I am doing besides the stuff that I have just tried myself

source/_components/mqtt.markdown
AlucardZero: In MQTT, `certificate` should point to a file containing the root of the certificate chain that signed the certificate you are using for SSL with Mosquitto.  Mine points at `identrust-root-and-lets-encrypt-x3.pem`, but if that same certificate is in Ubuntu's system store, pointing `certificate` at it will work and is probably better.

I contributed #962 that clarified what `certificate` is used for after I traced the code to see what exactly it does. The initialization code is https://github.com/home-assistant/home-assistant/blob/4692ea85b7e53ab21cefc9bfe4b73fc436d5238d/homeassistant/components/mqtt/__init__.py#L334 and the underlying documentation is https://pypi.python.org/pypi/paho-mqtt/1.1 (search for tls_set)

```
ca_certs
    a string path to the Certificate Authority certificate files that are to be treated as trusted by this client. If this is the only option given then the client will operate in a similar manner to a web browser. That is to say it will require the broker to have a certificate signed by the Certificate Authorities in ca_certs
```

I'm assuming Jammy has tested this and that setting `certificate` this way made it work better for him.  If that's true, I approve.

JammyDodger231: Thanks Alucard, I had tested multiple different variances in configuration, like passing the .pem certificate directly and nothing worked. This was the only method I found that would work on ubuntu 16.10 :+1:

Landrash: @JammyDodger231  assuming this follows @AlucardZero 's assumptions I'll merge this.

JammyDodger231: @Landrash It does :slightly_smiling_face: he knows more but what he wrote is accurate :+1:

Landrash:
Good feedback all around! Thanks 🎂

PR #1885, opened 2017-01-26
Added documentation for notify.twilio_call component

fakezeta:
**Description:**
Documentation for the twilio notification platform enables sending notifications via Voice Phone Call, powered by Twilio (https://twilio.com). Passed message will be read by Text-To-Speech service provided by Twilio.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#5580

### DIFF
--- a/source/_components/notify.twilio_call.markdown
+++ b/source/_components/notify.twilio_call.markdown
@@ -1,0 +1,56 @@
+yout: page
+title: "Twilio Call"
+description: "Instructions how to add user notifications to Home Assistant."
+date: 2017-01-27 00:09
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: twilio.png
+ha_category: Notifications
+ha_release: "0.36.1"
+---
+
+The `twilio` notification platform enables sending notifications via Voice, powered by [Twilio](https://twilio.com).
+Passed message will be read by Text-To-Speech service.
+
+This component is just an adaptation from the Twilio SMS notification platform and won't exist without it.
+
+To use this notification platform in your installation, add the following to your `configuration.yaml` file:
+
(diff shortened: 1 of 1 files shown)


danielperna84:
Just had a glimpse at the twilio site. There's something about a free API key, but they also have pricings (which makes sense for something like that). Hence I guess it would be a good addition for the documentation to mention possible costs that would be tied to the service.

robbiet480:
@danielperna84 I think most people will realize that pretty quick when you go to sign up for an account and it says you can't make calls without entering a credit card no?

danielperna84:
For the Amazon Polly TTS that was requested as well. Although I admit, that this case indeed may be different if it's not working for free at all. As stated above, from just looking at the site it seems like there's some free API (like for a lot of other services), but  the calling probably isn't included in that.
At least for me mentioning the financial aspects in the documentation would save me the time to examine what pricing models twilio offers. Could be just me, but I generally prefer more "verbose" documentation that tells me mostly everything I need to know instead of just point me into the right direction. :D

fakezeta:
Hi @danielperna84,

I added some information about the free trial limitation even if I agree with @robbiet480.
Describing the service too in detail means monitoring the offer and update the documentation accordingly.

my 2 cents

homeassistant:
Hi @fakezeta, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.github.io#1885).

Thanks!

fakezeta:
Done!

fabaff (approved):
Thanks :bird:

PR #1891, opened 2017-01-27
explain value extraction via template

felixfischer:
Getting values from a JSON variable via `state_format` did not work for me, but using `value_template` worked. I added some explanation on how to use it.

**Description:**

### DIFF
--- a/source/_components/mqtt.markdown
+++ b/source/_components/mqtt.markdown
@@ -280,7 +280,7 @@ It is also possible to extract JSON values by using a value template:
 ```yaml
 switch:
   platform: mqtt
-  value_template: '{{ value_json.somekey[0].value }}'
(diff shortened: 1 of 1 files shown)


source/_components/mqtt.markdown
```diff
@@ -275,6 +275,13 @@ switch:
   platform: mqtt
   state_format: 'json:somekey[0].value'
 ```
+It is also possible to extract JSON values by using a value template:
+
+```yaml
+switch:
+  platform: mqtt
+  value_template: '{{ value_json.somekey[0].value }}'
```
fabaff: Templates must be wrapped with `{% raw %}...{% endraw %}` otherwise they will be rendered and empty.

PR #1921, opened 2017-01-30
Update recorder.markdown

nodomain:
Updated to include the "native" mysqlclient package as documented in https://community.home-assistant.io/t/error-in-setting-up-recorder-with-mysql/7705/4 - verified that it works as well ;-)

**Description:**


nodomain:
You are right - obviously the hint regarding virtual env was missing which made it a pain for me ;)

DavidLP:
We could think about adding this as an additional option to the one before. Since installing the `mysqlclient` under windows can be troublesome.

Hoobie7:
For what it's worth I could never get the pymysql working reliably but mysqlclient seems to be working well for the 24 hours it's been up.

PR #1925, opened 2017-01-31
docs(docker-install): Add OSX port forwarding note

iDVB:
- Add note to people on OS X about how to properly forward ports so that you can browse to http://localhost:8123
- This tripped me and a co-worker up a fair bit so I'm hoping it will save others a good deal of googling.

### DIFF
--- a/source/getting-started/installation-docker.markdown
+++ b/source/getting-started/installation-docker.markdown
@@ -19,6 +19,8 @@ This will launch Home Assistant and serve the web interface from port 8123 on yo
 
 <p class='note'>
 When using boot2docker on OS X you are unable to map the local time to your Docker container. Replace `-v /etc/localtime:/etc/localtime:ro` with `-e "TZ=America/Los_Angeles"` (replacing America/Los_Angeles with [your timezone](http://en.wikipedia.org/wiki/List_of_tz_database_time_zones))
+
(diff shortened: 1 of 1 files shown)


source/getting-started/installation-docker.markdown
```diff
@@ -19,6 +19,8 @@ This will launch Home Assistant and serve the web interface from port 8123 on yo
 
 <p class='note'>
 When using boot2docker on OS X you are unable to map the local time to your Docker container. Replace `-v /etc/localtime:/etc/localtime:ro` with `-e "TZ=America/Los_Angeles"` (replacing America/Los_Angeles with [your timezone](http://en.wikipedia.org/wiki/List_of_tz_database_time_zones))
+
+Additionally, if your expectation is that you will be able to browse directly to `http://localhost:8123` on your OS X host, then you will also need to replace the `--net=host` switch with `-p 8123:8123`. This is currently the only way to forward ports on to your actual host(OS X) machine instead of the virtual machine inside xhyve. More detail on this can be found in [the docker forums](https://forums.docker.com/t/should-docker-run-net-host-work/14215/10).
 </p>
```
fabaff: Not an Apple user here. If there is no difference between `OS X` and `macOS` installations then we should use `macOS`. Like in the other docs.

PR #1936, opened 2017-02-01
Updated example to show how to set default color

Joeboyc2:
**Description:**


fabaff:
We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

PR #1978, opened 2017-02-05
Fix incorrect indention for sonos_say script

postlund:
**Description:**
Could not just copy and paste the script without getting an error during start of hass. This fixes that

### DIFF
--- a/source/_cookbook/sonos_say.markdown
+++ b/source/_cookbook/sonos_say.markdown
@@ -21,22 +21,22 @@ script:
     sequence:
      - service: media_player.sonos_snapshot
        data_template:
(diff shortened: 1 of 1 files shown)


source/_cookbook/sonos_say.markdown
```diff
@@ -21,22 +21,22 @@ script:
     sequence:
      - service: media_player.sonos_snapshot
        data_template:
-         entity_id: {% raw %}"{{ sonos_entity }}"{% endraw %}
```
fabaff: Any reason why you removed the `{% raw %}...{% endraw %}`?

PR #1995, opened 2017-02-07
iOS doc improvements

rbflurry:
**Description:**
This updates the iOS app pages to remove the need to load the notify component separately.


robbiet480:
Thanks! 🐬 🍪 💯

PR #2019, opened 2017-02-12
Split MQTT documentation

fabaff:
**Description:**
The MQTT documentation was stored and delivered as one page. The sections are now divided into single pages.


fabaff:
New PR for the larger re-organisation: #2055

PR #2037, opened 2017-02-14
Add wait_template for 0.39

pvizeli:
**Description:**

Add description for `wait_template` in script

### DIFF
--- a/source/getting-started/scripts.markdown
+++ b/source/getting-started/scripts.markdown
@@ -80,18 +80,13 @@ Wait until some things are complete. We support at the moment `wait_template` fo
 
 ```yaml
 # wait until media player have stop the playing
-wait_template: "{{ states.media_player.floor.states == 'stop' }}"
+wait_template: {% raw %}"{{ states.media_player.floor.states == 'stop' }}"{% endraw %}
 ```
 
 ```yaml
 # wait until a valve is < 10 or abort after 1 minutes.
-wait_template: "{{ states.climate.kitchen.attributes.valve < 10 }}"
-timeout: 00:10:00
-```
-
-```yaml
-# alternative
-wait_template: "{% if"
+wait_template: {% raw %}"{{ states.climate.kitchen.attributes.valve < 10 }}"{% endraw %}
+timeout: 00:01:00
 ```
 
 ### {% linkable_title Fire an Event %}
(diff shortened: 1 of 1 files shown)


source/getting-started/scripts.markdown
```diff
@@ -74,6 +74,25 @@ delay:
...
+
+```yaml
+# wait until a valve is < 10 or abort after 1 minutes.
+wait_template: "{{ states.climate.kitchen.attributes.valve < 10 }}"
+timeout: 00:10:00
+```
+
+```yaml
+# alternative
+wait_template: "{% if"
```
arsaboo: This is incomplete...

source/getting-started/scripts.markdown
```diff
@@ -74,6 +74,25 @@ delay:
...
+
+```yaml
+# wait until media player have stop the playing
+wait_template: "{{ states.media_player.floor.states == 'stop' }}"
+```
+
+```yaml
+# wait until a valve is < 10 or abort after 1 minutes.
+wait_template: "{{ states.climate.kitchen.attributes.valve < 10 }}"
+timeout: 00:10:00
```
arsaboo: Shouldn't it be `timeout: 00:01:00` for aborting after 1 minute?

source/getting-started/scripts.markdown
```diff
@@ -74,6 +74,25 @@ delay:
 # Valid formats include HH:MM and HH:MM:SS
 delay: {% raw %}'00:{{ states.input_slider.minute_delay.state | int }}:00'{% endraw %}
 ```
+### {% linkable_title Wait %}
+
+Wait until some things are complete. We support at the moment `wait_template` for waiting until a condition is `true`, see also on [Template-Trigger](/getting-started/automation-trigger/#template-trigger). It is possible to set a timeout after that will the script abort his execution. Timeout have same syntax as `delay`.
```
arsaboo: Unlike `delay` where you need to specify the wait time, `wait` uses a template to determine the wait time. You can pause the execution of the script until the condition is satisfied. At the moment, we support `wait_template` for waiting until a template condition is `true`, see also on [Template-Trigger](/getting-started/automation-trigger/#template-trigger). It is possible to set a timeout after which the script will abort execution. Timeout have the same syntax as `delay`.

(quick question - will `timeout` abort the entire script as suggested above or move to the next instruction)

pvizeli: Yeah it abort the hole script.

PR #2038, opened 2017-02-14
Add note about using trusted networks with reverse proxy.

Instagraeme:
**Description:**
Add cautionary note about using trusted networks with reverse proxy.


mweinelt:
The `x_forwarded_for` option basically already has that warning.

andrey-git:
I think the warning is still useful on each specific option.

PR #2009, opened 2017-02-11
Update to components/sensor.sensehat

nodinosaur:
• Have provided more information on how to install the SenseHAT for Virtual Environment installs
• Have included directions on making the SenseHAT data prettier

**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#<home-assistant PR number goes here>

PR #2051, opened 2017-02-15
Added documentation for sense hat new flag (is_hat_attached)

nodinosaur:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#<5883>

### DIFF
--- a/source/_components/sensor.miflora.markdown
+++ b/source/_components/sensor.miflora.markdown
@@ -19,9 +19,9 @@ To use your Mi Flora plant sensor in your installation, add the following to you
 
 ```yaml
 # Example configuration.yaml entry
-sensor
+sensor:
   - platform: miflora
-    mac: xx:xx:xx:xx:xx:xx
+    mac: "xx:xx:xx:xx:xx:xx"
     monitored_conditions:
       - temperature
 ```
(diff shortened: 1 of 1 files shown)


source/_components/sensor.sensehat.markdown
```diff
@@ -34,5 +35,119 @@ Configuration variables:
   - 'temperature'
   - 'humidity'
   - 'pressure'
+  is_hat_attached (Optional): True|False boolean; True declares the SenseHAT _is_ physically on the Raspberry Pi
+  
 
+#### Customizing the Sense HAT data
```
fabaff: You already added the details with #2009. Here only the new option should be included.

nodinosaur:
Yes, I know, but I also wrote back that the page was completely broken. That a section was missing and the order was wrong.
This is how the page should be rendered.
You're choice if you want to embed this one.

source/_components/sensor.sensehat.markdown
nodinosaur: Yes, I know, but I also wrote back that the page was completely broken. That a section was missing and the order was wrong.
This is how the page should be rendered.
You're choice if you want to embed this one.

nodinosaur: I should add, that that happened with the merger, not the way I wrote it. To correct it all, this version should be used. (_or let me fix / merge_)

nodinosaur: If you compare the original PR #2009 you will see that the text is different to what is on [/sensor.sensehat](https://home-assistant.io/components/sensor.sensehat/) So you will be able to judge for yourself that the merger failed somehow.
So this PR is essentially the same as #2009 with the addition of the new flag, and description.

PR #2052, opened 2017-02-15
Add documentation for LLAP Sensor

leppa:
**Description:**

Documentation for the submitted LLAP Sensor platform.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#6025


source/_components/sensor.llap.markdown
```diff
@@ -0,0 +1,86 @@
...
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: llap
+    sensors:
+      - name: "Room Temperature"
+        id: RT
+        prefix: "TEMP"
+        unit_of_measurement: "°C"
```
fabaff: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need.

There is a full sample below.

source/_components/sensor.llap.markdown
```diff
@@ -0,0 +1,86 @@
...
+comments: false
+sharing: true
+footer: true
+ha_category: Sensor
+ha_release: 0.39
+ha_iot_class: "Local Push"
+logo: wirelessthings.png
+---
+
+The `llap` platform supports wireless sensors which were produced by now defunct Ciseco / Wireless
```
fabaff: There are no line breaks needed. The renderer will take care about this. No need to change, I'm just telling

source/_components/sensor.llap.markdown
```diff
@@ -0,0 +1,86 @@
...
+    sensors:
+      - name: "Room Temperature"
+        id: RT
+        prefix: "TEMP"
+        unit_of_measurement: "°C"
+```
+
+Configuration variables:
+
+- **port** (*Optional*): Port on which to listen for UDP datagrams from the MessageBridge. Default
```
fabaff: In the intro you use `Message Bridge`

source/_components/sensor.llap.markdown
leppa: I usually put line breaks so that source file is also readable in plain text editors.

source/_components/sensor.llap.markdown
leppa: I can remove `unit_of_measurement`, but I wouldn't remove `prefix`. The reason is that in 99% of cases users will put some prefix, because except for the actual value the sensor also sends status messages (e.g., `AWAKE`, `BATTLOW`, etc.). It's optional only because someone might want to capture all messages from the sensor and process them by themselves (e.g., in template).

source/_components/sensor.llap.markdown
leppa: Yeah, missed the space. Will fix.

fabaff:
The Pull request in the main repository was closed. Please re-open this PR is your changes are ready to merge.

PR #2054, opened 2017-02-16
Fix the example camera url

tradiuz:
**Description:**


fabaff:
The examples works fine but it seems that it's necessary to note that the attributes have different names depending on the settings.

https://github.com/home-assistant/home-assistant.github.io/commit/558bc3607aa5f3e72043ef8afc576c0778d6630c

tradiuz:
Is there a reason to have two different attribute names depending on another variable?  That seems confusing.

Issue #1603, opened 2016-12-16
Clean up getting started


balloob:
Our getting started section has grown over time and the current structure is no longer a getting started guide but more a look at all the stuff you can do.

I think that we have to take a step back and see what is really important. Anything that is not the most important is distracting and should be moved away from the main menu/getting started. It looks too daunting now for new users.

So what is most important? Our Raspberry Pi image. We have 70% of our users using a Pi and that is awesome. I think the last time I checked 80% of them are on a Pi 3 so having it evolve around that will be even better.

So here are some improvements that I can see will improve our guide:
 - Limit getting started to the essentials:
   - Just the RPi image on a Pi 3
   - config 101 (but no advanced stuff like customize)
   - adding your first integrations
   - adding first automation. 
 - Create a Docs section and include it in top nav. This is where most of the getting started will move to.
 - Create special UI for docs. Currently our UI is not good for having long nav bars. Something like [this](https://www.polymer-project.org/1.0/start/) and [this](http://redux.js.org/) is beautiful. You can individually scroll the nav from the main content. I can see us having to implement a special layout that doesn't contain the normal header/sidebar just for docs sections.
 - Ecosystem is a great section, but let's move it to the documentation. We should also go through the getting started and find things that we should move to the ecosystem section. Like [autostart](https://home-assistant.io/getting-started/autostart/) and [Add to Android home screen](https://home-assistant.io/getting-started/android/), 
 - "Details" group and "advanced topics" group can stop existing on their own and also move to the docs header.

What do people think of this idea?

----

Extraction of items/tasks:

- [x] Rework "Getting started". Keep it focused on RPi and as simple as possible #2055
- [x] Create "Docs" section #2055
- [ ] Create improved layout for documentation #3477
- [x] Move Ecosystem to documentation #2055
- [x] Merge "Details" and "advanced topics" into docs #2055
- [x] Split MQTT documentation #2019 #2055
- [x] Create a Tools section. `hass`, scripts, Developer tools, etc #2114
- [x] Improve the FAQ sections of the different topics
- [x] Sync Component details and Docs (eg. Automation)

keatontaylor:
I approve, but I would encourage placing any advanced topic within the docs of the underlying component. This would seriously simplify navigation and I think help other see the big picture.

balloob:
I agree with that, I think that breaking component docs into subsections would help tremendously. We already do this for Z-Wave but I feel like MQTT could use that too

[...]

PR #2055, opened 2017-02-16
Re-organisation Documentation and Getting started

fabaff:
**Description:**

This the re-organization of the Home Assistant documentation as proposed in #1603.

- The "Getting started" steps are now again in a kind of wizard-style. Extremely minimalist and reduced to fit on a single screen with no scrolling needed to see the whole page.
- New "Docs" section which contains the information previously available in "Getting started"
- Ecosystem, Details, and the advanced topics are now part of "Docs"
- To make the maintenance easier, are now all files divided in a folder per topic in "Docs".
- MQTT splitted like Z-Wave and moved to "Docs"
- and everything else which I already forgot over the 30 commits

Todo:

- Rename section/titles in "Docs" if the name is not accurate enough
- Review the page if the previously used navigation link at the end of the pages is gone
- Check all links :disappointed: 
- Create a better layout of the "Docs" (see #1603 for navigation examples)

We need to discuss the handling of the HTTP 404 messages. This PR will break almost all links to the documentation. We introduced `redirect_from: ` a while back and could use it. Any thoughts?


fabaff:
I made the changes according the gitter conversation a couple of days back. The "Getting started" section was updated, redirects added, and we no longer use collections.

I will merge this as there are no more comments at the moment to address.

wroadd:
"Getting Started" section Index page contains wrong links

PR #2073, opened 2017-02-17
Improve documentation for device_class

armills:
**Description:**
Improve documentation for the new device_class feature.

### DIFF
--- a/source/_components/binary_sensor.markdown
+++ b/source/_components/binary_sensor.markdown
@@ -13,7 +13,7 @@ Binary sensors are gathering information about state of switches, contacts, pins
 
 Knowing that there are only two states allows Home Assistant to represent the sensor better in the frontend.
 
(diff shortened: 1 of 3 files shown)


source/_components/binary_sensor.markdown
```diff
@@ -13,7 +13,7 @@ Binary sensors are gathering information about state of switches, contacts, pins
 
 Knowing that there are only two states allows Home Assistant to represent the sensor better in the frontend.
 
-Most binary sensors support the `device_class:`  which let you specify the type of your sensor. The following types are supported:
+The display style of each entity can be modified in the [customize section](/getting-started/customizing-devices/). The following device classes are supported for binary sensors:
```
armills: One note is that right now most binary_sensor platform schemas support device_class. This is just the doc change, but this change will start steering users toward customize, where this really belongs. We should eventually deprecate the use of device_class directly in the platform config.

PR #2089, opened 2017-02-19
standardize use of chown as seen in other guides

kentcalero:
**Description:**

### DIFF
--- a/source/getting-started/installation-virtualenv.markdown
+++ b/source/getting-started/installation-virtualenv.markdown
@@ -44,7 +44,7 @@ This can be anywhere you want. As example we put it in `/srv`. You also need to 
 
 ```bash
 $ sudo mkdir /srv/homeassistant
-$ sudo chown homeassistant /srv/homeassistant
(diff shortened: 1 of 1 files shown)


andrey-git:
Step 1 says `sudo adduser --system homeassistant`
System users are created in `nogroup` group and not in a group matching their name.

kentcalero:
Assuming your are correct, just because they're not added to a group matching their name, it does not mean a group matching their name cannot exist.

Sections of guides reference chown homeassistant:homeassistant while other do not, we should standardize on a method of presentation.

Please advise.

Kent L. Calero

Sent from the field through my iPhone.

On Feb 19, 2017, at 3:13 AM, Andrey <notifications@github.com<mailto:notifications@github.com>> wrote:


Step 1 says sudo adduser --system homeassistant
System users are created in nogroup group and not in a group matching their name.

andrey-git:
I tested and I see that the group exists.
Lets indeed do this to standardize the guides

PR #2096, opened 2017-02-20
Update for PR #6114

justin8:
Update for module changes from home-assistant/home-assistant#6114

**Description:**
- Added mac field
- Updated example
- Added my TV to the supported models list as it works

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** home-assistant/home-assistant#6114


source/_components/media_player.samsungtv.markdown
```diff
@@ -25,6 +25,8 @@ To add a TV to your installation, add the following to your `configuration.yaml`
 media_player:
   - platform: samsungtv
     host: 192.168.0.10
+    port: 8001
+    mac: 5c:49:7d:9a:6f:5d
 ```
```
fabaff: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

justin8: Ok, I'll remove that part.

PR #2126, opened 2017-02-24
Remove unnecessary script usage

scottocs11:
**Description:**
Scripts are not necessary to perform this automation.

### DIFF
--- a/source/_cookbook/turn_on_light_for_10_minutes_when_motion_detected.markdown
+++ b/source/_cookbook/turn_on_light_for_10_minutes_when_motion_detected.markdown
@@ -16,37 +16,22 @@ This recipe will turn on a light when there is motion and turn off the light whe
 
 ```yaml
 automation:
-  alias: Turn on kitchen lights when there is movement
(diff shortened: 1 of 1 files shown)


scottocs11:
After posting an example, someone [pointed out](https://www.reddit.com/r/homeassistant/comments/5vt72e/im_posting_my_configuration_files_for_all_to_view/de5h3rs/) how much unnecessary coding is used in this example.

Landrash:
@scottocs11 That's a long living example for the beginning of HA.
@fabaff Where should this be merged?

fabaff:
`current` is ok.

joaoasilva:
This example is not right.
The light will turn off ALWAYS even if the sensor detects more movement after it goes to the 'off' state, doesn't reset the timer. The old example is the right logic and the proper implementation.

scottocs11:
@joaoasilva No, it will wait 10 minutes after the last movement to turn the light off. The automation to turn the light off checks to see if the motion sensor has been off for 10 minutes.

PR #2135, opened 2017-02-25
Documentation: sensor.dnsip

danielperna84:
**Description:**
Documentation for the new DNS IP sensor.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#6214

### DIFF
--- a/source/_components/sensor.dnsip.markdown
+++ b/source/_components/sensor.dnsip.markdown
@@ -34,7 +34,7 @@ Configuration variables:
 - **resolver** (*Optional*): The DNS server to target the query at.
 - **ipv6** (*Optional*): Set this to `true` or `false` if IPv6 should be used. Resolving the public IP will resolve to IP of the machine HASS is running on.
 - **resolver_ipv6** (*Optional*): The IPv6 DNS server to target the query at.
-- **scan_interval** (Optional): Defines number of seconds for polling interval. Default is 120 seconds.
+- **scan_interval** (*Optional*): Defines number of seconds for polling interval. Default is 120 seconds.
 
 #### {% linkable_title Extended example %}
 
--- a/source/_components/sensor.dnsip.markdown
+++ b/source/_components/sensor.dnsip.markdown
@@ -1,0 +1,49 @@
+---
+layout: page
+title: "DNS IP"
+description: "Instructions how to integrate timetable data for travelling in Germany within Home Assistant."
+date: 2017-02-25 11:05
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: Sensor
+logo: db.png
+ha_iot_class: "Cloud Polling"
+ha_release: 0.40
+---
+
+
+The `dnsip` sensor will expose an IP address, fetched via DNS resolution, as its value. There are two operational modes:
+
+1. When you enable the sensor with minimal configuration, it will query [OpenDNS](https://www.opendns.com/)' nameserver with the URI `myip.opendns.com`, which will resolve to your external IP address. You can use this to get notified if your internet connection has been interrupted and you've been assigned a new IP address.
+2. If you specify a `hostname`, a regular DNS lookup will be performed, providing you the IP the hostname resolves to.
+
+You may also override the nameserver that's being used by setting the `resolver` parameter to any nameserver you like.
+
+To enable this sensor, add the following lines to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: dnsip
+```
+
+Configuration variables:
+
+- **hostname** (*Optional*): The hostname for which to perform the DNS query.
+- **resolver** (*Optional*): The DNS server to target the query at.
+- **ipv6** (*Optional*): Set this to `true` or `false` if IPv6 should be used. Resolving the public IP will resolve to IP of the machine HASS is running on.
+
+This sensor stores a lot of attributes which can be accessed by other sensors eg. a [template sensor](/components/sensor.template/).
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  # Own public IPv4 address
+  - platform: dnsip
+  # IP address of home-assistant.io
+  - platform: dnsip
+    hostname: home-assitant.io
+    resolver: 8.8.8.8
+```
--- a/source/_components/sensor.dnsip.markdown
+++ b/source/_components/sensor.dnsip.markdown
@@ -8,7 +8,6 @@ comments: false
 sharing: true
 footer: true
 ha_category: Sensor
-logo: db.png
 ha_iot_class: "Cloud Polling"
 ha_release: 0.40
 ---
--- a/source/_components/sensor.dnsip.markdown
+++ b/source/_components/sensor.dnsip.markdown
@@ -33,6 +33,7 @@ Configuration variables:
 - **hostname** (*Optional*): The hostname for which to perform the DNS query.
 - **resolver** (*Optional*): The DNS server to target the query at.
 - **ipv6** (*Optional*): Set this to `true` or `false` if IPv6 should be used. Resolving the public IP will resolve to IP of the machine HASS is running on.
+- **resolver_ipv6** (*Optional*): The IPv6 DNS server to target the query at.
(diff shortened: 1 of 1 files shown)


source/_components/sensor.dnsip.markdown
```diff
@@ -0,0 +1,50 @@
+---
+layout: page
+title: "DNS IP"
+description: "Instructions how to integrate timetable data for travelling in Germany within Home Assistant."
```
mnoorenberghe: This description needs updating

source/_components/sensor.dnsip.markdown
```diff
@@ -0,0 +1,50 @@
...
+#### {% linkable_title Extended example %}
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  # Own public IPv4 address
+  - platform: dnsip
+  # IP address of home-assistant.io
+  - platform: dnsip
+    hostname: home-assitant.io
```
mnoorenberghe: Typo: assistant

danielperna84:
Thanks. Fixed now. 🙂

source/_components/sensor.dnsip.markdown
```diff
@@ -0,0 +1,50 @@
...
+footer: true
+ha_category: Sensor
+ha_iot_class: "Cloud Polling"
+ha_release: 0.40
+---
+
+
+The `dnsip` sensor will expose an IP address, fetched via DNS resolution, as its value. There are two operational modes:
+
+1. When you enable the sensor with minimal configuration, it will query [OpenDNS](https://www.opendns.com/)' nameserver with the URI `myip.opendns.com`, which will resolve to your external IP address. You can use this to get notified if your internet connection has been interrupted and you've been assigned a new IP address.
```
mnoorenberghe: Nit: myip.opendns.com is a hostname not a URI

mnoorenberghe: Nit: "You can use this to get notified if your internet connection has been interrupted and you've been assigned a new IP address.". This is partly explaining NAT and makes a few assumptions about how a given ISP/connection works so it feels a bit confusing to me. Perhaps you can just say "external/public" IP address and remove the last sentence?

source/_components/sensor.dnsip.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: dnsip
+```
+
+Configuration variables:
+
+- **hostname** (*Optional*): The hostname for which to perform the DNS query.
```
mnoorenberghe: Nit: State the default here

source/_components/sensor.dnsip.markdown
```diff
@@ -0,0 +1,50 @@
...
+- **resolver** (*Optional*): The DNS server to target the query at.
```
mnoorenberghe: Nit: state the default here

source/_components/sensor.dnsip.markdown
```diff
@@ -0,0 +1,50 @@
...
+- **ipv6** (*Optional*): Set this to `true` or `false` if IPv6 should be used. Resolving the public IP will resolve to IP of the machine HASS is running on.
```
mnoorenberghe: The grammar of "will resolve to IP of" could be improved.

mnoorenberghe: I don't understand why "Resolving the public IP will resolve to IP of the machine HASS is running on." is mentioned here. Could you elaborate on what you're trying to say?

source/_components/sensor.dnsip.markdown
```diff
@@ -0,0 +1,50 @@
...
+comments: false
+sharing: true
+footer: true
+ha_category: Sensor
+ha_iot_class: "Cloud Polling"
+ha_release: 0.40
+---
+
+
+The `dnsip` sensor will expose an IP address, fetched via DNS resolution, as its value. There are two operational modes:
```
mnoorenberghe: Isn't it just that there are two use cases? The modes really use the same code and logic just with a different hostname. I think it may be better to just cover the use cases with more explanation near the examples.

source/_components/sensor.dnsip.markdown
```diff
@@ -0,0 +1,50 @@
...
+- **scan_interval** (*Optional*): Defines number of seconds for polling interval. Default is 120 seconds.
+
+#### {% linkable_title Extended example %}
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  # Own public IPv4 address
+  - platform: dnsip
+  # IP address of home-assistant.io
```
mnoorenberghe: Nit: …using Google DNS

source/_components/sensor.dnsip.markdown
danielperna84: Well, the main purpose if the sensor is to get the own public IP. But it should be clear, that it can also be used like a regular `nslookup` (which usually is not capable of retrieving the public IP).
So yes, it's technically not two modes. But to me it feels like it does two different things. Don't know how to better describe it.

source/_components/sensor.dnsip.markdown
danielperna84: You're right. I'll change that.

danielperna84: I guess you're right. I wanted to provide an example for how this sensor is useful. And given there are people without that technical beackground, I thought this bit may be helpful.

source/_components/sensor.dnsip.markdown
danielperna84: The would be `myip.opendns.com`, and I think that would confuse people, since they don't know what this special hostname in conjuction with the default nameserver does.

source/_components/sensor.dnsip.markdown
danielperna84: Ok.

source/_components/sensor.dnsip.markdown
danielperna84: Since I'm not a native speaker: is there a "the" missing?

mnoorenberghe: See my next comment… I'm not sure since I don't understand what this sentence is about. Once you answer that I can suggest other wording here.

danielperna84: The IP you'll get in that case it not the one of the modem/gateway, which may be responsible for port forwarding or provide some other services. If f. ex. there's an OpenVPN Server running and you want to be notified about the IP you would have to connect to from external networks, the IP this sensor would get probably wouln't be the one people need in case HASS is running on a different machine.

source/_components/sensor.dnsip.markdown
danielperna84: Yes. Is that a bad thing? I wanted to provide an example that would actually work when just copy & pasting it.

mnoorenberghe: No, sorry, I just thought you should add that to the end of the sentence since not everyone knows that 8.8.8.8 is Google DNS

source/_components/sensor.dnsip.markdown
mnoorenberghe: Are you saying that ipv6 support doesn't work with `myip.opendns.com`? Or it doesn't work with any hostnames that are supposed to return your public IPv6 address?

danielperna84: It does work with `myip.opendns.com`. It's a technical detail. Lets say your modem has the IPv6 address `::123`, a server running OpenVPN has `::234` and the HASS machine has `::345`. This sensor would then, correctly get the value `::345`. With IPv4 you would get the IP the modem has externally and does it's NAT magic with. I am simply assuming, that people would expect the sensor value to be `::123` (the modem) instead of `::345`, and what they actually care about would be the `::234` of the OpenVPN server.

danielperna84:
Updated the documentation. Ok that way?

andrey-git:
Seems clear to me. @mnoorenberghe what do you think?

Landrash:
Looks clear to me also. Merging since parent pr has been merged.

Good job @danielperna84 :cocktail:

PR #2137, opened 2017-02-25
Clarified the Synology-section

roadrash2108:


mikegeeezy:
For DSM 6.1, I believe the Synology instructions also need a step to add the local/container port settings.

PR #2150, opened 2017-02-26
Move history_stats from "Improvements" to "New"

bokub:
**Description:**

The history_stats sensor platform is new in 0.39, so it should be in the "New platforms/components" list, not in the "Improvements" list.


bokub:
Reopened a clean PR on `home-assistant:current` instead of `home-assistant:next`.
This PR can be deleted

PR #2163, opened 2017-02-27
Update Vagrant documentation

shaftoe:
**Description:**

I tried https://home-assistant.io/docs/installation/vagrant/ instruction recently and it doesn't work out of the box, apparently one dependency script disappeared from the code base and broke the process. This patch should fix it, plus introduce an easier UX for interacting with the Vagrant box development environment.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#6236

### DIFF
--- a/source/_docs/installation/vagrant.markdown
+++ b/source/_docs/installation/vagrant.markdown
@@ -10,7 +10,7 @@ footer: true
 redirect_from: /getting-started/installation-vagrant/
 ---
 
-A `Vagrantfile` is available into `virtualization/vagrant` folder for quickly spinning up a Linux virtual machine running Home Assistant. This can be beneficial for those who want to experiment with Home Assistant and/or developers willing to easily test local changes and run test suite against them.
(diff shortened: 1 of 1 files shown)


shaftoe:
Pull request was approved and merged, I guess this should be OK too

source/_docs/installation/vagrant.markdown
```diff
@@ -44,46 +44,57 @@ $ git config --global core.autocrlf input
 ## {% linkable_title Create the Vagrant VM and start Home Assistant %}
 
 ```bash
-$ vagrant up
+$ ./provision.sh setup
 ```
 
-This will download and start a virtual machine using Virtualbox, which will internally setup the development environment necessary to start Home Assistant process and run test suite as well. After the VM has started successfully, the Home Assistant frontend will be accessible locally from your browser at [http://localhost:8123](http://localhost:8123)
+This will download and start a virtual machine using Virtualbox, which will internally setup the development environment necessary to start Home Assistant. The whole process might take up to 30' to complete, depending on Internet connection speed and workstation resources. After the VM has started successfully, the Home Assistant frontend will be accessible locally from your browser at [http://localhost:8123](http://localhost:8123)
```
Landrash: 30 of what? Minutes, seconds or years?

shaftoe: AFAIK the single quote char ' is generally used to identify minutes, while the double quote " seconds. Actually now that I'm digging into it, [according to wikipedia](https://en.wikipedia.org/wiki/Prime_(symbol)) it should be a "prime" char instead.
Anyway, I'll put _minutes_ just to avoid any confusion.

Landrash:
@shaftoe Thanks for the quick fix. 🍪

PR #2172, opened 2017-02-28
Adding in an example for a second hue hub.

CCOSTAN:
It wasn't super obvious to me how to add a second Hue Hub.  The information is in the light component but this makes it easier.

### DIFF
--- a/source/_components/light.hue.markdown
+++ b/source/_components/light.hue.markdown
@@ -29,6 +29,15 @@ light:
   host: DEVICE_IP_ADDRESS
 ```
 
+The light component supports multiple entries in <code>configuration.yaml</code> by appending a sequential number to the section: <code>light 2:</code>, <code>light 3:</code> etc.  When adding in a second hub, add the following lines to your `configuration.yaml`:
+
+```yaml
+# Example configuration.yaml entry
+light 2:
+  platform: hue
+  host: DEVICE_IP_ADDRESS
+```
+
 Configuration variables:
 
 - **host** (*Optional*): IP address of the device, eg. 192.168.1.10. Required if not using the `discovery` component to discover Hue bridges.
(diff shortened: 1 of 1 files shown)


source/_components/light.hue.markdown
```diff
@@ -29,12 +29,22 @@ light:
...
+  filename: phue2.conf
+```
+
 Configuration variables:
 
 - **host** (*Optional*): IP address of the device, eg. 192.168.1.10. Required if not using the `discovery` component to discover Hue bridges.
 
 - **allow_unreachable** (*Optional*): (true/false)  This will allow unreachable bulbs to report their state correctly.
-- **filename** (*Optional*): Make this unique if specifying multiple Hue hubs.
+- **filename** (*Optional*): Make this unique if specifying multiple Hue hubs. Necessary though if adding second Hub.
```
balloob: I don't think that that the added sentence adds any value compared to what is already there. Instead, rewrite the description to explain what it is used for instead of how it should be used.

source/_components/light.hue.markdown
```diff
@@ -29,12 +29,22 @@ light:
   host: DEVICE_IP_ADDRESS
 ```
 
+The light component supports multiple entries in <code>configuration.yaml</code> by appending a sequential number to the section: <code>light 2:</code>, <code>light 3:</code> etc.  When adding in a second hub, add the following lines to your `configuration.yaml`:
```
balloob: Specifying multiple platforms for a component is Home Assistant 101, we should not mention this in the platform docs here.

CCOSTAN: For me, I added this example in since if a user is having trouble adding in a second physical hub, they are going to go to the Hue Hub section of the documentation.  This will solve their pain.  The broader information in the light section might be overlooked. - It was by me. 

But no offense taken if you toss this change.  I just put it out there for you guys to decide. :)

balloob:
I'll close it as we should not repeat ourselves.

PR #2184, opened 2017-03-03
Dont Merge - Add /bin to PATH variable for virtualenv installations

cgtobi:
Add /bin to PATH variable for virtualenv installations to enable home assistant to find e.g. the ping command which is located in /bin. This was a problem on ubuntu which caused the WOL switch to not report the correct state of the computer.

See https://github.com/home-assistant/home-assistant/issues/5930 and https://community.home-assistant.io/t/wake-on-lan-switch-status-not-working/11700 for reference.

**Description:**

### DIFF
--- a/source/_docs/autostart/systemd.markdown
+++ b/source/_docs/autostart/systemd.markdown
@@ -48,7 +48,7 @@ Type=simple
 User=homeassistant
 # Make sure the virtualenv Python binary is used
 Environment=VIRTUAL_ENV="/srv/homeassistant"
-Environment=PATH="$VIRTUAL_ENV/bin:$PATH"
+Environment=PATH="$VIRTUAL_ENV/bin:/bin:$PATH"
 ExecStart=/srv/homeassistant/bin/hass -c "/home/homeassistant/.homeassistant"
 
 [Install]


Landrash:
There's nothing missing here. The `$PATH` variable is imported after your proposed change.
Checked on Debian Jessie and Stretch and change is not required.

I suggest you check the `$PATH` variable of the user your using to run HA since it's for some reason does not have the path you have install nmap included.

If you would like to write a specific Ubuntu fix then go ahead but we shouldn't complicate the example if not needed.

cgtobi:
From the context of the hass user:
```bash
$ printenv PATH
/srv/homeassistant/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
```

But when it's called by the systemd service `/bin` is missing from `$PATH`.

Besides, this has nothing to do with `nmap`. It was the `ping`command it couldn't find.

Landrash:
For some reason your `hass` user has an empty `$PATH` variable then. Most likely a change by Ubuntu since it's not how SystemD should handle it.

Tested it on my Arch & Manjaro systems and they both behave as it should.

cgtobi:
The `$PATH` variable is not empty. When I run `printenv PATH` from within hass I get the following.
```
"$VIRTUAL_ENV/bin:$PATH"
```

Landrash:
Since this is apparently a problem for more than you as I've seen in the chat I'll try to figure what is wrong.
This file most likely won't be changed and an exception will be mentioned for the distro that have this problem.

Using your solution breaks the standard that systemd services should be created after and as far as I'm told it something messed up with Ubuntu in this case.

cgtobi:
As far as I understand the [systemd documentation](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines) things like `$PATH` wont be resolved as in bash environments. So it shouldn't be an Ubuntu-related problem but simply a wrong configuration of the systemd unit.

Landrash:
@cgtobi So your reasoning is that Ubuntu is right and everything else wrong..?

cgtobi:
No, I said it shouldn't be an Ubuntu-related problem. I just try to understand what's going on.

balloob:
The issue with the wake on lan has been resolved, I don't think that this PR is required anymore.

cgtobi:
Which wake on lan issue do you refer to? There are at least three from the past two weeks or so. home-assistant/home-assistant#6397, home-assistant/home-assistant#6345 and  home-assistant/home-assistant#5930.

Landrash:
Closed after discussion with me on Gitter. Actually a problem after @cgtobi dove a bit deeper into it and talked to the dev's of systemd. New pr coming.

PR #2188, opened 2017-03-03
Document notify.webostv icon configuration

andersonshatch:
**Description:**
Add configuration for notify.webostv icons

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** home-assistant/home-assistant#6380

### DIFF
--- a/source/_components/notify.webostv.markdown
+++ b/source/_components/notify.webostv.markdown
@@ -33,6 +33,7 @@ Configuration variables:
 - **host** (*Required*): The IP of the LG WebOS Smart TV, e.g. 192.168.0.10
 - **name** (*Required*): The name you would like to give to the LG WebOS Smart TV.
 - **filename** (*Optional*): The filename where the pairing key with the TV should be stored. This path is relative to Home Assistant's config directory. It defaults to `webostv.conf`.
+- **icon** (*Optional*): The path to an image file to use as the icon in notifications -- if provided, this image will override the Home Assistant logo.
 
 A possible automation could be:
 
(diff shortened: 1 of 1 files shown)


source/_components/notify.webostv.markdown
```diff
@@ -33,6 +33,7 @@ Configuration variables:
 - **host** (*Required*): The IP of the LG WebOS Smart TV, e.g. 192.168.0.10
 - **name** (*Required*): The name you would like to give to the LG WebOS Smart TV.
 - **filename** (*Optional*): The filename where the pairing key with the TV should be stored. This path is relative to Home Assistant's config directory. It defaults to `webostv.conf`.
+- **icon** (*Optional*): The path to an image file to use as the icon in notifications -- if provided, this image will override the Home Assistant logo.
```
Landrash: Suggest you remove the double dashes and make it two separate sentences.

source/_components/notify.webostv.markdown
```diff
@@ -49,3 +50,14 @@ automation:
       data:
         message: "You should open a window! (Livingroom Co2: {{ states.sensor.netatmo_livingroom_co2.state }}ppm)"
 ```
+
+The icon can be overridden for individual notifications by providing a path to an alternative icon image to use:
+
+```yaml
```
Landrash: Since it's fairly common for folks to copy the examples I think you should make it a full automation example as the one above.

andersonshatch:
Thanks for the review @Landrash -- I've addressed both of your suggestions now.

balloob:
This is pointed at current and not at next. Should be merged after we release 0.40

PR #2220, opened 2017-03-08
added scan_interval

ih8gates:
**Description:**
Added scan_interval to docs. Default scan_interval for climate platform causes you to be rate-limited by Honeywell API.

### DIFF
--- a/source/_components/climate.honeywell.markdown
+++ b/source/_components/climate.honeywell.markdown
@@ -24,12 +24,13 @@ climate:
   scan_interval: 600
 ```
 <p class='note'>
-Scan interval is expressed in seconds. Omitting scan_interval may result in too-frequent polling and cause you to rate-limited by Honewell.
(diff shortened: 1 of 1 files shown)


Landrash (changes requested):
`scan_interval is not described among configuration variables.

Example should always be minimal since they are often copied and not include any optional variables.

By all means lets have an example and in this case we can include it in the minimal config since it seems to be a sane default.

My personal preference would be to override the sane default instead of having to set it.

ih8gates:
The scan_interval is inherited (from climate?) if it is not specified in the component configuration.  The default value sets scan_interval at 120 sec. Setting scan_interval in the component settings works properly and avoids the API limiting imposed by Honeywell.

See also: https://github.com/home-assistant/home-assistant/issues/5597

Landrash:
Added it as a variable for clarity since it's not clear in the climate component.

Landrash (approved):

PR #2223, opened 2017-03-09
Remove unnecessary environment variable setup from systemd service

cgtobi:
The environment variables are not needed and are actually overriding the system PATH variable and were causing trouble in certain circumstances.

**Description:**

### DIFF
--- a/source/_docs/autostart/systemd.markdown
+++ b/source/_docs/autostart/systemd.markdown
@@ -49,7 +49,6 @@ After=network.target
 [Service]
 Type=simple
 User=%i
-# Make sure the virtualenv Python binary is used
 ExecStart=/srv/homeassistant/bin/hass -c "/home/homeassistant/.homeassistant"
 
 [Install]
--- a/source/_docs/autostart/systemd.markdown
+++ b/source/_docs/autostart/systemd.markdown
@@ -48,7 +48,7 @@ After=network.target
 
 [Service]
 Type=simple
-User=homeassistant
+User=%i
 # Make sure the virtualenv Python binary is used
 ExecStart=/srv/homeassistant/bin/hass -c "/home/homeassistant/.homeassistant"
 
--- a/source/_docs/autostart/systemd.markdown
+++ b/source/_docs/autostart/systemd.markdown
@@ -39,7 +39,7 @@ ExecStart=/usr/bin/hass
 WantedBy=multi-user.target
 ```
 
-If you've setup Home Assistant in `virtualenv` following our [python installation guide](https://home-assistant.io/getting-started/installation-virtualenv/) or [manual installation guide for raspberry pi](https://home-assistant.io/getting-started/installation-raspberry-pi/), the following template should work for you.  If Home Assistant install is not located at `/srv/homeassistant`, please modify the `Enviroment=` and `ExecStart=` lines appropriately.  
+If you've setup Home Assistant in `virtualenv` following our [python installation guide](https://home-assistant.io/getting-started/installation-virtualenv/) or [manual installation guide for raspberry pi](https://home-assistant.io/getting-started/installation-raspberry-pi/), the following template should work for you.  If Home Assistant install is not located at `/srv/homeassistant`, please modify the `ExecStart=` line appropriately.  
 
 ```
 [Unit]
@@ -49,7 +49,7 @@ After=network.target
 [Service]
 Type=simple
 User=%i
-ExecStart=/srv/homeassistant/bin/hass -c "/home/homeassistant/.homeassistant"
+ExecStart=/srv/homeassistant/bin/hass -c "/home/%i/.homeassistant"
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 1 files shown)


cgtobi:
@Landrash please take a look now.

brahmafear:
@cgtobi Whoa...  Isn't that going to break this for VENV users?  I thought the service needed either the environment variables or to pre-start the activate script for the VENV.

cgtobi:
I'm an VENV user myself and stumbled across this when debugging another issue. I works totally fine for my and I was discussing this with @Landrash over the past days. He came to pretty much the same result.

Landrash:
@brahmafear [This](https://github.com/home-assistant/hassbian-scripts/blob/dev/etc/systemd/system/home-assistant%40homeassistant.service) is the one used with HASSbian.
This change is fairly similar and I would appreciate some more try it out and verify that it functions as predicted.

brahmafear:
@cgtobi @Landrash Are you sure hass isn't running globally when you set it up that way?  Without the activate script and without resetting environment variables, there doesn't seem to be any mechanism for using the VENV.

brahmafear:
@cgtobi Let me phrase it this way -- is there a downside to adding this line before ExecStart?

`ExecStartPre=source /srv/homeassistant/bin/activate`

cgtobi:
@brahmafear I couldn't see why. I this case it might actually be worth putting the path into a variable.
Maybe something like this:
```
Environment=VIRTUAL_ENV=/srv/homeassistant
ExecStartPre=source ${VIRTUAL_ENV}/bin/activate
ExecStart=${VIRTUAL_ENV}/bin/hass -c "/home/%i/.homeassistant"
```

This is completely untested, not sure if it would work that way.

brahmafear:
@cgtobi @Landrash Always worth further research....  Just found this on @cgtobi's link:

'The virtualenv is "baked into the Python interpreter in the virtualenv". This means you can launch python or console_scripts directly in that virtualenv and don't need to manage PATH etc yourself.'

Probably will keep my setup with ExecStartPre in place, but seems like omitting that line as well as the PATH settings in original commit will work fine.

cgtobi:
What I am not sure about is whether the activation of the virtual environment would actually work. It could be that it is not persistent for the service. Does that make sense?

brahmafear:
@cgtobi Hmm.  Should be?  This is the first line of my hass executable:
`#!/srv/hass/bin/python3`
so it is getting the correct run environment.

cgtobi:
@brahmafear That is probably what they were talking about when they called the virtual environment 'baked in', I'd assume.

Landrash:
This looks good to me and I'll merge it tomorrow if there's no further discussions.

siebert:
I can report that setting PATH did break the wake_on_lan switch, because $PATH can't be used in the Environment statement and the result was that no binaries could be found via PATH. See http://stackoverflow.com/questions/35439123/systemd-environment-directive-to-set-path for details.

jjmontesl:
I didn't know it. But indeed, running a script binary from a virtualenv will use the python installation in that virtualenv, without the need of activating the virtualenv first

I'm unsure it this will allow forked processes to use the virtualent, though.

brahmafear:
Withdrawing my suggestion of the ExecStartPre statement as I am seeing a bunch of these messages in daemon.log:
`Mar  9 22:21:04 pi systemd[1]: [/etc/systemd/system/homeassistant@hass.service:8] Executable path is not absolute, ignoring: source /srv/hass/bin/activate`

So it seems systemd isn't using that directive.  This has been referenced in [this](https://community.home-assistant.io/t/auto-start-on-raspberry-pi/7207/14) community discussion.  

So it seems that neither the ExecStartPre nor setting the environment variables are necessary, but @jjmontesl has a good point about whether forked processes will inherit the venv.

dale3h:
The primary responsibility of the `/srv/homeassistant/bin/activate` script is to set the priority of `PATH` to `/srv/homeassistant/bin`. And as previously mentioned, the `bin/hass` file already has the path to the virtualenv's python interpreter set as the shebang. Therefore, setting the include path should be unnecessary in the service file.

Landrash:
@cgtobi Let's just remove the unnecessary variables and leave it as a set path and user. The user in question can then change it after their own preference and it's easier to understand without the `%i` variable.

Great work for figuring this out :)

cgtobi:
This should be it then, right?

dale3h:
I could be mistaken, but I feel that leaving out `%i` entirely is the best way to go about this. 99.9% of our users do not run multiple instances on the same machine, so hardcoding the user to `homeassistant` seems like it will keep things simple and easy to understand. If we do, I think we will have to leave the service filename as `home-assistant.service`.

I am not a systemd master, so I am not entirely sure.

dale3h:
Regardless of what we do, we ***really*** need to get AIO and HASSbian on the same settings/filenames/paths/users/etc.

cgtobi:
@dale3h That's fine by me. More systemd experienced users can change it to their likings.

Issue #2224, opened 2017-03-09
NGINX and scenegen missing in topics


danielperna84:
When you go to the [Ecosystem](https://home-assistant.io/docs/ecosystem/) site, there are a ton of topics on the right side. Far at the bottom there are the ones for the ecosystem. Looking at the repo, there are markdown-files for NGINX and scenegen. These are however not present in the topic-list. Not knowing much about the internals of all this I assume, it's because those two don't have their own subfolder with further details about the documentation. Directly navigating to the sites (https://home-assistant.io/docs/ecosystem/nginx/, https://home-assistant.io/docs/ecosystem/scenegen/) is working. So I assume this is a bug.

Edit: Synology is missing as well.

Edit: After searching through this I found where this is done: https://github.com/home-assistant/home-assistant.github.io/blob/next/source/_includes/asides/docs_navigation.html
Thought there was some scripting that adds links automatically. But since it's hardcoded I might as well fix it myself.

fabaff:
I missed them when I created the menu.

PR #2225, opened 2017-03-09
Add HASS Configurator to ecosystem

danielperna84:
**Description:**
Since my [HASS Configurator](https://github.com/danielperna84/hass-poc-configurator) has reached a stable state, I feel it's time to add it to the ecosystem section of the HASS site.
I don't think the amount of content justifies splitting the documentation up into several files. Hence the single-file-approach.
This has one drawback though, caused by the possible bug in Issue #2224 . It won't be listed in the topics right now. I assume as soon as the bug is fixed, this will be resolved automatically.

Any comments, suggested changes, rephrasing etc.? 🙂


source/_docs/ecosystem/hass-configurator.markdown
```diff
@@ -102,7 +102,7 @@ Since the configurator script on its own is no service, you will have to take so
 2. If your system is using systemd (that's usually what you'll find on a Raspberry PI), there's a [template file](https://github.com/danielperna84/hass-poc-configurator/blob/master/hass-poc-configurator.systemd) you can use and then apply the same process to integrate it as mentioned in the [HASS documentation](https://home-assistant.io/docs/autostart/systemd/). If you use this method you have to set the `BASEPATH` variable according to your environment.
 3. If you have [supervisor](http://supervisord.org/) running on your system, [hass-poc-configurator.supervisor](https://github.com/danielperna84/hass-poc-configurator/blob/master/hass-poc-configurator.supervisor) would be an example configuration you could use to control the configurator.
 4. A tool called [tmux](https://tmux.github.io/), which should be pre-installed with recent AIO installers.
```
Landrash: Tmx is also pre-installed with the HASSbian image.:)

Landrash:
Looks good and can be merged. Up to you if you would like to mention the HASSbian image along side AiO installer under tmux. :)

danielperna84:
Uhm, I actually though it was the AiO that had tmux. Could it be I'm mixing those two up and tmux actually is not part of the AiO and part of HASSbian instead?

Landrash:
Checked the install script for AiO and I can't find any mention of tmux. I think it's only in later versions of HASSbian :)

danielperna84:
Ok, took that out. Thanks for investigating. :)

balloob:
If it is considered stable, you might want to consider renaming your repo to no longer have 'proof of concept' in the name. GitHub will auto redirect all old urls to the new name.

danielperna84:
I didn't know that. Thanks for the hint. 🙂

danielperna84:
Uhm, could it be that I should have used the current-branch as the target for this PR? My documentations for components/platforms always ended up online when the PR was against current and the new HASS release came out. This time though it seems to be missing.

balloob:
Yes, you should have pointed it at `current` to go out right away, `next` to go out with next release. We cut release branch a couple of days before the actual release

danielperna84:
Ah, ok. Then I'll just leave it this way since it's not urgent. Didn't notice the cut was made a few days before.

PR #2228, opened 2017-03-10
Fix notify_if__new_ha_release.markdown

fanaticDavid:
**Description:**

A template requires the use of `data_template` instead of `data` for the template to be rendered (instead of resulting in an error).


Landrash:
Thank you :cookie:

PR #2231, opened 2017-03-10
Added documentation for Ring binary sensor

tchellomello:
**Description:**
Added documentation for Ring binary sensors

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#6520

https://community-home-assistant-assets.s3.amazonaws.com/original/2X/f/f32ab302af330212fd4e6b69194a9e6961cd7131.png

**Configuration**
```yaml
#configuration.yaml
binary_sensor:
  - platform: ring
    username: !secret ring_username
    password: !secret ring_password
    monitored_conditions:
      - ding
      - motion
```

### DIFF
--- a/source/_components/binary_sensor.ring.markdown
+++ b/source/_components/binary_sensor.ring.markdown
@@ -1,0 +1,38 @@
+---
+layout: page
+title: "Ring"
+description: "Instructions on how to integrate your Ring.com devices within Home Assistant."
+date: 2017-03-10 10:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: ring.png
+ha_category: Binary Sensor
+ha_release: 0.40
+---
+
+The `ring` binary sensor allows you to integrate your [Ring.com](https://ring.com/) devices in Home Assistant.
+
+Currently only doorbells are supported by this sensor.
+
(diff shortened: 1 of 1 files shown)


source/_components/binary_sensor.ring.markdown
```diff
@@ -0,0 +1,38 @@
...
+    monitored_conditions:
+      - ding
+      - motion
+```
+
+Configuration variables:
+
+- **username** (*Required*): The username for accessing your Ring account.
+- **password** (*Required*): The password for accessing your Ring account.
+- **monitored_conditions** array (*Required*): Conditions to display in the frontend. The following conditions can be monitored.
```
Landrash: Are both of these required? If not I think It should be marked as optional and to be removed from the minimal example.

If they generally are always monitored I don't see a reason to have them included by default by the component and remove the configuration variable.

tchellomello: Hello @Landrash, 

By default there are not included or monitored. The user must specify which sensor to use. 

https://github.com/tchellomello/home-assistant/blob/binary_sensor_ring/homeassistant/components/binary_sensor/ring.py#L46

The user can choose only one or 2 sensors. They are not dependent.

PR #2232, opened 2017-03-11
Documentation for 'recorder' sometimes incorrectly mentions 'history'.

herm:
See title.


herm:
Sorry, I selected the wrong base branch first and had to fix this.

Landrash:
First good job and nice to have it fixed, then a small nit pick.
Avoid descriptions such as see title, in general it's good practice to fill out fields as requested if not only for consistency but also it makes it easier for folks using screen readers and similar devices.

Thank you :cookie:

PR #2236, opened 2017-03-12
Add missing verify_ssl parameter to the Ubiquiti Unifi WAP component docs

jquatier:
Add missing verify_ssl parameter to the Ubiquiti Unifi WAP component. I noticed the code handles it but the docs don't have it.


finish06:
Please merge this pull as the new component will fail if end users do not update the configuration with a SSL boolean result.

PR #2268, opened 2017-03-16
Fix ambiguous wording

qcryptic:
The line made it seem like there may be more optional parameters, but there isn't any.

**Description:** Changed line to make it more obvious there is only 2 optional parameters.

NOTE: The table on this page overlaps the navigation links to the right on my screen. Not sure how to fix that, but someone who does should look into it.


qcryptic:
This table also needs to be fixed... I am not too sure how to fix that myself...

table

Landrash:
Looks good and can be merged. Unsure about what you think is wrong with the table. Your screenshot is an example of the screen being to small to display it properly and nothing we can easily fix in this PR.

PR #2269, opened 2017-03-16
Added SSL guide to cookbook

mf-social:
Added SSL guide to cookbook.  Contains an image, which I have put in the images folder.

Hope this is ok?


source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+
+This will tell you your current external IP address
+
+Type the external IP address in to the url bar with http:// in front and :8123 after like so (12.12.12.12 is my example!):
+
+```text
+http://12.12.12.12:8123
+```
+
+Can you see your HA instance?  Awesome! If not, your router may not support ' loopback' - try the next step anyway and if that works, and this one still doesn't, just remember that you cannot use loopback, so will have to use internal addresses when you're on your home network.  More on this later on if it's relevant to you.
```
bdurrer: The fact that you put _loopback_ in quotes suggests that it has a meaning which a lot of users don't understand.
I guess it should either be explained or left out.

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+
+Just to verify this isn't some kind of witchcraft that is actualy using your internal network, pick up your phone, disconnect it from your wifi so that you are on your mobile data and not connected to the home network, put the same url in the browser on your phone.
```
bdurrer: It's _actually_, two L :smile:
Maybe transform to a checklist instead of one big sentence

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+sharing: true
+footer: true
+ha_category: Infrastructure
+---
+
+This guide was added by mf_social on 16/03/2017 and was valid at the time of writing.  This guide makes the following assumptions:
+
+ * You can access your HA instance across your local network, and access the device that it is on via SSH from your local network.
+ * You know the internal IP address of your router and can access your router's configuration pages.
+ * You have already set up a password for your HA instance, following the advice on this page: [http](https://home-assistant.io/docs/configuration/basic/)
```
bdurrer: The fact that a password should be set up BEFORE opening the router's firewall cannot be stressed enough.
I think there should be a fat warning note on the beginning of the page.

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+Port Range - 80
+Local IP - YOUR-HA-IP
+Local Port - 80
+Protocol - Both
+```
+
+Remember to save the new rule.
+
+<p class='note'>
+In cases where your ISP blocks port 80 you will need to change the port forward options to forward port 443 from outside to port 443 on your HA device.  Please note that this will limit your options for automatically renewing the certificate, but this is a limitation because of your ISP setup and there is not a lot we can do about it!
```
bdurrer: I know what we are talking about here, but I still had a hard time reading that note (but then, I don't have an suggestion to change it besides writing shorter sentences in general 😏 ) .

Also I do wonder how a "noob" could detect or know that port 80 is blocked? But I guess when it gets longer it would require an extra troubleshooting section

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+</p>
+
+now SSH in to the device your HA is running on.
+
+
+<p class='note'>
+If you're running the 'standard' setup on a raspberry pi the chances are you just logged in as the 'pi' user.  If not, you may have logged in as the HA user.  There are commands below that require the HA user to be on the sudoers list.  If you are not using the 'standard' pi setup it is presumed you will know how to get your HA user on the sudoers list before continuing.  If you are running the 'standard' pi setup, from your 'pi' user issue the following command (where <hass> is the HA user):
```
bdurrer: Technical Question: Why does the user need to be on the sudoers list? I think it's not a good idea to give homeassistant that priviledge, since it's running the web services.
EDIT: Ah certbot needs root in --standalone mode. Bummer.

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+```bash
+$ ls /etc/letsencrypt/live/
+```
+
+This should show a folder named exactly after your DuckDNS url.
+
+Our HA user needs access to files within the letsencrypt folder, so issue the following commands to change the permissions.
+
+```bash
+$ sudo chmod 755 /etc/letsencrypt/live/
```
bdurrer: Sorry about being a pain in the ** 😁 : Why does it need 755, I guess 744 would be enough?

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+
+```text
+Service name - ha_ssl
+Port Range - 443
+Local IP - YOUR-HA-IP
+Local Port - 8123
+Protocol - Both
+```
+
+If during step 4 you had to use port 443 instead of port 80 to generate your certificate, you should delete that rule now.
```
bdurrer: Maybe rewrite or reorder to line 301 to make clear which rule you mean with "delete that rule"

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+
+```yaml
+http:
+  api_password: YOUR_PASSWORD
+  ssl_certificate: /etc/letsencrypt/live/examplehome.duckdns.org/fullchain.pem
+  ssl_key: /etc/letsencrypt/live/examplehome.duckdns.org/privkey.pem
+  base_url: examplehome.duckdns.org
+```
+  
+You may wish to set up other options for the http component at this point, these extra options are beyond the scope of this guide but can be found on the http component page here: [http](https://home-assistant.io/components/http/)
```
bdurrer: Style hint: More full stops would improve the text 😉

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+
+In cases where you need to access via the local network only (which should be few and far between) you can access it with the following URL (note the added S after http):
+
+```text
+https://YOUR-HA-IP:8123
+```
+
+...and accepting the browsers warning that you are connecting to an insecure site.  This warning occurs because your certificate expects your incoming connection to come via your DuckDNS URL.  It does not mean that your device has suddenly become insecure.
+
+Some cases such as this are where your router does not allow 'loopback' or where there is a problem with incoming connections due to technical failure.  In these cases you can still use your internal connection and ignore the warnings.
```
bdurrer: That's way too technical, without explaining the technical terms. I suggest to just remove it.

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+
+Go to your router's configuration pages and delete the ha_test rule.
+
+You chould now have two rules in relation to HA for your port forwards, named:
+
+```text
+ha_ssl and ha_letsencrypt
+```
+
+If you have any more for HA you should delete them now.  If you only have ha_ssl this is probably because during step 4 you had to use port 443 instead of port 80, so we deleted the rule during step 5.
```
bdurrer: "any more" of what?

source/_cookbook/SSL guide.md
```diff
@@ -0,0 +1,570 @@
...
+
+You are now part of one of two groups:
+
+ * If you have BOTH rules you are able to set up auto renewals of you certificates.
+ * If you only have one, you will have to manually change the rule when you want to update your certificate, and then change it back afterwards.
+ 
+Please remember whether you are a ONE-RULE person or a BOTH-RULE person for step 8!
+ 
+LetsEncrypt certificates only last for 90 days.  When they have less than 30 days left they can be renewed.  Renewal is a simple process.
```
bdurrer: That might be helpful further up to explain what the guide wants to tell us with "renewal"

source/_cookbook/SSL guide.md
bdurrer: I guess that could be avoided by adding an extra user and create wrapper scripts which do `su `to that extra user... well yeah, thats probably too complicated, the guide is already very extensive 👍

source/_cookbook/SSL guide.md
mf-social: I agree, I'll add a warning when I get a sec.

source/_cookbook/SSL guide.md
mf-social: Fair point, I was kinda hoping in the context that it would just 'make sense', any suggestions for an explanatory sentence?

source/_cookbook/SSL guide.md
mf-social: Ha, I didn't use spell check for the entire guide, so if you only found one spelling mistake I'm well happy! 

I'll correct when I get a sec.

source/_cookbook/SSL guide.md
mf-social: Yeah, this is kinda what I meant in the opening section about not covering every eventuality, and why I said 'novice' rather than noob. If you can think of a better way to word this bit I'm all ears :-)

source/_cookbook/SSL guide.md
mf-social: Yeah, there are ways around this, but as a catch-all for a guide it seems like the easiest and isn't that insecure in the grand scheme of things.

source/_cookbook/SSL guide.md
mf-social: Noted, I'll reorder when I get a sec.

source/_cookbook/SSL guide.md
mf-social: Which bit?

source/_cookbook/SSL guide.md
mf-social: Rules, is it really that unclear? I think when it is 'presented' and formatted it will look right, but if it doesn't I'll reword it. Not sure what to though?

source/_cookbook/SSL guide.md
mf-social: Sorry, I don't get this one, can you clarify please? 

Thanks for all the feedback so far :-D

bdurrer: On 229 the talk is about automatic renewal, but I guess some readers wonder why that is necessary.
But I don't know where a better place is than down here :)

source/_cookbook/SSL guide.md
bdurrer: Yeah it might be the formatting. Speaking of, I suggest you put the rule names as `code`.
PS: Line 366 has a typo: "You _chould_ now"

source/_cookbook/SSL guide.md
bdurrer: It's again about the loopback. Which I don't know how to explain in two or less sentences 😄

source/_cookbook/SSL guide.md
bdurrer: errm no, I don't have a better explanation. I guess that problem won't happen often anyway.

source/_cookbook/SSL guide.md
mf-social: Warning added.

source/_cookbook/SSL guide.md
mf-social: Spelling corrected.

source/_cookbook/SSL guide.md
mf-social: Typo corrected - will have to see how clear it is when it is rendered and if there are any issues with clarity I will fix with a further PR.

source/_cookbook/SSL guide.md
mf-social: In all honesty I have no idea!  I am not so familiar with the unix permissions system tbh, and I think I've only ever used 755 and 777.  Looking it up I guess 744 would be enough, but I haven't tested it.

mf-social:
Added a few bits and corrected a few bits, should be good to go...

Issue #1083, opened 2016-10-03
Search Feature


fabaff:
Our documentation website has no search feature. A lot of people are using `site:home-assistant.io` with Google but a local search functionality would be preferred. There is a existing [feature request](https://community.home-assistant.io/t/search-option-on-main-website/3179) which could serve as a reference for already discussed stuff.

One option which came up was [Jekyll search using lunr.js](http://jekyll.tips/jekyll-casts/jekyll-search-using-lunr-js/). We guess that there are other solutions available that works with GitHub Pages.

cdax:
I checked out the [Jekyll Cast](http://jekyll.tips/jekyll-casts/jekyll-search-using-lunr-js/) that outlines how something like this might be implemented, and I'd like to start working on this. I've been hoping to check out lunr.js for a while now and this looks like a great opportunity.

I'd like some inputs on how this should look on the site? I'm thinking something as simple as the search box on [Django's documentation website](https://docs.djangoproject.com/en/1.10/)?

fabaff:
Simple is good :+1: 

At some point we need to discuss the position but I guess this can wait.

balloob:
An alternative could be to have a query to DuckDuckGo / Google with site:home-assistant.io.

Since the index will be of significant size, we should make sure it won't get automatically downloaded on every page if we have a search box on every page 👍

thebouv:
Just a drive-by suggestion (because I was looking through a bunch of `hacktoberfest` issues), but you might want to consider http://readthedocs.io for hosting your docs. Provides search et al, they host docs for tons of projects, and you won't have to reinvent the wheel.

balloob:
@thebouv our dev docs are 👍  https://dev-docs.home-assistant.io/en/dev/

This would be for the website with all the documentation for the different platforms.

Which makes me think, maybe a different (or both) thing that we can do, is add search to the components page. Have it filter the components based on a search input.

bdurrer:
Sorry to resurrect dead issues. I'm not sure why this one was closed.
A search feature over all components would really boost newbs like me getting started.

A static search index would be good enough for most usecases, so this might be an "easy" solution: http://robin.radic.nl/github-pages-search/

balloob:
@bdurrer you're right, it's not done yet. Although you did add component search which helped a lot!

Another thing we might want to add is https://community.algolia.com/docsearch/

It seems easy to add?

balloob:
[...]

PR #2275, opened 2017-03-16
Feature component search

bdurrer:
**Description:**
Adds a search box to the component overview page and enables the user to quickly find components based on their `title `and `ha_category`.
* It uses isotope's advanced filter feature
* It's tested on Firefox, Chrome and IE11. 
* It's responsive. 
* It works with international component names like V**ä**sttrafik Public Transport
* Includes some minor improvements to the JavaScript in general.

This is related to issue #1083. However, this PR only applies to the component page, not the complete website. It's meant to complement the category selection.
Making the whole page searchable would require a different approach (e.g. an json file which is parsed with fuse.js).

### DIFF
--- a/sass/custom/_paulus.scss
+++ b/sass/custom/_paulus.scss
@@ -330,6 +330,21 @@ p.note {
       -moz-transition-property: -moz-transform, opacity;
       transition-property: transform, opacity;
   }
+  
+      
+  .component-search{
+    margin-bottom: 24px;
+  
+    input{
+      width: 100%;
+      padding: 10px;
+      
+      background-color: #fefefe;
+      border-radius: 2px;
+      border: 1px solid;
+      border-color: #7c7c7c #c3c3c3 #ddd;
+    }
+  }
 }
 
 @media only screen and (max-width: $lap-end) {
--- a/source/components/index.html
+++ b/source/components/index.html
@@ -52,24 +52,33 @@ Support for these components is provided by the Home Assistant community.
       <a href='#other' class="btn">Other</a>
     </div>
   </div>
-  <div class="grid__item five-sixths lap-one-whole palm-one-whole hass-option-cards" id="componentContainer">
-    {% for component in components %}
-      {% if component.ha_category %}
-      {% assign sliced_version = component.ha_release | split: '.' %}
-      {% assign minor_version = sliced_version[1]|plus: 0 %}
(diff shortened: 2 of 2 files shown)


balloob:
This is awesome!

Would you be able to rebase on top of the `current` branch? That way we can get this to everyone when we merge instead of waiting for the next release (1,5 weeks from now).

bdurrer:
PHU! That was the scariest thing I did with GIT recently :simple_smile:
Still looking okay and works on my browsers.

One thing I dislike and couldn't come up with a good solution: Deleting in the search box until it's empty is the equivalent to the "all" filter. While that's semantically correct, the all filter's performance is terrible with 640 elements, it takes several seconds on IE and looks like the search somehow broke down. 
I don't know if we just accept that or maybe switch to "featured" when the search box is empty. The downside is, that "feature" is simply not the correct result set for an empty search

fabaff:
@balloob was talking about to improve the loading process in the near future as there are too many requests and too much files. Going more for a on-demand approach than loading everything.

Let's merge it and wait for feedback.

balloob:
Maybe we can filter out everything when an empty search is passed in ?

balloob:
Put the search feature on social media: https://twitter.com/home_assistant/status/842758527009669120 👍

bdurrer:
@fabaff I think the loading process is not the issue. Its a slow browser with a slow javascript engine. IE simply can't handle 640 elements being made visible and isotope probably isn't made for that amount either. Even Chrome takes 1-2 seconds to render the `all `filter.

@balloob Hm yeah, that would work I guess. According to my measurement one character search is slow too, but at least tolerable.

But I guess Fabian is right, let's wait and take a look for yourself on the website. I can push the empty-search later, it's a no-brainer.

balloob:
@bdurrer since you are currently touching this code, I have an idea to make the page load super fast: for all the image objects, instead of specifying a `src` attribute we specify a `data-src` attribute. Then when isotope is showing the cards, it will go through all image elements that are not visible that are having a `data-src` attribute and set that as `src`.

bdurrer:
That's what I would have done too. But first I'll test some css options, I think it's even easier solvable background-image, browser skip those nowadays [in certain cases].

or even create html elements on the fly and use a json as data repository. 🤔 

I created a new issue #2282 for this.

PR #2280, opened 2017-03-17
added mandatory information ( contract # )

godloth:
**Description:**
Added :- **contract** (required since HA 0.40) Your contract number with Hydro-Quebec
and
    contract: 'CONTRACT#'  (NOTE:works with the ' ' ,insert your contract in between)

### DIFF
--- a/source/_components/sensor.hydroquebec.markdown
+++ b/source/_components/sensor.hydroquebec.markdown
@@ -22,6 +22,7 @@ sensor:
   - platform: hydroquebec
     username: MYUSERNAME
     password: MYPASSWORD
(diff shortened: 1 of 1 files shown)


source/_components/sensor.hydroquebec.markdown
```diff
@@ -32,6 +33,7 @@ Configuration variables:
 
 - **username** (*Required*): Username used to log into the Hydro-Québec site.
 - **password** (*Required*): Password used to log into the Hydro-Québec site.
+- **contract** (required since HA 4.0) Your contract number with Hydro-Quebec
```
bdurrer: Maybe write it with acute (Qu**é**bec) like on the properties above?

Issue #2282, opened 2017-03-17
Improve resource loading on components overview


bdurrer:
As discussed in #2275 , we should improve the loading process of images.

bdurrer:
This was solved with PR #2287 and can be closed 😄

PR #2287, opened 2017-03-18
Issue 2282 performance

bdurrer:
**Description:**
This is a solution for the issue #2282.

The main reason for the components overview page being slow:
- Loading all images even when only a few are shown
- DOM Manipulation over a large set (600+) of elements
- element placement is done with JS and absolute positioning (masonry)

**Solution:**
- Use a lightwight rendering engine for the tiles (mustachejs)
- Let the browser do the placment: Use flexbox layout
- reduce DOM Manipulation (no show/hide on specific items, classes are only added on the container)
- reduce page size by compressing unnecessary white space
- reduced page size by using an JSON object as data source


bdurrer:
You might not have realized it, but that "compress pages" applies to all pages. I forgot to mention that 🙈 , oops. Anyway, it's a good thing to do for all pages.
But keep that in mind please, in case you ever see something weird happen in the future when doing super-weirdo-crazyness-stuff and it suddendly fails 😉 
The compress-template comes with debug mode etc. (see [repo](https://github.com/penibelst/jekyll-compress-html))

PR #2334, opened 2017-03-26
Added documentation and logo for eddystone_temperature platform.

citruz:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** home-assistant/home-assistant#6789


rytilahti:
Just a question, the backend doesn't use a separate scan binary but depends on giving extra caps for python binary?

citruz:
Correct, it reads the data directly from the bluetooth socket. Thats why the extra caps are needed.

Landrash:
Cleaned up some minor bits and merging.
@citruz  Looks really good and hope this component can be expanded upon in the future.
If you really wanna go that extra mile you could check what changes needs to be done to the HASSbian image and include that in this documentation. (Python in a venv etc.)

PR #2361, opened 2017-03-31
Updated sensor.min_max to include missing configuration variable

Genestealer:
Updated example and added extra missing  **unit_of_measurement** configuration variable.

**Description:**


fabaff:
> Updated example and added extra missing unit_of_measurement configuration variable.

We keep the configuration sample limited to the required setting to avoid confusion for the users. There is no [`_unit_of_measurement`](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/sensor/min_max.py#L47) configuration variable

PR #2368, opened 2017-04-01
Doc for nadtcp media player component

mwsluis:
**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** https://github.com/home-assistant/home-assistant/pull/7955


source/_components/media_player.nad7050.markdown
```diff
@@ -0,0 +1,42 @@
...
+```yaml
+# Example configuration.yaml entry
+media_player:
+  - platform: nad7050
+    name: NAD D 7050
+    host: 192.168.0.112
+    min_volume: -60
+    max_volume: -20
+    volume_step: 5
+```
```
fabaff: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

PR #2378, opened 2017-04-04
Document new token auth option for Hook switch

KlaasH:
**Description:**

Reflect new option to configure Hook switch with API token rather than username/password.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#6922


Landrash:
@fabaff Should this be merged in since the change seems to be live already?

PR #2380, opened 2017-04-05
fixed broken links and syntax.

bbrendon:
**Description:**


Landrash:
Reverted the name since we should try to keep naming the same as to the page we're linking to.
Thank you for finding it:)

PR #2400, opened 2017-04-09
inserted important hint concering glances server config

linuxlurak:
**Description:**
I think many users will stumble over the option to configure glance as webserver (-w) or server (-s). This will speed up your configuration time.

### DIFF
--- a/source/_components/sensor.glances.markdown
+++ b/source/_components/sensor.glances.markdown
@@ -32,6 +32,16 @@ $ curl -X GET http://IP_ADDRESS:61208/api/2/mem/free
 
 For details about auto-starting `glances`, please refer to [Start Glances through Systemd](https://github.com/nicolargo/glances/wiki/Start-Glances-through-Systemd).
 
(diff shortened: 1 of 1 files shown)


source/_components/sensor.glances.markdown
```diff
@@ -32,6 +32,16 @@ $ curl -X GET http://IP_ADDRESS:61208/api/2/mem/free
...
 For details about auto-starting `glances`, please refer to [Start Glances through Systemd](https://github.com/nicolargo/glances/wiki/Start-Glances-through-Systemd).
 
+To enable a glances server e.g. on Ubuntu edit `/etc/default/glances` this way:
+```shell
+# Default is to launch glances with '-s' option.
+DAEMON_ARGS="-w"
+
+# Change to 'true' to have glances running at startup
+RUN="true"
+```
```
fabaff: Instead of duplicating the Glances documentation, a link like for systemd is sufficient and we don't need to take care if the information are up-to-date.

fabaff:
As Ubuntu is using `systemd` as well, I think that we need details how to run a daemon in an obsolete way.

PR #2414, opened 2017-04-11
Forecasts for weather underground

pezinek:
**Description:**

This PR allows forecasts for the Weather Underground component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#7062


source/_components/sensor.wunderground.markdown
```diff
@@ -100,19 +103,133 @@ Configuration variables:
...
   - **wind_kph**: Current wind speed in kph
   - **wind_mph**: Current wind speed in mph
+  - **wind_1d_kph** [<sup>[1d]</sup>](#1d): Forecasted wind speed in kph
+  - **wind_1d_mph** [<sup>[1d]</sup>](#1d): Forecasted wind speed in mph
   - **wind_string**: Text summary of current wind conditions
 
-All the conditions listed above will be updated each 5 minutes with exception of `alerts` that will be updated each 15 minutes by default.
+All the conditions listed above will be updated every 5 minutes.
+
+<p class='note'>
```
rpitera: Might be easier to follow if the forecast options were broken out into their own section independent of the standard monitoring options.

pezinek: OK, I've removed the notes and replaced them with with "plain text". 
Dissecting the forecast options to special section would be problematic as they logically belong to the monitored conditions list. But I'm open to further suggestions.

source/_components/sensor.wunderground.markdown
```diff
@@ -100,19 +103,133 @@ Configuration variables:
...
+    <a name="12h">[12h]</a> - For 12 hour forecasts, you could replace
+    the ```_1d_``` part of the sensor name to get forecast for different
+    period/daytime, e.g. ```weather_2n``` will give you forecast for
+    tomorrow night. Valid values for day are ```1``` to ```4``` and
+    valid values for daytime are ```d``` or ```n```.
+</p>
+
+<p class='note'>
+    <a name="1d">[1d]</a> - For daily forecasts, you could replace the number
+    in ```_1d_``` part of the sensor name only. Valid values are from
```
pezinek: Fixed.

pezinek:
Updated to address the suggestions from the review.

source/_components/sensor.wunderground.markdown
rpitera: I think after reading the "additional examples" this will work fine.  I would just strongly suggest adding a section header "Forecast" after the words "All the conditions listed above will be updated every 5 minutes." to separate them for readability.  Otherwise, I am so anxious to see this merged!  Thank you.

source/_components/sensor.wunderground.markdown
```diff
@@ -136,25 +136,27 @@ Configuration variables:
...
+### 12 hour forecasts
 
-<p class='note'>
-    <a name="1d">[1d]</a> - For daily forecasts, you could replace the number
-    in ```_1d_``` part of the sensor name only. Valid values are from
-    ```1``` to ```4```.  That will give you forecast for different day.
-</p>
+Monitored conditions marked with <a name="12h">[12h]</a> are 12 hour 
+forecasts. To get a forecast for different period/daytime replace the
+`_1d_` part of the sensor name.  e.g. `weather_2n` will give you forecast for
```
Landrash: This one is a mix of 12h and 1d. On purpose of a mistake?

pezinek: This is on purpose, in fact what I call 12h forecast here is called daily forecast by the WU API, the problem is that for some conditions they do provide more fine grained granularity and provide values for day and night, and for some they don't, so I just wanted to make the sensor naming a bit more uniform and let users use the same convention, unless they really do care about what happens outside during nighttime.

source/_components/sensor.wunderground.markdown
pezinek: OK, added the "Forecasts" header, how does it look now ?

rpitera: Great!  Thanks; I think it helped.

pezinek:
Hi, is there something else expected from me before this PR could be approved ?

rpitera:
I think there is just the one change that Landrash suggested.

pezinek:
Actually I believe I addressed all the issues Landrash  pointed out either by fixing the code or by explaining in the comments. Would you mind to point me again to the issues you believe I still have to address ?

rpitera:
The only thing I can see in red is the comment around 115 - "-All the conditions listed above will be updated each 5 minutes with exception of `alerts` that will be updated each 15 minutes by default."

pezinek:
That one is intentional change and not a mistake, as I explained in the comments, so I believe this PR is ready for approval, unless you strongly disagree with my explanation ;-)

rpitera:
`That one is intentional change and not a mistake, as I explained in the comments, so I believe this PR is ready for approval, unless you strongly disagree with my explanation ;-)`

I've already approved.  Just waiting on @Landrash now.

Landrash:
Looks good and can be merged when parent pr is merged.

pezinek:
@Landrash: Thanks !

rpitera:
@pezinek: And thanks to you for this!  I'd almost given up hope of getting this into HA.

PR #2415, opened 2017-04-12
Update raspberry-pi-all-in-one.markdown

DubhAd:
Remove the leading command prompt from the installer command, since it keeps causing confusion in people who're not experienced with command lines.

The latest thread showing this is https://community.home-assistant.io/t/all-in-one-installer-and-hassbian-not-working-help/15413 - however it's not the first and I'm pretty sure I remember catching a few of these on Gitter too.


Landrash:
Where possible the HA documentation always has a leading command prompt to avoid confusion. Having variation in our documentation and separating us from a common way of doing command line documentation is a bad idea.

PR #2422, opened 2017-04-13
add files to gitignore

niklaswa:


eiddor:
Thanks for the update - One of these days, I need to update the cookbook (and sample file) to make .gitignore only reflect files that you want to backup, vs. files that you want to ignore.  Seems safer that way.

PR #2447, opened 2017-04-17
Clarify zone tracking in iOS app

patrickeasters:
Clarify the process for enabling zone tracking in the iOS apps.

**Description:**
When I was initially setting this up, I interpreted the docs as saying `track_ios: true` needed to be in the actual zone configuration. After looking in the gitter chat, I saw a few other folks make this same mistake. I can see the iOS app being something a new user would setup early on, so making this very clear seems like it would be beneficial to all.

Any thoughts @robbiet480? Thanks for all your work on the app!

### DIFF
--- a/source/_docs/ecosystem/ios/location.markdown
+++ b/source/_docs/ecosystem/ios/location.markdown
@@ -34,8 +34,23 @@ At launch, Home Assistant for iOS sets up geofences for all zones in your Home A
 
 ### Configuration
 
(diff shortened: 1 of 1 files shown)


patrickeasters:
Well, there's talk of changing the default behavior to track all zones, so this may need to be changed or scrapped.

robbiet480:
It's gonna be scrapped, build is going up now and will be out within the next 36-48 hours. Thanks for trying to help!

PR #2448, opened 2017-04-17
make this all one detached command

mrdoornbos:
This allows that switch user to be done unattended and makes this script-able.

**Description:**


Landrash:
Commands are separated for readability and for the end users to be able to get a grasp of how the system is set-up.

I could just make everything one-liners but that would be horrible for teaching folks how their own system works.

PR #2453, opened 2017-04-18
Update docs for ssh changes

gurumitts:
**Description:**

This PR will fix issues users with multiple Lutron Caseta devices were having. The root cause of the issue was a difference in IDs being reported by the Lutron Integration report and the values received by the SSH /devices call. This has been resolved by fixing the underlying library to only use SSH. 

Breaking changes:
As a side effect of moving to ssh. The username and password are no longer required.  These value have been removed from the code.  

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#7165


Landrash:
@gurumitts Does this mean the component no longer requires any authentication?

fabaff:
In the configuration check username and password are marked as optional.

gurumitts:
I didn't want to break anything that was already configured.  The username and password are no longer needed.  I switched to using only ssh and it has a hardcoded key that cannot be changed.  Only the IP address is required.

Landrash:
Thank you 🍰

PR #2469, opened 2017-04-22
I needed to install an additional application

Joeboyc2:
I needed to install an additional application for this to work so have added the command above.

This may not be the case for all, but is helpful to have all commands for completeness (at least i think so)

**Description:**

### DIFF
--- a/source/_components/tradfri.markdown
+++ b/source/_components/tradfri.markdown
@@ -20,6 +20,7 @@ For this to work, you need to install a modified lib-coap library:
 
 ```bash
 $ sudo apt-get install libtool
+$ sudo apt-get install autoconf
 
 $ git clone --depth 1 --recursive -b dtls https://github.com/home-assistant/libcoap.git
 $ cd libcoap


source/_components/tradfri.markdown
```diff
@@ -20,6 +20,7 @@ For this to work, you need to install a modified lib-coap library:
 
 ```bash
 $ sudo apt-get install libtool
+$ sudo apt-get install autoconf
```
fabaff: Wouldn't it be an idea to switch to `sudo apt-get install build-essential` as I'm not sure if Debian/Ubuntu/armbian are shipping the compiler?

Joeboyc2: Possibly.... I'm afraid I'm not too familiar with the requirements to agree/disagree

I'm just going by the steps I personally had to follow

fabaff: Let's go with your addition and adjust if needed in the future.

PR #2476, opened 2017-04-22
Fix telegram link and remove old page

### LINKED ISSUE #2474: Telegram Polling website bad link
Heyo
So the link on the bottom of [this page](https://home-assistant.io/components/telegram_bot.polling/) this page redirects to [https://home-assistant.io/components.notify.telegram](https://home-assistant.io/components.notify.telegram) however it should be [https://home-assistant.io/components/notify.telegram](https://home-assistant.io/components/notify.telegram)

Cheers

MartinHjelmare:
**Description:**
* Fix link to notify.telegram
* Remove old webhooks page.
* Fix name of telegram bot event type.

fixes #2474


fabaff:
`source/_components/telegram_webhooks.markdown` should be marked with `ha_category: Deprecated` and a note added. Something like that:

```html
<p class='note warning'>
**This platform has been deprecated in favor of the Telegram Bot "[Webhooks](/components/telegram_bot.webhooks/)" platform and will be removed in the future. Please use the "webhooks" platform.**
</p>
```

There may be still some users out there who are using older releases of Home Assistant.

MartinHjelmare:
Ah ok, I'll fix.

fabaff:
True, we don't really do the "deprecated-waiting-removing" thing as the release cycle are too short. In the latest release the obsolete platform/component is gone ;-)

We didn't spent much thoughts on the category name for the "old" stuff. If you have a better idea, please go ahead and change it.

With redirects we could perhaps get rid of the `Deprecated` category. I would like that: One category less on the components page. If a users what to get details about an obsolete platform/component (link in the forum or somewhere else) he/she is redirected to the new platform/component. We wouldn't break the links and instead of duplicating the content it could just be moved.

balloob:
Sooo, awkward, I just removed that page because I thought it was no longer in use 🤔  should I restore it?

balloob:
btw, not going to merge this PR because there are a lot of changes planned for telegram component (got another cleanup). Closing this for now.

PR #2492, opened 2017-04-23
bind HTTPS to ipv6 port in NGINX example configuration

EyMaddis:
**Description:**

The nginx configuration does not bind port 443 with IPv6 (but does so for 80).

Another thing I could change is the fact that (in my setup), the `sites-available/hass` file must not include the `http {... }` wrapping. I was not sure whether that is specific to my setup (fresh nginx installation via `apt-get`) so I did not change it yet.

### DIFF
--- a/source/_docs/ecosystem/nginx.markdown
+++ b/source/_docs/ecosystem/nginx.markdown
@@ -111,6 +111,7 @@ http {
 
         # These shouldn't need to be changed
         listen 443 default_server;
+        listen [::]:443 default_server ipv6only=on;
         add_header Strict-Transport-Security "max-age=31536000; includeSubdomains";
         ssl on;
         ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
--- a/source/_docs/ecosystem/nginx.markdown
+++ b/source/_docs/ecosystem/nginx.markdown
@@ -89,8 +89,7 @@ http {
         server_name example.com;
 
         # These shouldn't need to be changed
-        listen 80 default_server;
-        listen [::]:80 default_server ipv6only=on;
+        listen [::]:80 default_server ipv6only=off;
         return 301 https://$host$request_uri;
     }
 
@@ -110,8 +109,7 @@ http {
 
 
         # These shouldn't need to be changed
-        listen 443 default_server;
-        listen [::]:443 default_server ipv6only=on;
+        listen [::]:443 default_server ipv6only=off; # if your nginx version is >= 1.9.5 you can also add the "http2" flag here
         add_header Strict-Transport-Security "max-age=31536000; includeSubdomains";
         ssl on;
         ssl_protocols TLSv1 TLSv1.1 TLSv1.2;


source/_docs/ecosystem/nginx.markdown
```diff
@@ -111,6 +111,7 @@ http {
 
         # These shouldn't need to be changed
         listen 443 default_server;
+        listen [::]:443 default_server ipv6only=on;
```
mweinelt: Unless you are on a very old machine you should be able to just
```
listen [::]:443 ssl http2 default_server;
```

As `sysctl net.ipv6.bindv6only` defaults to `0` and so binding to ipv6 will also bind to the ipv6-mapped ipv4 address.

EyMaddis: I am on my Raspberry Pi 3 running a Debian Jessie derivate (OSMC) and that does not work.

> invalid parameter "http2" in /etc/nginx/sites-enabled/default:44

Also I tested it without `http2` and it does not bind to IPv4. 

I verified `net.ipv6.bindv6only = 0`

mweinelt: As per https://trac.nginx.org/nginx/ticket/345 we might need to set `ipv6only=off` then.

When running Jessie the necessary nginx version is only available in the backports I'm afraid.

EyMaddis: I would argue that the current solution is the same as the port handling of port 80 (see above that line in the same file).
Binding to v6 individually works on every version, so I do not really see a reason to change the PR.

mweinelt: Just tested and confirmed working, I don't feel strongly about saving a single line.
```
listen [::]:443 ssl default_server ipv6only=off;
```

We should however note that http2 can be enabled given a recent enough nginx version.

source/_docs/ecosystem/nginx.markdown
```diff
@@ -110,7 +109,7 @@ http {
 
 
         # These shouldn't need to be changed
-        listen 443 default_server;
+        listen [::]:443 default_server ipv6only=off; # if your nginx version is >= 1.9.5 you can also add the "http2" flag here
```
mweinelt: I'm still quite certain this will make 443 a plain http port. Are you sure leaving out the `ssl` flag here is correct?

EyMaddis: Right below is `ssl on` which does the trick.

mweinelt: Ok, cool!

EyMaddis: So is this approved? Can you merge the changes please?

mweinelt: Not a member, just a contributor.

lgtm

EyMaddis: Ah, I see. But I guess you can approve your review, can't you? :)

PR #2496, opened 2017-04-24
Update telegram_command events

Landrash:
Update documentation referring to telegram_command events.

**Description:**
Fix for https://github.com/home-assistant/home-assistant.github.io/issues/2495
@fabaff Can you verify that this is correct before merge?


MartinHjelmare:
It's correct. I included it in #2476, but that's a bigger change so we can merge this first.

PR #2501, opened 2017-04-24
Creation of howto for backup to dropbox

riemers:
**Description:**
A new page for a dropbox sync. Don't know if i have all the bits and pieces correct when it comes to the documentation, so feel free to give me a slap in the right direction if its not 100%. Assuming the page can find a place here.


riemers:
I just noticed, its missing a link to the page itself. Perhaps a subsection for bakup to place the github/dropbox under, there is also another one that uses usb drive i want to create.

PR #2511, opened 2017-04-25
Updated Api.ai logo

arsaboo:
The current logo was not visible against our white background (on the components page) or the gray background (elsewhere on the website).


fabaff:
Even better would be one with a transparent background :-P

PR #2527, opened 2017-04-28
added example config

fredsmith:
Add example configuration.


fabaff:
I will merge it but keep in mind that it will only show up after 0.44.

PR #2541, opened 2017-05-01
Update install path for openzwave config in venv

bruhautomation:
**Description:**

Updates install path set by Hassbian 1.2


balloob:
I wonder if we should keep both paths for people that run the old version?

bruhautomation:
Probably not a bad idea, but I'm not totally sure when this changed. This was just changed in Hassbian 1.2, right? So AIO and <Hassbian 1.1 are still using `/srv/homeassistant/python-openzwave/openzwave/config`?

Landrash:
@bruhautomation Hassbian uses the new path since that's the one used with either the git install or by the pip python_openzwave package.

PR #2542, opened 2017-05-01
added pi zero w blogpost

adgelbfish:


Landrash:
Nice work @adgelbfish

PR #2555, opened 2017-05-03
Change turn on/off lights example to use 'state'

andrewfoster:
The example in the current documentation does not appear to work in hass 0.43.2. A trigger containing 'to' without a 'from' doesn't fire, however one that uses 'state' does.

**Description:**


Landrash:
Thank you :cookie:

PR #2571, opened 2017-05-05
Make manual installation docs consistent with AIO

aaronwolen:
**Description:**
Minor change to the manual installation instructions that places the virtualenv in the same location used by the AIO installer.


Landrash:
Manual install is consistent with HASSbian by design. The AIO chose another path but this won't be changed.

PR #2577, opened 2017-05-05
Update docs for PyPI python-openzwave install.

JshWright:
PYOZW is now installed automatically from PyPI, simplifying the installation process significantly.


source/_docs/z-wave.markdown
```diff
@@ -14,57 +14,17 @@ redirect_from: /getting-started/z-wave/
...
-/usr/local/lib/python3.4/dist-packages/libopenzwave-0.3.0b8-py3.4-linux-x86_64.egg/config
+$ sudo apt-get install libudev-dev
 ```
 
-If you followed along with setting up a virtual environment, your path will be:
-
-```bash
-/srv/homeassistant/python-openzwave/openzwave/config
-```
+_Note: The Home Assistant docker image and the All In One installer have support for Z-Wave already built-in!_
```
Landrash: So does HASSbian.

JshWright: Good point, I'll add that as well

PR #2587, opened 2017-05-07
Rename "Enviro pHat" to "Raspberry Pi Enviro pHat"

ntwb:
When searching the components for "Pi" or "Raspberry" I'd expect the Enviro pHat to be included in those search results.

### DIFF
--- a/source/_components/sensor.envirophat.markdown
+++ b/source/_components/sensor.envirophat.markdown
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: "Enviro pHAT"
+title: "Raspberry Pi Enviro pHAT"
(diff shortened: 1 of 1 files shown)


Landrash:
This change isn't in line with the rest of the components. The components are in general named after the product and the one's we named raspberry pi is part of the actual hardware of the pi.

ntwb:
Sounds fair enough, why does the code review request changes though?

If you want a change you will need to tell me what changes you're happy to make.
If you don't want any changes then close the PR 😄

Landrash:
@ntwb Sorry, my mistake.  Started out with another suggestion.

PR #2588, opened 2017-05-07
Improve Vera component docs

DavidDeSloovere:
Saw a forum post when I was trying to figure this out with the same question, so at least 2 people didn't know how to figure out the entity id. Sure it's easy if you know HA, but for someone who's just starting this might help.

Icon url is the same as in this page: https://home-assistant.io/docs/configuration/devices/

### DIFF
--- a/source/_components/vera.markdown
+++ b/source/_components/vera.markdown
@@ -1,7 +1,7 @@
 ---
 layout: page
 title: "Vera"
-description: "Instructions how to setup Vera hubs within Home Assistant."
+description: "Instructions how to setup Vera hubs and configure devices within Home Assistant."
 date: 2015-03-23 20:04
 sidebar: true
 comments: false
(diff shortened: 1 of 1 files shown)


source/_components/vera.markdown
```diff
@@ -46,3 +48,6 @@ vera:
   lights: [15, 17, 19, 21, 22, 24, 26, 43, 64, 70, 87]
 ```
 
+### {% linkable_title Using z-wave devices in automation %} 
+
+If you want to use a z-wave device from the Vera controller in Home Assistant automation, you'll need the entity id. In the Home Assistant UI you'll find all entities listed under the <img src='/images/screenshots/developer-tool-states-icon.png' alt='service developer tool icon' class="no-shadow" height="38" /> icon of the Developer Tools section. Look for entities that contain 'Vera Device Id' in their attributes, you'll find the entity id on the left.
```
Landrash: Z-Wave shoudl be used instead of z-wave.

DavidDeSloovere: You are correct. I had fixed the spelling in an older piece of text but failed to write it correctly in my own :-|

PR #2593, opened 2017-05-07
[WIP]Update plant.markdown

Landrash:
**Description:**
Add further information to the plant sensor.

### DIFF
--- a/source/_components/plant.markdown
+++ b/source/_components/plant.markdown
@@ -18,7 +18,7 @@ To use your `plant` sensor in your installation, add the following to your `conf
 ```yaml
 # Example configuration.yaml entry
 plant:
-  entity_id:
+  sunflower:
     sensors:
       moisture: sensor.mqtt_plant_moisture
       battery: sensor.mqtt_plant_battery
@@ -54,6 +54,7 @@ Configuration variables:
 ## {% linkable_title Examples %}
 ### Using plain MQTT sensor to get the data
 This is a practial example that uses a multiple of `MQTT sensors` to supply the readings used by the `plant` sensor.
+Another good source of this data would be the [Mi Flora](https://home-assistant.io/components/sensor.miflora/) component. 
 
 
 ```yaml
--- a/source/_components/plant.markdown
+++ b/source/_components/plant.markdown
@@ -11,7 +11,12 @@ ha_category: Other
 ha_release: 0.44
 ---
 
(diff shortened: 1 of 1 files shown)


source/_components/plant.markdown
```diff
@@ -27,9 +32,42 @@ plant:
     min_temperature: 15
 ```
 
-## Using plain MQTT sensor to get the data
+Configuration variables:
+
+- **simulated_plant** (*Required*):
```
frog32: This is an entity_id, I did not try but should be anything that also works elsewhere.

source/_components/plant.markdown
```diff
@@ -27,9 +32,42 @@ plant:
...
+    - **moisture** (*Optional*):  
+    - **battery** (*Optional*):
+    - **temperature:** (*Optional*):
+    - **conductivity:** (*Optional*):
+    - **brightness:** (*Optional*):
+  - **min_moisture** (*Optional*):
+  - **max_moisture** (*Optional*):
+  - **min_battery** (*Optional*):
+  - **min_conductivity** (*Optional*):
+  - **min_temperature** (*Optional*):
```
frog32: Add max_moisture, max_conductivity, min_brightness, max_brightness and max_temperature.
Small detail that I am not sure if it should be included: Temperature is float while the others are positive integers.

source/_components/plant.markdown
```diff
@@ -27,9 +32,42 @@ plant:
...
+
+## {% linkable_title Examples %}
+### Using plain MQTT sensor to get the data
+This is a practial example that uses a multiple of `MQTT sensors` to supply the readings used by the `plant` sensor.
+
```
frog32: Maybe add a reference to https://home-assistant.io/components/sensor.miflora/ since it is the sensor that delivers all those measurements. If we have more sensors in the future we can add them as well.

source/_components/plant.markdown
```diff
@@ -11,9 +11,14 @@ ha_category: Other
...
 
+This`plant`component lets you merge moisture, conductivity, light intensity, temperature and battery level for a plant into a single UI element. It also supports setting minimum and maximum values for each measurement and will change its state to "problem" if it is not within those limits.
+
+To use your `plant` sensor in your installation, add the following to your `configuration.yaml` file:
+
 ```yaml
 plant:
-  simulated_plant:
+  entity_id:
```
frog32: Since this is an example we should use a real plant name instead. This might be confusing.

PR #2617, opened 2017-05-11
Add Openzwave dependency

Kane610:
**Description:**
There is a dependency missing that is needed when compiling openzwave


fabaff (approved):
Thanks :bird:

PR #2618, opened 2017-05-12
LIFX is polling

amelchio:
**Description:**

LIFX bulbs do push their state when booting but not during normal operation.


Landrash:
Thank you :cookie:

PR #2635, opened 2017-05-15
Kodi specific services to call Kodi API methods  

azogue:
**Description:**
Documentation and examples for new Kodi-specific services, including:
 - ~Doc for new service: `kodi_execute_addon` to run a Kodi Addon with optional parameters. Results of the Kodi API call, if any, are redirected in a Home Assistant event: `kodi_execute_addon_result`. ~
 - Doc for new service: `kodi_run_method` to run a Kodi JSONRPC API method with optional parameters. Results of the Kodi API call are redirected in a Home Assistant event: `kodi_run_method_result`. 
 - Add `timeout` parameter to yaml config (needed to make slow queries to the JSONRPC API, default timeout is set to 5s).
 - Examples for using the new services: one simple yaml example to turn on the TV attached to Kodi; and other, more complex, to query the Kodi library and make a dynamic input select.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#7603


source/_components/media_player.kodi.markdown
```diff
@@ -51,3 +52,184 @@ Add music to the default playlist (i.e. playlistid=0).
...
+            addonid: script.json-cec
+            params:
+                command: activate
+```
+
+#### Dynamic input select with Kodi content
+
+A more complex example, to configure a `input_select` that can play in Kodi the latest entries added when selecting them.
+
+It consists of 3 scripts that query the Kodi API to return the contents of the Kodi library, and an AppDaemon app with which the events are heard with the results of the API queries.
```
Landrash: Love the changes but I'm unsure if this last part should be in the component documentation. 
I would prefer it being in the ecosystem section.
@fabaff what to you think?

azogue: You mean the complex example, right? Where and how do you think it would be more convenient to place it?

(I've removed the `kodi_execute_addon` service, as discussed in the home-assistant/home-assistant#7603)

armills: Agree with @Landrash myself. I like having the simple example, but the big one I'd prefer elsewhere. Maybe in the cookbook with a link here?

azogue: @Landrash , @armills 
I have moved the complex example to the cookbook, but I'm not sure about metadata, links and so on, what do you think?

source/_cookbook/automation_kodi_dynamic_input_select.markdown
```diff
@@ -0,0 +1,156 @@
...
+  Media Select:
+    control: hidden
+    entities:
+      - input_select.kodi_results
+      - script.get_recent_movies
+      - script.get_recent_episodes
+      - script.get_pvr_channels
+```
+
+The AppDaemon app:
```
Landrash: Could you add in some more about the application and maybe some filler text about how to install it?
Not needed but considering this can be the first exposure to AppDaemon for the reader.

armills: This would be good, but it can come in a follow-up PR. Please link it to this PR if you can add some more text.

PR #2640, opened 2017-05-16
Transition updated

boojew:
According to this https://home-assistant.io/cookbook/automation_sun/ (and my own experience) transition appears to only be supported on Hue bulbs - so I have updated the Transition note.

**Description:**

### DIFF
--- a/source/_components/light.markdown
+++ b/source/_components/light.markdown
@@ -33,7 +33,7 @@ Turns one light on or multiple lights on using [groups]({{site_root}}/components
 | Service data attribute | Optional | Description |
 | ---------------------- | -------- | ----------- |
 | `entity_id` | no | String or list of strings that point at `entity_id`s of lights. Else targets all.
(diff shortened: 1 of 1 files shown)


Danielhiversen:
It seams that several light platforms support transition: https://github.com/home-assistant/home-assistant/search?utf8=%E2%9C%93&q=transition&type= ?

Landrash:
@Danielhiversen is correct.
I've changed the language to be a bit more ambiguous.
This ok with you @boojew ?

boojew:
What do you guys think about actually listing the supported platforms or is that a bad idea as it wont get updated in a timely manner?

Thanks

Landrash:
I would prefer the platforms specifying this themselves.
Need to look it over a bit and see if I can come up with a good solution.

Landrash:
I'm at a loss.
The list you've changed is a bad place to update since it's unlikely that it will be updated.
Some platforms has references to the functionality but not all so assuming that the platforms will have the information is also bad. Question is if we should go trough them all and add the information in question.

I'll merge this as I've changed it and open an issue instead.

Issue #1887, opened 2017-01-27
Development break with ruby 2.4


adrianlzt:
Gem yajl-ruby should be set to version 1.3.0 to [be able to compile](https://github.com/brianmario/yajl-ruby/pull/165) with ruby 2.4.

Jekyll gives a warning:
```
/usr/lib/ruby/gems/2.4.0/gems/jekyll-3.2.1/lib/jekyll/hooks.rb:57: warning: constant ::Fixnum is deprecated
```

And maybe the docs should change:
``rake preview``by ``bundle exec rake preview`` and ``rake generate`` with ``bundle exec rake generate`` to solve this error:
```
Gem::LoadError: You have already activated rake 12.0.0, but your Gemfile requires rake 10.5.0. Prepending `bundle exec` to your command may solve this.
```

balloob:
Pull request welcome 👍

fabaff:
#2644 was merged and should fix this issue as well.

PR #2644, opened 2017-05-16
Updated pygments.rb to latest to make it build under ruby 2.4

auduny:
**Description:**

On latest OS X, default ruby is 2.4. Older version of Pygments.rb depends on a version of yajl-ruby that is not buildable on latest ruby 2.4. Updated pygments to latest version. Issue #1887 seems to refer to this issue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 group :development do
   gem 'rake', '~> 10.0'
   gem 'jekyll', '~> 3.0'
-  gem 'pygments.rb', '~> 0.6.3'
+  gem 'pygments.rb', '~> 1.1.2'
   gem 'rdiscount', '~> 2.0'
   gem 'RedCloth', '~> 4.2'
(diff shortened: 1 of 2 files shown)


Landrash:
Does this change affect the availability of building on any of the other platforms?

auduny:
@Landrash Nope I do not think so. I just built it on Ubuntu 16.04 with ruby 2.3.1 Seems to be completely fine.

Landrash:
OK with you @fabaff ?

tiktok7:
Just found this problem the hard way.. tried 2.4 then found this when I was going to log the issue. It might be kinder to newbies to merge the change... :wink:

fabaff:
Works on an older version of ruby too.

```bash
$ ruby -v
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-linux]
```

balloob:
We should probably try to update all our gems.

Issue #2310, opened 2017-03-21
More metadata for components and filter in searches


Zac-HD:
I can't be the only person who has used [the component docs](https://home-assistant.io/components/) as a shopping list - it's a great resource, and you can be confident that supported devices are (a) supported by homeassistant and therefore (b) not locked down to the point of uselessness.

However, it would be even better if components were tagged with and searches could be filtered by a number of attributes (stored in eg a `tags:` dict in the header).

- discoverable: true/false
- requires_internet: always/setup/never
- official_api: true/false
- iot_device_security: ok/weak/broken (if relevant)

These are example properties, but they illustrate the idea.  Thoughts?

bdurrer:
The discoverable components can easily be found on the discovery component page.
For the others... one would need to crawl through those 603 components and add the required metadata.

Personally I don't see all too much value in the _suggested_ properties, but I see your point. A lot of components are assigned to an category but would fit into multiple categories. E.g. an camera being a _sensor_ and, well, a _camera_
... so yeah, tagging in general would help a lot


I would happily extend the search feature, but I'm not the one going to add tags to the components 😄

Zac-HD:
Sounds good!  I agree that there should be some more discussion about what tags we should use (they're just examples).  And if someone else can write the search function, I'm willing to do the tagging.  It would be good to have Travis check that the header is in order though, to ensure that the tags stay up-to-date and new components are tagged too.

petersohn:
I would be happy to be able to search by IoT class.

balloob:
Yeah, we also already fill in the IoT class, so that should be easy.

Another one I would like to see is a url-only search option to search by version. Right now for each release blog post I link to the "released in current version" section, but each new release will have old release blog posts link to the wrong components.

bdurrer:
@balloob I'll add that. Is there any reason to keep the three `#added_in_...` tags, if the overview links to the version hashes instead?

balloob:
No, we can drop those categories if we can just generate links for the last 3 versions.

bdurrer:
So, here is the thing: adding more metadata to the search does not improve it's quality. searching for "polling" looks like garbage, because you can't see _why_ you got those results.

There are two ways I see right now:

1. add grouped items into the sidebar and maybe add some headers (eg versions, categories, IoT class)
2. make the search "smart" with meta tags, eg you can enter `iot: polling` to only filter for iot classes.

Problems:
[...]

PR #2646, opened 2017-05-16
Issue 2310: Release filter on component page

bdurrer:
**Description:**
As discussed in issue #2310 
This PR removes the "dynamic" hashtags "current/previous/2 version ago" and replaces them with versioned tags in the pattern `#version/{number}`. That way "permalinks" to specific versions can be created.
This is works for older versions too. However, a long time ago in a galaxy far, far away, not all components had a "introduced in version X" meta tag and linking to those versions has not the desired effect 😄 


balloob:
🙇  this is amazing.

PR #2663, opened 2017-05-19
Fix pacaur invocation in Arch installation instructions

zoresvit:
Pacaur does not work with root privileges for security reasons:

```
$ sudo pacaur -S home-assistant
[sudo] password for zoresvit: 
:: you cannot perform this operation as root
```
It will only ask for a password interactively when invoking `makepkg`.


Landrash:
Your correct! Thank you. :cookie:

PR #2688, opened 2017-05-22
Small bug in initial developer docs modifying a variable out of scope

jessexoc:
Fixes `NameError: name 'count' is not defined` when going through the development:events tutorial


Landrash:
Thank you :cookie:

PR #2689, opened 2017-05-22
Update development_environment.markdown

bzillins:
**Description:**


balloob:
I just realized that instructions are the same for both.

PR #2692, opened 2017-05-23
Added example for raw/endraw

nodomain:
Added "raw/endraw" example with my learnings from https://github.com/home-assistant/home-assistant.github.io/pull/2685. I found the docs not intuitive enough and eventually got it by looking at the source code...

**Description:**
The docs now show an explicit example of how to avoid parsing of YAML variables in docs.

### DIFF
--- a/source/developers/website.markdown
+++ b/source/developers/website.markdown
@@ -88,7 +88,14 @@ You can use the default markdown syntax to generate syntax highlighted code. For
 
 Note that you can replace `yaml` next to \`\`\` with the language that is within the block.
 
-Please add `{% raw %} ... {% endraw %}` around variables and other things that should not be rendered on the website:
+When you're writing code that is to be executed on the terminal, prefix it with `$`.
+
+### {% linkable_title Templates %}
+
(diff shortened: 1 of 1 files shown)


source/developers/website.markdown
```diff
@@ -94,7 +94,24 @@ When you're writing code that is to be executed on the terminal, prefix it with
...
+  - alias: "Example notification"
+    trigger:
+      - platform: state
+        entity_id: sensor.example
+    action:
+      - service: notify.notify
+        data:
+          title: "Test"
+          message: >-
+            {% raw %}{% if is_state("sensor.example", "test") %}
```
fabaff: The issue is that this way the template will not be rendered but you still can't see where `{% raw %}` and `{% endraw %}` are placed without looking at the source of the page. We tried to solve already but there was no smart way to do it. We decided not to create an image as the provided link show a couple of examples.

nodomain: LOL. Got ya. I should stop contributing in the morning commute train before my coffee 😂

PR #2699, opened 2017-05-23
Add more info about the /api/history/period route

bokub:
**Description:**

The [documentation](https://home-assistant.io/developers/rest_api/#get-apihistoryperiodlttimestamp) about the `/api/history/period` route is too light:

 - we don't know if `<timestamp>` is the beginning or the end of the period
 - the `end_time` parameter is in the code but nowhere in the doc
 - we don't know the default period length

I added some precisions


bokub:
**P.S: Rendered preview:**

----

capture

fabaff:
@bokub, feel free to submit extensions/additions to the existing documentation against the `current` branch. This way it would be online immediately.

PR #2706, opened 2017-05-24
Proposed update to yaml config

SilvrrGIT:
I have not read the docs extensively but the yaml config example given seems incorrect.  It have adjusted to to what I believe is the correct config for the automation example given.  I am not sure if things changed with the new Automation editor but I don't want people struggling with configs due to bad documentation.

**Description:**


bbrendon:
thanks. when i got started i found a lot of typos in the yaml i got from the docs. Hope you got it right and that it helps other people.

PR #2711, opened 2017-05-25
Remove platform specific information from light component

amelchio:
**Description:**

The text is rewritten to not favor any particular platform and the
`light.hue_activate_scene` service call is moved to the Hue documentation.


Landrash:
Looks good. Thank you :cookie:

PR #2713, opened 2017-05-25
Updating documentation for REST switch with template example

cribbstechnologies:
**Description:**

I've updated the documentation to include an example with `is_on_template` and fixed a typo in the current docs


source/_components/switch.rest.markdown
```diff
@@ -30,10 +30,34 @@ Configuration variables:
...
+```json
+{"is_active": "true"}
+```
+
+
+```yaml
+switch:
+  - platform: rest
+    resource: http://<address>/led_endpoint
+    body_on: '{"active": "true"}
```
Landrash: Is the ' there on purpose or are we lacking one?

source/_components/switch.rest.markdown
```diff
@@ -30,10 +30,34 @@ Configuration variables:
...
+    body_off: '{"active": "false"}
```
Landrash: Is the ' there on purpose or are we lacking one?

source/_components/switch.rest.markdown
```diff
@@ -30,10 +30,34 @@ Configuration variables:
...
+- **is_on_template** (*Optional*): A [template](/topics/templating/) that determines the state of the switch from the value returned by the GET request on the resource url. This template should compute to a boolean (True or False). Default is equivalent to `'{% raw %}{{ value_json == body_on }}{% endraw %}'`. This means that by default, the state of the switch is on if and only if the response to the GET request matches `body_on`.
 
+<p class='note info'>
+When Home Assistant is trying to update the state of the device by issuing a GET to the endpoint, the value returned from the endpoint is made available to the template as `value_json`. Values within the JSON can be referenced using dot notation:
+`{"state": "off"}` would be referenced as `{% raw %}{{value_json.state}}{% endraw $}`
+</p>
```
fabaff: The info part looks like the re-phrased details from `is_on_template` and we should start to add details to the platforms which are already mentioned in the template documentation. If it's unclean them the template documentation should be expanded.

source/_components/switch.rest.markdown
```diff
@@ -30,10 +30,34 @@ Configuration variables:
...
 <p class='note warning'>
 Make sure that the URL matches exactly your endpoint or resource.
 </p>
 
+
+## {% linkable_title Example %}
+
+### {% linkable_title Switch with templated value %}
+
+This example shows a switch that uses a [template](/topics/templating/) to allow Home Assistant to determine its state. In this example the REST endpoint returns this JSON response with true indicating the switch is on
```
fabaff: Period is missing at the end.

source/_components/switch.rest.markdown
cribbstechnologies: I specifically asked @balloob if I should add this info in some overarching place and he said if it wasn't in the component people wouldn't find it. I've pushed this block into the is_on_template definition but it looks crowded now.

fabaff: I don't know how many platforms are supporting templates nowadays almost all contains a link to the template documentation. The details for accessing/processing incoming data is a topic that concerns very platform and is not limited to REST.

cribbstechnologies:
I didn't mean to do this and don't know what caused it.

cribbstechnologies:
The link to the template page is already there. I removed the extra
explanation but I've now answered three questions in the community about
REST-related components and how to parse the values in the template. It's
not readily obvious to newcomers that value_json is how the data will be
returned and not just an example. I'm unsure how this should best be
addressed.

PR #2774, opened 2017-06-05
Add service_url config option to volvooncall

palp:
**Description:**
Doc change for https://github.com/home-assistant/home-assistant/pull/7919

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#7919


palp:
I snipped the app traffic to check for other URLs and found China has a different one as well, so I added an example for that.

PR #2822, opened 2017-06-15
Create notify.group.markdown

PuckStar:
**Description:**
Took a while before I found out, everything must be in lower case here.
Even if you have it with capitals in the individual services.


balloob:
Thanks!

PR #2828, opened 2017-06-16
Add network_key configuration

hamid-elaosta:
Network key can be set here rather than in the (difficult to find) configuration `options.xml`. It takes the same comma separated "0xNN" string syntax as previous.

**Description:**

### DIFF
--- a/source/_docs/z-wave.markdown
+++ b/source/_docs/z-wave.markdown
@@ -91,18 +91,6 @@ To add a Z-Wave device to your system, go to the Services menu and select the `z
 
 Security Z-Wave devices require a network key before being added to the network using the `zwave.add_node_secure` service. You must set the *network_key* configuration variable to use a network key before adding these devices.
 
-Edit your `options.xml` file:
-
-```bash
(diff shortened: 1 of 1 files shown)


dale3h:
Looks good! 🎃

arsaboo:
Just a quick question - If I already have network_key in the options.xml, can I just remove it from there and paste it in the configuration.yaml under zwave?
```yaml
zwave:
  usb_path: /dev/ttyACM0
  new_entity_ids: true
  network_key: !secret zwave_network_key
```

dale3h:
@arsaboo I am not entirely sure, but my gut tells me yes, you can do this. Please give it a shot and let us know!

arsaboo:
So, I commented out the network_key from options.xml and everything still works.

hamid-elaosta:
Did you also add it to configuration as well or just comment it out? If the latter, I assume (although maybe @dale3h can comment), that it might continue to work but future devices cannot be added securely.

I found that without a key, you will get an error/warning on startup pointing out that you do not have a key.

arsaboo:
I did add it to the configuration, as indicated in my previous post.

PR #2834, opened 2017-06-18
Update camera.rpi_camera.markdown

ibenrodriguez:
fix typo - removed the leading dash before the keyword platform at line 24.
   platform: rpi_camera

**Description:**
removed the leading dash before the keyword platform at line 24.
   platform: rpi_camera


Landrash:
It's not a typo actually.  If your using more than one platform you indent it and add the dashes.
Quite a bit of the documentation is written with cut and past in mind. Assuming the end user will forget things like this help minimize the amount of support the community has to provide.

PR #2835, opened 2017-06-18
Update device_tracker.bluetooth_tracker.markdown

ibenrodriguez:
fix a typo: remove leading - (hyphen or dash) prior to the platform entry at line 23.

**Description:**
fix a typo: remove leading - (hyphen or dash) prior to the platform entry at line 23.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant)


ibenrodriguez:
Thanks Fabian. There must be another issue I can track down because the bluetooth device tracker isn't working on the raspberry pi 3. I was getting an error on the config for that line but now, even with no config error, the bluetooth devices are not showing up with the passive scans.  Must be something else...

WARNING (MainThread) [homeassistant.setup] Setup of device_tracker is taking over 10 seconds.

PR #2859, opened 2017-06-20
Add description of name

mjj4791:
Added a description  of the name configuration item and its impact on the entity names used by homeassistant

### DIFF
--- a/source/_components/weather.buienradar.markdown
+++ b/source/_components/weather.buienradar.markdown
@@ -22,7 +22,6 @@ To add the buienradar weather to your installation, add the following to your `c
 # Example configuration.yaml entry
 weather:
   - platform: buienradar
-  - name: buienradar
 ```
 
 Configuration variables:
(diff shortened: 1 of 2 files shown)


mjj4791:
No parent pr 
The component is live; just clarification after user feedback

Michiel

source/_components/weather.buienradar.markdown
```diff
@@ -22,10 +22,13 @@ To add the buienradar weather to your installation, add the following to your `c
 # Example configuration.yaml entry
 weather:
   - platform: buienradar
+  - name: buienradar
```
Landrash: Since the name is an optional parameter is shouldn't be included in the most basic configuration example.

mjj4791:
OK; name removed from minimal config example

mjj4791:
**Question:**
I changed the **logo** and **ha_release** in a previous commit.
I now see the logo in the documentation live now, but the component does not show up ad added in 0.47..... (when docs were created it was set to 0.44 and never got updated to the actual version the component got introduced, so now it is not showing up in the recently added....)

mjj4791:
OK; closing this pr and creating a new one

PR #2860, opened 2017-06-20
Clarification on contents of authorized_keys file Update debugging.markdown

ozarchie:
**Description:**

### DIFF
--- a/source/hassio/debugging.markdown
+++ b/source/hassio/debugging.markdown
@@ -13,7 +13,7 @@ The following debug tips and tricks are for people who are running the Hass.io i
 
 ## {% linkable_title SSH access to the host %}
 
-Create an `authorized_keys` file in the root of your SD card with your public key. Once the device is booted, you can access your device as root over SSH on port 22222.
+Create an `authorized_keys` file in the root of your SD card with your public key. Once the device is booted, you can access your device as root over SSH on port 22222. The contents of the 'authorized_keys' file should be the contents of your 'id_rsa.pub' file.
 
 Windows instructions how to generate and use private/public keys with Putty are [here][windows-keys]. Instead of the droplet instructions, add the public key as per above instructions.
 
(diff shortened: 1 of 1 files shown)


source/hassio/debugging.markdown
```diff
@@ -13,7 +13,7 @@ The following debug tips and tricks are for people who are running the Hass.io i
 
 ## {% linkable_title SSH access to the host %}
 
-Create an `authorized_keys` file in the root of your SD card with your public key. Once the device is booted, you can access your device as root over SSH on port 22222.
+Create an `authorized_keys` file in the root of your SD card with your public key. Once the device is booted, you can access your device as root over SSH on port 22222. The contents of the 'authorized_keys' file should be the contents of your 'id_rsa.pub' file.
```
Landrash: The filename is set when the key files are created and `id_rsa.pub` would assume the end user didn't change this during setup.

ozarchie: Sometimes these keys are created by third party subscriptions e.g. github(.rsa.pub). many users now 'subscribe' using their github and/or other public keys.

Landrash: That's true but I'm just a tad unsure if you really should specify the file name since it's not necessarily correct?

ozarchie:
Oh OK. I thought that id_rsa.sub was a placeholder for the actual name. My
mistake. Cheers John

source/hassio/debugging.markdown
ciotlosm: If you really don't want to reuse a file, than we should include in the docs the complete workflow to create a specific key for the hass.io installation.

As hass.io won't have any display output, I would also recommend moving this inside the "Installation" section. I didn't notice this at start, but I also just opened a pull request for something similar: https://github.com/home-assistant/home-assistant.github.io/pull/2870.

PR #2870, opened 2017-06-22
Add instructions around authorized_keys

ciotlosm:
As you don't have display output and as you can't ssh using username&password I recommend you add this in the installation steps.

I know there is the option to use the ssh server in the addons, but that is only available after you blindly wait up to 10 minutes with no clue if it's going to work.

**Description:**


Landrash:
@pvizeli Any work on this PR or should it be closed?

pvizeli:
We do not maintain the developer access to host. It stay on developer page and that is okay. I see a lot bugs that only come while user do stuff and break the supervisor.

PR #2877, opened 2017-06-23
Fix binary_sensor.template example

sdague:
The binary_sensor.template was return states of "On" and "Off",
however that doesn't seem to work in 0.47. You need True / False,
which makes sense, given that's what the single line evals are going
to return. Update the existing example, and add another one about
computing composite occupancy from device tracker and motion sensors.


sdague:
Appologies for some extra white space fixing, my editor does a whitespace clean on save so the random characters get purged. It usually makes patches better, but on docs where whitespace at end of lines is sometimes forgotten, it gets dorky.

balloob:
Jinja2 templates need to be wrapped in `{% raw %}` tags or the jekyll build will fail (that's why CI failed). Fixed it in 2189bbda0301d73227f14053744aba7206386d1b 👍

PR #2914, opened 2017-07-02
Additional documentation

open-homeautomation:
Added documentation for additional parameters introduced by a separate change request (already included in 0.48) 

KNX should be moved from the "DIY" section as KNX isn't a DIY solution, but an industry standard used for building automatisation.

**Description:**

### DIFF
--- a/source/_components/cover.knx.markdown
+++ b/source/_components/cover.knx.markdown
@@ -22,15 +22,8 @@ To use your KNX covers in your installation, add the following to your `configur
 # Example configuration.yaml entry
 cover:
   - platform: knx
-    name: Bedroom
     updown_address: 9/0/0
     stop_address: 9/0/1
-    setposition_address: 2/62
-    getposition_address: 2/64
-    setangle_address: 2/63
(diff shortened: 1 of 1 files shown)


source/_components/cover.knx.markdown
```diff
@@ -22,13 +22,23 @@ To use your KNX covers in your installation, add the following to your `configur
...
   - platform: knx
+    name: Bedroom
     updown_address: 9/0/0
     stop_address: 9/0/1
+    setposition_address: 2/62
+    getposition_address: 2/64
+    setangle_address: 2/63
+    getangle_address: 2/65
+    invert_position: True
+    invert_angle: True
```
fabaff: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

open-homeautomation:
Additional attributes have been removed now.

PR #2940, opened 2017-07-06
Fix Lack of Space after Android Chrome

SConaway:
**Description:**
Fix Lack of Space after Android Chrome in source/_faq/after-upgrading.markdown


Landrash:
Thank you :cookie:

PR #2945, opened 2017-07-07
Update downloader documentation with `subdir` data attribute

rollbrettler:
**Description:**
According to [the code](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/downloader.py#L31) it is also possible to download a file into a subdirectory of **download_dir**


balloob:
Thanks! 👍

PR #2947, opened 2017-07-07
Optional variables added to sensor.cups.markdown

trux66:
Sensor configuration supports two additional variables which are not listed but would be extremely helpful to anyone running their print server on a host other than the one running Home Assistant.

Added two lines to the Configuration variables section to show host and port optional values.

**Description:**


Landrash:
@trux66 Is this something that works now?

PR #2963, opened 2017-07-12
Initial media_extractor docs

fabaff:
**Description:**
Initial media_extractor docs.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#8369


arsaboo:
The docs suggest that an example should be there, but it seems it is missing. It will be good to provide an example so that people know how to use this service.

PR #2967, opened 2017-07-12
LIFX: make broadcast address configurable

amelchio:
**Description:**

The LIFX broadcast address can now be set.

While at it, remove `server` from the configuration example because listing it implies that it is needed while in reality it rarely is. If you have more than one network interface you probably know what you are doing and can figure out how to set it.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#8453

### DIFF
--- a/source/_components/light.lifx.markdown
+++ b/source/_components/light.lifx.markdown
@@ -21,11 +21,11 @@ _Please note, the `lifx` platform does not support Windows. The `lifx_legacy` pl
 # Example configuration.yaml entry
 light:
   - platform: lifx
-    server: 192.168.1.10
 ```
 Configuration variables:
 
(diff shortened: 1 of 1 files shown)


Landrash:
@amelchio I would prefer you keep server and change it's description. It's always better to have all the variables documented.

amelchio:
Well, it is still documented (with a changed description) but just not listed in the YAML example. It's a bit unclear to me whether you missed that or you actually do think that all variables should be in the example?

I think too many variables in the examples is not always good. I have seen several users copy/paste the example, skip the description and then ask "what should I set `server` to" even though the description says to omit if unsure.

How about I add a second example so we have the basic and advanced setups separated and people can copy/paste the simple one?

Landrash:
@amelchio That was me missing that you moved it. My mistake and I'm sorry about that :cocktail:

PR #2972, opened 2017-07-13
Replace Samsung with Sharp Aquos

halkeye:
**Description:**

Update the docs to say aquos tv instead of samsung smart tv


Landrash:
Thank you :cookie:

halkeye:
\o/ Yay I'm helping.

PR #2982, opened 2017-07-15
Added documentation for the env_var tag in configuration files

devspacenine:
Added documentation for the !env_var tag in configuration files, including a default value parameter as described in the pull request below.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** home-assistant/home-assistant#8484

### DIFF
--- a/source/_docs/configuration/yaml.markdown
+++ b/source/_docs/configuration/yaml.markdown
@@ -61,3 +61,20 @@ sensor:
     state_topic: sensor2/topic
 ```
 
+### {% linkable_title Using Environment Variables %}
+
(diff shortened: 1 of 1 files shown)


Landrash:
@devspacenine Just for discussion. Wouldn't it be better to include this information [here](https://home-assistant.io/docs/configuration/secrets/)?

fabaff:
I tend to agree partially with @Landrash but I would suggest to go an additional step. The usage of `!env_var` is an advance feature thus we should place it under "Advanced Configuration" more as its own section than attaching it to "Storing Secrets".

PR #2988, opened 2017-07-16
Add documentation on using credstash as a secret store

justin8:
**Description:** Add documentation for [credstash support](https://github.com/home-assistant/home-assistant/pull/8494)

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#8494

### DIFF
--- a/source/_docs/configuration/secrets.markdown
+++ b/source/_docs/configuration/secrets.markdown
@@ -89,3 +89,26 @@ Please enter password for encrypted keyring:
 <p class='note warning'>
   If you are using the Python Keyring, [autostarting](/getting-started/autostart/) of Home Assistant will no longer work.
 </p>
+
+### {% linkable_title Storing passwords in a keyring managed by your OS %}
+
+Using [Credstash](https://github.com/fugue/credstash) is an alternative way to `secrets.yaml`. They can be managed from the command line via the credstash script.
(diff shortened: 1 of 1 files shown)


Landrash:
I think this would be better suited for the [ecosystem](https://home-assistant.io/docs/ecosystem/) to avoid diluting the documentation with non base functionality?

justin8:
It appears that ecosystem is for functionality that is outside of core.

This works similar to keyring support for secrets, which is in core and in the paragraph above this documentation.

fabaff:
I guess that sooner or later we need to create a section for the [scripts](https://home-assistant.io/docs/tools/scripts/). I would be fine if we place it inside the "Storing Secrets" docs for now and move the keyring and the credstash later.

justin8:
+1 for that. It needs some cleanup, but that should be a separate CR. Right now it places it with the most logically similar information

Landrash:
Agree wth fabaff. Merging as is and in the future this might get moved,

MartinHjelmare:
The parent PR is not merged yet.

Landrash:
My mistake, sorry about that.

balloob:
About to merge it, will revert the revert 👍

PR #2994, opened 2017-07-16
Change incorrect domain to apple_tv

postlund:
**Description:**
Minor typo in last change.


schmittx:
These docs still show Apple TV as a media player platform but the blog release for 0.49.0 shows Apple TV as it's own component (separate from media player). Which configuration is correct?

PR #3017, opened 2017-07-19
Kodi turn_on_action and turn_off_action as scripts documentation

azogue:
**Description:**

Update Kodi documentation with new script syntax for `turn_on_action` and `turn_off_action` configuration parameters.
Add a few examples of the new usage, including the migration equivalencies.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#8558


armills:
Can you maybe create this with a new branch? It seems that somehow Travis got locked up on these branches.

I think the problem is that the compiler was actually trying to parse the templates for the website. Can you update it to use the raw syntax as used here? https://github.com/home-assistant/home-assistant.github.io/blob/c071586d1d2fbb01ef8b8ddd909751f099cd92e0/source/_docs/configuration/templating.markdown

azogue:
ok, @armills, didn't know about the `{% raw %}` syntax. I'll make it again

PR #3019, opened 2017-07-19
Added steps to run installer by hand

haXs:
Had an issue out of the box, got some help on Discord to run the installer manually and was asked to put it into the docs

**Description:**


Landrash:
Since it currently seems to fail more often for some folks this seems like a good compromise.
Can't figure out why it's failing since it works 20/20 times for me.

haXs:
For some reason, my Pi did not obtain a network address until I unplugged and re-plugged my cable, so it never had an IP to complete the install and it never tries again if it failed the first time

PR #3049, opened 2017-07-24
Add info about preferable way to add device to z-wave network

minchik:
This update should prevent issues like this https://github.com/home-assistant/home-assistant/issues/7076 in future


Landrash:
Merging into next since next release aint that far away.

PR #3070, opened 2017-07-27
Add note about discovery for sabnzbd.

Hellowlol:
**Description:**
Add that you need to enter sabnzbd API Key in the configurator, this will enable a minimal setup. The user should use the yml for more advanced stuff.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** https://github.com/home-assistant/home-assistant/pull/8634


source/_components/sensor.sabnzbd.markdown
```diff
@@ -33,6 +33,8 @@ sensor:
     - 'disk_free'
 ```
 
+If SABnzbd is discovered you can enter your API Key in Configurator and `sensor.sabnzbd_status` will be added.
```
Landrash: More information that about that this is limited should probably added.

Hellowlol:
@Landrash what about something like this?

PR #3105, opened 2017-07-30
value_template: Leave this out when using the appdaemon to set the value

vrs01:
**Description:**


Landrash:
We should avoid diluting the documentation for home-assistant with third party components where possible.
This documentation should be with the documentation provided by Appdeamon.
@fabaff  Would you agree?

vrs01:
ok I will put it into the documentation of appdaemon.

PR #3106, opened 2017-07-30
Template sensor: Don't put  'value_template'

vrs01:
When you want to set the value of a template sensor in Homeassistant, leave out the 'value_template' line in the configuration of the sensor.

**Description:**


vrs01:
not shure if this is the right way to use the template sensor.

PR #3119, opened 2017-08-02
Update device-specific.markdown

waxhell:
Use entity_ids for Minimote example rather than obsolete object_id


Landrash:
Copy of https://github.com/home-assistant/home-assistant.github.io/pull/3038

Landrash:
Can @waxhell & @mcginnie come to a consensus of how you would like this change implemented since there's a discrepancy on how you want the names. Since you both have submitted PR's I'm ok with either of you reviewing these and suggesting a solution.

waxhell:
zwave.XXXX should be required when using "entity_id" -- the other PR shouldn't work properly if tested.

Landrash:
I'll take your word for it :slightly_smiling_face: 
Thank you :cookie:

PR #3157, opened 2017-08-09
Added documentation to HA PR#8505

mar-schmidt:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#8505

### DIFF
--- a/source/_components/sensor.command_line.markdown
+++ b/source/_components/sensor.command_line.markdown
@@ -95,7 +95,7 @@ logger:
 
 ### {% linkable_title Details about the upstream Home Assistant release %}
 
-You can see directly in the frontend (**Developer tools** -> **About**) what release of Home Assistant you are running. The Home Assistant releases are available on the [Python Package Index](https://pypi.python.org/pypi). This makes it possible to get the current release.
+You can see directly in the frontend (**Developer tools** -> **About**) see what release of Home Assistant you are running. The Home Assistant releases are available on the [Python Package Index](https://pypi.python.org/pypi). This makes it possible to get the current release.
 
 ```yaml
 sensor:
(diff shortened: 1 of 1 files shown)


source/_components/sensor.command_line.markdown
```diff
@@ -144,3 +144,17 @@ sensor:
     command: "python3 /path/to/script/arest-value.py"
     unit_of_measurement: "°C"
 ```
+
+### {% linkable_title Usage of templating in command %}
+
+command: supports inclusion of templates. Which means you can use various states from other sensors in your commands.
```
Landrash: Formatting looks a bit off here.

source/_components/sensor.command_line.markdown
```diff
@@ -144,3 +144,17 @@ sensor:
...
+This is good for example if you need to include a state of a specific sensor as an argument to your cmd script.
```
Landrash: Avoid using short hand. In this case `cmd` should be extended to command.

source/_components/sensor.command_line.markdown
```diff
@@ -144,3 +144,17 @@ sensor:
...
+
+```yaml
+# Example entry for `configuration.yaml` (if applicable):
+sensor:
+  - platform: command_line
+    name: yr wind direction cmd
```
Landrash: I suggest you simplify the name to `wind direction` for the example since new users might not know what yr is.

source/_components/sensor.command_line.markdown
mar-schmidt: Not quite sure what you are referring to here, however, I have revised the text and changed small parts in my new commit

PR #3166, opened 2017-08-10
Add HipChat notify service.

BioSehnsucht:
**Description:**
Documentation for HipChat notification component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#8918

I'm making an assumption in the docs that the `ha_release` will be 0.51. I didn't find any recommendation for what you should put when it's not released yet.


BioSehnsucht:
I just noticed that a 0.51 PR was created about the same time I was making these PR's. Let me know if I should bump the `ha_release` in the documentation page to something else (or make it something other entirely).

source/_components/notify.hipchat.markdown
```diff
@@ -0,0 +1,53 @@
...
+title: "HipChat"
+description: "Instructions how to add HipChat notifications to Home Assistant."
+date: 2017-08-10 15:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: hipchat.png
+ha_category: Notifications
+ha_release: "0.51"
```
Landrash: Bump to 0.52 as suggested :+1:

source/_components/notify.hipchat.markdown
```diff
@@ -0,0 +1,53 @@
...
+---
+
+
+The `hipchat` platform allows you to deliver notifications from Home Assistant to [HipChat](https://hipchat.com/).
```
Landrash: Replace deliver with send?

source/_components/notify.hipchat.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+To enable the HipChat notification in your installation, add the following to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+notify:
+  - name: NOTIFIER_NAME
+    platform: hipchat
+    token: ABCDEFGHJKLMNOPQRSTUVXYZ
+    room: '1234567'
```
Landrash: Are the quotes around the number necessary?

source/_components/notify.hipchat.markdown
```diff
@@ -0,0 +1,53 @@
...
+```
+
+Configuration variables:
+
+- **name** (*Optional*): Setting the optional parameter `name` allows multiple notifiers to be created. The default value is `notify`. The notifier will bind to the service `notify.NOTIFIER_NAME`.
+- **token** (*Required*): The HipChat API token to use for sending HipChat notifications.
+- **room** (*Required*): The default room to post to if no room is explicitly specified when sending the notification message.
```
Landrash: Suggest you remove the word message since it's redundant.

source/_components/notify.hipchat.markdown
```diff
@@ -0,0 +1,53 @@
...
+- **color** (*Optional*): Setting color will override the default color for the notification. By default not setting this will post to HipChat using the default color yellow. Valid options are 'yellow', 'green', 'red', 'purple', 'gray', 'random'.
+- **notify** (*Optional*): Setting notify will override the default notify (blink application icon, chime, or otherwise call attention) setting for the notification. By default this is false. Valid options are true and false.
```
Landrash: Wrap the available values in quotes for readability.
> `true` , `false`.

source/_components/notify.hipchat.markdown
BioSehnsucht: The API treats it as a string rather than a number, and I suppose there's the possibility that a room ID might be non-numeric (though I haven't seen any).

source/_components/notify.hipchat.markdown
BioSehnsucht: It should be true / false (bool) not 'true' / 'false' (str), so I didn't put in quotes. I can change it, or is there a better third option for making it clear without implying str?

BioSehnsucht: I guess it may be irrelevant since it should be cast appropriately to bool by voluptuous? In which case I'll make that change to use quotes. The gotcha though would be that magic str to bool doesn't happen when passing data JSON to notify service, probably, so should clarify in that section that it should be bool.

BioSehnsucht:
@Landrash I've made changes based on your suggestions.

BioSehnsucht:
Updated docs to reflect changes made to use a PyPi library.

PR #3212, opened 2017-08-20
Update switch.broadlink.markdown

mattie47:
**Description:**

Have changed "type" from Optional, to "Required for some models". I don't know if this is non-standard for documentation...

Regardless, my SP3 would not work without this field configured.  I spent a lot of time reading various threads on what to troubleshoot.

This post: https://community.home-assistant.io/t/broadlink-rm-pro-and-a1-sensor/3678/305 also implied the friendly_name is required but I found this not to be the case, and instead "Broadlink Switch" was simply displayed.

After much googling and trying various things, I believe the following two pip packages were required for broadlink.py to work:

sudo pip3 install pycrypto voluptuous

Should I also add this to the documentation?

Thanks,

Matt


fabaff:
> Should I also add this to the documentation?

No, `voluptuous` is a core requirement of Home Assistant and `pycrypto` is a requirement of `python-broadlink` if needed.

PR #3246, opened 2017-08-26
Documentation of the Xiaomi Philips Light component

syssi:
**Description:**

Documentation of the Xiaomi Philips Light component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** https://github.com/home-assistant/home-assistant/pull/9087

### DIFF
--- a/source/_components/light.xiaomi_philipslight.markdown
+++ b/source/_components/light.xiaomi_philipslight.markdown
@@ -15,7 +15,7 @@ ha_iot_class: "Local Polling"
 
 The `xiaomi_philipslight` platform allows you to control the state of your Xiaomi Philips LED Ball Lamp and Xiaomi Philips LED Ceiling Lamp.
 
-Current supported features are `on`, `off`, `set_cct` (colortemp) , `set_bright` (brightness).
+Currently, the supported features are `on`, `off`, `set_cct` (colortemp) , `set_bright` (brightness).
 
 Please follow the instructions on [Retrieving the Access Token](/xiaomi/#retrieving-the-access-token) to get the API token to use in the `configuration.yaml` file.
 
--- a/source/_components/light.xiaomi_philipslight.markdown
+++ b/source/_components/light.xiaomi_philipslight.markdown
@@ -1,0 +1,105 @@
+---
+layout: page
+title: "Xiaomi Philips Light"
+description: "Instructions how to integrate your Xiaomi Philips Lights within Home Assistant."
+date: 2017-08-26 08:45
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: philips.png
+ha_category: Light
+ha_iot_class: "Local Polling"
+---
+
+The `xiaomi_philipslight` platform allows you to control the state of your [Xiaomi Philips LED Ball Lamp and Xiaomi Philips LED Ceiling Lamp.
+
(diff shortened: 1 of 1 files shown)


syssi:
+New-component/platform

balloob:
Please add `ha_version: 0.53` to your page metadata.

Did you base your page on another page? (and thus we have another page missing a ha_version tag)

syssi:
I just removed the ha_version because I was unsure. The page is based on vacuum.xiaomi.markdown.

source/_components/light.xiaomi_philipslight.markdown
```diff
@@ -0,0 +1,106 @@
...
+ha_category: Light
+ha_version: 0.53
+ha_iot_class: "Local Polling"
+---
+
+The `xiaomi_philipslight` platform allows you to control the state of your Xiaomi Philips LED Ball Lamp and Xiaomi Philips LED Ceiling Lamp.
+
+Current supported features are `on`, `off`, `set_cct` (colortemp) , `set_bright` (brightness).
+
+## {% linkable_title Getting started %}
```
fabaff: I moved the instructions from the vacuum platform to the component. So a link to https://home-assistant.io/components/xiaomi/#retrieving-the-access-token would be sufficient.

source/_components/light.xiaomi_philipslight.markdown
```diff
@@ -0,0 +1,106 @@
...
+light:
+  - platform: xiaomi_philipslight
+    name: Xiaomi Philips Smart LED Ball
+    host: 192.168.130.67
+    token: your-led-ball-token-here
+  - platform: xiaomi_philipslight
+    name: Xiaomi Philips Smart LED Ceiling Lamp
+    host: 192.168.130.68
+    token: your-ceiling-lamp-token-here
+```
```
fabaff: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_components/light.xiaomi_philipslight.markdown
syssi: The new location of the token instructions isn't appropriate. The page "https://home-assistant.io/components/xiaomi/#retrieving-the-access-token" belongs to the xiaomi gateway. A device with another protocol and another security mechanism (key != token). We mixed up two independent platforms now.

syssi:
Can you move section "Retrieving the Access Token" to a better place or make a suggestion?

source/_components/light.xiaomi_philipslight.markdown
fabaff: Same app, same name... Sure, we can improve the location but I don't see an issue here as the intro make it clear that it's for platforms.

Maintaining three times the same instructions which will be outdated probably soon and are extensive doesn't make sense and will lead to frustrated users. We saw that with the Raspberry platforms in the past.

syssi:
The new place is okay for me. I will suggest a change again if the community users will mix up the xiaomi zigbee gateway (keys) and wifi (tokens) too often which is mentioned on the same page now.

syssi:
https://community.home-assistant.io/t/xiaomi-gateway-not-discovered/25030/21

PR #3247, opened 2017-08-26
Documentation of the Xiaomi Vacuum component improved

syssi:
- IOT class added.
- Section added with a simplified way to extract the token from rooted android devices.


syssi:
+Enhancement

PR #3255, opened 2017-08-27
Update isy994.markdown

boojew:
The custom device section needs some serious updates. In its current state, it is very confusing. I have added some clarity - and will hopefully rewrite it completely in the future.


frenck:
Thanks @boojew 🥇

PR #3276, opened 2017-08-29
Where to find token on iOS devices themselves.

bbbenji:
**Description:**
I have added the location of DB files directly on iOS devices. This is useful if the user is jailbroken and doesn't want to make a backup just to retrieve tokens. The user will be able to SCP the files directly from their device or transfer them using a file browser app.


fabaff (approved):
Thanks 🐦

PR #3278, opened 2017-08-30
Update alarm_control_panel.egardia.markdown

jeroenterheerdt:
**Description:**
Updated documentation to reflect changes in workings of Egardia component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** home-assistant/home-assistant#9225


source/_components/alarm_control_panel.egardia.markdown
```diff
@@ -67,4 +67,20 @@ alarm_control_panel:
  python3 [path/to/]egardiaserver.py &
 ```
  **Do not forget the trailing & here; otherwise you will be having a hard time logging back into the machine.**
+ You can also use one of the methods mentioned here: https://www.dexterindustries.com/howto/run-a-program-on-your-raspberry-pi-at-startup/ or device your own. A sample shell script that can be used the init.d method and configures logging as well is shown below:
```
fabaff: I think that it would help to provide a sample for systemd instead of the old init.d system.

jeroenterheerdt: great idea. I will add it.

jeroenterheerdt:
Systemd example added.

fabaff:
Needs to go into `current` after the release of 0.53.

PR #3291, opened 2017-09-01
Update Z-Wave documentation to mention Wake

DubhAd:
Adding a note that battery powered devices need to be awake to update their settings, since this trips up many folks new to Z-Wave.

### DIFF
--- a/source/_docs/z-wave.markdown
+++ b/source/_docs/z-wave.markdown
@@ -101,9 +101,9 @@ An easy script to generate a random key:
 cat /dev/urandom | tr -dc '0-9A-F' | fold -w 32 | head -n 1 | sed -e 's/\(..\)/0x\1, /g'
 ```
 
-### {% linkable_title Updating Device Settings %}
+### {% linkable_title Battery Powered Devices %}
 
-When using the Z-Wave control panel to update device settings, please remember that you'll need to Wake any battery devices first. Details for how to do that vary from device to device, and will be covered in the manual. The manual will also say how long the device will be awake for, which may only be a few seconds.
+Battery powered devices need to be awake before you can use the Z-Wave control panel to update their settings. How to wake your device is device specific, and some devices will stay awake for only a couple of seconds. Please refer to the manual of your device for more details.
 
 ### {% linkable_title Events %}
 
--- a/source/_docs/z-wave.markdown
+++ b/source/_docs/z-wave.markdown
@@ -101,6 +101,10 @@ An easy script to generate a random key:
 cat /dev/urandom | tr -dc '0-9A-F' | fold -w 32 | head -n 1 | sed -e 's/\(..\)/0x\1, /g'
 ```
 
+### {% linkable_title Updating Device Settings %}
+
+When using the Z-Wave control panel to update device settings, please remember that you'll need to Wake any battery devices first. Details for how to do that vary from device to device, and will be covered in the manual. The manual will also say how long the device will be awake for, which may only be a few seconds.
+
 ### {% linkable_title Events %}
 
 #### {% linkable_title zwave.network_complete %}


source/_docs/z-wave.markdown
```diff
@@ -101,6 +101,10 @@ An easy script to generate a random key:
 cat /dev/urandom | tr -dc '0-9A-F' | fold -w 32 | head -n 1 | sed -e 's/\(..\)/0x\1, /g'
 ```
 
+### {% linkable_title Updating Device Settings %}
+
+When using the Z-Wave control panel to update device settings, please remember that you'll need to Wake any battery devices first. Details for how to do that vary from device to device, and will be covered in the manual. The manual will also say how long the device will be awake for, which may only be a few seconds.
```
frenck: Some spelling, grammar, and readability issues here.
I suggest replacing it with:

> When using the Z-Wave control panel to update device settings of devices that are battery powered, you'll need to awake that device first. How to wake your device is device specific, and some devices will stay awake for only a couple of seconds. Please refer to the manual of your device for more details.

DubhAd: I think "wake that device" is more appropriate than "awake that device" (or at least, I know very few people that would consider using awake that way, though awaken maybe).

> When using the Z-Wave control panel to update device settings of devices that are battery powered, you'll need to wake that device first. How to wake your device is device specific, and some devices will stay awake for only a couple of seconds. Please refer to the manual of your device for more details.

Or possibly:

> When using battery powered devices, before you can use the Z-Wave control panel to update device settings, you'll need to wake that device first. How to wake your device is device specific, and some devices will stay awake for only a couple of seconds. Please refer to the manual of your device for more details.

That at least makes it slightly more likely that people will notice the key detail of it referring to battery powered devices.

frenck: @DubhAd 
Just looked it up, both wake and awake are perfectly fine to use in this case, so I'm fine with either one.

I like where is going, bringing up the battery powered devices to the beginning of the sentence makes a lot of sense. I would move the z-wave control panel part slightly more to the back since it is now kind of stuck...

How about taking your suggestion a little further:
> Battery powered devices need to be wakened up before you can use the Z-Wave control panel to update device settings. How to wake your device is device specific, and some devices will stay awake for only a couple of seconds. Please refer to the manual of your device for more details.

frenck: Maybe we'll need to update the heading title as well? Maybe make the heading more towards the "battery powered devices setting" or something...

DubhAd: Using awake in that way is more formal/archaic than wake, though I'd use "awaken that device", rather than "awake that device". Mind you, I last did formal grammar "some years" ago.

So, I'd either suggest:

> Battery powered devices need to be awake before you can use the Z-Wave control panel to update their settings. How to wake your device is device specific, and some devices will stay awake for only a couple of seconds. Please refer to the manual of your device for more details.

or

> Battery powered devices need to be woken up before you can use the Z-Wave control panel to update their settings. How to wake your device is device specific, and some devices will stay awake for only a couple of seconds. Please refer to the manual of your device for more details.

As for the title, I think "Battery Powered Devices"? Then over time it can be expanded to include the other things to keep in mind (such as the fact that they don't relay for the mesh).

frenck: I'm fine with both suggestions 👍 
As for the title, I agree.

DubhAd: I'll work out how to submit an updated change based on that then. Thanks!

frenck:
@DubhAd Thank you for this awesome first contribution to our documentation 👍 
I really appreciate how you took on the discussion, even though this is a small change/addition, the result is perfect and valuable 🥇 

Looking forward to more PR's from you 🎉  😃

PR #3322, opened 2017-09-07
Describe the state_topic setting for battery tracking using Owntracks

eithe:
I wasn't sure what the /tablet/tablet section of the state_topic part meant so had to google a bit to figure it out. With this change, users will immediately understand what they should set as the topic.

**Description:**


fabaff:
I guess that we need to add a note that the username can be anything if no auth is used.

PR #3338, opened 2017-09-10
Xiaomi light description updated

syssi:
The current explanation doesn't fit. It looks like there are zigbee sub-devices called "xiaomi lights" which can be controlled via the xiaomi gateway. In reality there is just the internal night light of the gateway which can be controlled. No additional lights are supported.


syssi:
Superseeded by https://github.com/home-assistant/home-assistant.github.io/pull/3607

PR #3349, opened 2017-09-11
Added information about Hass.io update process.

mikegchambers:
**Description:**  Hass.io users end up on this page when an update is available, but then can't see what they have to do.  This is because of the the updater component and the inevitable delay in updates getting into the hass.io build.

This update informs users about an updates transition from HA into Hass.io, how to install it, and suggests disabling the updater component to alleviate 'notification stress'.

### DIFF
--- a/source/hassio/index.markdown
+++ b/source/hassio/index.markdown
@@ -30,7 +30,7 @@ The advantages of using Hass.io:
 
 ### {% linkable_title Upgrading %}
 
-Hass.io users can update Home Assistant via the ‘Hass.io' page in the UI.  However please note that Home Assistant updates take time to roll into the Hass.io builds.  Therefore there is often a slight delay between the availability of a Home Assistant update and an update being available in Hass.io, be patient.  When a Hass.io update is available it will be shown as available on the ‘Hass.io' page in your UI.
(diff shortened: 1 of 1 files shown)


source/hassio/index.markdown
```diff
@@ -13,6 +13,10 @@ Hass.io turns your Raspberry Pi (or another device) into the ultimate home autom
 
 [Go to the installation instructions &raquo;][install]
 
+ <p class='note'>
+Home Assistant updates take time to roll into the Hass.io builds.  Therefore there is often a slight delay between the availability of a Home Assistant update and an update being available in Hass.io, be patient. See: [Upgrading](#upgrading)
+ </p>
+
```
fabaff: Either this note or the "Updating" section. I would prefer the "Updating" section. Both contains the same information.

mikegchambers:
Yep.  Removed the 'note' from the top.

source/hassio/index.markdown
mikegchambers: Yep, removed this change and kept the other.

mikegchambers:
@fabaff - New to all this. Did I do that right?  I tried to remove the first part of the change and keep the second bit.  - Mike

PR #3356, opened 2017-09-11
Added link to Hassio

mikegchambers:
**Description:** A Hassio link was added to the 'note' section about avoiding permission errors, but I think we need to add a link up top too so that folks can go see the upgrading section of the Hassio documentation.  Thanks


frenck:
@mikegchambers Great first contribution to our documentation 🥇 

I've replaced `HASSIO` by `Hass.io` since that is more consistent.

mikegchambers:
👍🏻

PR #3358, opened 2017-09-12
Add information on the purge_days

DubhAd:
That the purge happens every other day is only documented in the code, and in issue #6302. This is to raise the visibility of this, and help people understand that if they restart frequently the purge will never happen.


frenck:
@DubhAd Thanks! 🥇 

I've learned something today 😉

PR #3362, opened 2017-09-13
Sensor Worx Landroid

nilzen:
**Description:**

Worx Landroid mower sensor


source/_components/sensor.worx-landroid.markdown
```diff
@@ -0,0 +1,32 @@
...
+  platform: worx-landroid
+  host: 192.168.0.10
+  pin: 1234
+```
+
+Configuration variables:
+
+- **host** (*Required*): The ip address or host name of the mower.
+- **pin** (*Required*): The pin code for the mower.
+- **allow_unreachable** (*Optional*): This will allow the mower to be outside of wifi range without raising and error (default: True).
```
happyleavesaoc: s/and/an

PR #3364, opened 2017-09-14
Note that Netgear Orbi uses port 80 for its API

powdahound:
Nothing more to add. :)


frenck:
Great first addition to our documentation @powdahound! 🥇

PR #3381, opened 2017-09-16
Add Jekyll plugin for Configuration Variables

dale3h:
**Description:**
This is the first step in standardizing our documentation. With the addition of this plug we will be able to ensure that all component and platform pages have the same format for their configuration variables.

With this addition I have included some extra styles on the heading tags that add a thin border above them, to help break the page up for easier reading.

**Advantages:**
- **Configuration Variables** section is now easily linked to.
- Each configuration variable gets its own link for easy reference.
- `<h2>` tags now have a thin border above them for easier reading.
- Standardized configuration format.
- Easily document if a configuration variable is required, optional, inclusive, exclusive.
- Easily document the type of value that is expected for each configuration variable.
- Easily document the default value of a configuration variable.
- Nested configuration variables are allowed.
- Specific types are automatically linked to their respective documentation.

**Other Changes:**
- I have only included updates to five of the documentation pages:
  - Template Binary Sensor
  - Template Cover
  - Template Light
  - Template Sensor
  - Template Switch
- These pages previously contained a very significant amount of invalid configurations.


dale3h:
@balloob The only concerns I really have are where I added/updated the styles. I put them in `sass/custom/_paulus.scss`. Can you please confirm that the additions and changes to this file are acceptable?

Everything works perfectly on my local environment. I have also double-checked 20+ other pages to make sure that the styling changes did not have any significant impact on existing design(s).

At some point I'd really like to have a list of value types that the **Configuration Variables** section can link to, i.e.: `string`, `int`, `float`, `bool`, `list`, `map`, `entity_id`, `device_class`, `hostname`/`ip_address`, `template`, `action`, `time`, `data`, etc.

I thought about adding an **Allowed values:** field for platforms like LimitlessLED (`bridges.groups.types`) that only allow a limited set of values, but I will address that topic when I come to a component or platform like that. I would prefer to keep clutter down as much as possible.

Any discussion and feedback is very much welcomed on these changes and the plan moving forward. Our documentation is a bit of a disaster right now, and honestly, it is currently our weakest link. It is my goal to turn that around and get it organized in a reasonable fashion.

balloob:
lol yeah adding them to paulus.scss is fine. I added that file because I kept forgetting how custom CSS was called 😇

plugins/configuration.rb
```diff
@@ -0,0 +1,92 @@
...
+      (REQUIRED_VALUES.include? value) ? REQUIRED_VALUES[value] : REQUIRED_VALUES[false]
+    end
+
+    def render_config_vars(vars, nested = false)
+      result = Array.new
+      result << "<dl class='#{nested ? "nested" : ""}'>"
+
+      result << vars.map do |key, attr|
+        markup = Array.new
+        markup << "<dt><a class='title-link' name='#{slug(key)}' href='\##{slug(key)}'></a> #{key}</dt>"
```
balloob: Ruby supports multi-line strings.

```ruby
markup = <<TPL
Here your fancy
And multi-line string including interpolation: #{slug(key)}
TPL
```

dale3h: The reason I broke it up like this is because of all of the `if` statements. I could break it into an array, and then build from there, but I do not know how to do the `if` statements inside of a heredoc.

balloob: Oh I didn't even see the if statements.

balloob:
Ok to merge when conflicts resolved.

dale3h:
Closing this PR and opening a new one. I don't think I resolved the conflict the correct way.

PR #3388, opened 2017-09-17
Xiaomi Gateway: Allow static config. of a gateway without discovery

syssi:
https://github.com/home-assistant/home-assistant/pull/9464


syssi:
The change is already merged. I don't know why this PR is still open.

PR #3408, opened 2017-09-19
Added missing scan_interval option

neonbunny:


pvizeli:
That is home-assistant 1x1. That work on all platform with polling stuff behind...

neonbunny:
I would argue that the average user (i.e. me) will not know, nor should need to know, the internal component hierarchy. However, I can also see that duplication in the documentation would be messy. Feel free to dismiss this this PR.

fabaff:
This is covered in the docs and the "average user" doesn't need to change it in most cases, otherwise the search feature in the docs will help.

PR #3410, opened 2017-09-20
Correct typo

NicoHood:
**Description:**
As in the table below play_media is correct.


NicoHood:
@fabaff I committed this to branch current, I hope that is correct. Please consider to also merge it into next. I dont know your exact workflow, I just saw that both branches are in use.

PR #3415, opened 2017-09-22
Add Jekyll plugin for Configuration Variables

dale3h:
**Description:**
This is the first step in standardizing our documentation. With the addition of this plug we will be able to ensure that all component and platform pages have the same format for their configuration variables.

With this addition I have included some extra styles on the heading tags that add a thin border above them, to help break the page up for easier reading.

**Advantages:**
- **Configuration Variables** section is now easily linked to.
- Each configuration variable gets its own link for easy reference.
- ~~`<h2>` tags now have a thin border above them for easier reading.~~
- Standardized configuration format.
- Easily document if a configuration variable is required, optional, inclusive, exclusive.
- Easily document the type of value that is expected for each configuration variable.
- Easily document the default value of a configuration variable.
- Nested configuration variables are allowed.
- Specific types are automatically linked to their respective documentation.

**Other Changes:**
- I have only included updates to five of the documentation pages:
  - Template Binary Sensor
  - Template Cover
  - Template Light
  - Template Sensor
  - Template Switch
- These pages previously contained a very significant amount of invalid configurations.

### DIFF
--- a/plugins/configuration.rb
+++ b/plugins/configuration.rb
@@ -1,0 +1,92 @@
+module Jekyll
+  class ConfigurationBlock < Liquid::Block
+    TYPE_LINKS = {
+      'action'       => '/docs/automation/action/',
+      'device_class' => '/components/binary_sensor/',
+      'template'     => '/docs/configuration/templating/',
+    }
+
+    REQUIRED_VALUES = {
+      true        => 'Required',
+      false       => 'Optional',
+      'required'  => 'Required',
+      'optional'  => 'Optional',
+      'inclusive' => 'Inclusive',
+      'exclusive' => 'Exclusive'
+    }
+
+    def initialize(tag_name, text, tokens)
+      super
+    end
+
+    def slug(key)
+      key.downcase.strip.gsub(' ', '-').gsub(/[^\w-_]/, '')
+    end
+
+    def type_class(type)
+      ((type.is_a? Array) ? type.join(' ') : type).downcase
+    end
+
+    def type_link(type)
+      if type.include? ','
+        type = type.split(',')
+      end
+
+      if type.is_a? Array
+        return (type.map { |t| type_link(t) }).join(' | ')
+      end
+
+      type = type.strip
+      (TYPE_LINKS.include? type.downcase) ? "[%s](%s)" % [type, TYPE_LINKS[type.downcase]] : type
+    end
+
+    def required_value(value)
+      if !!value == value
+        return REQUIRED_VALUES[value]
+      end
+
+      value = value.strip.downcase
+      (REQUIRED_VALUES.include? value) ? REQUIRED_VALUES[value] : REQUIRED_VALUES[false]
+    end
+
+    def render_config_vars(vars, nested = false)
+      result = Array.new
+      result << "<dl class='#{nested ? "nested" : ""}'>"
+
+      result << vars.map do |key, attr|
+        markup = Array.new
+        markup << "<dt><a class='title-link' name='#{slug(key)}' href='\##{slug(key)}'></a> #{key}</dt>"
+        markup << "<dd>"
+        markup << "  <p class='desc'>"
+        markup << "    <span class='type'>(<span class='#{type_class(attr['type'])}'>#{type_link(attr['type'])}</span>)</span> " if attr.key? 'type'
+        markup << "    <span class='required'>(#{required_value(attr['required'])})</span> " if attr.key? 'required'
+        markup << "    <span class='description'>#{attr['description']}</span>" if attr.key? 'description'
+        markup << "  </p>"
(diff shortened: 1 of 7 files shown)


plugins/configuration.rb
```diff
@@ -0,0 +1,92 @@
+module Jekyll
+  class ConfigurationBlock < Liquid::Block
+    TYPE_LINKS = {
+      'action'       => '/docs/automation/action/',
+      'device_class' => '/components/binary_sensor/',
```
balloob: This is not unique to the binary sensor. Better would be to link to the component that the platform that is using device_class belongs to?

plugins/configuration.rb
```diff
@@ -0,0 +1,92 @@
+module Jekyll
+  class ConfigurationBlock < Liquid::Block
+    TYPE_LINKS = {
+      'action'       => '/docs/automation/action/',
```
balloob: This should link to the script syntax page https://home-assistant.io/docs/scripts/

plugins/configuration.rb
```diff
@@ -0,0 +1,92 @@
...
+      if type.is_a? Array
+        return (type.map { |t| type_link(t) }).join(' | ')
+      end
+
+      type = type.strip
+      (TYPE_LINKS.include? type.downcase) ? "[%s](%s)" % [type, TYPE_LINKS[type.downcase]] : type
+    end
+
+    def required_value(value)
+      if !!value == value
```
balloob: Use [`.is_a?`](https://apidock.com/ruby/Object/is_a%3F)

plugins/configuration.rb
```diff
@@ -0,0 +1,92 @@
+module Jekyll
+  class ConfigurationBlock < Liquid::Block
+    TYPE_LINKS = {
+      'action'       => '/docs/automation/action/',
+      'device_class' => '/components/binary_sensor/',
+      'template'     => '/docs/configuration/templating/',
+    }
+
+    REQUIRED_VALUES = {
+      true        => 'Required',
```
balloob: These can never happen because by the time you test against this hash, you already assume it's a string.

plugins/configuration.rb
```diff
@@ -0,0 +1,92 @@
...
+    def render_config_vars(vars, nested = false)
+      result = Array.new
+      result << "<dl class='#{nested ? "nested" : ""}'>"
+
+      result << vars.map do |key, attr|
+        markup = Array.new
+        markup << "<dt><a class='title-link' name='#{slug(key)}' href='\##{slug(key)}'></a> #{key}</dt>"
+        markup << "<dd>"
+        markup << "  <p class='desc'>"
+        markup << "    <span class='type'>(<span class='#{type_class(attr['type'])}'>#{type_link(attr['type'])}</span>)</span> " if attr.key? 'type'
```
balloob: These lines are too long. Let's stick with

```rb
if attr.key? 'type'
  markup << ""
end
```

sass/custom/_paulus.scss
```diff
@@ -309,19 +308,23 @@ article.post, article.page, article.listing {
 
   h2 {
     font-size: 1.5em;
-    margin-top: 2em;
+    margin: 1.5em 0 1rem;
```
balloob: This applies to ALL headers. Let's do a re-design of the website in another PR please.

plugins/configuration.rb
dale3h: I have added the ability to specify the component/platform in the form of `{% configuration binary_sensor.template %}` tag. If not specified, the component and platform will automatically be detected from `page['slug']`. This will allow us to use `%{component}` in the `TYPE_LINKS`.

plugins/configuration.rb
dale3h: I changed the logic up to allow custom strings instead of using a hash.

plugins/configuration.rb
```diff
@@ -0,0 +1,114 @@
...
+      super
+      @component, @platform = text.strip.split('.', 2)
+    end
+
+    def slug(key)
+      key.downcase.strip.gsub(' ', '-').gsub(/[^\w\-]/, '')
+    end
+
+    def type_class(type)
+      ((type.is_a? Array) ? type.join(' ') : type).downcase
```
balloob: Funny that you're wrapping your `type.is_a?` expression in parentheses because you are omitting the parentheses on the function params (`type.is_a?(Array)`)

dale3h: :joy: This is my first Ruby project. I still don't know when to, and when not to, use parenthesis!

balloob: I usually use them always as I like consistency. Not using them can get you into situations like this.

dale3h: Good to know. I will stick to that as well. I've always been confused (I think since Visual Basic 3!) why it is sometimes optional to use parenthesis.

PR #3430, opened 2017-09-24
Update ecobee.markdown

etsinko:
**Description:**
removed hold_temp configuration variable. It is not used in the code since revision b732174def5894708e90c7d8708e971410a43afd

### DIFF
--- a/source/_components/ecobee.markdown
+++ b/source/_components/ecobee.markdown
@@ -55,7 +55,6 @@ ecobee:
 Configuration variables:
 
 - **api_key** (*Required*): Your ecobee API key. This is only needed for the initial setup of the component. Once registered it can be removed. If you revoke the key in the ecobee portal you will need to update this again and remove the ecobee.conf file in the `.homeassistant` configuration path.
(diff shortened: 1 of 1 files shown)


etsinko:
Hi,
please see that hold_temp is currently not used anywhere:
1. it is passed into ecobee platform during setup:
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/climate/ecobee.py#L52

2. It is passed into thermostat during creation:
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/climate/ecobee.py#L57

3. Ecobee thermostat stores it as a class attribute (But never uses it!!)
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/climate/ecobee.py#L116

fabaff:
So, the proper way would be to update the code and then docs. Till now it's valid to define `hold_temp` in the configuration even if the value is not used. I assume that [5586](https://github.com/home-assistant/home-assistant/pull/5586) was incomplete.

PR #3446, opened 2017-09-26
Updated docs for iOS actionable notification since 0.54 breaking change

mikeoh90:
My first contribution, so apologies if I've not gone about it the right way! I've updated iOS actionable notifications docs to show category identifiers must be lower case since version 0.54 and upper case pre-0.54 and that action identifiers should remain uppercase. Updated examples to show category identifiers as lower case.

**Description:**


fabaff:
This was changed with #3426 already but not deployed.

PR #3461, opened 2017-09-28
[WIP] Standards/Style Guide for Documentation of Home Assistant

dale3h:
## NOT READY TO MERGE

**Description:**
I have started a list of documentation standards/styling guide, and would love to open the topic up to everyone for discussion.

1. Is there anything here that seems absolutely absurd and should not be a standard?
1. Is there anything you would like to add that is missing?
1. Do you have any recommendations on how to lay it out in an orderly fashion?

This document/section will be expanding/changing quite a bit over the coming weeks (and possibly months), as we make our way through each of the documentation pages. Unfortunately, each page needs to be updated manually and cannot be batch processed. However, this also gives the opportunity to address and fix any syntax errors, logic errors, etc that we come across.

Please keep in mind that right now this is just "my thoughts" dropped into a quick Markdown file. This is why I am sharing it with all of you, so that we can address the rules of documentation as a community moving forward.


onerealfunnyguy:
some how some way, those writting the docs that know this stuff need to phrase it in a way that us noobs can understand so more people with lower skill levels have access to setting up a running system, vs talking to skilled linux users and admins

phillprice:
One thing I've found as a new convert to HA is that I miss seeing simple things like what it will look like in the UI if I do something I know it would be tricky to keep UI changes upto date with each release. 

There is tooling available through selenium and image checkers that could spot ui changes between builds and flag that a change in the docs may be required if that could help.

For instance in here there's not a single picture of how a light shows up in the ui, what happens when you click it to bring up the panel etc.

https://home-assistant.io/components/light.hue/

frenck:
I think we should add some style guide to Markdown in general.
For example: http://guides.dealerdirect.io/code-styling/markdown/

andrey-git:
Using the new config format is important, but beyond that I think the rules will cause more harm than good unless there is hound-like CI to alert on them.

Imagine (I would assume a common) usecase of someone seeing a typo, a missing parameter, or a not-clear-enough explanation. They use the "edit this page" button and send a quick PR. Then a week later a reviewer asks them to fix their oxford comma. I would assume half of the people won't bother.

ajfriesen:
How about having a template for the different parts of documentation?
At least for the components, I could imagine that.

Just define some headings which most components share like:

1. How to enable component
2. Requierements
2. Configuration Variables
2. Platform Services
    1. Service A
    2. Service B
    3. Service n
3. Retrieving the Access Token
4. Specific information for HASSIO
5. Examples

In the different sections we could write down some rules from your list or some best practice approaches. (If you start a new component, this could help.)
If a section is empty everybody knows there is help needed. If the section should be empty we could mark it as not necessary.

This is just a quick example and needs some work, but I think you get the idea.

PR #3484, opened 2017-10-01
Recorder: MariaDB

Landrash:
**Description:**
Adds information about using the MariaDB database engine with the `recorder` component.

### DIFF
--- a/source/_components/recorder.markdown
+++ b/source/_components/recorder.markdown
@@ -112,7 +112,7 @@ For MariaDB you may have to install a few dependencies. On the python side we us
 $ sudo apt-get install mariadb-server libmariadbclient-dev
 $ pip3 install mysqlclient
 ```
-If you are in a virtual environment, don't forget to activate it before installing the pymysql package.
+If you are in a virtual environment, don't forget to activate it before installing the `mysqlclient` package.
 
 ```bash
 pi@homeassistant:~ $ sudo su homeassistant -s /bin/bash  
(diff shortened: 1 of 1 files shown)


source/_components/recorder.markdown
```diff
@@ -102,6 +104,27 @@ If you only want to hide events from e.g. your history, take a look at the [`his
 
 Not all Python bindings for the chosen database engine can be installed directly. This section contains additional details which should help you to get it working.
 
+### {% linkable_title MariaDB %}
```
fabaff: As MariaDB is drop-in replacement for MySQL I would like to suggest to keep only one section.

Landrash: The problem is that it isn't since the dependencies are different. Sure we could only have one section but it would be harder to search for in that case and be a tad more convoluted for a new user.

source/_components/recorder.markdown
```diff
@@ -102,6 +104,27 @@ If you only want to hide events from e.g. your history, take a look at the [`his
...
+
+For MariaDB you may have to install a few dependencies. On the python side we use the `mysqlclient`.
+
+```bash
+$ sudo apt-get install mariadb-server libmariadbclient-dev
```
fabaff: I assume that the you only need the server package if you want to run the database locally.

source/_components/recorder.markdown
fabaff: It's `libmariadbclient-dev` vs. `libmysqlclient-dev`.

source/_components/recorder.markdown
Landrash: That is true. Fixing.

Landrash:
@fabaff Slimmed down the change. Would this be preferable? 
Order is Alphabetical in the section of text.

PR #3498, opened 2017-10-02
Added supported model to Panasonic Viera component page

GeoffreyWesthoff:
**Description:**
Added compatible Panasonic Viera model to documentation

What works:

- Skipping forwards and backwards (Cannot stop due to missing pause button)
- Turning TV off
- Volume control

What does not:

- Cannot track media 
- Pausing content 
- Turning on TV using MAC

### DIFF
--- a/source/_components/media_player.panasonic_viera.markdown
+++ b/source/_components/media_player.panasonic_viera.markdown
@@ -26,6 +26,7 @@ Currently known supported models:
 - TX-L42ET50
 - TX-P50GT60E
 - TX-65EXW784
+- TX-32AS520E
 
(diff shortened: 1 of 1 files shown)


Landrash:
Unsure on how well supported the other models are or if the limitations are similar.
@fabaff Should we extend this page furter or keep is simple as it currently is?

GeoffreyWesthoff:
I can also confirm that calling the `media_player` domain with `media_play_pause` will pause and resume the TV content

Landrash:
@GeoffreyWesthoff If that's the case we can merge this as is. Thank you.

fabaff:
I think that we can let it be simple. If something not working then we could go down the same path as for the Samsung TVs.

PR #3502, opened 2017-10-03
Add Information to Push Notifications component

ajfriesen:
**Description:**

Add information for verifying your domain with HASSIO, which is not obvious to all users.


ajfriesen:
I do not get it.
Do I just to delete the conflict markers because I do not see any conflicts!?

PR #3551, opened 2017-10-07
Remove some stale info from Z-Wave

armills:
**Description:**
This PR removes some info about old version from the Z-Wave docs. It's been long enough now to remove it.

### DIFF
--- a/source/_docs/z-wave.markdown
+++ b/source/_docs/z-wave.markdown
@@ -18,8 +18,6 @@ Before configuring the Z-Wave setup, please take a moment and read [this article
 
 ### {% linkable_title Installation %}
 
-As of version 0.45, Home Assistant automatically installs python-openzwave from PyPI as needed.
-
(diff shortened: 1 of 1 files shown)


source/_docs/z-wave.markdown
```diff
@@ -18,8 +18,6 @@ Before configuring the Z-Wave setup, please take a moment and read [this article
 
 ### {% linkable_title Installation %}
 
-As of version 0.45, Home Assistant automatically installs python-openzwave from PyPI as needed.
```
armills: After removing the version reference, I was originally going to move this below the installation steps shown. As it's written right now, the section starts off sounding like home assistant will automatically install what it needs, but then we give further instructions below. (Bad for skimming)

However, I'm thinking the line is unnecessary entirely. Home Assistant always auto-installs requirements. It's not really necessary to bring up here. We still have info about first run install time below anyway.

frenck: Agreed.

frenck:
Thanks, @armills, that indeed is an improvement.

PR #3554, opened 2017-10-08
Update docs with emulated_hue_hidden

Danielhiversen:
**Description:**
Add docs for this pr: https://github.com/home-assistant/home-assistant/pull/9382

### DIFF
--- a/source/_components/emulated_hue.markdown
+++ b/source/_components/emulated_hue.markdown
@@ -96,7 +96,7 @@ homeassistant:
   customize:
     light.bedroom_light:
       # Don't allow light.bedroom_light to be controlled by the emulated Hue bridge
-      emulated_hue: false
+      emulated_hue_hidden: false
     light.office_light:
       # Address light.office_light as "back office light"
       emulated_hue_name: "back office light"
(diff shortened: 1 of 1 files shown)


nikc0069:
Emulated_hue_hidden: false in the doco implies it is not hidden. Before this change the config was emulated_hue: false to not add to emulated hue. Now you mark hidden as false to hide it, whereas logic dictates that if hidden is false, it is not hidden. Therefore the comment in the doco needs to make it clear which is which.

kipwittchen:
Based on this: https://github.com/home-assistant/home-assistant/pull/9382#issuecomment-335155004

I believe this pull request as written is incorrect. It should be emulated_hue: false becomes emulated_hue_hidden: true

source/_components/emulated_hue.markdown
```diff
@@ -96,15 +96,15 @@ homeassistant:
   customize:
     light.bedroom_light:
       # Don't allow light.bedroom_light to be controlled by the emulated Hue bridge
-      emulated_hue: false
+      emulated_hue_hidden: false
```
kipwittchen: Should this be emulated_hue_hidden: true ?

PR #3562, opened 2017-10-08
Minor updates to emulated_hue and OpenCV documents

arsaboo:
**Description:** Minor cleanup and clarifications to the docs

### DIFF
--- a/source/_components/emulated_hue.markdown
+++ b/source/_components/emulated_hue.markdown
@@ -104,7 +104,7 @@ homeassistant:
 
 The following are attributes that can be applied in the `customize` section:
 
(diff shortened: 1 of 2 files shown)


Danielhiversen:
Maybe also set `emulated_hue_hidden: True` here
https://github.com/home-assistant/home-assistant.github.io/pull/3562/files#diff-4753543f88ce12c6813f94231100d275R99

arsaboo:
@Danielhiversen I am thinking that it may confuse readers. Right now, we have `emulated_hue_hidden: false` in both places. Happy to change it if you like it that way.

PR #3568, opened 2017-10-09
Initial glossary

### LINKED ISSUE #3263: Docs direly need a glossary
HA docs are full of confusing buzzwords, like "entity", "component", "platform", "service", etc. It direly needs an visible/very accessible glossary explaining these terms, ideally linked to thru at least few first doc chapters, to let novices grasp them well enough.

fabaff:
**Description:**
A very simple glossary that uses a [data file](https://jekyllrb.com/docs/datafiles/).

Something like [glossarizer](https://github.com/PebbleRoad/glossarizer) could replace the  

**Related issue (if applicable):** fixes #3263

### DIFF
--- a/source/_data/glossary.yml
+++ b/source/_data/glossary.yml
@@ -1,13 +1,13 @@
 - topic: Platform
   description: A platform makes the connection to a specific software or hardware platform. The `pushbullet` platform works with the service from pushbullet.com.
 - topic: Component
-  description: "The components provides the core logic for the functionality. Like `notify` provides sending notifications."
+  description: "The components provide the core logic for the functionality in Home Assistant. Like `notify` provides sending notifications."
 - topic: Service
-  description: "Services are called to perform actions."
+  description: "[Services](/docs/scripts/service-calls/) are called to perform actions."
 - topic: Event
   description: When somethings happen.
 - topic: Entity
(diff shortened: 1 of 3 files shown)


source/_data/glossary.yml
```diff
@@ -0,0 +1,17 @@
+- topic: Platform
+  description: A platform makes the connection to a specific software or hardware platform. The `pushbullet` platform works with the service from pushbullet.com.
+- topic: Component
```
arsaboo: We should move `component` above `platform` to maintain the hierarchy. A component can have multiple platforms. Also, this may be a good place to clear up the confusion between platform and component. For example, Abode component provides the underlying core logic and infrastructure that can be used by individual platforms (e.g., Alarm, sensors, etc. )

dale3h: Based on the sort order, I think `component` will appear well above `platform`. Right now it appears to be sorted in alphabetical order by `topic`.

fabaff: `| sort: 'topic'` will do the sorting.

This is hopefully just a start. Collecting the details and provide an easy to understand description is the hard part. Nice formatting can be done later :wink:

fabaff:
OK, let's merge it.

PR #3573, opened 2017-10-09
Update MQTT broker docs for clarity

DubhAd:
Adding words to note that the reason a "run your own" broker is the most private is because the embedded broker doesn't support TLS/SSL.
Also updating to use the new configuration tags.


fabaff:
hbmqtt supports SSL/TLS. Thus, the embedded broker support it but requires configuration.

DubhAd:
Ah, their documentation makes no reference to it, but now you've mentioned it I had a poke at the source and see it. Will update.

fabaff:
It does...[Features](http://hbmqtt.readthedocs.io/en/latest/#features) and [configuration](http://hbmqtt.readthedocs.io/en/latest/references/broker.html#broker-configuration).

DubhAd:
Darn it. Clearly my ability to RTFM is failing :/

Danielhiversen:
This pr broke the travis tests: https://travis-ci.org/home-assistant/home-assistant.github.io/builds/286374091

fabaff:
@Danielhiversen I know...should be fixed soon.

PR #3578, opened 2017-10-10
Modbus Sensor: Clarify how to use scan_interval 

ziotibia81:
**Description:**

Many forum posts and issues ask how to change the scan_interval for modbus sensor.
This patch clarify the configuration parameters.


fabaff:
I wouldn't consider two thread as "many forum posts". I guess that here is make sense to add `scan_interval`. Usually adding the same information over and over again doesn't help as people will still not read it.

PR #3583, opened 2017-10-10
device_tracker.mqtt configuration.yaml is wrong

tsvi:
The topic to be looked up must be in quotes for the example to work.

**Description:**
The current documentation's example does not wrong. By putting the topic in quotes the mqtt device tracker actually does work.


fabaff (approved):
It looks more like that the tracker is not working because the topics are starting with `/`. I see no difference if the topic is wrapped in quotes or not. But to be on the safe side let's wrap it.

Thanks :bird:

PR #3593, opened 2017-10-12
Script - wait_template 'trigger.entity_id' support

cdce8p:
**Description:**
Added `trigger.entity_id` and `variables` support for wait_template

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** home-assistant/home-assistant#9807

### DIFF
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -90,6 +90,30 @@ wait_template: {% raw %}"{{ states.climate.kitchen.attributes.valve < 10 }}"{% e
 timeout: 00:01:00
 ```
 
+When using `wait_template` within an automation `trigger.entity_id` is supported for `state`, `numeric_state` and `template` triggers, see also [Available-Trigger-Data](/docs/automation/templating/#available-trigger-data).
+
+{% raw %}
+```yaml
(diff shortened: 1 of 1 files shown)


frenck:
@cdce8p Could you please rebase this one onto the `next` branch?

source/_docs/scripts.markdown
```diff
@@ -98,7 +98,7 @@ wait_template: "{{ is_state(trigger.entity_id, 'on') }}"
 ```
 {% endraw %}
 
-It is also possible to use dummy variables, e.g. in scripts, when using `wait_template`.
+It is also possible to use dummy variables, e.g., in scripts, when using `wait_template`.
```
cdce8p: Check out [Link](https://jakubmarian.com/comma-after-i-e-and-e-g/). Both versions are possible. I just tend to like the British one better.

cdce8p:
@frenck I would suggest to leave it at `current` and just merge it after the next version of HA is public. Since it is only an extension to the current functions and not a new platform, I figured that `next`would not be right here.

source/_docs/scripts.markdown
frenck: It OK to like the British one better, but we have a policy on having American English only.

PR #3594, opened 2017-10-12
add alternative command

apo-mak:
**Description:**
if you have set dns local-name "hassio.example.local" or "hassio" is resolving to a local address like 192.x.x.x or 172.x.x.x or 10.x.x.x.x the command give strange results. alternatively the 172.17.0.2 command has fix the problem.


pvizeli:
That work only inside hassio and this `172.17.x` is the docker default address. In a next update we don't listen our API on that address, so you need use the internal address `172.30.32.2`. I would maintain, that he should use the ssh add-on to set stuff inside hass.io also if there have host access.

PR #3467, opened 2017-09-30
Rewriting the Z-Wave documentation

DubhAd:
This is a starting point for rewriting the Z-Wave documentation, to reflect all the undocumented changes, addition of the control panel, etc.

Discussion and feedback appreciated ;)

PR #3644, opened 2017-10-16
Fix travis build

Danielhiversen:
**Description:**

This fixes the travis build failure from pr #3467


Danielhiversen:
@fabaff Any idea why travis did not ran for #3467 ? 
It seems that  travis is not running for pr against the `next` branch?

fabaff:
> It seems that travis is not running for any pr?

True, I can remember what the trigger/requirement was for a Travis run at the moment. No real access to the internet.

PR #3659, opened 2017-10-17
Add default groups info to group component doc

milanvo:
**Description:**

Add info regarding default groups to group component docs

### DIFF
--- a/source/_components/group.markdown
+++ b/source/_components/group.markdown
@@ -92,7 +92,7 @@ Notice in the example below that in order to refer to the group "Living Room", y
 
 ## {% linkable_title Default groups %}
 
-Some components automatically creates special groups, which contains component entities. These groups are named like `group.all_...`, for example:
+Some components automatically create special groups containing component entities. These groups are named like `group.all_...`, for example:
 - `group.all_switches`
 - `group.all_lights`
 - `group.all_devices`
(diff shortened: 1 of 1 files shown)


source/_components/group.markdown
```diff
@@ -89,3 +89,34 @@ Notice in the example below that in order to refer to the group "Living Room", y
       - group.living_room
       - group.bedroom
 ``` 
+
+## {% linkable_title Default groups %}
+
+Some components automatically creates special groups, which contains component entities. These groups are named like `group.all_...`, for example:
```
mnoorenberghe: I think the first sentence would be a bit clearer written as:
"Some components automatically create special groups containing component entities."

source/_components/group.markdown
```diff
@@ -89,3 +89,34 @@ Notice in the example below that in order to refer to the group "Living Room", y
...
+- `group.all_switches`
+- `group.all_lights`
+- `group.all_devices`
+- `group.all_scripts`
+- `group.all_automations`
+
+You can see list of these groups in  **State** <img src='/images/screenshots/developer-tool-states-icon.png' class='no-shadow' height='38' /> page of the **Developer Tools**.
+Default groups appears in the HOME tab, if not overridden by user views and groups. Default groups has attribute `hidden: true` set by default, so you must [customize](/docs/configuration/customizing-devices/) it
```
mnoorenberghe: Nit: Remove the "s" from "appears"

mnoorenberghe: I think 
> Default groups has attribute `hidden: true` set by default

would be clearer as:
> Default groups are hidden by default

mnoorenberghe: Replace "it" with "them" since we're talking about more than one group

source/_components/group.markdown
```diff
@@ -89,3 +89,34 @@ Notice in the example below that in order to refer to the group "Living Room", y
...
+to be visible in your custom groups and views.
+
+```yaml
+# Example configuration.yaml for customize and include default groups in user view
```
mnoorenberghe: Replace "for customize and" with "to"

mnoorenberghe: Replace "user" with "a custom"

milanvo:
@mnoorenberghe thanks for review. I am not english native speaker, so I appreciate your help. Corrections are applied.

PR #3691, opened 2017-10-20
remove ssl and port in configuration.yaml

phillprice:
**Description:**

Make clear to remove SSL and port in configuration.yaml to avoid 502 errors

### DIFF
--- a/source/_addons/nginx_proxy.markdown
+++ b/source/_addons/nginx_proxy.markdown
@@ -11,7 +11,7 @@ footer: true
 
 Setup a SSL proxy with NGINX and redirect port 80 to 443. Make sure you have generated a certificate before you start this add-on. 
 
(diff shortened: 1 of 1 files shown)


source/_addons/nginx_proxy.markdown
```diff
@@ -9,7 +9,7 @@ sharing: true
 footer: true
 ---
 
-Setup a SSL proxy with NGINX and redirect port 80 to 443. Make sure you have generate certificate before you start this add-on.
+Setup a SSL proxy with NGINX and redirect port 80 to 443. Make sure you have generated a certificate before you start this add-on. Remove any SSL and port details in your configuration.yaml to avoid a HTTP 502 error.
```
frenck: `port details` might cause confusion, maybe consider rewording that part.

PR #3692, opened 2017-10-20
Removed last_boot and replaced with since_last_boot

riemers:
Checked the source code of system monitor, cannot find last_boot anymore, hence removed it in favor of the one still in used 'since_last_boot'. Noticed it since the example recorder didn't remove it for me.


fabaff:
See https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/sensor/systemmonitor.py#L31

PR #3697, opened 2017-10-20
Update the list of Danish region on CLA signature page

DarkFox:
**Description:**
Updated the list of Danish regions, as it is outdated. The old danish "amter" (counties) were reshuffled into [5 regions](https://en.wikipedia.org/wiki/Regions_of_Denmark) in 2007.


fabaff:
Feel feel to submit PR which are not related to a change in the main repo and touch existing documentation  against `current`. This way, the changes will deployed as soon as the site is rebuilt.

PR #3700, opened 2017-10-21
update init.d daemon script to specify HASS_BIN

fermulator:
As per:
 * https://home-assistant.io/docs/autostart/init.d/
, coupled with:
 * https://home-assistant.io/docs/installation/virtualenv/

It's obvious that the user needs to replace "RUN_AS" with the correct user.
What can be made more clear is to include a "HASS_BIN" shell variable, with a similar instruction to update that if needed.

Note that I opted NOT to change the HASS_BIN to the full path /usr/bin/hass, as the previous revision presumably assumes that the `hass` binary is in the $PATH env for whatever user is running it ... (didn't want to change that).

**Description:**


fabaff (approved):
Thanks :bird:

PR #3722, opened 2017-10-22
toonopafstand.eneco.nl is no longer available

apo-mak:
**Description:**
the toonopafstand.eneco.nl give the above message but the library look like is  keep using the domain with no problems reported.
i change the word "website" to domain so users are not confuse  when the click it .


fabaff (approved):
Thanks :bird:

PR #3699, opened 2017-10-21
Renaming API.AI to Dialogflow

fabaff:
**Description:**
Renaming API.AI to Dialogflow.

Successor of #3597

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#10006

PR #3763, opened 2017-10-24
API.AI is now Dialogflow

pattyland:
https://blog.dialogflow.com/post/apiai-new-name-dialogflow-new-features/

I guess for backward compatibility we should not change the configuration.yaml entries.

**Description:**

### DIFF
--- a/source/_components/apiai.markdown
+++ b/source/_components/apiai.markdown
@@ -1,7 +1,7 @@
 ---
 layout: page
-title: "Api.AI"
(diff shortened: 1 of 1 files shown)


pattyland (commented):
Missed a few appearances

fabaff:
Duplicate of #3699. The renaming is already present in `next`.

Component was renamed which is a breaking change. Users have to update their configuration anyway thus it's much easier to do it at one point in time than insidious.

PR #3771, opened 2017-10-25
Update Toon docs

Boltgolt:
**Description:**

Added more information to the Toon docs, and updated the titles so the components don't all show up as "Toon" in the component search.


Boltgolt:
I see @fabaff also noticed the title issue and fixed it [here](https://github.com/home-assistant/home-assistant.github.io/pull/3752/files), but i think these titles are a lot more discriptive of what the component actially does.

source/_components/climate.toon.markdown
```diff
@@ -13,6 +13,15 @@ logo: toon.png
 ha_iot_class: "Cloud Polling"
 ---
 
-The `toon` climate platform allows you to interact with your Toon device.
+The `toon` climate platform allows you to interact with your Toon thermostat. For compatibility reasons the states are different from the normal "Toon programma's", they are mapped as follows:
```
frenck: I like the addition but maybe we should keep the documentation as English as possible.
I like to suggest: `For compatibility reasons, the states in Home Assistant are different from the states displayed on your Toon device and are mapped as follows:`

source/_components/climate.toon.markdown
```diff
@@ -1,8 +1,8 @@
 ---
 layout: page
-title: "Toon"
-description: "Instructions how to integrate Toon Climate devices within Home Assistant."
-date: 2017-10-21 12:00
+title: "Toon Thermostat"
+description: "Instructions how to integrate Toon Thermostats within Home Assistant."
```
frenck: `Thermostats` is not part of the title or product name and should be written lower caps.

source/_components/switch.toon.markdown
```diff
@@ -1,8 +1,8 @@
 ---
 layout: page
-title: "Toon"
-description: "Instructions how to integrate Toon Climate devices within Home Assistant."
-date: 2017-10-21 12:00
+title: "Toon Smart Plugs"
+description: "Instructions how to integrate Toon Smart Plugs within Home Assistant."
```
frenck: `Smart Plugs` is not part of the product name and should be written in lower caps.

frenck:
@Boltgolt, Good job on the additions, nevertheless I have some small advise and concerns. Could you take a look?

Boltgolt:
@frenck Thanks for the quick review, i pushed the changes as suggested.

fabaff:
Those changes should have been made against `current`. #3752 introduced some changes.

PR #3772, opened 2017-10-25
Better documentation for state attributes

armills:
**Description:**
Not that it's easy to find, but we'll at least have this practice documented.


andrey-git:
Maybe add that the instructions are for new platform, and new components should use  `state_attributes`

PR #3789, opened 2017-10-26
Created new documentation for GC100 component and platforms

davegravy:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#10159

### DIFF
--- a/source/_components/binary_sensor.gc100.markdown
+++ b/source/_components/binary_sensor.gc100.markdown
@@ -19,11 +19,11 @@ To enable this sensor, you first have to set up [gc100](/components/gc100/), and
 binary_sensor:
   - platform: gc100
     ports:
-      '3:1': Doorchime
-  
+      - '3:1': Doorchime
+      - '3:2': Garage Obstruction
 ```
 
 Configuration variables:
 
-- **ports** (*Required*): A list of module-address to name mappings in the format 'x:y': name, where x is module #, y is address 
+- **ports** (*Required*): A list of module-address to name mappings in the format 'x:y': name, where x is module #, y is address.
 
--- a/source/_components/binary_sensor.gc100.markdown
+++ b/source/_components/binary_sensor.gc100.markdown
@@ -1,0 +1,29 @@
+---
+layout: page
+title: gc100 Binary Sensor
+description: "Instructions on how to set up an gc100 binary sensor within Home Assistant."
+date: 2017-10-27 17:26
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: System Monitor
+ha_release: 0.56.3
+ha_iot_class: "Local Polling"
+---
+
+To enable this sensor, you first have to set up [gc100](/components/gc100/), and add the following lines to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+binary_sensor:
+  - platform: gc100
+    ports:
+      '3:1': Doorchime
+  
+```
+
+Configuration variables:
+
+- **ports** (*Required*): A list of module-address to name mappings in the format '<x>:<y>': <name>, where <x> is module #, <y> is address 
+
--- a/source/_components/binary_sensor.gc100.markdown
+++ b/source/_components/binary_sensor.gc100.markdown
@@ -7,8 +7,8 @@ sidebar: true
 comments: false
 sharing: true
 footer: true
-ha_category: System Monitor
-ha_release: 0.56.3
+ha_category: Binary Sensor
+ha_release: 0.57
 ha_iot_class: "Local Polling"
 ---
 
--- a/source/_components/binary_sensor.gc100.markdown
+++ b/source/_components/binary_sensor.gc100.markdown
@@ -25,5 +25,5 @@ binary_sensor:
 
 Configuration variables:
 
-- **ports** (*Required*): A list of module-address to name mappings in the format '<x>:<y>': <name>, where <x> is module #, <y> is address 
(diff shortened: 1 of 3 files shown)


rytilahti:
I think it would make sense to add somewhere in the docs information what on earth is gc100 anyway :-)

davegravy:
I tried to stay consistent with apcupsd which I used as a template for this - it has a link to the apcuspd project but doesn't give much detail in the home assistant docs itself. Maybe it's just well known enough to not need it.

I link to Global Cache, the manufacturer, but can certainly expand a bit if it's desired.

source/_components/switch.gc100.markdown
```diff
@@ -0,0 +1,31 @@
+---
+layout: page
+title: gc100 Switch
+description: "Instructions on how to set up an gc100 switch within Home Assistant."
+date: 2017-10-27 17:26
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: System Monitor
```
fabaff: Should be `Switch`.

source/_components/gc100.markdown
```diff
@@ -0,0 +1,29 @@
...
+hardware device which has an array of relays, RS232 serial ports, and flexible ports which can be programmed to be either digital inputs or IR blaster outputs. There are a variety of submodels of the gc100 which have different amounts of each I/O type.
+
+Currently, only relays and ports configured to be digital inputs are supported in Home Assistant.
+
+To enable this device, add the following lines to your `configuration.yaml`:
+
+```yaml
+# Example configuration.yaml entry
+gc100:
+```
```
fabaff: Please add the required keys to the sample.

source/_components/gc100.markdown
```diff
@@ -0,0 +1,29 @@
...
+layout: page
+title: "gc100"
+description: "Instructions on how to integrate gc100 with Home Assistant."
+date: 2017-10-26 17:20
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: Hub
+ha_release: 0.56.3
```
fabaff: It will go into 0.57 perhaps but not a bugfix release.

source/_components/binary_sensor.gc100.markdown
```diff
@@ -0,0 +1,29 @@
...
+
+To enable this sensor, you first have to set up [gc100](/components/gc100/), and add the following lines to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+binary_sensor:
+  - platform: gc100
+    ports:
+      '3:1': Doorchime
+
```
fabaff: No blank line needed.

source/_components/binary_sensor.gc100.markdown
```diff
@@ -0,0 +1,29 @@
...
+binary_sensor:
+  - platform: gc100
+    ports:
+      '3:1': Doorchime
+  
+```
+
+Configuration variables:
+
+- **ports** (*Required*): A list of module-address to name mappings in the format 'x:y': name, where x is module #, y is address
```
fabaff: Period is missing.

upsert:
Has this been tested on the sensor inputs found on the lower model as well as the relay outputs? I have a GC-100-06 which does not have any relays but I have never tried using the ports for anything other than sending IR. Looking at the datasheet from the manufacturer Global Caché, it is not really clear what voltages you would supply to trigger the sensor inputs. 24 VDC contact closure switches? I am assuming you have done some testing to find out.

Regarding using documentation from other components as a model, a lot of components have very terse and vague documentation that could use improvement, especially for novice users. I think it is better to supply background information and context where appropriate to encourage new users to try a component.

I assume you have seen [iTach Remote](https://home-assistant.io/components/remote.itach/) component that supports sending IR commands on the same device? Would be nice to have everything supported under one component but the iTach stuff looks to be taking a different approach with a C++ library and Python wrapper.

source/_components/gc100.markdown
```diff
@@ -0,0 +1,31 @@
...
+date: 2017-10-26 17:20
+sidebar: true
+comments: false
+sharing: true
+footer: true
+ha_category: Hub
+ha_release: 0.57
+---
+
+[gc100](https://www.globalcache.com/products/gc-100/) can be integrated into Home Assistant. gc100 is a TCP-controllable
```
upsert: How about:
```
The Global Caché [GC-100](https://www.globalcache.com/products/gc-100/) can be integrated into Home Assistant. The GC-100 is a TCP-controllable...
```
This would make it clear that this component is about a product from Global Caché. I suggest also using uppercase **GC-100** throughout to be consistent with the manufacturer's branding (except if talking about the Home Assistant component `gc100`).

davegravy:
@upsert Thanks for the feedback and questions. I only have the GC-100-12 and all my testing is on it, but the GC-100-18 and 18R appear to be identical except for the chassis / mounting. As for the -06 the API specification didn't have anything model-specific that I encountered so I would expect it should work.

My doorbell is connected to one of these digital inputs and it was a long time ago that I hooked it up, but I definitely remember being confused by the documentation and needing to do some trial and error to understand the scheme. https://www.globalcache.com/files/docs/QuickStart_GC-100-12_distrib.pdf may help. If you do run into trouble I can try to reverse engineer my doorbell :)

I searched before starting this project but didn't come across that platform until AFTER I created this pull request. I've already been in touch with the author and expressed interest in one day "fusing" our components. I'm pretty useless with C++ though, so it'd have to be python-based or I'd need to learn. For now users will have to decide if they want IR or sensors + relays.

PR #3792, opened 2017-10-26
Translation Guideline Docs

c727:
this is a first start for translation guidlines based on notes by @armills

### DIFF
--- a/source/developers/frontend_translation.markdown
+++ b/source/developers/frontend_translation.markdown
@@ -56,4 +56,4 @@ In the following PR is used for pull request at GitHub
 1. Don't discuss in already merged PRs if a translation is good or bad. Instead open an [issue](https://github.com/home-assistant/home-assistant-polymer/issues) or new PR for discussion or improvements.
 2. Don't add phrases that don't exist in `en.json`.
 3. Translated phrases have to start with capital letter.
-4. Make a note in your PR's description if you translated a phrase with an other word that fits (in your opinion) better. For example instead of translating `States` you have choosen the word for `Overview`
+4. Make a note in your PR's description if you translated a phrase with an other word that fits (in your opinion) better. For example instead of translating `States` you have chosen the word for `Overview`
--- a/source/developers/frontend_translation.markdown
+++ b/source/developers/frontend_translation.markdown
@@ -25,7 +25,7 @@ translations/zh-Hans-SG.json
 
 Language subtags will inherit from their parent, so `en-US.json` should only contain the entries that will be different from `en.json`. All language tags inherit from `en.json` as master translation and fallback for incomplete translations.
 
-Language tags follow [BCP47](https://tools.ietf.org/html/bcp47). In most cases you only need the [ISO 639-1 language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) and eventually also the [country code](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements). Also include the native language name (has to start with capital letter) in the file.
+Language tags follow [BCP47](https://tools.ietf.org/html/bcp47). In most cases you only need the [ISO 639-1 language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) and eventually also the [country code](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements). Also add language tag, native language name (has to start with capital letter) and country code (if used) inside the file.
 
 Example for French (`fr.json`):
 ```
@@ -60,4 +60,5 @@ In the following PR is used for pull request at GitHub.
 #### {% linkable_title Rules for translating %}
 1. Don't discuss in already merged PRs if a translated phrase is appropriate or not. Instead open an [issue](https://github.com/home-assistant/home-assistant-polymer/issues) or new PR for discussion or improvements.
 2. Don't submit incomplete translations (also don't include untranslated or unchanged phrases) or add phrases that don't exist in `en.json`.
-3. Translated phrases have to start with capital letter. (Currently there is no conclusion if every word in a non-sentence has to start with capital letter)
+3. Use the text in value for your translation - not the key. Example: For `"states": "Overview"` translatable `Overview`.
+4. Translated phrases have to start with capital letter. (Currently there is no conclusion if every word in a non-sentence has to start with capital letter)
--- a/source/developers/frontend_translation.markdown
+++ b/source/developers/frontend_translation.markdown
@@ -10,55 +10,23 @@ footer: true
 ha_release: 0.57
 ---
 
-This will guide you how to translate the Home Assistant frontend to another language. Currently there are only a few translatable phrases - more will be added soon.
+We don't accept translations via GitHub anymore so please use our [project site at lokalise.co](https://lokalise.co/project/XXX/).
 
-Translations are defined in the [translations](https://github.com/home-assistant/home-assistant-polymer/tree/master/translations) folder of [home-assistant-polymer repository](https://github.com/home-assistant/home-assistant-polymer) at GitHub. The files are JSON formatted and named after their language tags.
+The translation of the Home Assistant frontend is still WIP. More phrases will be available soon.
 
-Examples:
-```
-translations/de.json
-translations/en.json
-translations/en-US.json
-translations/zh-Hans.json
-translations/zh-Hans-SG.json
-```
+#### {% linkable_title How to start %}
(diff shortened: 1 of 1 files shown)


source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,59 @@
...
+{
+  "language": {
+    "fr-CA": "français (CA)"
+  },
+  ...
+}
+```
+
+### {% linkable_title Special charackters %}
+The most common ones you may need are `\'` for `'` and `\n` for `new line`. Check this [table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Using_special_characters_in_strings) for more special charackters. PS: You don't have to HTML escape your charackters.
```
arsaboo: typo: charackters - characters.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,59 @@
...
+
+Example:
+```
+translations/en.json
+translations/en-US.json
+translations/de.json
+```
+
+Language subtags will inherit from their parent, so `en-US.json` should only contain the entries that will be different from `en.json`. All language tags inherit from `en.json` as the master translation and fallback for incomplete translations.
+You can get the correct tag and native name (has to start with capital letter) from the [list of ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). If necessary also add the corresponding [country code](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements).
```
armills: Still haven't reached complete consensus on this, but I think we want to reference BCP47 here. https://tools.ietf.org/html/bcp47 From what I've seen this is the reference that browsers follow. Open to other arguments here, but at first glance, since we're choosing the default based on the browsers reported language, we should inherit the language tag standard from whatever browsers report.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,59 @@
...
+{
+  "language": {
+    "fr-CA": "Français (CA)"
+  },
+  ...
+}
+```
+
+### {% linkable_title Special characters %}
+The most common ones you may need are `\'` for `'` and `\n` for `new line`. Check this [table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Using_special_characters_in_strings) for more special characters. PS: You don't have to HTML escape your characters.
```
armills: Looks like the standard I linked to originally was for Javascript, not JSON. Key point is that single quotes don't need to be escaped for JSON, only double quotes. Having a hard time finding a better quick reference than this, but `\n` and `\"` are the common escapes that might be needed. http://www.json.org/

Also, we could probably drop `\n`. I'm having a hard time thinking of a string in the hass frontend that would contain a newline.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,59 @@
...
+In the following PR is used for pull request at GitHub
+
+#### {% linkable_title Before you submit %}
+1. Check if your language already exists in translations(https://github.com/home-assistant/home-assistant-polymer/tree/master/translations).
+2. Check if there is already an [open PR](https://github.com/home-assistant/home-assistant-polymer/pulls) for your language which you can review.
+
+#### {% linkable_title Rules for translating %}
+1. Don't discuss in already merged PRs if a translation is good or bad. Instead open an [issue](https://github.com/home-assistant/home-assistant-polymer/issues) or new PR for discussion or improvements.
+2. Don't add phrases that don't exist in `en.json`.
+3. Translated phrases have to start with capital letter.
```
armills: Just want to call this out for discussion. We should talk about case conventions. I think certainly starting with a capital letter is safe, but I think the majority of the strings should be title instead of sentence case, unless it's an actual sentence.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,64 @@
...
+  },
+  ...
+}
+```
+
+### {% linkable_title Special characters %}
+Use `\\` for backslash (`\`) and `\"` for double quotes (`"`). Don't escape forward slashes (`/`) or HTML encode characters.
+
+### {% linkable_title Submit a translation %}
+In the following PR is used for pull request at GitHub.
```
fabaff: Remove that and use "Pull request" where needed.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+#### {% linkable_title Before you submit a translation %}
+1. Check if your language tag already exists in translations(https://github.com/home-assistant/home-assistant-polymer/tree/master/translations).
```
fabaff: The links needs to be fixed.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,64 @@
...
+2. Check if there is already an [open PR](https://github.com/home-assistant/home-assistant-polymer/pulls) for your language tag which you can review or discuss.
+3. Only submit translations with language and country code (`fr-CA.json`) if there there is already an corresponding translation with language code only (`fr.json`)
+
+#### {% linkable_title Rules for translating %}
+1. Don't discuss in already merged PRs if a translated phrase is appropriate or not. Instead open an [issue](https://github.com/home-assistant/home-assistant-polymer/issues) or new PR for discussion or improvements.
+2. Don't submit incomplete translations (also don't include untranslated or unchanged phrases) or add phrases that don't exist in `en.json`.
+3. Use the text in value for your translation - not the key. Example: For `"states": "Overview"` translate `Overview`.
```
fabaff: Somebody who is not familiar with JSON will not know that a `name-value` pair or a `key-value` pair is. I would like to suggest to rewrite that in a non-developer way.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,64 @@
...
+translations/zh-Hans.json
+translations/zh-Hans-SG.json
+```
+
+Language subtags will inherit from their parent, so `en-US.json` should only contain the entries that will be different from `en.json`. All language tags inherit from `en.json` as master translation and fallback for incomplete translations.
+
+Language tags follow [BCP47](https://tools.ietf.org/html/bcp47). In most cases you only need the [ISO 639-1 language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) and eventually also the [country code](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements). Also add language tag, native language name (has to start with capital letter) and country code (if used) inside the file.
+
+Example for French (`fr.json`):
+```
```
fabaff: We could format it as JSON by appending `json`.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,64 @@
...
+In the following PR is used for pull request at GitHub.
+
+#### {% linkable_title Before you submit a translation %}
+1. Check if your language tag already exists in translations(https://github.com/home-assistant/home-assistant-polymer/tree/master/translations).
+2. Check if there is already an [open PR](https://github.com/home-assistant/home-assistant-polymer/pulls) for your language tag which you can review or discuss.
+3. Only submit translations with language and country code (`fr-CA.json`) if there there is already an corresponding translation with language code only (`fr.json`)
+
+#### {% linkable_title Rules for translating %}
+1. Don't discuss in already merged PRs if a translated phrase is appropriate or not. Instead open an [issue](https://github.com/home-assistant/home-assistant-polymer/issues) or new PR for discussion or improvements.
+2. Don't submit incomplete translations (also don't include untranslated or unchanged phrases) or add phrases that don't exist in `en.json`.
```
armills: I think incomplete translations are OK, as long as only the actually translated keys are included in the JSON. Should be a moot point once we move to a proper platform though.

source/developers/frontend_translation.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+### {% linkable_title Submit a translation %}
+In the following PR is used for pull request at GitHub.
+
+#### {% linkable_title Before you submit a translation %}
+1. Check if your language tag already exists in translations(https://github.com/home-assistant/home-assistant-polymer/tree/master/translations).
+2. Check if there is already an [open PR](https://github.com/home-assistant/home-assistant-polymer/pulls) for your language tag which you can review or discuss.
+3. Only submit translations with language and country code (`fr-CA.json`) if there there is already an corresponding translation with language code only (`fr.json`)
+
+#### {% linkable_title Rules for translating %}
```
balloob: Should we add that translations should be provided by native speakers, not by people using Google Translate?

balloob:
We should add a note that our names "Home Assistant" and "Hass.io" should not be translated into other languages. Our name is the same in each language.

armills:
I think we should extend what @balloob said above to apply to all proper nouns, (i.e. other project names, etc.) at least until we find an exception that makes us rethink it.

PR #3804, opened 2017-10-27
Update python_script.markdown

ztalbot2000:
Using Homeassistant 0.56.2, the latest at this time, the component python_script.py defines the DOMAIN as 'python_script' and the FOLDER 'python_scripts'.  As is you will encounter the error: 'Unable to find component python_script'.  It took awhile to figure out that figure out that the DOMAIN and FOLDER should be linked and a quick change to the documentation was all that was needed.

**Description:**

### DIFF
--- a/source/_components/python_script.markdown
+++ b/source/_components/python_script.markdown
@@ -25,7 +25,9 @@ This component allows you to write Python scripts that are exposed as services i
 
 ## {% linkable_title Writing your first script %}
 
- - Add to `configuration.yaml`: `python_script:`
+ - Add to `configuration.yaml`: 
+   `python_script:`
+   `  python_scripts:`
  - Create folder `<config>/python_scripts`
  - Create a file `hello_world.py` in the folder and give it this content:
 


arsaboo:
Are you sure this is required? I only have the following in my config:
```yaml
python_script:
```

And have all the scripts inside the folder `python_scripts` and it works for me.

fabaff:
Same here `0.56.2` and `dev` works both fine. As we set the validation to `extra=vol.ALLOW_EXTRA)` would your change pass and put `python_scripts` with `None` in the configuration dict.

ztalbot2000:
For something so simple, I spent days trying to determine the problem.  I
took apart the component python_script to figure out what I documented.
Now I try and go back to reproduce it and I cannot!  I even went an
restored my configuration.yaml using TimeCapsule for every day of the last
week and it still works.
I uninstalled everything and reinstalled and again it still works.

I thought I deleted anything that could possibly be left behind in some
configuration and I cannot get it to fail anymore.  I do know that as I
documented it, it does work and I believe it to be as it was intended.  I
guess what you decide to do with my document change is up to you.  At least
my commit reasons will show others the way if they encountered what I did.

Take care,
John Talbot

fabaff:
I will close it. We can re-open it anytime if the error can be reproduced.

PR #3854, opened 2017-10-31
Documentation for retry on error in InfluxDB

janLo:
**Description:**

This adds an optional max_retries parameter to the InfluxDB component
to specify if and how often the component should try to send the data
if the connection failed due to an IOError.

The sending will be scheduled for a retry in 20 seconds as often as the
user specified. This can be handy for flaky getwork connections between
the DB and Homeassistant or outages like daily DSL reconnects.


frenck:
👍  Can be merged as soon as the parent PR is merged.

PR #3856, opened 2017-11-01
Remove 'development' from http and update frontend

fabaff:
**Description:**
Remove 'development' from http and update frontend.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#10267

### DIFF
--- a/source/_components/frontend.markdown
+++ b/source/_components/frontend.markdown
@@ -18,6 +18,13 @@ This offers the official frontend to control Home Assistant.
 frontend:
 ```
 
+Configuration variables:
+
+- **themes** (*Optional*): Allow to define different themes. See below for further details.
+- **extra_html_url** (*Optional*): List of addtional [resources](/developers/frontend_creating_custom_ui/) to load.
+- **development_repo** (*Optional*): Allow to point to a repository. Useful for Frontend development.
+
+
 ## {% linkable_title Themes %}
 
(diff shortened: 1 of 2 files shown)


source/_components/frontend.markdown
```diff
@@ -18,6 +18,13 @@ This offers the official frontend to control Home Assistant.
 frontend:
 ```
 
+Configuration variables:
```
andrey-git: Isn't there some new format for config variables?

source/_components/frontend.markdown
```diff
@@ -18,6 +18,13 @@ This offers the official frontend to control Home Assistant.
...
+
+- **themes** (*Optional*): Allow to define different themes. See below for further details.
+- **extra_html_url** (*Optional*): List of addtional [resources](/developers/frontend_creating_custom_ui/) to load.
+- **development_repo** (*Optional*): Allow to point to a repository. Useful for Frontend development.
```
andrey-git: Allow to point to a directory containing frontend file instead of taking them from a pre-built pypi package. Useful for Frontend development.

source/_components/frontend.markdown
fabaff: There is but at the moment it's more do it quick than do it nice. As I need to update it anyways, I can add it.

PR #3871, opened 2017-11-02
Release 0 57

balloob:
🎉


Landrash:
Documentation for the changes to Tradfri seems to be missing.
Following the `next` branch will cause the Trådfri component to fail to initialize if the api_key is set in the configuration.

PR #3872, opened 2017-11-02
Update xiaomi (miio) retrieve token documentation

drgx:
**Description:**
Add easier way to retrieve miio token by using nodejs and [miio](https://github.com/aholstenson/miio) package

### DIFF
--- a/source/_components/vacuum.xiaomi_miio.markdown
+++ b/source/_components/vacuum.xiaomi_miio.markdown
@@ -139,6 +139,29 @@ method and requires a `key` (16 alphanumeric chars), which can be obtained
 easily via a hidden menu item at the Mi-Home app.
 </p>
 
+#### {% linkable_title Using Nodejs %}
+Follow the pairing process using your phone and Mi-Home app. After the pairing process competed, install Nodejs to your computer [link](https://nodejs.org/en/download/)
+
+After Nodejs installed. Try to install [miio](https://github.com/aholstenson/miio/blob/master/docs/management.md) buy using npm command
+```bash
+$ npm install --save miio
+```
+After the `miio` package installed properly, simply run
+```bash
+$ miio --discover
+```
+
+And it will simply return the list of all miio devices information that connected in the same network
+Example result:
+```
+Device ID: 48765421
+Model info: zhimi.airpurifier.m1 (air-purifier)
+Address: 192.168.100.9 (zhimi-airpurifier-m1_miio48765421)
+Token: token-as-hex-here via auto-token
+Support: At least basic
+```
+
+
 #### {% linkable_title Windows and Android %}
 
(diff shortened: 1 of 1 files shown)


arsaboo:
Those instructions are not working...can you please double check:
```
arsaboo@aloknuc:~$ sudo npm install --save miio
[sudo] password for arsaboo:
arsaboo@1.0.0 /home/arsaboo
└─┬ miio@0.13.0
  └─┬ mkdirp@0.5.1
    └── minimist@0.0.8

npm WARN arsaboo@1.0.0 No description
npm WARN arsaboo@1.0.0 No repository field.
arsaboo@aloknuc:~$ miio --discover
miio: command not found
```

arsaboo:
Looks like the installation command should be `npm install -g miio` as indicated here https://github.com/aholstenson/miio/blob/master/docs/missing-devices.md

syssi:
What about using python-miio (the underlying library of the component) and the provided tools of the upstream: https://github.com/rytilahti/python-miio#miio-extract-tokens

drgx:
@arsaboo yahh.. it should be `npm install -g miio` why the hell i'm typing `sudo npm install --save miio` .... i think' I'm drunk :confused:... sorry will fix it ASAP
@syssi I never try `python-miio` lib, but i will try it and do another pull request when it's works

drgx:
@arsaboo is this pull-request acceptable? or i need to create using `python-miio`?

rytilahti:
The described method of using the automatic discovery (which is what `miio --discover` afaik does) is also doable with python-miio (`mirobo discover --handshake 1`), but this way of extracting tokens only works on specific set of devices, and sometimes only when the device has not been previously connected with the official app. This in turn requires pointing those users to extract the tokens from backups / sqlite databases which is what the current documentation here explains: https://home-assistant.io/components/vacuum.xiaomi_miio/#retrieving-the-access-token .

The python-miio documentation is currently being reworked in this incomplete PR https://github.com/rytilahti/python-miio/pull/93 which I hope can be used as a reference for an easy obtaining tokens: https://github.com/rytilahti/python-miio/pull/93/files#diff-3b734457dc9f0169aa39ce25f6b0c89aR36 . Unfortunately I do not have access to iOS devices, but I would happily accept contributions.

Considering homeassistant is using python-miio for supporting those devices (and its therefore installed for those users), it _may_ make sense to avoid pointing users to install nodejs-based miio, which is to my experience not a simple feat to do e.g. on raspbian. Just my 0.05€ :-)

drgx:
Thank you for the explanation @rytilahti. Yah based on your explenation, I think it's more make sense using `python-miio` since the users don't need to install nodejs. (IMO nodejs is easier becuase every day I work with nodejs  :rofl:) 
I will close this pull request since there is a better solution using `python-miio`. :+1:

rytilahti:
Ok! I will try to get the documentation ready and to be shipped hopefully this weekend, then we can modify the documentation to point to that :-)

PR #3883, opened 2017-11-04
documentation for Lutron Caseta update

mdonoughe:
**Description:**

This is the documentation update to go with home-assistant/home-assistant#10286.

I'm not sure I'm doing the right thing with the pairing script, but I made some improvements over the original:
- I rewrote it in Python, so it should be more accessible. The old script only worked in Bash on Linux (not even Bash on other platforms) and only with certain tools installed. At least for Home Assistant, it seems more likely that the user can execute a Python script.
- The CA certificate is retrieved from the bridge. Previously, this was a manual step.
- The files we need are generated as separate files so the user doesn't need to split them apart afterwards.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#10286


ontl:
The sample config in the documentation expects "caseta.key" and "caseta.crt", but it looks like the script in the zip generates "client.key" and "client.crt" instead. Recommend changing the get_lutron_cert.py script to generate "caseta.xxx".

Thanks for all your work on this fix!

mdonoughe:
I thought I'd changed all of those, but apparently I only changed the documentation. It should be better now.

PR #3887, opened 2017-11-04
Fix home-assistant release for component

ttroy50:
Component was actually releases in v0.57 of home-assistant


fabaff:
Could be fixed in `current`.

PR #3888, opened 2017-11-04
fix ha-release tag for component

ttroy50:
Fixes the ha-release for the component


fabaff:
The right file to fix it would be `source/_components/climate.ephember.markdown` in the `current` branch.

PR #3890, opened 2017-11-04
Added instructions to save Ring video file using automation

arsaboo:
**Description:**
Added instructions to use an automation to save Ring video files (which may be easier for most users).


arsaboo:
Not sure why this this was closed. It simplifies the docs.

Issue #3929, opened 2017-11-08
[Roadmap] Update installation guideline


c727:
**Background**
* discussion on Discord and GitHub #3918 
* https://www.youtube.com/watch?v=s5_dtRoFYyU&feature=youtu.be&t=368

**Tag to use for your PR**
[Install]

* Remove unnecessary stuff
* lead new users to Hass.io, if they reach limitations they should check the alternatives

Todo
- [ ] restructure navigation bar/ sidebar
- [ ] what are the limitations of hass.io, what isn't working

DubhAd:
Rather than directing people just to HassIO, I'm strongly in favour of guiding and informing. Let people know about the "top 3" install methods (HassIO, Hassbian, VENV) and why they might want to chose them (as per Ryan's [draft here](https://github.com/home-assistant/home-assistant.github.io/pull/3918), but more complete).

If we just go "Use HassIO" then there's a strong risk we'll lose the mid ground, only retaining those who just want a Wink/SmartThings/etc replacement, and those who're Docker developers.

c727:
@DubhAd 
The Hassbian docs need some rework before we can add it :stuck_out_tongue_closed_eyes:

PR #3931, opened 2017-11-08
[Install] Refresh Venv guide

c727:
#3929

### DIFF
--- a/source/_docs/installation/virtualenv.markdown
+++ b/source/_docs/installation/virtualenv.markdown
@@ -74,9 +74,9 @@ $ sudo usermod -G dialout -a homeassistant
 This can be anywhere you want.  We chose to put it in `/srv`. You also need to change the ownership of the directory to the user you created above.
 
 ```bash
-$ sudo mkdir /srv/home-assistant
-$ sudo chown homeassistant:homeassistant /srv/home-assistant
-$ virtualenv -p python3 /srv/home-assistant
+$ sudo mkdir /srv/homeassistant
+$ sudo chown homeassistant:homeassistant /srv/homeassistant
+$ virtualenv -p python3 /srv/homeassistant
 ```
 
 ## {% linkable_title Install or update Home Assistant %}
@@ -90,7 +90,7 @@ $ sudo su -s /bin/bash homeassistant
 The `su` command means 'switch' user. We use the '-s' flag because the `homeassistant` user is a system user and doesn't have a default shell by default (to prevent attackers from being able to log in as that user).
 
 ```bash
-$ source /srv/home-assistant/bin/activate
+$ source /srv/homeassistant/bin/activate
 (home-assistant)$ pip3 install --upgrade homeassistant
 ```
 
@@ -99,11 +99,11 @@ $ source /srv/home-assistant/bin/activate
 There are two ways to launch Home Assistant. If you are **in** the virtualenv, you can just run `hass` and it will work as normal. If the virtualenv is not activated, you just use the `hass` executable in the `bin` directory mentioned earlier. There is one caveat... Because Home Assistant stores its configuration in the user's home directory, we need to be the user `homeassistant` user or specify the configuration with `-c`.
 
 ```bash
-$ sudo -u homeassistant -H /srv/home-assistant/bin/hass
+$ sudo -u homeassistant -H /srv/homeassistant/bin/hass
 ```
 
 The `-H` flag is important. It sets the `$HOME` environment variable to `/home/homeassistant` so `hass` can find its configuration.
 
 ## {% linkable_title Starting Home Assistant on boot %}
 
-The [autostart instructions](/getting-started/autostart/) will work just fine, just be sure to replace `/usr/bin/hass` with `/srv/home-assistant/bin/hass` and specify the `homeassistant` user where appropriate.
+The [autostart instructions](/getting-started/autostart/) will work just fine, just be sure to replace `/usr/bin/hass` with `/srv/homeassistant/bin/hass` and specify the `homeassistant` user where appropriate.
--- a/source/_docs/installation/virtualenv.markdown
+++ b/source/_docs/installation/virtualenv.markdown
@@ -35,20 +35,20 @@ $ sudo pip3 install --upgrade virtualenv
 ```bash
 $ cd $HOME
 $ mkdir home-assistant
-$ virtualenv -p python3 home-assistant
+$ virtualenv -p python3 homeassistant
 ```
 
 ## {% linkable_title Step 3: Install or update Home Assistant %}
 
 ```bash
-$ source $HOME/home-assistant/bin/activate
-(home-assistant)$ pip3 install --upgrade homeassistant
+$ source $HOME/homeassistant/bin/activate
+(homeassistant)$ pip3 install --upgrade homeassistant
 ```
 
 ## {% linkable_title Step 4: Run Home Assistant %}
 
 ```bash
-$ $HOME/home-assistant/bin/hass
+$ $HOME/homeassistant/bin/hass
 ```
 
 ### {% linkable_title Advanced guide %}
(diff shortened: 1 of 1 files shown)


DubhAd:
I'd be in favour of using the same location for the virtual environment as is used for Hassbian. Doing that makes it easier to help people regardless of how they've installed it.

source/_docs/installation/virtualenv.markdown
```diff
@@ -1,20 +1,27 @@
...
+description: "How to install Home Assistant in a virtual environment."
 date: 2016-4-16 16:40
 sidebar: true
 comments: false
 sharing: true
 footer: true
 redirect_from: /getting-started/installation-virtualenv/
 ---
+<p class='note'>
+Note: Beginners should check our [Getting started guide](/getting-started/) first. This is for users that require a flexible installation or want to run Home Assistant on any platform.
```
fabaff: Remove `Note:`. Again, `platform`...no no differentiation about hardware or software. Better leave it out. The note addresses the beginner and shouldn't confuse them.

source/_docs/installation/virtualenv.markdown
```diff
@@ -1,20 +1,27 @@
 ---
 layout: page
-title: "Installation in virtualenv"
-description: "Instructions how to install Home Assistant in a virtual environment."
+title: "Installation: Python virtual envenvironment"
```
fabaff: What's the advantage of starting to prefix the title with the section where the doc is located?

Also, there is a typo in `envenvironment`.

source/_docs/installation/virtualenv.markdown
```diff
@@ -23,9 +30,31 @@ $ sudo apt-get install python3-pip python3-dev
 $ sudo pip3 install --upgrade virtualenv
 ```
 
-### {% linkable_title Step 1: Create a Home Assistant user & group %}
+## {% linkable_title Step 2: Setup virtualenv %}
+
+```bash
+$ cd $HOME
+$ mkdir homeassistant
+$ virtualenv -p python3 homeassistant
```
fabaff: I would prefer `$ python3 -m venv homeassistant`. As this is the same way it's described in the official Python documentation.

source/_docs/installation/virtualenv.markdown
```diff
@@ -1,20 +1,27 @@
...
 date: 2016-4-16 16:40
 sidebar: true
 comments: false
 sharing: true
 footer: true
 redirect_from: /getting-started/installation-virtualenv/
 ---
+<p class='note'>
+Note: Beginners should check our [Getting started guide](/getting-started/) first. This is for users that require a flexible installation or want to run Home Assistant on any platform.
+<p class='note'>
```
fabaff: That's will lead to invalid HTML after rendering if the checks fail.

source/_docs/installation/virtualenv.markdown
```diff
@@ -23,33 +30,54 @@ $ sudo apt-get install python3-pip python3-dev
...
-This step is optional, but it's a good idea to give services like Home Assistant their own user. It gives you more granular control over permissions, and reduces the exposure to the rest of your system in the event there is a security related bug in Home Assistant. This is a reasonably Linux oriented step, and will look different on other operating systems (or even other Linux distributions).
+It's a good idea to give services like Home Assistant their own user. It gives you more granular control over permissions, and reduces the exposure to the rest of your system in the event there is a security related bug in Home Assistant. This is a reasonably Linux oriented step, and will look different on other operating systems (or even other Linux distributions).
 
 ```bash
 $ sudo adduser --system homeassistant
 $ sudo addgroup homeassistant
 ```
 
-Home Assistant stores its configuration in `$HOME/.homeassistant` by default, so in this case, it would be in `/home/homeassistant/.homeassistant`
+Home Assistant stores it's configuration in `$HOME/.homeassistant` by default, so in this case, it would be in `/home/homeassistant/.homeassistant`.
```
MartinHjelmare: It should be `its`. Here it should be a possessive pronoun.

source/_docs/installation/virtualenv.markdown
```diff
@@ -59,48 +87,21 @@ $ sudo su -s /bin/bash homeassistant
...
 (homeassistant)$ pip3 install --upgrade homeassistant
 ```
 
-And that's it... you now have Home Assistant installed, and you can be sure that every bit of it is contained in `/srv/homeassistant`.
+## {% linkable_title Run Home Assistant (Basic guide step 4) %}
 
-### {% linkable_title Finally... Run Home Assistant %}
-
-There are two ways to launch Home Assistant. If you are **in** the virtualenv, you can just run `hass` and it will work as normal. If the virtualenv is not activated, you just use the `hass` executable in the `bin` directory mentioned earlier. There is one caveat... Because Home Assistant stores its configuration in the user's home directory, we need to be the user `homeassistant` user or specify the configuration with `-c`.
+There are two ways to launch Home Assistant. If you are **in** the virtualenv, you can just run `hass` and it will work as normal. If the virtualenv is not activated, you just use the `hass` executable in the `bin` directory mentioned earlier. There is one caveat... Because Home Assistant stores it's configuration in the user's home directory, we need to be the user `homeassistant` user or specify the configuration with `-c`.
```
MartinHjelmare: `stores its configuration` not `stores it's configuration`

source/_docs/installation/virtualenv.markdown
```diff
@@ -59,48 +87,21 @@ $ sudo su -s /bin/bash homeassistant
...
 ```bash
 $ sudo -u homeassistant -H /srv/homeassistant/bin/hass
 ```
 
-The `-H` flag is important. It sets the `$HOME` environment variable to `/home/homeassistant` so `hass` can find its configuration.
-
-### {% linkable_title Upgrading Home Assistant %}
-
-Upgrading Home Assistant is simple, just repeat steps 3, 5 and 6.
+The `-H` flag is important. It sets the `$HOME` environment variable to `/home/homeassistant` so `hass` can find it's configuration.
```
MartinHjelmare: Same as above.

source/_docs/installation/virtualenv.markdown
c727: 😢 thanks... I should refresh my english

source/_docs/installation/virtualenv.markdown
```diff
@@ -94,13 +94,13 @@ $ source /srv/homeassistant/bin/activate
 
 ## {% linkable_title Run Home Assistant (Basic guide step 4) %}
 
-There are two ways to launch Home Assistant. If you are **in** the virtualenv, you can just run `hass` and it will work as normal. If the virtualenv is not activated, you just use the `hass` executable in the `bin` directory mentioned earlier. There is one caveat... Because Home Assistant stores it's configuration in the user's home directory, we need to be the user `homeassistant` user or specify the configuration with `-c`.
+There are two ways to launch Home Assistant. If you are **in** the virtualenv, you can just run `hass` and it will work as normal. If the virtualenv is not activated, you just use the `hass` executable in the `bin` directory mentioned earlier. There is one caveat... Because Home Assistant stores its configuration in the user's home directory, we need to be the user `homeassistant` user or specify the configuration with `-c`.
```
MartinHjelmare: There's an extra `user` before `homeassistant`.

source/_docs/installation/virtualenv.markdown
```diff
@@ -1,20 +1,27 @@
...
 
-Virtualenvs are pretty easy to setup. This example will walk through one method of setting one up (there are certainly others). We'll be using Debian in this example (as many Home Assistant users are running Raspbian on a Raspberry Pi), but all of the Python related steps should be the same on just about any platform.
+Virtualenvs are pretty easy to setup. We'll be using Debian in this example (as many Home Assistant users are running Raspbian on a Raspberry Pi), but all of the Python related steps should be the same on just about any platform.
 
-### {% linkable_title Step 0: Install some dependencies %}
+### {% linkable_title Basic guide %}
+
+The basic guide is for testing Home Assistant. Also check the advanced guide below for productional instances.
```
MartinHjelmare: "...for instances used in production."

PR #3933, opened 2017-11-08
[Install] Update Alternative install

c727:
#3929

### DIFF
--- a/source/_docs/installation.markdown
+++ b/source/_docs/installation.markdown
@@ -10,6 +10,10 @@ footer: true
 redirect_from: /getting-started/installation/
 ---
 
+<p class='note'>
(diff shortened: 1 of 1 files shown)


source/_docs/installation.markdown
```diff
@@ -10,6 +10,10 @@ footer: true
 redirect_from: /getting-started/installation/
 ---
 
+<p class='note'>
+Note: Beginners should check our [Getting started guide](/getting-started/) first. This is for users that require advanced installtions
```
fabaff: Remove `Note:`, it wil be rendered as a note. Also, there is a typo in `installtions` and the period at the end is missing.

PR #3951, opened 2017-11-10
Test badge text in translations and use placeholders

c727:
@armills


c727:
@frenck seems someone already added this to the docs

armills:
Ah, I think that was my mistake. Pace has been pretty fast lately.

PR #3961, opened 2017-11-11
truth table for condition sun

flowolf:
truth table to illustrate sunset condition with and without offsets.


frenck:
@flowolf Thank you for this addition! 🏅

PR #4009, opened 2017-11-18
Fix line endings on rainbird switch

MartinHjelmare:
**Description:**
At some point this file got windows line endings. Fixed now.


frenck:
I've added it to the 0.58 milestone since I think it should be fixed in the next release.

PR #4037, opened 2017-11-21
Update group.markdown to show that you can customize order of groups.

isabellaalstrom:
**Description:**

I updated the docs for groups to make sure it's clear that you can use customize to reorder your groups.

## Checklist:


arsaboo:
We should clarify that `order` only works with groups that are not defined. So, `groups` that are defined are appear in the order they are listed.

PR #4038, opened 2017-11-21
Add reference to HA version and what's new in the latest release.

JerryWorkman:
Provide information to users so that they can determine the impact of the latest version on their system. Updates many times require changes to configuration, renaming of entities (z-wave for example), and deprecation of features,


frenck:
@JerryWorkman That is a pretty useful addition! Thank you!

PR #4041, opened 2017-11-22
Optimise images.

grischard:
Saved 12.5 MB out of 56.7 MB. 27.5% per file on average (up to 99.6%)

## Description:

For every file, [ImageOptim](https://github.com/ImageOptim/ImageOptim) compared the different tools and picked the smallest file that was produced by or by a combination of `advpng`, `pngout` and `zopfli` for PNGs, `gifsicle` for GIFs, `svgo` for SVGs or `jpegoptim` and `mozjpeg` for JPEGs.

This includes a rather ironic 74.8% change on `source/images/blog/2016-08-optimizing-web-app/performance-diagram.png` which went from 38237 bytes to 9620 bytes.

No related issues, no code change, no functionality change, no visual differences in images (compression is lossless).


grischard:
That makes sense. There's a few stray images in `source/demo/` and `source/favicon.png` and `source/static/` too, and it looks like the only two third-party sources are  jwplayer/jwplayer and octopress/classic-theme. I'll submit separate PRs there, and update the PR here.

grischard:
I actually can't find the source for the jwplayer glow theme. It looks like those files haven't been touched since 2014. What do you think we should do?

grischard:
So both https://github.com/jwplayer/jwplayer/pull/2541 and https://github.com/octopress/classic-theme/pull/6 have been opened, and haven't received any attention :). I'm happy to re-optimise the images and submit a new PR here if and when we update those if upstream ignores them.

@fabaff do you see this as blocking to merge this PR, or could it be merged like that?

grischard:
Thanks!

PR #4071, opened 2017-11-25
Update to avoid putting HA user in sudoers

DubhAd:
None of the steps require that HA is added to sudoers or given a password. Rewriting those steps to remove that, to improve overall security.

### DIFF
--- a/source/_docs/ecosystem/certificates/lets_encrypt.markdown
+++ b/source/_docs/ecosystem/certificates/lets_encrypt.markdown
@@ -240,21 +240,7 @@ In cases where your ISP blocks port 80 you will need to change the port forward 
 
 Now SSH in to the device your Home Assistant is running on.
 
-<p class='note'>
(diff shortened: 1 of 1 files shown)


source/_docs/ecosystem/certificates/lets_encrypt.markdown
```diff
@@ -474,37 +454,12 @@ $ crontab -e
  
  
 #### Option 2:
```
fabaff: Did you remove the whole section on purpose?

DubhAd: Yes, because the section is entirely about adding homeassistant to sudoers.

fabaff: Not according to the diff. `Option 2:` covers the renewal process through an automation rule.

frenck:
@DubhAd Could you take a look at the feedback provided by @fabaff, that would be awesome. This PR is running a bit stale...

DubhAd:
Yup, will do. Got distracted by other things.

frenck:

PR #4097, opened 2017-11-29
Template trigger does not accept entity_id

amelchio:
**Description:**

This reverts #4018 which appears to give wrong advice.

Fixes home-assistant/home-assistant#10838

CC @DubhAd

## Checklist:

  - [X] Branch: Fixes, changes and adjustments should be created against `current`. New documentation for platforms/components and features should go to `next`. 

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -138,10 +138,6 @@ The US Naval Observatory has a [tool](http://aa.usno.navy.mil/data/docs/AltAz.ph
 
 ### {% linkable_title Template trigger %}
 
-<p class='note warning'>
-  If your template trigger has no `entity_id` listed, then it is evaluted on *every* state change for *every* entity. Instead you should create a [template sensor](/components/sensor.template/) or [template binary sensor](/components/binary_sensor.template/) and use that in your automation.
-</p>
-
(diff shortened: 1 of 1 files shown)


DubhAd:
The update came from advice given on Discord by the devs (@dale3h amongst others, I think @balloob too) who'd said that template triggers were evaluated on every state change for every entity if no entity_id was specified. If that information is now "old news" that's good to know.

iamtpage:
See here: 
https://github.com/home-assistant/home-assistant/blob/0f8e48c26df1ceb03c030b988a1d3d9bc3492f5e/homeassistant/helpers/template.py#L47

When a template string is defined and it contains entities, the template sensor automatically extracts them and adds them to the triggered entity list: 
https://github.com/home-assistant/home-assistant/blob/0f8e48c26df1ceb03c030b988a1d3d9bc3492f5e/homeassistant/helpers/template.py#L58

If a template string is defined and it contains no entities, it then proceeds to add all entities to the triggered entity list here: 
https://github.com/home-assistant/home-assistant/blob/0f8e48c26df1ceb03c030b988a1d3d9bc3492f5e/homeassistant/helpers/template.py#L49

I think thats where the confusion is coming from. Regardless, the documentation does not hold true to what is actually going on in the code. So either a PR needs to be submitted to change the code, or this PR needs to be approved to change the documentation.

PR #4152, opened 2017-12-07
Added update instructions

Naesstrom:
Added back the instructions for updating the AIO installer

**Description:**

## Checklist:


DubhAd:
I think this would also have to include a warning to upgrade Python, and build a new Python 3.5.3 (or later) or Python 3.6 venv.

fabaff:
We don't want that people continue to use the AIO installer and support them because Hass.io and HASSbian are under active development. The instructions are still available in the log if one really needs then.

https://github.com/home-assistant/fabric-home-assistant is read-only thus there will be no changes/updates.

frenck:
The warning is pretty clear about this. Home Assistant will no longer support the AIO installation method and it is advise to move to another method.

This PR adds some level of legacy support, which should be avoided.

PR #4191, opened 2017-12-11
Update virtual environment instructions

balloob:
Greatly simplify the virtual environment instructions by:
 - Not putting a title between every command, so it's easy to see what we will do
 - Actually talking just about installing it on a virtual environment instead of becoming more of a Raspbian guide. There is already a standalone installation page for that.
 - Remove the autostart things. We don't know the OS/environment this is getting installed.

Feedback welcome.

### DIFF
--- a/source/_docs/installation/virtualenv.markdown
+++ b/source/_docs/installation/virtualenv.markdown
@@ -9,99 +9,35 @@ sharing: true
 footer: true
 redirect_from: /getting-started/installation-virtualenv/
 ---
-<p class='note'>
-Beginners should check our [getting started guide](/getting-started/) first.
-</p>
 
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
I partly agree with @Landrash, and I think we could link to the raspbian guide for a more complete example, or keep keep the advanced guide at the bottom of the page, and just condense the guide at the top here without headers.

fabaff:
From my point of view, everything that is not related to the `venv` should be removed to avoid confusion for users, including adding users to group (which is already covered at other places), but it wouldn't harm to link to other guides or "next steps" like autostart. 

"[Manual installation on a Raspberry Pi](https://home-assistant.io/docs/installation/raspberry-pi/)" covers an advance setup and is/was duplicating a lot of information of the `venv` instructions. Or perhaps it was vice-versa.

balloob:
So people have the option to select from a wide range of installation methods. I am not sure if it helps to link to other OSes, if people were interested in that they would probably have clicked on the card for their OS?

Linking to other things is also hard because we don't know what OS they are on. It can even be Windows.

balloob:
I've added a link at the bottom to the other guides.

balloob:
I'm open for suggestions on where to move the non-virtual env related stuff. I am not familiar enough with the instructions to "guess" the platform it works on.

PR #4200, opened 2017-12-12
Add problem device class

armills:
**Description:**
Docs PR for new problem device class.

Polymer PR: https://github.com/home-assistant/home-assistant-polymer/pull/726
Backend PR: https://github.com/home-assistant/home-assistant/pull/11130


balloob:
This will also need to be added here: https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/binary_sensor/__init__.py#L22

PR #4257, opened 2017-12-20
added hassio shutdown and reboot comment

oliverscu:
Added a note about hassio shutdown & reboot to reduce chances that users pull power on a working system and corrupt their SD cards

**Description:**

## Checklist:


frenck:
I'm not sure about this PR.
It is a quickstart guide... I do think it is generally bad practice to add "notes" to it. I could add a ton of notes like this, which I won't for the same reason I think we should not add this one.

frenck:
I gave it another thought. The SSH add-on is optional. So it is not even guaranteed a user will have this, furthermore, the add-on needs to be installed first.

I'm closing this one for the above-stated reasons.

PR #4326, opened 2018-01-01
added option example

CCOSTAN:
it wasn't clear to me.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/media_player.cast.markdown
+++ b/source/_components/media_player.cast.markdown
@@ -17,16 +17,26 @@ ha_iot_class: "Local Polling"
 
 Google Cast devices like Android TVs and Chromecasts will be automatically discovered if you enable [the discovery component]({{site_root}}/components/discovery/). There is a issue where Chromecasts can only be discovered if your device is connected to the same subnet as your Chromecast.
 
-Chromecast platform can also be forced to load by adding the following lines to your `configuration.yaml`:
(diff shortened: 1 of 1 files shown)


source/_components/media_player.cast.markdown
```diff
@@ -23,6 +23,7 @@ Chromecast platform can also be forced to load by adding the following lines to
 # Example configuration.yaml entry
 media_player:
   - platform: cast
+    host: 192.168.1.10 #this is an optional variable
```
fabaff: We try to keep the configuration samples minimal. The `host` entry is mentioned a couple of line down the description of the variables.

CCOSTAN: agreed but as someone who just implemented it, it was not easily identifiable how to add the single variable.  at first, I tried indenting and then with a dash in a list..  There were no good examples.  

I understand the desire to keep the configuration minimal but for me, an example of one of the options was needed to understand the format.  This is also why I just added the single option to show the format.

CCOSTAN: @fabaff I can add a second 'advanced' example at the end of the page with both options if you think that would be cleaner for new users.

fabaff: Sure, we do that at a couple of places. If it helps then we should add it.

PR #4332, opened 2018-01-03
Update binary_sensor.xiaomi_aqara.markdown

gerard33:
Remove unnecessary spaces in table, to avoid text in table crossing the text in the right frame


gerard33:
@fabaff, hmm the table is still crossing the text in the right frame...
Maybe the number of columns should be reduced by combining some data in the columns?

PR #4402, opened 2018-01-13
Added a "tools" section

Guyanthalas:
Tools section to include information on tools that are known to work well with the YAML syntax. Hopefully we can point to this in FAQ's when assisting new people who are having trouble understanding how to write in YAML.

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/configuration/yaml.markdown
+++ b/source/_docs/configuration/yaml.markdown
@@ -10,7 +10,7 @@ footer: true
 redirect_from: /getting-started/yaml/
 ---
 
-Home Assistant uses the [YAML](http://yaml.org/) syntax for configuration. YAML might take a while to get used to but is really powerful in allowing you to express complex configurations.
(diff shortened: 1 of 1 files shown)


source/_docs/configuration/yaml.markdown
```diff
@@ -93,3 +93,13 @@ lights: !include lights.yaml
 ```
 
 More information about this feature can also be found at [splitting configuration](/docs/configuration/splitting_configuration/).
+
+### {% linkable_title Tools %}
+
+The community has found certain tools to be of use when editing YAML files.
+
+#### Notepad++
```
arsaboo: Why just Notepad++? There are other more capable options, e.g., Atom, vscode.

fabaff: We shouldn't start to list editors. There are way to many to keep track and to maintain in the future.

Guyanthalas: @arsaboo For sure, but I only have knowledge of the one I use. I figured that people could add the ones that we know work well. A starting point, if you will. :)

@fabaff Makes sense. I just see MANY people having trouble with marking up YAML. This would allow us to point newer users to a collection of solid tools, and we could update FAQ's accordingly to point to the list. That way we can point people in the right direction if they are having trouble with the correct spacing.

c727:
we already recommend VSCode as multi-platform solution in our getting started. Also n++ does only syntax highlighting - no syntax or config validation.

PR #4423, opened 2018-01-15
Removed relative_time

DubhAd:
Apparently it was removed some time ago (bug from 0.30.1 saying it didn't work then), but the docs weren't updated.


arsaboo:
Indeed, `relative_time` did not work in templates starting somewhere around 0.30. See https://github.com/home-assistant/home-assistant/issues/3892 for discussion.

PR #4438, opened 2018-01-16
Pulled through notes of other path options

DubhAd:
Added notes that for Hass.io there may be multiple path options depending on the device.


nicxvan:
LGTM

Landrash:
My Z-wave USB device is mounted to `ttyAMA0` and so is my previous one. Changing this to one for a specific radio seems like a idea.

source/_docs/z-wave/installation.markdown
```diff
@@ -159,17 +159,21 @@ To enable Z-Wave, plug your Z-Wave USB stick into your Raspberry Pi 3 and add th
...
 ```yaml
 zwave:
-  usb_path: /dev/ttyAMA0
+  usb_path: /dev/ttyACM0
 ```
 
-For some devices the `/dev/ttyAMA0` device is not detected by udev and is therefore not mapped by Docker. To explicitly set this device for mapping to Home-Assistant, execute the following command using the ssh add-on:
+Depending on your Z-Wave device it may instead be `/dev/ttyAMA0` (eg Razberry board) or `/dev/ttyUSB0` (eg HUBUZB-1). 
+
+For some devices it is not detected by udev and is therefore not mapped by Docker. To explicitly set this device for mapping to Home-Assistant, execute the following command using the ssh add-on:
```
frenck: This is no longer true with the release of Hass.io 0.80. All serials devices are mapped automatically.

DubhAd: Ah, really? I'll update then ;)

frenck:
Resolved merge conflict, can be merged as soon as the build passes.

PR #4439, opened 2018-01-16
sync usb_path to other docs

Guyanthalas:
The hassio zwave documents reference usb_path: /dev/ttyACM0, so this document should be in sync. After some discussion it appears that this could be variable, in which case there should be a good way to figure out the CORRECT answer rather than using a guess as example code without stating it as such. Possibly linking back to the linux "how to find" section?

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/z-wave/installation.markdown
+++ b/source/_docs/z-wave/installation.markdown
@@ -162,14 +162,6 @@ zwave:
   usb_path: /dev/ttyACM0
 ```
 
-For some devices the `/dev/ttyACM0` device is not detected by udev and is therefore not mapped by Docker. To explicitly set this device for mapping to Home-Assistant, execute the following command using the ssh add-on:
-
-```bash
-$ curl -d '{"devices": ["ttyACM0"]}' http://hassio/homeassistant/options
-```
-
-After that, you need to change `usb_path` to `/dev/ttyACM0`.
-
 ### {% linkable_title RancherOS %}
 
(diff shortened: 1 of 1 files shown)


nicxvan:
I just dealt with this process myself.
I added the steps that I used here: https://community.home-assistant.io/t/help-setting-up-aeotec-z-stick-gen5-z-wave-plus-usb-stick/39175/17?u=nicxvan

If you want I can look at adding some of this information. I just wasn't sure where to do that. I do have another open PR that adds some info about enabling the serial interface for RPI.

source/_docs/z-wave/installation.markdown
```diff
@@ -159,16 +159,16 @@ To enable Z-Wave, plug your Z-Wave USB stick into your Raspberry Pi 3 and add th
 
 ```yaml
 zwave:
-  usb_path: /dev/ttyAMA0
+  usb_path: /dev/ttyACM0
 ```
 
-For some devices the `/dev/ttyAMA0` device is not detected by udev and is therefore not mapped by Docker. To explicitly set this device for mapping to Home-Assistant, execute the following command using the ssh add-on:
+For some devices the `/dev/ttyACM0` device is not detected by udev and is therefore not mapped by Docker. To explicitly set this device for mapping to Home-Assistant, execute the following command using the ssh add-on:
```
frenck: No longer true for Hass.io 0.80, since it will automatically map all devices.

Guyanthalas:
I've removed the language that no longer applies.

frenck:
Great thx! Merging...

PR #4492, opened 2018-01-22
Add esphomelib link

OttoWinter:
**Description:**

Add a link to [esphomelib](https://github.com/OttoWinter/esphomelib), which is a library that aims to make using Home Assistant with ESP32s really easy, in the documentation of the Light MQTT-JSON platform. 

Note: this can probably be considered somewhat of a self-promotion, but I strongly believe this library can help lots of makers save lots of time. Additionally, this library could also be used for other MQTT components such as sensors or switches, but plastering this "ad" all over those documentations is most likely annoying/too much - and implementing those components yourself, in contrast to lights, is relatively easy.

## Checklist:


OttoWinter:
And thank you for the fix!

fabaff:
There is no problem with promoting useful things. 

If you like then you could create a post in our blog. In the past we already did a couple of posts which where showing details about a specific topic that didn't fit well into the documentation. Because they where "cross-component/platform", too detailed, or use just a nice use case.

PR #4520, opened 2018-01-26
IHC component

dingusdk:
This PR is changes to the PR I have already done, so I am not sure it should be to the "next" branch. (I guess so since the previous PR is not current yet?)

This changes the wrong "host" in configuration to "url". (Sorry I did not see this, but another user testing this was kind to let me know)


frenck:
@dingusdk Thanks!

I did some minor tweaking and placed it on the list to pick up / merge into the next release.

PR #4557, opened 2018-01-31
Changed note about base url

JamesMcClelland:
Changed the location to be above the fold

**Description:**

I have noticed a few threads in the home assistant community where people have made this mistake of not configuring their base url when trying to get TTS to work, I think its quite easy to miss since you can just add the two line config at the top and it will appear to give you something.

Not a major change but thought it might help others :+1: 

home-assistant/home-assistant#<home-assistant PR number goes here>

## Checklist:

### DIFF
--- a/source/_components/tts.markdown
+++ b/source/_components/tts.markdown
@@ -22,6 +22,10 @@ tts:
   - platform: google
 ```
 
+<p class='note'>
+If you are running Home Assistant over SSL or from within a container, you will have to setup a base URL (`base_url`) inside the [http component](/components/http/).
+</p>
+
 The following optional parameters can be used with any platform. However the TTS component will only look for global settings under the configuration of the first configured platform:
 
 | Parameter           | Default | Description                                                                                                                                                                                                                                                                                                                                                                               |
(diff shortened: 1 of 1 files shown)


source/_components/tts.markdown
```diff
@@ -22,6 +22,10 @@ tts:
   - platform: google
 ```
 
+<p class='note'>
+If you are running Home Assistant over SSL or from within a container, you will have to setup a base URL (`base_url`) inside the [http component](/components/http/).
```
DubhAd: If a reverse proxy is in use, `base_url` isn't needed. I'm not sure though how to call that out, without adding confusion though.

frenck: I agree with @DubhAd. You could point to it without saying why.
Like:

```text
Depending on your setup, you might need to set a base URL (`base_url`) inside the [http component](/components/http/).
```

JamesMcClelland: With this change I had just moved what was currently in place as I thought that was already the correct wording :smile:  I can change it to your suggestion but I'm not sure how to add more commits to an open pull request or if I should just create a new pull request. Any advice would be helpful :+1: Cheers

frenck: Simply choose the "Files changed" tab here on GitHub (top of the PR), hit the edit button (pencil) button on the file header, and update 👍

frenck (dismissed):

frenck:

frenck:
Hmmm, the build fails on something not in this branch?
Need to pull in changes from current maybe?

frenck (approved):

frenck:
Rebased, merging it in.

PR #4559, opened 2018-01-31
Added note about error

DubhAd:
Adding a specific note about the error that indicates the library is missing


source/_docs/z-wave/installation.markdown
```diff
@@ -232,3 +232,11 @@ sudo chmod g+rw /dev/ttyAMA0
 ### {% linkable_title Device path changes %}
 
 If your device path changes when you restart, see [this guide](http://hintshop.ludvig.co.nz/show/persistent-names-usb-serial-devices/) on fixing it.
+
+### {% Unable to install python_openzwave %}
```
balloob: This is invalid Markdown. It should be`{% linkable_title … %}`

frenck: My bad, I should have spotted that one. Sorry!

PR #4571, opened 2018-02-03
Add market restriction note

ReneNulschDE:
**Description:**
Add a note that the component can integrate cars in Europe and Africa only currently.

## Checklist:

### DIFF
--- a/source/_components/mercedesme.markdown
+++ b/source/_components/mercedesme.markdown
@@ -21,6 +21,10 @@ This component provides the following platforms:
  - Sensors - such as fuel status, service interval, remaining km...
  - Device tracker - to track location of your car
 
+<p class='note warning'>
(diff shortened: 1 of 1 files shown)


source/_components/mercedesme.markdown
```diff
@@ -21,6 +21,10 @@ This component provides the following platforms:
  - Sensors - such as fuel status, service interval, remaining km...
  - Device tracker - to track location of your car
 
+<p class='note warning'>
+  The component can integrate cars out of the European and Africa markets only.
```
Landrash: Shouldn't this be *Africa**n*** markets?

ReneNulschDE: Thank you for your valid comment. I've created a new PR to provide a fix.

PR #4574, opened 2018-02-04
Had issues installing as the "wheel" package was missing on dietpi :)

rniemand:
**Description:**

## Checklist:

### DIFF
--- a/source/_docs/installation/raspberry-pi.markdown
+++ b/source/_docs/installation/raspberry-pi.markdown
@@ -64,6 +64,7 @@ $ source bin/activate
 Once you have activated the virtual environment you will notice the prompt change and then you can install Home Assistant.
 
 ```bash
+(homeassistant) homeassistant@raspberrypi:/srv/homeassistant $ pip install wheel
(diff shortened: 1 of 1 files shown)


source/_docs/installation/raspberry-pi.markdown
```diff
@@ -64,6 +64,7 @@ $ source bin/activate
 Once you have activated the virtual environment you will notice the prompt change and then you can install Home Assistant.
 
 ```bash
+(homeassistant) homeassistant@raspberrypi:/srv/homeassistant $ pip install wheel
```
DubhAd: For consistency, this should be `pip3`. It's also worth calling out that it's only necessary for DietPi.

rniemand: Agreed, as pointed out this is for a different flavor of the OS, but it is a nice to have imho as it opens up a whole lot of new SBC's to run hass. In my case this was an Orange Pi One and hass runs like a beast!

Landrash: This page is called raspberry-pi. If it's for other SBC's then we need to create a separate page for those :)

PR #4597, opened 2018-02-07
Added note about display in Overview

DubhAd:
It's come up quite a few times, with people asking why some scripts have a toggle, and some don't. This is to add a little clarity about it (I'm **assuming** that `wait` also results in a toggle - it'd be good to get that verified).


frenck:
Thank you @DubhAd

PR #4603, opened 2018-02-08
Make distinction between hass.io and homeassistant when installing

gpbenton:
**Description:**

Following discussion in this thread https://community.home-assistant.io/t/difference-home-assistant-hass-io/42097 I decided to try and help clarify the getting started page to emphasize difference between hass.io and Home Assistant for new users.

The suggested text emphasizes hass.io as the standard way of installing Home Assistant, but makes the alternatives more prominent.

## Checklist:

### DIFF
--- a/source/getting-started/index.markdown
+++ b/source/getting-started/index.markdown
@@ -9,9 +9,14 @@ sharing: true
 footer: true
 ---
 
-The goal of this getting started guide is to install Hass.io on a Raspberry Pi 3. Hass.io is our own all in one solution that turns your Raspberry PI into the ultimate home automation hub.
+Home Assistant is a Python program and can be installed and run as any other Python program.  However, configuring a whole system can be intimidating for the novice, so there are a number of ways to do most of the hard work of installing and configuring the Home Assistant.
 
 Follow this guide if you want to easily get started with Home Assistant, or if you have no or little Linux experience. For advanced users, check our [alternative installation methods](/docs/installation/).
+
+## {% linkable_title HASS.IO %}
+HASS.IO is the easiest way of installing Home Assistant.  It is our own all in one solution that turns your Raspberry PI into the ultimate home automation hub, but can also be run on other operating systems.
+
+The goal of this getting started guide is to install Hass.io on a Raspberry Pi 3. 
 
 ### {% linkable_title Hardware requirements %}
 
--- a/source/getting-started/index.markdown
+++ b/source/getting-started/index.markdown
@@ -11,7 +11,19 @@ footer: true
 
 Home Assistant is a Python program and can be installed and run as any other Python program.  However, configuring a whole system can be intimidating for the novice, so there are a number of ways to do most of the hard work of installing and configuring the Home Assistant.
 
-Follow this guide if you want to easily get started with Home Assistant, or if you have no or little Linux experience. For advanced users, check our [alternative installation methods](/docs/installation/).
+## {% linkable_title Which Installation Method Should I Choose? %}
+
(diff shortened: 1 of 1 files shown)


frenck:
Thank you for the change, nevertheless, this part has been recently re-written to remove the "old" methods.

@balloob Rewrote it, so maybe I leave it up to him to decide.

balloob:
I don't want to merge this. The goal of the getting started guide is not to show all the different options, the goal of a getting started guide is to literally get people started. Our target audience is no longer the people that know programming, Python and are comfortable with a shell. Our target audience is everyone. And those people will be very confused with all the different options. They need a simple guide to follow.

The old guide tried to do it all and by doing so accomplished nothing. Beginning users were researching Docker, Arch and Vagrant to run their Home Assistant, thought it was too complicated and gave up. Never even gave any option a try.

The second paragraph tells people that there are other options and I think that this is enough.

flamingm0e:
>The second paragraph tells people that there are other options and I think that this is enough.

But it ISN'T enough.  If you look at the forums, reddit, and discord, there is an astounding number of people that do not know the difference, and ask about the 'hassio tab in their hassbian install', or 'add-ons' on their system that has been running for a year.  We see the same repetitive questions over and over regarding the confusion between them.

mf-social:
Agreed, the problem here is that to newcomers  there's no distinction between hassio and homeassistant and they believe that it's just two different words for the same thing (like sofa and couch).

The naming convention of hassio didn't help but we're stuck with that now.

There needs to be a simple explanation of what homeassistant is, what hassio is, and why they're different.

There must be 20 - 30 threads a week where newcomers are asking for a hassio tab on their manual install or some other question that should have been ironed out before they even hit download (or opened a terminal).

These 20 - 30 people are the exact non-techie people you're trying to appeal to.

This PR would've helped them in the future.

balloob:
If the problem is that we don't clearly explain how Hass.io and Home Assistant relate, then we should solve that. That's not solved by showering newcomers with all these different options that they should not care about and only confuse.

mynkow:
This is insane. I am trying to get my head around this for the last 24h and I still have no idea what is the difference between HomeAssistant(HA) and hass.io.

Here is the chronology of my experience. 

1. I want to try HA. OK, go and download it.
2. Ohh, there is a docker container. I have installed it. It is working. Now what?
3. I have watched the first youtube video and... hmmm... this guy is installing addons from a tab which I am missing. Why is that?
4. You noob, you have installed a legacy version of HA. There is a newer version with that tab ;)
5. From that point I have started questioning myself:
 - Can I run and use HA without going through the hass.io route?
 - People are stating that HA and hass.io are two different things. But after all you end up with the same think which is HA, don'y you?
 - Do I need addons? Can I install the addons manually if I have HA? I guess not because addons are controlled by a docker supervisor. Then why should I want to use HA vs hass.io. Just make a table which compares the two things side by side.

I have literally spent 24 hours and it is frustrating and time consuming.

One more thing. @balloob mentioned that the getting started must throw the people directly to a working thing. However, I think that most people want to try HA before buying a rasPI and the docs for those people are limited.

Hope this helps. I am going to read more to understand what is the diff between HA and hass.io... Wish me luck!

ZenithDK:
I have to agree with the other comments here.
I've run Gentoo and Debian and Ubuntu for years on end and I know how to administer a system, but this whole hass.io thing has me confused as well.
This FAQ entry is only somewhat helpful, I think a table explaining the differences and what "add-ons" even are would be very helpful also: https://www.home-assistant.io/faq/#home-assistant-vs-hassio

PR #4614, opened 2018-02-10
Clarify distribution of random values

robmarkcole:
Clarify that values drawn produce a discrete uniform distribution, and not for example a normal (or bell shaped) distribution. I intend to implement a random physics sensor which will return a bell distribution and this edit will avoid confusion.

### DIFF
--- a/source/_components/sensor.random.markdown
+++ b/source/_components/sensor.random.markdown
@@ -14,7 +14,7 @@ ha_release: 0.32
 ---
 
 
-The `random` sensor platform is creating random sensor values (integers) out of a given range. This can be useful if you want to test automation rules. It generates a new value every time it is polled.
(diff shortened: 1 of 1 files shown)


arsaboo:
I like the clarification, but I am wondering if it will confuse readers. How many users may be aware of `uniform distribution` :)

I am thinking that most users have a uniform distribution in mind when they say `random`. May be we can keep this as is and add the distribution details to the `random_normal` that you want to create in the future. 

I will wait for someone else to weigh-in before deciding the way forward.

robmarkcole:
Perhaps I could reword it? It wasn’t obvious to me to expect a uniform distribution, and I read somewhere Explicit is better than implicit :)

arsaboo:
To be honest, what you have currently is perfectly fine and we can merge as-is. I am just looking for another set of eyes here :)

PR #4635, opened 2018-02-11
Update component_loading.markdown

jonasius:
Maybe there should be an info where the components can be found if Home Assistant is installed in an venv.

**Description:**

## Checklist:

### DIFF
--- a/source/developers/component_loading.markdown
+++ b/source/developers/component_loading.markdown
@@ -14,8 +14,8 @@ A component will be loaded on start if a section (ie. `light:`) for it exists in
  * `<config directory>/custom_components/<component name>`
  * `homeassistant/components/<component name>` (built-in components)
     
-    If Home Assistant is installed in an **venv**:
-    * `homeassistant/lib/python3.5/site-packages/homeassistant/components/<component name>`
+    If Home Assistant is installed in an **venv** you find the components folder under the following subfolder:
+    * `lib/python3.5/site-packages/homeassistant/components/<component name>`
 
 Once loaded, a component will only be setup if all dependencies can be loaded and are able to setup. Keep an eye on the logs to see if your component could be loaded and initialized.
 
--- a/source/developers/component_loading.markdown
+++ b/source/developers/component_loading.markdown
@@ -11,17 +11,13 @@ footer: true
 
 A component will be loaded on start if a section (ie. `light:`) for it exists in the `configuration.yaml` file. A component can also be loaded if another component is loaded that depends on it. When loading a component Home Assistant will check the following paths:
 
- * `<config directory>/custom_components/<component name>`
+ * `.homeassistant/custom_components/<component name>`
  * `homeassistant/components/<component name>` (built-in components)
     
 If Home Assistant is installed in an [`venv`](/docs/installation/virtualenv/) you find the components folder under the following subfolder: `lib/python3.5/site-packages/homeassistant/components/<component name>`
 
 Once loaded, a component will only be setup if all dependencies can be loaded and are able to setup. Keep an eye on the logs to see if your component could be loaded and initialized.
 
-<p class='note warning'>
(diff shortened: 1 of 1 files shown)


source/developers/component_loading.markdown
```diff
@@ -13,6 +13,9 @@ A component will be loaded on start if a section (ie. `light:`) for it exists in
 
  * `<config directory>/custom_components/<component name>`
  * `homeassistant/components/<component name>` (built-in components)
+    
+    If Home Assistant is installed in an **venv**:
+    * `homeassistant/lib/python3.5/site-packages/homeassistant/components/<component name>`
```
fabaff: That path is mentioned already on line 15. Also, the path for a `venv` will most likely be completely different than the one you provided..

jonasius: I installed Home Assistant according to the introductions, so the full path is: `/srv/homeassistant/lib/python3.5/site-packages/homeassistant/components/<component name>` why do you think the path will be completely different?

Maybe on the component loading page should be an notice that the homeassistant base folder is inside the `venv`? Or is that too obvious?

fabaff: Because the [`venv` guide](https://home-assistant.io/docs/installation/virtualenv/) doesn't suggest a location. The Raspbian guide uses `/srv/homeassistant`. I use `/opt/homeassistant` and `~/home/repos/home-assistant` that one because I don't have to deal with permissions. 

> Or is that too obvious?

It seems not. Otherwise you wouldn't have to suggest an adjustment. Let's add a note that about a possible location.

frenck (dismissed):
@jonasius Please add the note as suggested by @fabaff . Thanks already! 👍

jonasius:
What do you think about this path and description. The main problem for me was that the folders are located in lib/python3.5/site-packages/...

fabaff (dismissed):

source/developers/component_loading.markdown
```diff
@@ -9,17 +9,15 @@ sharing: true
...
 ---
 
-A component will be loaded on start if a section (ie. `light:`) for it exists in the config file. A component can also be loaded if another component is loaded that depends on it. When loading a component Home Assistant will check the following paths:
+A component will be loaded on start if a section (ie. `light:`) for it exists in the `configuration.yaml` file. A component can also be loaded if another component is loaded that depends on it. When loading a component Home Assistant will check the following paths:
 
- * `<config directory>/custom_components/<component name>`
+ * `.homeassistant/custom_components/<component name>`
  * `homeassistant/components/<component name>` (built-in components)
+    
+If Home Assistant is installed in an [`venv`](/docs/installation/virtualenv/) you find the components folder under the following subfolder: `lib/python3.5/site-packages/homeassistant/components/<component name>`
```
DubhAd: Worth making the Python version a place holder, since some people will be running 3.6, and 3.7 will be out soon too.

jonasius: Like `3.x`?

DubhAd: That and something like `replace 3.x with your Python version` just in case people aren't paying attention, and go looking for a directory literally called 3.x

source/developers/component_loading.markdown
```diff
@@ -9,17 +9,15 @@ sharing: true
 footer: true
 ---
 
-A component will be loaded on start if a section (ie. `light:`) for it exists in the config file. A component can also be loaded if another component is loaded that depends on it. When loading a component Home Assistant will check the following paths:
+A component will be loaded on start if a section (ie. `light:`) for it exists in the `configuration.yaml` file. A component can also be loaded if another component is loaded that depends on it. When loading a component Home Assistant will check the following paths:
 
- * `<config directory>/custom_components/<component name>`
+ * `.homeassistant/custom_components/<component name>`
```
frenck: I do not agree on this change, since it is not Hass.io compatible.
The previous version works better.

frenck:
@jonasius Any plans on moving this forward?

frenck:
Closing this PR, the developers' documentation has been moved to:

<https://github.com/home-assistant/developers.home-assistant>

PR #4652, opened 2018-02-12
Provide an example of how to make a 'when dark' condition by ORing su…

Human:
…nrise and sunset conditions.

See issue#12308 at https://github.com/home-assistant/home-assistant/issues/12308

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/scripts/conditions.markdown
+++ b/source/_docs/scripts/conditions.markdown
@@ -114,6 +114,16 @@ condition:
   after_offset: "-1:00:00"
 ```
 
+```yaml
+condition:
+    condition: or  # 'when dark' condition: either after sunset or before sunrise
+    conditions:
+      - condition: sun
+        after: sunset
+      - condition: sun
+        before: sunrise
+```
+
 Here is a truth table to clarify the parameters with and without offset:
 
 | command                            |        night | at sunrise  | daytime | at sunset  |


Human:
Resolves issue #12308

Human:
@mf-social Would you like to handle the review?

ryanwinter:
The truth table is still misleading, or just downright wrong.

arsaboo:
@ryanwinter Create a separate PR for the truth table please..thanks!

ryanwinter:
I'm wondering if the current code makes sense. It uses midnight as an additional state change point, but this isn't obvious, or something anyone would need.

I actually think the current table reflects what it should do. Let me think of a code change to propose instead.

mf-social:
^ I think that's intentional, otherwise it will always be true.

If you make it go past midnight, at any given daylight time it is after sunrise today and before sunrise tomorrow, at any given darkness time it is after sunset today and before sunset tomorrow.

There has to be a cutoff and midnight makes perfect sense IMO.

ryanwinter:
I think that the cutoff should be the next sun event which will trigger the end state change, rather than an arbitrary time which was likely picked as midnight to simplify logic. I mean why not 3am, midday, 8:57pm?

mf-social:
Because it is after sunrise / sunset on the *day* in question, a day runs from midnight to midnight, not 3 - 3 or any other time. 🤔

ryanwinter:
Yes, but a night runs from sunset to sunrise, and we are talking about sun related conditions.

mf-social:
Yes, and sun related conditions do not exclusively cater for when it is dark. They currently cater for whether the sun is up or down on that particular **day**. Day as in Monday, Wednesday etc, not 'the time it was (or will be)  light' 

Like I say, it works well for me in that format, but if you think it's better another way then that's cool. I was just thinking it will be a massive breaking change for people's long-standing automations so you might want to garner some more opinions before spending time rewriting the logic.

ryanwinter:
Well I agree this could be a breaking change, people (including myself) have either written logic to work around this weird effect, or moved to sun elevation instead to bypass the confusing logic which contradicts the documentation.

Let me look at a coding solution first and see if this makes sense.

PR #4537, opened 2018-01-28
Documentation of new HomeMatic lock section.

PH89:
**Description:**
Adds the documentation for the new HomeMatic lock section

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#11980

## Checklist:

- [x] Branch: Fixes, changes and adjustments should be created against `current`. New documentation for platforms/components and features should go to `next`.
- [x] The documentation follow the [standards][standards].

[standards]: https://home-assistant.io/developers/documentation/standards/

PR #4663, opened 2018-02-13
Create lock.homematic.markdown

krys1976:
According to new homematic lock device behaviour. KeyMatic is being used as a lock device as well.


c727:
duplicate of #4537

PR #4690, opened 2018-02-17
Added how to find the web UI.

DubhAd:
As the getting started guide is now totally Hass.io focussed, there's no longer anywhere to tell people how to reach the web UI.

### DIFF
--- a/source/_docs/configuration.markdown
+++ b/source/_docs/configuration.markdown
@@ -11,6 +11,8 @@ footer: true
 
 When launched for the first time, Home Assistant will write a default configuration file enabling the web interface and device discovery. It can take up to a minute for your devices to be discovered and appear in the user interface.
 
+The web interface can be found at `http://ip.ad.dre.ss:8123/` - for example if your Home Assistant system has the IP address `192.168.0.40` then you'll find the web interface as `http://1921.68.0.40:8123/`.
+
 The location of the folder differs between operating systems:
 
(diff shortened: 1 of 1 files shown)


c727:
- we should define a URL scheme that we use everywhere, eg in getting started we use `http://hassio.local:8123`
- 2nd: do we need this explained on every subpage again and again? :stuck_out_tongue_closed_eyes:

DubhAd:
That URL scheme works great for Hass.io, and not at all for everything else ;)

As a result, yes, for all the install guides and relevant jumping off points we _should_ be telling people how to get to the UI. After all, we can't assume that people will read the Getting Started guide, particularly now it's Hass.io focussed and many who don't use it skip it.

source/_docs/configuration.markdown
```diff
@@ -11,6 +11,8 @@ footer: true
 
 When launched for the first time, Home Assistant will write a default configuration file enabling the web interface and device discovery. It can take up to a minute for your devices to be discovered and appear in the user interface.
 
+The web interface can be found at `http://ip.ad.dre.ss:8123/` - for example if your Home Assistant system has the IP address `192.168.0.40` then you'll find the web interface as `http://1921.68.0.40:8123/`.
```
arsaboo: Typo `http://1921.68.0.40:8123/` -> `http://192.168.0.40:8123/`

PR #4691, opened 2018-02-17
Note about rename node being broken

DubhAd:
Given https://github.com/home-assistant/home-assistant/issues/12430, adding a note that Rename Node is now broken, and pointing to the blog post on how to rename nodes.

Yes, I'm frustrated ;)


DubhAd:
@balloob Can we get this approved, or feedback why not, please. This is still coming up multiple times a day and still causing chaos with the user base.

PR #4706, opened 2018-02-19
Update addon_communication.markdown

danielwelch:
**Description:**

When developing my first add-on that relied on `hassio.addon_stdin`, it was pretty unclear how to identify my add-on via a service call. After looking at the source, I understand that the github repo url is hashed and think this is a great solution, but do with it was documented a little better. I tried to add some explanation of what these variables actually are. `{SLUG}` was pretty obvious, but `{REPO}` required me to search around in the source code.

At the very least, the typo `{REPO}-{SLUG}` and following examples should be corrected to {REPO}_{SLUG} ([corresponding source](https://github.com/home-assistant/hassio/blob/587047f9d648b8491dc8eef17dc6777f81938bfd/hassio/addons/data.py#L110). 

## Checklist:

- [X] Branch: Fixes, changes and adjustments should be created against `current`. New documentation for platforms/components and features should go to `next`.


c727:
thanks for your help on this, I added some changes :+1:

PR #4718, opened 2018-02-20
Added 1ch MiLight dimmer support.

rubenverhoef:
**Description:**
Added documentation for a 1ch MiLight dimmer.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#12558

## Checklist:

### DIFF
--- a/source/_components/light.limitlessled.markdown
+++ b/source/_components/light.limitlessled.markdown
@@ -18,7 +18,7 @@ ha_release: pre 0.7
 
 ### {% linkable_title Setup %}
 
-Before configuring Home Assistant, make sure you can control your bulbs or LEDs with the MiLight mobile application. Discover your bridge(s) IP address. You can do this via your router or a mobile application like Fing ([android](https://play.google.com/store/apps/details?id=com.overlook.android.fing&hl=en) or [iTunes](https://itunes.apple.com/us/app/fing-network-scanner/id430921107?mt=8)). Keep in mind that LimitlessLED bulbs are controlled via groups. You can not control an individual bulb via the bridge, unless it is in a group by itself. Note that you can assign an `rgbw`, `rgbww` and `white` group to the same group number, effectively allowing 12 groups (4 `rgbww`, 4 `rgbw` and 4 `white`) per bridge.
+Before configuring Home Assistant, make sure you can control your bulbs or LEDs with the MiLight mobile application. Discover your bridge(s) IP address. You can do this via your router or a mobile application like Fing ([android](https://play.google.com/store/apps/details?id=com.overlook.android.fing&hl=en) or [iTunes](https://itunes.apple.com/us/app/fing-network-scanner/id430921107?mt=8)). Keep in mind that LimitlessLED bulbs are controlled via groups. You can not control an individual bulb via the bridge, unless it is in a group by itself. Note that you can assign an `rgbw`, `rgbww`, `white` and `dimmer` group to the same group number, effectively allowing 12 groups (4 `rgbww`, 4 `rgbw`, 4 `white` and 4 `dimmer`) per bridge.
 
 To add `limitlessled` to your installation, add the following to your `configuration.yaml` file:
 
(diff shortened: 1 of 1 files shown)


source/_components/light.limitlessled.markdown
```diff
@@ -18,7 +18,7 @@ ha_release: pre 0.7
...
+Before configuring Home Assistant, make sure you can control your bulbs or LEDs with the MiLight mobile application. Discover your bridge(s) IP address. You can do this via your router or a mobile application like Fing ([android](https://play.google.com/store/apps/details?id=com.overlook.android.fing&hl=en) or [iTunes](https://itunes.apple.com/us/app/fing-network-scanner/id430921107?mt=8)). Keep in mind that LimitlessLED bulbs are controlled via groups. You can not control an individual bulb via the bridge, unless it is in a group by itself. Note that you can assign an `rgbw`, `rgbww`, `white` and `dimmer` group to the same group number, effectively allowing 12 groups (4 `rgbww`, 4 `rgbw`, 4 `white` and 4 `dimmer`) per bridge.
```
amelchio: That must add up to 16 groups then? It's a nitpick but some user is bound to be confused by that, so please fix it :)

source/_components/light.limitlessled.markdown
```diff
@@ -74,6 +77,9 @@ Refer to the [light]({{site_root}}/components/light/) documentation for general
   - When using a legacy WiFi bridge (before v6), you can observe on the MiLight mobile application, you can not select a specific brightness or temperature - you can only step each property up or down. There is no indication of which step you are on. This restriction, combined with the unreliable nature of LimitlessLED transmissions, means that setting white bulb properties is done on a best-effort basis. The only very reliable settings are the minimum and maximum of each property.
   - *Temperature*: Wifi bridge v6 supports 101 temperature steps; older versions only 10.
   - *Brightness*: Wifi bridge v6 supports 101 brightness steps; older versions only 10.
+- **Dimmer**
+    - This type is for a single color LED dimmer like the 1CH MiLight dimmer module or similar. Version 6 Wifi bridges are supported and not sure if legacy Wifi bridges(before v6) are working with this type.
```
amelchio: Please remove " and not sure if legacy Wifi bridges(before v6) are working with this type" since it won't work in HA at least (no legacy command set was added).

source/_components/light.limitlessled.markdown
rubenverhoef: I don't know exactly, is it possible to have 16 groups? I think so, because the remote_style is different for this dimmer?
I just have one milight dimmer which i want to control with home-assistant. :)

source/_components/light.limitlessled.markdown
rubenverhoef: So only version 6 of the bridge is supported, and support for the older versions is dropped? Because the rest of the documentation is talking about the legacy wifi bridges... I only have an V6 bridge with one LED controller.

amelchio: The old bridges only support lights that existed at that time, AFAIK.

This is similar to RGBWW which has a note "(Only supported on v6 bridges)" so actually you could remove the entire last sentence and add that note in a similar way, for consistency.

happyleavesaoc: Broadly speaking, `python-limitlessled` and therefore HASS support legacy (v5 and earlier) and current (v6) bridges. But support for the 1CH dimmer is only available for v6 bridges (since your PR added a V6 command set, not a Legacy command set). v5 doesn't have dimmers anyway, I don't think.

source/_components/light.limitlessled.markdown
happyleavesaoc: Yes, he means just update the part where it says "12 groups" to read "16 groups".

rubenverhoef: Great, :) i will do that

source/_components/light.limitlessled.markdown
rubenverhoef: Okay, thanks for the fast explanation. I will edit this tomorrow and update the PR.

PR #4725, opened 2018-02-21
Filter Sensor

dgomes:
**Description:**

Documentation for the filter component 

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#12650

## Checklist:


source/_components/sensor.filter.markdown
```diff
@@ -0,0 +1,102 @@
...
+comments: false
+sharing: true
+footer: true
+ha_category: Sensor
+ha_release: 0.12
+ha_iot_class: "Local Push"
+logo: home-assistant.png
+---
+
+The `filter` component supports sensors which process the **numeric states** from other entities. `filter` will apply a signal processing algorithm to a serie of states and produce a `new state` given the choosen algorithm.
```
c727: series
produces/ generates/ calculates

source/_components/sensor.filter.markdown
```diff
@@ -0,0 +1,102 @@
...
+    name: filtered relative humidity 
+    filter: lowpass
+    options:
+      time_constant: 10
+  - platform: filter
+    entity_id: sensor.temperature
+    name: filtered temperature
+    filter: outlier
+    options:
+      constant: 0.10
```
c727: is this time_constant? it's not listed again in the details below

source/_components/sensor.filter.markdown
dgomes: How about:

```
`filter` will apply a signal processing algorithm to a series of states and generate a `new state` given the chosen algorithm.
```

I changed to *series* (plural), but each filter works with a single serie (singular). I'm no english speaking native, but it made sense to me. Tell me how to proceed.

source/_components/sensor.filter.markdown
dgomes: no... it's **radius**... will correct

PR #4727, opened 2018-02-21
Consistent documentation for user_id / chat_id

alexbarcelo:
Using a more consistent usage regarding the interaction between `target`, `user_id` and `chat_id`.

The `target` can be both a `user_id` or a `target_id`, leaving the `user_id` (which was the original one). But the `chat_id` parameters should be a `chat_id` identificator, for consistency.

Also updated the description for `target` parameter on Telegram calls.


DubhAd (approved):
👍

PR #4757, opened 2018-02-25
Update sensor.mqtt.markdown to add initial slashes for ESPEasy

ax42:
The sensors topic required the initial slashes to function in my installation.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/sensor.mqtt.markdown
+++ b/source/_components/sensor.mqtt.markdown
@@ -167,14 +167,19 @@ sensor:
 
 ### {% linkable_title Get sensor value from a device with ESPEasy %}
 
-Assuming that you have flashed your ESP8266 unit with [ESPEasy](https://github.com/letscontrolit/ESPEasy). Under "Config" set a name ("Unit Name:") for your device (here it's "bathroom"). A "Controller" for MQTT with the protocol "OpenHAB MQTT" is present and the entries ("Controller Subscribe:" and "Controller Publish:") are adjusted to match your needs. In this example the topics are prefixed with "home". Also, add a sensor in the "Devices" tap with the name "analog" and "brightness" as value. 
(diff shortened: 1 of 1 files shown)


source/_components/sensor.mqtt.markdown
```diff
@@ -142,9 +142,9 @@ Assuming that you have flashed your ESP8266 unit with [ESPEasy](https://github.c
 As soon as the unit is online, you will get the state of the sensor.
 
 ```bash
-home/bathroom/status Connected
+/home/bathroom/status Connected
```
fabaff: For me, this looks like that it's depend on the ESPEasy configuration.

ax42: Possible (I'm travelling now, can't test it), however the default ESPEasy config seems to include the slash (as do most MQTT examples I've seen), and having the slash in the example would have saved me ~2 hours of debugging.

fabaff: There was no ESPEasy default configuration for Home Assistant at the time when the example was added. A `/` as prefix will increase the number of topic level with no real benefit. Also, the description says that `home` is used and not `/home`.

ax42: Default publish channel for ESPEasy now is `/%sysname%/%tskname%/%valname%`.  In addition, the HA docs say 

> and the entries (“Controller Subscribe:” and “Controller Publish:”) are adjusted to match your needs.

The "adjusted to match your needs" by default gives a topic with a slash (as shown above), in which case the configuration example in the docs will fail to work.  I found this particularly difficult to debug as I needed to first install another app to read MQTT messages.  

I'm done here.  I feel I have made the point that the documentation could be improved, especially for people like myself who knows their way around tech however don't know MQTT.  Feel free to reject the PR or accept it -- I don't feel this discussion is helpful any more as it feels to me you are only looking for reasons to reject.  Your point about "There was no default config when the example was added" is an example of this, as there clearly is now a default ESPEasy config, and (to my mind) the documentation should be updated to stay in line.

frenck: @fabaff Could you please take a look at the comment from @ax42?

frenck:
Rebased the PR in order to fix the build.

source/_components/sensor.mqtt.markdown
fabaff: The change by @ax42 is still incomplete.

I will update it and make it clear how the topic should look like.

fabaff (approved):

PR #4773, opened 2018-02-27
HA version 0.64.0 voluptuous config check.

JerryWorkman:
**Description:**

## Checklist:

### DIFF
--- a/source/developers/development_validation.markdown
+++ b/source/developers/development_validation.markdown
@@ -28,15 +28,7 @@ Some things to keep in mind:
 - Use the constants defined in `const.py`
 - Import `PLATFORM_SCHEMA` from the parent component and extend it
 - Preferred order is `required` first and `optional` second
-- Starting with Home Assistant version 0.64.0 voluptuous now requires default values for config keys to be valid values. For instance:
-```
-vol.Required(CONF_SWITCH, default=None): cv.string,
-```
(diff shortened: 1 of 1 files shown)


source/developers/development_validation.markdown
```diff
@@ -28,6 +28,15 @@ Some things to keep in mind:
 - Use the constants defined in `const.py`
 - Import `PLATFORM_SCHEMA` from the parent component and extend it
 - Preferred order is `required` first and `optional` second
+- Starting with Home Assistant version 0.64.0 voluptuous now requires default values for config keys to be valid values. For instance:
+```
+vol.Required(CONF_SWITCH, default=None): cv.string,
```
fabaff: If a value is required then there shouldn't be a default. We want to fail if the value is missing.

fabaff (changes requested):
According to the origin PR, it's Custom components only.

source/developers/development_validation.markdown
JerryWorkman: Good point, but it was allowed with the previous version of Voluptuous. So the comment should be to remove the default parameter for vol.required but also insure that vol.optional has valid values.

frenck:
Rebased this PR in order to fix the build.

fabaff (approved):

PR #4800, opened 2018-03-02
Add page on whitelist_external_dirs

robmarkcole:
**Description:**
I feel whitelist_external_dirs badly needs its own page as this comes up on the forums so often. Is it worth commenting on permissions, since I know there are sometimes issues that directories need the permissions modified? (I've not direct experience of that however)

## Checklist:

### DIFF
--- a/source/_docs/configuration/whitelist.markdown
+++ b/source/_docs/configuration/whitelist.markdown
@@ -10,7 +10,7 @@ footer: true
 redirect_from: /getting-started/basic/#accessing-directories
 ---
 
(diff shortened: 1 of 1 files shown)


fabaff:
> Is it worth commenting on permissions

No. Users shouldn't start to modify permissions of folder because it could lead to serious information leakage if they do it wrong and just want to go the easy way.

> I think this should go under the Advanced Configuration category?

No need to go there. We have it already under "Setup basic information". If it's not enough then we  should make it more clear there.

PR #4802, opened 2018-03-02
Adds www folder doc

robmarkcole:
**Description:**
The use of the www folder is not documented anywhere from what I can tell. 

As a separate issue it is a bit confusing that you create a ```www``` folder but then access the ```local``` path.

## Checklist:

### DIFF
--- a/source/_docs/configuration/www_folder.markdown
+++ b/source/_docs/configuration/www_folder.markdown
@@ -1,0 +1,13 @@
+---
+layout: page
+title: "Sharing files"
+description: "Share files via the www folder."
+date: 2018-03-02 12:50
+sidebar: true
+comments: false
+sharing: true
+footer: true
+redirect_from: /getting-started/basic/#www-folder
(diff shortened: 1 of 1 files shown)


c727:
use 1x ` instead of 3x

Is this linked anywhere?

skalavala:
Thanks for taking the initiative to update the docs. I don't think the statement "Note that `www` or one of its parent directories must be added to the whitelist_external_dirs." is true. Once you add `image.jpg` to the `www` folder, it can be accessed without having to add `whitelist_external_dirs:` option to the `configuration.yaml`. Can you please double check? I maybe wrong if something has changed changed. But it would be good to double check.

robmarkcole:
@skalavala you are correct, I've just double checked and `www` is safe.
@c727 I've checked [the docs](https://home-assistant.io/developers/documentation/create_page/) but theres nothing on how to link the article, will keep searching

PR #4892, opened 2018-03-11
Provided note about python versions

DubhAd:
Synology Package Center provides only 3.5.1, nothing newer :(


fabaff (approved):
Thanks :bird:

PR #4900, opened 2018-03-11
Added version note

DubhAd:
Since Synology offer an outdated Python, providing a note about how to install 0.64.x


fabaff (approved):
Thanks :bird:

PR #4918, opened 2018-03-14
Adds rel nofollow to external links

### LINKED ISSUE #4583: Add rel="nofollow" to external links
Too many outbound links will hurt our ranking. By adding this we also make sure that we don't "endorse" the other website for the search ranking algo's.

frenck:
**Description:**

Adds `rel="external nofollow"` to all external links on the website.

Fixes #4583 

## Checklist:

### DIFF
--- a/plugins/no_follow.rb
+++ b/plugins/no_follow.rb
@@ -1,0 +1,32 @@
+# Jekyll Auto Nofollow Plugin
+# Automatically adds rel='external nofollow' to outgoing links.
+
+require 'jekyll'
+require 'nokogiri'
+
+module Jekyll
+    module NoFollow
+        def nofollow(content)
+            dom = Nokogiri::HTML.fragment(content)
+
+            # Find all links
+            dom.css('a').each do |link|
+
+                # All external links start with 'http', skip when this one does not
+                next unless link.get_attribute('href') =~ /\Ahttp/i
+
+                # Play nice with links that already have a rel attribute set, skip it
+                next if link.get_attribute('rel')
+
+                # Play nice with our own links
+                next if link.get_attribute('href') =~ /\Ahttps?:\/\/\w*.?home-assistant.io/i
+
+                # Add rel attribute to link
+                link.set_attribute('rel', 'external nofollow')
+            end
+            dom.to_s
+        end
+    end
+end
+
+Liquid::Template.register_filter(Jekyll::NoFollow)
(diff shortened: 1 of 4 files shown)


plugins/no_follow.rb
```diff
@@ -0,0 +1,32 @@
...
+module Jekyll
+    module NoFollow
+        def nofollow(content)
+            dom = Nokogiri::HTML.fragment(content)
+
+            # Find all links
+            dom.css('a').each do |link|
+
+                # All external links start with 'http', skip when this one does not
+                next unless link.get_attribute('href') =~ /\Ahttp/i
```
balloob: This assumption is not correct. For the release notes I use absolute urls because I copy paste same notes between GitHub and our site.

plugins/no_follow.rb
```diff
@@ -0,0 +1,32 @@
...
+
+                # Play nice with links that already have a rel attribute set, skip it
+                next if link.get_attribute('rel')
```
balloob: I don't know about this, that means one can cheat our system

plugins/no_follow.rb
frenck: Ok, will try to work around that one...

balloob: nevermind, just saw the home assistant check a few lines below

plugins/no_follow.rb
frenck: Another take: It allows for exceptions and keeps the documented Jekyll behavior as is.
(Lets put everybody in prison to prevent crimes... 😉)

plugins/no_follow.rb
frenck: Still not an issue right?

This rule is about skipping relative links...

plugins/no_follow.rb
balloob: Can we just append ' external nofollow` if it exists?

frenck: Sounds like a plan

frenck: Done.

frenck:
Latest change seems to add a lot of build time, need to investigate a little more. Adding WIP status.

balloob:
I'm not surprised if it takes longer, it has to parse all HTML -> DOM and back

frenck:
It actually wasn't that bad... ~500 sec for the full site generation (including the dom parsing and the installation/compilation of Nokogiri). After the last change, it seems impacted a lot.

balloob (approved):

PR #4922, opened 2018-03-15
Fix default values in REST switch

drndos:
According to https://github.com/home-assistant/home-assistant/blob/master/homeassistant/components/switch/rest.py
* Resource does not have any default value.
* Method is set in lowercase "post"
* Default name is not "REST Binary Switch" but "REST Switch"
* Default body_on is "ON" not true
* Default body_off is "OFF" not false

**Description:**

## Checklist:

- [/] Branch: Fixes, changes and adjustments should be created against `current`. New documentation for platforms/components and features should go to `next`.
- [/] The documentation follow the [standards][standards].

### DIFF
--- a/source/_components/switch.rest.markdown
+++ b/source/_components/switch.rest.markdown
@@ -1,7 +1,7 @@
 ---
 layout: page
 title: "RESTful Switch"
-description: "Instructions how to integrate REST switches into Home Assistant."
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @drndos! Approved 🥇 

PR #4951, opened 2018-03-18
Tweaked automations

DubhAd:
Added comment about why we don't use `-` before the first `service:` line.

Removed the `-` from the last example's service_template


frenck (approved):

frenck:
🎖

PR #4957, opened 2018-03-19
Updated with new path.

CCOSTAN:
**Description:**

## Checklist:

### DIFF
--- a/source/_components/feedreader.markdown
+++ b/source/_components/feedreader.markdown
@@ -67,4 +67,4 @@ hass.bus.listen(EVENT_FEEDREADER, event_listener)
 
 To get started developing custom components, please refer to the [developers](/developers) documentation
 
-For a drop in packaged complete example of Feedreader, you can use the [PodCast notifier](https://github.com/CCOSTAN/Home-AssistantConfig/blob/master/packages/hasspodcast.yaml).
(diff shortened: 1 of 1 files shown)


NovapaX (commented):
The examples present in in the documentation cover about the exact same topic as the linked package. Besides: that package link to custom-ui resources, so i'ts not really 'drop in'. And it's not exactly a great example (lots of commented config)

I think we should remove this as it could lead to more user confusion (and broken links in our documentation)
If needed the inline examples could be extended with another notification example, but currently they communicate the possibilities pretty well.

P.S. I just visited your GitHub page, all I can say is: WOW! 🥇

frenck (approved):

frenck:
For now, it is a fix, since the current is invalid.
Therefore, I'm accepting this PR.

@NovapaX Feel free to propose a change in a separate ticket or PR.

CCOSTAN:
@NovapaX - Thanks for the comments. Glad you liked the Repo.. I try hard to keep it useful.  I made some changes to my package based on your suggestions.  I removed all the extra commented lines from the bottom (I forgot they were there) and rearranged the `service` actions to put the local resource items to the bottom.  I also added a single `remove everything below this line` comment in the file for users who want to use it as a true 'drop in ' `package`.  Thanks for the suggestions.  Definitely made the result better!

Hopefully we can keep the links since i do believe my repo and the podcast are great resources for new users and little things that we can do to expose them is beneficial to all.

PR #4966, opened 2018-03-20
Add documentation for homekit controller component

mjg59:
**Description:**
Documentation for Homekit accessories.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) :** home-assistant/home-assistant#13346

## Checklist:

### DIFF
--- a/source/_components/homekit_controller.markdown
+++ b/source/_components/homekit_controller.markdown
@@ -1,0 +1,28 @@
+---
+layout: page
+title: "Homekit controller support"
+description: "Instructions how to integrate your Homekit devices within Home Assistant."
+date: 2018-03-19 21:04
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: apple-homekit.png
+ha_category: Hub
+ha_release: 0.66
+ha_iot_class: "Local Polling"
+---
+
+[HomeKit](https://developer.apple.com/homekit/) controller integration for Home Assistant allows you to connect HomeKit accessories to Home Assistant. This component should not be confused with the [HomeKit](homekit) component, which allows you to control Home Assistant devices via HomeKit.
+
+There is currently support for the following device types within Home Assistant:
+
+- [Light](../light.homekit_controller)
+
+The component will be automatically configured if the [`discovery:`](components/discovery/) component is enabled. For each detected Homekit accessory, a configuration prompt will appear in the web front end. Use this to provide the HomeKit PIN. It can be disabled by adding the following configuration:
+
+```yaml
+discovery:
+  ignore:
+    - homekit
+```
--- a/source/_components/homekit_controller.markdown
+++ b/source/_components/homekit_controller.markdown
@@ -20,7 +20,7 @@ There is currently support for the following device types within Home Assistant:
 - [Light](../light.homekit_controller)
 - [Switch](../switch.homekit_controller)
 
-The component will be automatically configured if the [`discovery:`](components/discovery/) component is enabled and an enable entry added for homekit:
+The component will be automatically configured if the [`discovery:`](components/discovery/) component is enabled and an enable entry added for HomeKit:
 
 ```yaml
 discovery:
--- a/source/_components/homekit_controller.markdown
+++ b/source/_components/homekit_controller.markdown
@@ -1,7 +1,7 @@
 ---
 layout: page
-title: "Homekit controller support"
-description: "Instructions how to integrate your Homekit devices within Home Assistant."
+title: "HomeKit controller support"
+description: "Instructions how to integrate your HomeKit devices within Home Assistant."
 date: 2018-03-19 21:04
 sidebar: true
 comments: false
@@ -9,7 +9,7 @@ sharing: true
 footer: true
 logo: apple-homekit.png
 ha_category: Hub
-ha_release: 0.66
+ha_release: 0.67
 ha_iot_class: "Local Polling"
 ---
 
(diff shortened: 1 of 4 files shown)


source/_components/homekit_controller.markdown
```diff
@@ -0,0 +1,22 @@
+---
+layout: page
+title: "Homekit controller support"
+description: "Instructions how to integrate your Homekit devices within Home Assistant."
+date: 2018-03-19 21:04
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: homekit.png
```
NovapaX: we already have a `apple-homekit.png` resource, I suggest you use that.

source/_components/homekit_controller.markdown
```diff
@@ -0,0 +1,22 @@
...
+ha_category: Hub
+ha_release: 0.66
+ha_iot_class: "Local Polling"
+---
+
+[Homekit](https://developer.apple.com/homekit/) controller integration for Home Assistant allows you to connect Homekit accessories to Home Assistant
```
NovapaX: We should add some kind of note to explain how this component differs from the existing HomeKit component? So the purpose of both is clear to the users.

source/_components/homekit_controller.markdown
```diff
@@ -0,0 +1,22 @@
...
+
+There is currently support for the following device types within Home Assistant:
+
+- [Light](../light.homekit_controller)
+
+The component will be automatically configured if the [`discovery:`](components/discovery/) component is enabled. For each detected Homekit accessory, a configuration prompt will appear in the web front end. Use this to provide the Homekit PIN.
```
NovapaX: Is there a way to disable Homekit discovery? Or ignore devices? If so, that should be described.

NovapaX (changes requested):

source/_components/homekit_controller.markdown
mjg59: Done

source/_components/homekit_controller.markdown
mjg59: Done

source/_components/homekit_controller.markdown
mjg59: Done

source/_components/homekit_controller.markdown
NovapaX: I don't see it yet. Forgot to push?

mjg59: Oops! Really done.

NovapaX (dismissed):

source/_components/homekit_controller.markdown
```diff
@@ -0,0 +1,28 @@
...
+There is currently support for the following device types within Home Assistant:
+
+- [Light](../light.homekit_controller)
+
+The component will be automatically configured if the [`discovery:`](components/discovery/) component is enabled. For each detected Homekit accessory, a configuration prompt will appear in the web front end. Use this to provide the HomeKit PIN. It can be disabled by adding the following configuration:
+
+```yaml
+discovery:
+  ignore:
+    - homekit
```
arsaboo: Should we call this `homekit_controller` or something? It can easily confused with `homekit` component

NovapaX: yes, agree. maybe homekit_accessory  ?

Thinking about it. Maybe the platform should be called 'homekit_accesory' instead of 'homekit_controller'

source/_components/homekit_controller.markdown
```diff
@@ -0,0 +1,28 @@
...
+ha_iot_class: "Local Polling"
+---
+
+[HomeKit](https://developer.apple.com/homekit/) controller integration for Home Assistant allows you to connect HomeKit accessories to Home Assistant. This component should not be confused with the [HomeKit](homekit) component, which allows you to control Home Assistant devices via HomeKit.
+
+There is currently support for the following device types within Home Assistant:
+
+- [Light](../light.homekit_controller)
+
+The component will be automatically configured if the [`discovery:`](components/discovery/) component is enabled. For each detected Homekit accessory, a configuration prompt will appear in the web front end. Use this to provide the HomeKit PIN. It can be disabled by adding the following configuration:
```
arsaboo: Also, what about those that don't use `discovery`? Please specify the minimum config for this component

NovapaX:
Another open end here: where to notify about the dependencies and possible long build time.
(that depends on when those dependencies are pulled in. If they are pulled in after the user entered the PIN in the configurator, that would be the place to warn about that.)

source/_components/homekit_controller.markdown
mjg59: This is the name that comes from netdisco, so we can't change that without changing the behaviour of netdisco, and that might break other users of netdisco. I went both ways on the component name - it's implementing a homekit controller but it's for controlling homekit accessories, so I could switch that.

source/_components/homekit_controller.markdown
mjg59: The metadata needed for HomeKit is only present in the mdns broadcasts, so it requires discovery.

mjg59:
Hm, good point. I think it'll happen the first time a HomeKit accessory is detected and dispatched to this component, which will happen before there's any opportunity for user interaction.

NovapaX:
If that is the case, and it cannot be changed, I have my doubts about enabling this platform by default. But further discussion about that should be done in the backend PR.

frenck (changes requested):
Marked PR with "Request changes" to prevent accidental merge.

@mjg59 Please resolve the comments/reviews/discussion 😃  👍

source/_components/homekit_controller.markdown
marthoc: If I may interject my 2 cents as a person who frequently helps on the forums. Now that there’s two HomeKit components they should really both be renamed. The current one to homekit_bridge (as that’s what it emulates) and this to homekit_accessories (since that’s what it supports adding to HA). There’s already confusion re the differences between Homekit and Homebridge, users can’t tell them apart since they both expose devices to HomeKit.

mjg59:
Implementation has now been changed so the component isn't enabled by default.

frenck (approved):

frenck:
@mjg59 Approved.

Can be merged as soon as the build passes (since the parent has been merged already).

PR #4977, opened 2018-03-21
Current

nielstron:
**Description:**
The outlier filter is wrongly described as band-stop filter while it is an band-pass filter (See https://en.m.wikipedia.org/wiki/Band-pass_filter for more information).
Also outliers and outlier detection is often interpreted slightly different (for example also concerning outliers within the accepted value range) so this should be renamed to band-pass on the long term.

## Checklist:

### DIFF
--- a/source/_components/sensor.filter.markdown
+++ b/source/_components/sensor.filter.markdown
@@ -99,7 +99,7 @@ The returned value is rounded to the number of decimals defined in (`precision`)
 
 ### {% linkable_title Band-pass %}
 
-The Band-pass filter (`bandpass`) cuts out any value outside a specific range.
+The Band-pass filter (`outlier`) cuts out any value outside a specific range.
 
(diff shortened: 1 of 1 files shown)


source/_components/sensor.filter.markdown
```diff
@@ -97,9 +97,9 @@ LowPass(state) = A * previous_state + B * state
 
 The returned value is rounded to the number of decimals defined in (`precision`).
 
-### {% linkable_title Outlier %}
+### {% linkable_title Band-pass %}
 
-The Outlier filter (`outlier`) is a basic Band-stop filter, as it cuts out any value outside a specific range.
+The Band-pass filter (`outlier`) cuts out any value outside a specific range.
```
arsaboo, line 102: The filter is `outlier`. So, let us not rename the filter. Feel free to update the description.

PR #4990, opened 2018-03-22
Template corrections

DubhAd:
Updated the templates to be standards compliant.

### DIFF
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -73,7 +73,7 @@ delay:
 ```yaml
 # Waits however many minutes input_number.minute_delay is set to
 # Valid formats include HH:MM and HH:MM:SS
-delay: {% raw %}'00:{{ states.input_number.minute_delay.state | int }}:00'{% endraw %}
+delay: {% raw %}'00:{{ states('input_number.minute_delay')|int }}:00'{% endraw %}
 ```
 ### {% linkable_title Wait %}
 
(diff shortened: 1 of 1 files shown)


source/_docs/scripts.markdown
```diff
@@ -73,20 +73,20 @@ delay:
 ```yaml
 # Waits however many minutes input_number.minute_delay is set to
 # Valid formats include HH:MM and HH:MM:SS
-delay: {% raw %}'00:{{ states.input_number.minute_delay.state | int }}:00'{% endraw %}
+delay: {% raw %}'00:{{ states('input_number.minute_delay')|int }}:00'{% endraw %}
```
dale3h: * All examples containing Jinja2 templates should be wrapped **outside** of the code markdown with the `{% raw %}` tag.
* Use double quotes (`"`) for:
  * `friendly_name`
  * Single-line templates:
    * `value_template`
    * `level_template`
    * `icon_template`
    * Children of `data_template`

source/_docs/scripts.markdown
```diff
@@ -73,20 +73,20 @@ delay:
...
 ```
 ### {% linkable_title Wait %}
 
 Wait until some things are complete. We support at the moment `wait_template` for waiting until a condition is `true`, see also on [Template-Trigger](/docs/automation/trigger/#template-trigger). It is possible to set a timeout after which the script will abort its execution if the condition is not satisfied. Timeout has the same syntax as `delay`.
 
 ```yaml
 # wait until media player have stop the playing
-wait_template: {% raw %}"{{ states.media_player.floor.state == 'stop' }}"{% endraw %}
+wait_template: {% raw %}"{{ is_state('media_player.floor', 'stop') }}"{% endraw %}
```
dale3h: * All examples containing Jinja2 templates should be wrapped **outside** of the code markdown with the `{% raw %}` tag.

source/_docs/scripts.markdown
```diff
@@ -73,20 +73,20 @@ delay:
...
 ```
 
 ```yaml
 # wait until a valve is < 10 or abort after 1 minutes.
-wait_template: {% raw %}"{{ states.climate.kitchen.attributes.valve < 10 }}"{% endraw %}
+wait_template: {% raw %}"{{ states.climate.kitchen.attributes.valve|int < 10 }}"{% endraw %}
```
dale3h: * All examples containing Jinja2 templates should be wrapped **outside** of the code markdown with the `{% raw %}` tag.

source/_docs/scripts.markdown
```diff
@@ -73,20 +73,20 @@ delay:
...
 timeout: 00:01:00
```
dale3h: This should be wrapped in single quotes: `timeout: '00:01:00'`.

dale3h (changes requested):

dale3h:
I committed a few more changes directly to the branch.

dale3h (approved):

PR #5001, opened 2018-03-24
Rewording, adding links

DubhAd:
The current doc is written as if the listed options are the only options, no Bluetooth, iOS app, etc. It also pushes people quite hard at OwnTracks over MQTT.

While this won't solve the iOS app problem (it's not a component for some reason), it should hopefully help people realise that OwnTracks isn't the only option ;)


frenck (approved):

frenck:
Nice work @DubhAd!
I've tweaked it a little bit (grammar/spelling).

PR #5015, opened 2018-03-25
Highlighting the hosting of files

DubhAd:
The use of `www/` is hidden in the middle of the page, it's impossible to direct people to it, and trivial to overlook. It's also wrong for Hass.io ;)

Updating to make it possible to link to it, and adding details on where to find it on Hass.io.

Also adding a note about needing to restart after creating the folder (there's been lots and lots of feedback about needing to do so)


fabaff (approved):
Thanks :bird:

PR #5017, opened 2018-03-25
Clarity updates

DubhAd:
Added that you can reload some pieces without a restart. Added reference to ensuring you run the check config as the user you run Home Assistant as (so that the right configuration is found)


fabaff (approved):
Thanks :bird:

PR #5020, opened 2018-03-25
Drop note about Pi 3 2.4GHz only WiFi

scop:
**Description:**

The (currently unsupported) B+ has 5GHz support.

## Checklist:


source/getting-started/index.markdown
```diff
@@ -41,7 +41,7 @@ We will need a few things to get started with installing Home Assistant. Links b
 
 1. Put the SD card in your SD card reader.
 1. Open Etcher, select the Hass.io image and flash it to the SD card.
-1. WiFi setup only: open the file `system-connections/resin-sample` (from the `resin-boot` volume on the SD card) with a text editor. Change `ssid` to be your network name and `psk` to be your password. Note that the Raspberry Pi 3 is a 2.4GHz WiFi device, so do not try to connect it to a 5GHz network.
```
fabaff, line 44: The note is not about the Raspberry Pi B+. Thus, it's still valid.

scop: The note is about "Raspberry Pi 3", which to me includes all Pi 3 models, including B and B+.

NovapaX: ok. Then it should be changed to something 'Raspberry Pi 3B (standard version, not the +-version)'

scop: I think that's too much detail, and that just removing the note would be a better idea.

NovapaX: Then I think we should just replace it with:
`Note: When connecting to a 5Ghz-only network, make sure your device supports that. If you don't know what that means, then you probably have nothing to worry about.`

scop: WFM

scop: Just to make it clear, because @NovapaX came up with the improved wording, I think it would be a good idea for them to submit a superseding PR to get the authorship logged right.

frenck: Feel free to take make the change @scop.
Members of the Home Assistant project (like me, @fabaff and @NovapaX) make suggestions on a daily basis and are not seeking for the credit on every single change/PR 😉

frenck (changes requested):
Awaiting change.

scop:
Hmm, this PR is incorrectly requesting merge to next, current would be the intention. I'll close this and open a new one, as it seems it's not possible to change the merge target afterwards.

NovapaX:
For future reference: https://blog.github.com/2016-08-15-change-the-base-branch-of-a-pull-request/

PR #5030, opened 2018-03-27
Added note about venv upgrade

DubhAd:
Added a note that if you've upgraded Python you'll need to rebuild the venv.


fabaff (dismissed):
Thanks :bird:

fabaff (approved):

PR #5039, opened 2018-03-28
Add details for values if the user should replace it (fixes #4999)

### LINKED ISSUE #4999: Request: Expansion and Clarification of the Documentation Standards
Referenced in the pull request template is the [Documentation Standards](https://home-assistant.io/developers/documentation/standards/). Reading through it however still leaves a lot of room for improvisation. 

My specific issue was in #4978 where I was was unsure if you were supposed to keep the surrounding marks or not. The context was of the example code block used `*marks around the part*` that I had to replace with my own bits. Some documentation uses `[brackets]` some use `"double quotes"` some use `'single quotes'` some don't use anything at all surrounding what you're supposed to replace with your own parts.

I personally prefer putting `[brackets]` around that kind of thing like the [Google Assistant](https://home-assistant.io/components/google_assistant/) docs, but when I put in #4978 I changed it to that but it was refused to be accepted. 

Regardless what the end result is, this is a kind of thing that really should be decided as a uniform decision across all the documentation.

fabaff:
**Description:**
Add details for values if the user should replace it.

**Related issue (if applicable):** fixes #4999

## Checklist:

### DIFF
--- a/source/developers/documentation/standards.markdown
+++ b/source/developers/documentation/standards.markdown
@@ -28,6 +28,8 @@ To ensure that the documentation for Home Assistant is consistent and easy to fo
   * Use `[string, int]` for configuration variables that accept multiple types.
 * Use YAML sequence syntax in the sample code if it is supported.
 * All examples should be formatted to be included in `configuration.yaml` unless explicitly stated.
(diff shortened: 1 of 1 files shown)


DubhAd (approved):
👍

PR #5063, opened 2018-03-31
Update hue.markdown

sebk-666:
Added some clarifying remarks regarding the configuration syntax.

### DIFF
--- a/source/_components/hue.markdown
+++ b/source/_components/hue.markdown
@@ -29,14 +29,20 @@ If you want to enable the component without relying on the [discovery component]
 hue:
   bridges:
     - host: DEVICE_IP_ADDRESS
+      allow_unreachable: true
+      allow_hue_groups: false
 ```
 
 Configuration variables:
 
-- **host** (*Optional*): IP address of the device, eg. 192.168.1.10. Required if not using the `discovery` component to discover Hue bridges.
+- **host** (*Optional* before 0.66): IP address of the device, eg. 192.168.1.10. Required if not using the `discovery` component to discover Hue bridges.
 - **allow_unreachable** (*Optional*): (true/false)  This will allow unreachable bulbs to report their state correctly.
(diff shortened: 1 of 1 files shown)


source/_components/hue.markdown
```diff
@@ -29,15 +29,21 @@ If you want to enable the component without relying on the [discovery component]
 hue:
   bridges:
     - host: DEVICE_IP_ADDRESS
+      allow_unreachable: true
+      allow_hue_groups: false
```
fabaff, line 33: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

fabaff (changes requested):

source/_components/hue.markdown
sebk-666: I understand that, but in this case it's very misleading, as at least I was assuming that:
```
hue:
  bridges:
    - host: 192.168.10.5
    - allow_unreachable: true
    - allow_hue_groups: true
```
was a valid and senseful configuration (due to all parameters being optional, this was valid before 0.66, but didn't make any sense).
So I wanted to give an example that shows the omitted dashes for additional parameters belonging to the same bridge.

I could add this as a second example and leave the original, minimal example as it is, if you prefer this option.

Also, up to 0.66, the host parameter was optional too, so according to your logic, it shouldn't have been in the example ;)

frenck:
I do not agree on this PR.
It feels a bit like using a band-aid or something.

I agree with @fabaff, no optional parameters in the initial example.
Feel free to provide additional/specific use-cases or examples.

The rest of this PR fiddles around with notes, that should be in the release notes as a breaking change.
I find it a generally bad practice to add exceptions to the documentation for specific versions.

I'm going to close this one because of the above reasoning.

PR #5064, opened 2018-03-31
Escape special characters in credits

trisk:
**Description:**
Full names on GitHub can contain control characters for HTML or Markdown. Escape these before they reach the Markdown parser for the credits page.


fabaff (approved):
Thanks :bird:

PR #5072, opened 2018-03-31
Properly quote things so ON/OFF don't become True/False

Skaronator:
Source:
https://github.com/home-assistant/home-assistant/blob/89c7c80e42c769857389a3edda2a874b63f974f5/homeassistant/components/light/mqtt.py#L61-L62


fabaff (approved):
Thanks :bird:

PR #5074, opened 2018-03-31
Match current code

Skaronator:
Source: https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/light/__init__.py#L460


Skaronator:
Just notice that the current value in Home Assistant isn't really correct.

DubhAd (approved):
👍

PR #5089, opened 2018-04-02
Update hue.markdown

sebk-666:
- Removed the "(Optional)" tag from the host-Parameter, because it is mandatory now
- Added another configuration example to clarify how to specify multiple configuration variables; the examples in the "multiple bridges..." section do show this already, but this section might be skipped by the majority of users owning only a single bridge.


DubhAd (approved):
👍

PR #5105, opened 2018-04-04
Update device_tracker.owntracks_http.markdown

flavio20002:
Track battery with a template sensor.

### DIFF
--- a/source/_components/device_tracker.owntracks_http.markdown
+++ b/source/_components/device_tracker.owntracks_http.markdown
@@ -34,3 +34,17 @@ Open OwnTracks and go to Connection preferences:
  - Identification: Turn **Authentication** on, username `homeassistant` and password is your API password that you use to login to Home Assistant.
  
 Host example: If I host my Home Assistant at `https://example.duckdns.org`, my name is Paulus and my phone is a Pixel I would set the host to be `https://example.duckdns.org/api/owntracks/paulus/pixel`. This will result in an entity with an ID of `device_tracker.paulus_pixel`. You can pick any name for the user and the device.
+
+You can track the battery of you device, by defining a template sensor:
+
+```yaml
+# Example configuration.yaml entry for battery monitoring
+device_tracker:
+    - platform: template
+    sensors:
+      pixel_battery:
+        value_template: '{{ states.device_tracker.paulus_pixel.attributes.battery|int }}'
+        unit_of_measurement: "%"
+        entity_id: 
+        - device_tracker.paulus_pixel
+```


source/_components/device_tracker.owntracks_http.markdown
```diff
@@ -34,3 +34,17 @@ Open OwnTracks and go to Connection preferences:
...
  
 Host example: If I host my Home Assistant at `https://example.duckdns.org`, my name is Paulus and my phone is a Pixel I would set the host to be `https://example.duckdns.org/api/owntracks/paulus/pixel`. This will result in an entity with an ID of `device_tracker.paulus_pixel`. You can pick any name for the user and the device.
+
+You can track the battery of you device, by defining a template sensor:
+
+```yaml
+# Example configuration.yaml entry for battery monitoring
+device_tracker:
+    - platform: template
+    sensors:
```
DubhAd, line 44: The indenting here is off - platform should have four less spaces before it.

source/_components/device_tracker.owntracks_http.markdown
```diff
@@ -34,3 +34,17 @@ Open OwnTracks and go to Connection preferences:
...
+      pixel_battery:
+        value_template: '{{ states.device_tracker.paulus_pixel.attributes.battery|int }}'
```
DubhAd, line 46: You should consider using a default for when the battery level isn't defined:
`'{{ states.device_tracker.paulus_pixel.attributes.battery|default(0)|int }}'``

source/_components/device_tracker.owntracks_http.markdown
```diff
@@ -34,3 +34,17 @@ Open OwnTracks and go to Connection preferences:
...
+        unit_of_measurement: "%"
+        entity_id: 
+        - device_tracker.paulus_pixel
+```
```
DubhAd, line 50: I'd suggest (and it is just a suggestion) adding an icon template to make a pretty icon that reflects the battery level:

```      icon_template: >
        {% set battery_level = states.device_tracker.person1_tablet.attributes.battery|default(0)|int %}
        {% set battery_round = (battery_level / 10) |int * 10 %}
        {% if battery_round >= 100 %}
          mdi:battery
        {% elif battery_round > 0 %}
          mdi:battery-{{ battery_round }}
        {% else %}
          mdi:battery-alert
        {% endif %}```

(that one came from work done by @skalavala )

That'll then match up with battery level reports by the various components that support it.

source/_components/device_tracker.owntracks_http.markdown
fabaff, line 46: Please check https://www.home-assistant.io/developers/documentation/standards/#templates

PR #5108, opened 2018-04-04
Remove ref to watchers

robmarkcole:
`watchers` is not required so is removed


frenck (approved):

frenck:
Thanks @robmarkcole 🏅

PR #5112, opened 2018-04-04
Added troubleshooting section

DubhAd:
Note about how to (usually) resolve issues with certbot permissions

### DIFF
--- a/source/_components/http.markdown
+++ b/source/_components/http.markdown
@@ -95,3 +95,13 @@ If you want to use Home Assistant to host or serve static files then create a di
 <p class='note'>
   If you've had to create the `www/` folder for the first time, you'll need to restart Home Assistant.
 </p>
+
+## {% linkable_title Troubleshooting %}
+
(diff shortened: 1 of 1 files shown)


frenck:
Shouldn't this be more user specific?
Simply giving the user, group and world the executable bit feels a bit "easy way out".

DubhAd:
It is, but you can only access a file that you know the full path and name of that way. It also works for any user that somebody's running Home Assistant as, and saves trying to take people through identifying the group of the user they run Home Assistant as, and changing the group ownership of the directories.

So, "easy way out" yes - but for the people running it too.

frenck:
@DubhAd Thanks for the explanation. 
From a security perspective, I do not agree.
From a user perspective, I understand.

I'll let it up to another reviewer to decide.

dgomes:
its really easy, but I don't think that should be our stance (advising a less secure solution)

A PR with better error handling (indicating not enough permissions to access the file) would be better

frenck:
Closing this PR, because of above-stated reasoning.

PR #5147, opened 2018-04-10
Revert "Add example file attachment parameters"

frenck:
Reverts home-assistant/home-assistant.github.io#5134

As it seems, it should not have been merged already!


DubhAd (approved):
👍

PR #5154, opened 2018-04-11
Update the motion activated light example

RiRomain:
**Description:**
Make use of the new timer component for the motion activated light example.
It allow the user to cancel the timer if needed, extend it, pause it, ...

## Checklist:

### DIFF
--- a/source/_cookbook/turn_on_light_for_10_minutes_when_motion_detected.markdown
+++ b/source/_cookbook/turn_on_light_for_10_minutes_when_motion_detected.markdown
@@ -16,23 +16,27 @@ This recipe will turn on a light when there is motion and turn off the light whe
 
 ```yaml
 automation:
-- alias: Turn on kitchen light when there is movement
-  trigger:
-    platform: state
-    entity_id: sensor.motion_sensor
-    to: 'on'
-  action:
-    service: homeassistant.turn_on
-    entity_id: light.kitchen_light
+  - alias: Turn on kitchen light when there is movement
+    trigger:
+      - platform: state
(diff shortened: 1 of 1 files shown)


DubhAd:
I'd personally prefer to see it as an additional example, rather than replacing the existing one.

RiRomain:
It might be confusing, no? Beginner usually don't really know what they need and might be confused by which of the two they should use and why.
I think the previous (current) example have no advantage over the timer.

frenck:
I think this is confusing as well and agree with @DubhAd on making this an additional example as well.

The timer allows for additional actions (like you described, extend it, pause it e.d.).
But that's not the point of this example. It added an extra moving part (the timer) without any additional functionality.

So if you decide to add it as an additional example, I'd suggest on enriching it with a reason to use the timer instead of the current method (e.g., add something to pause/unpause it).

frenck (changes requested):
See comments above.

RiRomain:
> without any additional functionality

I cannot really agree with that, the solution with the timer allow to use any trigger. While the current solution only allow state trigger.

I use an mqtt trigger, which doesn't have a state but only a trigger event, it simply doesn't work with the state trigger and an automation based on its own state
```
    trigger:
      - platform: mqtt
        topic: home/movementsensor1
        payload: 1
```

DubhAd:
So add it as a second way of achieving the goal. For people new to HA, the `for` option is simpler and self contained. Then your suggestion can become the _advanced_ method.

RiRomain:
This might be confusing, having 2 automation offering the same feature. Let's leave the current one.

PR #5163, opened 2018-04-12
Added further notes about dialout group

DubhAd:
Added note about checking the account is in the dialout group, and how to add it if not

Added note about the device path being an example


fabaff (approved):
:bird:

PR #5181, opened 2018-04-15
Propose edits to DuckDNS docs

robmarkcole:
I've found this docs confusing. `lets_encrypt.accept_terms` should appear at the top of the list as it's the first entered variable. Also this should become `required` as typically `optional` config variables are not displayed in the example config, but it is here. I've also noticed that seconds are not optional, but are required. I've also added a comment about the need to use `https`, as this wasn't obvious to me (when accessing locally). Please see [this thread](https://community.home-assistant.io/t/effortless-encryption-with-lets-encrypt-and-duckdns/42492/33).

### DIFF
--- a/source/_addons/duckdns.markdown
+++ b/source/_addons/duckdns.markdown
@@ -26,7 +26,7 @@ featured: true
 
 Configuration variables:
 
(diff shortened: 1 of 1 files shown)


fabaff (approved):
Thanks :bird:

PR #5191, opened 2018-04-15
Adding list of supported components

DubhAd:
Only the blog post says what components are supported, making it confusing for people. Adding that information to the docs.


fabaff (approved):
:bird:

NovapaX:
Doing a search for `def unique_id(self)` in the sources gives me at least 60 components that support the entity registry now.....

PR #5197, opened 2018-04-16
Add info on whitelisting to File sensor

robmarkcole:
Needs to be added as causing issues for users https://community.home-assistant.io/t/0-67-whitelisted-dir-issues/50305/5


arsaboo (approved):
Thanks 💯

PR #5199, opened 2018-04-16
Describe zip-file REQUIREMENT syntax

NovapaX:
**Description:**

## Checklist:

### DIFF
--- a/source/developers/component_deps_and_reqs.markdown
+++ b/source/developers/component_deps_and_reqs.markdown
@@ -41,3 +41,12 @@ If you need to make changes to a requirement to support your component, it's als
 * `git clone https://github.com/balloob/pychromecast.git`
 * `pip install ./pychromecast`
 * `hass --skip-pip`
+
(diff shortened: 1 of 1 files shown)


fabaff:
We shouldn't add those details. The default should be that `REQUIREMENT` point to a package which is available on PyPI because other ways slow down the tests. Adding information about way around this may encourage people not to follow this suggestion.

NovapaX:
I wrote this more with custom components in mind....

NovapaX:
I updated this to make clear this is not the desired approach for components included in home-assistant. I still think it's useful information for those creating custom components or for testing new builds/branches of upstream packages.

fabaff (approved):
:bird:

fabaff:
The developer documentation was moved. I will manually add the changes at the right location.

NovapaX:
Thanks for that! (I'm a bit short on time lately and my development time has reduces to zero unfortunately)

PR #5280, opened 2018-04-30
Adding note about mDNS

DubhAd:
For when mDNS doesn't work, people are stuck. Adding a note about how to work around that.


fabaff (approved):
:bird:

PR #5284, opened 2018-05-01
Calling out that devices are still added

DubhAd:
It's a common source of confusion that with `track_new_devices: False` devices are still added to `known_devices.yaml`. This adds a note to say that this is to be expected.


fabaff (approved):
Thanks :bird:

PR #5294, opened 2018-05-02
Added explicit call out about Z-Wave

DubhAd:
Called out the need to disable the auto-start when you're using Z-Wave


ludeeus:
The same issues are present if the user rely on `discovery:` for their components.

fabaff (approved):
Thanks :bird:

PR #5297, opened 2018-05-02
Add specific instructions for Meraki device tracker

dudeofawesome:
The current instructions aren't really complete, and require you to have both an understanding of how the Home Assistant API works, as well as be able to read through the code for the Meraki component to find the API endpoint you're supposed to use.


fabaff (approved):
Thanks :bird:

PR #5305, opened 2018-05-04
Re-ordered the automation blocks

DubhAd:
The automations were as the editor produces them, so not logically ordered or clear to read. I've re-written them to match the rest of the documentation.


fabaff (approved):
Thanks :bird:

PR #5308, opened 2018-05-04
Added note about instant updates

DubhAd:
Added note about instant updates and the Controlled command classes that are required for it to work.


fabaff (approved):
:bird:

PR #5314, opened 2018-05-06
Allow HomeKit name to be customized

schmittx:
**Description:**
Allow name shown in HomeKit to be customized by specifying `name` within `entity_config`.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#14159

## Checklist:

### DIFF
--- a/source/_components/homekit.markdown
+++ b/source/_components/homekit.markdown
@@ -23,22 +23,6 @@ The `HomeKit` component allows you to forward entities from Home Assistant to Ap
   If you are upgrading Home Assistant from `0.65.x` and have used the HomeKit component, some accessories may not respond or may behave unusually. To fix these problems, you will need to remove the Home Assistant Bridge from your Home, stop Home Assistant and delete the `.homekit.state` file in your configuration folder and follow the Homekit [setup](#setup) steps again.
 </p>
 
-{% raw %}
-```yaml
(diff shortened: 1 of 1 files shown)


cdce8p (approved):

cdce8p (changes requested):
Just though about this. Can we move the example config and the change for the filter description to a new PR? That way we avoid potential merge conflicts.

You would need to fetch the current version first, since I did some small changes and rebased the branch.

schmittx:
Yeah, honestly I kind of forgot that I made the other changes along with the name customize stuff. My bad.

I'll split this out later tonight, we can merge the example config and filter description first then the name customize details in a separate PR.

cdce8p:
Would work. I just haven't done it yet, so that we have the example here, but I guess the commit history works as well. Will do the change in a minute.

The new PR would need to be against the `current` branch though. So the changes wont be in there until they are merge in three weeks. That's why I wanted to split them, just in case somebody else does some other changes.

cdce8p:
So, I reverted the changes that should go into the new PR.
Might be best if you take a look at https://github.com/home-assistant/home-assistant.github.io/pull/5314/commits/2e1200318536e545783051e7fb240b90ed623763 and reverse it.

Will merge this shortly.

cdce8p (approved):

PR #5316, opened 2018-05-06
synology reverse proxy: add template patch

gsemet:
instruction from orobardet/dsm-reverse-proxy-websocket works with recent version of DSM. Current documentation does not seem to work


fabaff (approved):
Thanks :bird:

gsemet:
Thanks !

PR #5318, opened 2018-05-06
Change to advertise esphomeyaml

OttoWinter:
**Description:**

The MQTT discovery docs pointed to [esphomelib](https://github.com/OttoWinter/esphomelib). But since that was added in the docs I released another, easier-to-use project called [esphomeyaml](https://esphomelib.com/esphomeyaml/index.html). Basically, it creates custom firmwares for ESP8266/ESP32 nodes from YAML configuration files.

So on the MQTT discovery site, I'd like to advertise esphomeyaml. There's another link to esphomelib on the MQTT JSON Light page. But, as the people navigating to that component page probably prefer customizability, I think keeping the link esphomelib there is fine.

## Checklist:


fabaff (approved):
:bird:

PR #5347, opened 2018-05-11
Restore link to dev docs

DubhAd:
Now that it's "hidden" at the bottom of the page we've had a few questions about whether there's any developer docs. Suggesting restoring it to the top so that people don't assume there's none ;)

### DIFF
--- a/source/_includes/custom/navigation.html
+++ b/source/_includes/custom/navigation.html
@@ -13,6 +13,7 @@
   <li><a href="/components/">Components</a></li>
   <li><a href="/docs/">Docs</a></li>
   <li><a href="/cookbook/">Examples</a></li>
+  <li><a href="https://developers.home-assistant.io/">Developers</a></li>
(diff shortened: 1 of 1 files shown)


frenck:
This was discussed in the chat extensively.
The main reasoning for removing this link, is the "user" focus, instead of the developer.
Our developers should be able to find it just fine. Maybe that is misguessed?

@balloob, since you've recently changed this, what do you think?

DubhAd:
Looks like at least some of the developers are struggling 🤷‍♂️

balloob:
I think that we should not add it to the top. The main website is indeed user focused. There are a bunch of links already:

 - First link in the sidebar of the "need help" page
 - A link at the bottom of every page.
 - There is a link inside `CONTRIBUTING.md` on the main repo.

frenck:
Closing this PR. See reasoning above.

PR #5371, opened 2018-05-15
Update sensor.scrape.markdown

flavio20002:
**Description:**

Added scan_interval. I tested with this snippet, with and without the scan_interval

  - platform: scrape
    resource: https://odditymall.com/random
    name: Random Test
    scan_interval: 900
    select: "h2 > a"

## Checklist:

### DIFF
--- a/source/_components/sensor.scrape.markdown
+++ b/source/_components/sensor.scrape.markdown
@@ -36,6 +36,7 @@ Configuration variables:
 - **authentication** (*Optional*): Type of the HTTP authentication. Either `basic` or `digest`.
 - **username** (*Optional*): The username for accessing the website.
 - **password** (*Optional*): The password for accessing the website.
+- **scan_interval** (*Optional*): Defines number of seconds for polling interval (defaults to 60 seconds).
 
 ## {% linkable_title Examples %}
 


source/_components/sensor.scrape.markdown
```diff
@@ -36,6 +36,7 @@ Configuration variables:
 - **authentication** (*Optional*): Type of the HTTP authentication. Either `basic` or `digest`.
 - **username** (*Optional*): The username for accessing the website.
 - **password** (*Optional*): The password for accessing the website.
+- **scan_interval** (*Optional*): Defines number of seconds for polling interval (defaults to 60 seconds).
```
fabaff, line 39: There is no need to add this. It described [here](https://www.home-assistant.io/docs/configuration/platform_options/#scan-interval) already and is valid for platform which are doing polling.

fabaff (changes requested):

flavio20002:
There is need to add this...I explain you why: other platforms like command line sensor has this inside the documentation. Moreover the scan interval is very important, otherwise, the website could ban your IP for too much traffic. 

Look at: https://www.home-assistant.io/components/sensor.command_line/ and https://www.home-assistant.io/components/binary_sensor.command_line/

Moreover, in this page https://www.home-assistant.io/docs/configuration/platform_options/#scan-interval it's not written the correct default, that is 60 seconds for the Scrape component. A lot of user get confused about that.

@fabaff Let me add it in one of the examples at least...

PR #5372, opened 2018-05-15
Remove simplepush.io

c727:
**Description:**

website is offline and apps are removed from store
home-assistant/home-assistant#14358


fabaff (approved):
:bird:

PR #5378, opened 2018-05-16
Add HomeKit support for media players

schmittx:
**Description:**
Adds support for media players in HomeKit.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#14446

## Checklist:

### DIFF
--- a/source/_components/homekit.markdown
+++ b/source/_components/homekit.markdown
@@ -95,6 +95,11 @@ homekit:
                 required: false
                 type: string
                 default: ''
+              mode:
+                description: Operation mode of switch within HomeKit. Valid modes are `on_off`, `play_pause`, `play_stop`, and `toggle_mute`. If the entity does not support the specified `mode`, the first supported `mode` will be used. Only applicable for `media_player` entities.
+                required: false
+                type: string
+                default: Depends on `supported_features` of entity.
 {% endconfiguration %}
 
 <p class='note'>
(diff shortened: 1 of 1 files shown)


source/_components/homekit.markdown
```diff
@@ -95,6 +102,11 @@ homekit:
                 required: false
                 type: string
                 default: ''
+              mode:
+                description: Operation mode of switches within HomeKit. Valid modes are `on_off`, `play_pause`, `play_stop`, and `toggle_mute`. The entity will ignore any `mode` which is not supported. If not specified, all supported modes will be added. Only applicable for `media_player` entities.
```
cdce8p: What do you think of:
Operation **modes** of switches within **the Home app**. Valid modes are `on_off`, `play_pause`, `play_stop`, and `toggle_mute`. ~~The entity will ignore any `mode` which is not supported.~~ ~~If not specified, all supported modes will be added.~~ Only applicable for `media_player` entities.

source/_components/homekit.markdown
```diff
@@ -95,6 +102,11 @@ homekit:
...
+                required: false
+                type: list
+                default: `on_off`, `play_pause`, `play_stop`, `toggle_mute`
```
cdce8p: The default would change to: '`<All supported modes>`'

cdce8p (changes requested):

source/_components/homekit.markdown
schmittx: I agree, but I don't think a code block should be used since it's not the actual entry. I prefer just plain text.

source/_components/homekit.markdown
schmittx: I agree, but I think we should stay consistent and use **HomeKit** rather than **the Home app**, because HomeKit support is not exclusive to Apple's Home app.

source/_components/homekit.markdown
cdce8p: I wanted to use the same syntax for placeholders as above for `entity_id`. A normal `string` might be misleading.

cdce8p (approved):
Ready to be merged once parent PR is merged and test pass.

PR #5398, opened 2018-05-19
Added format to two digits

DubhAd:
Formatted the digit for the delay to two digits, rather than a random one or two. There have been multiple reports of the example in the doc not working for 1 through 9, and it needing formatted to two digits.


fabaff (approved):
Thanks :bird:

PR #5428, opened 2018-05-25
Remove 'mention my name' as I no longer active develop this component

aequitas:
**Description:** Due to time constraints I no longer actively develop on HA. I don't want to give users false hope for integrating their hardware if I can't commit to that promise.


DubhAd (approved):
👍

PR #5431, opened 2018-05-25
Update device_tracker.fritz.markdown

mtdcr:
Added zlib1g-dev, which is needed to link lxml. Rephrased to make clear that one either needs to install lxml from a repository or build it from sources.


fabaff (approved):
Thanks :bird:

PR #5432, opened 2018-05-25
Added notes about the states

DubhAd:
A common question is "what's the state of the device tracker", so I've added a summary


fabaff (approved):
Thanks :bird:

PR #5446, opened 2018-05-27
Update about the only known non-working add-on

DubhAd:
Based upon details from Frenck in the forum, updating about the only known non-working add-on


Vasiley (dismissed):
Thanks Again

frenck (approved):

PR #5449, opened 2018-05-28
Update for IP and ports

DubhAd:
Updated to use RFC5737 IP address (documentation example IP range). 

Updated some incorrect assumptions about HTTPS only able to use port 443 and SSH only able to use 22


Vasiley (approved):
:+1:

PR #5458, opened 2018-05-30
Improving the Securing docs

DubhAd:
This probably needs some back and fore to work out what the projects want here, but I thought I'd give it a first pass through to turn this from a "wall of things" that seem equally important into something hopefully more digestible. Particularly as some of the things look like you should do all of them, but they're actually mutually exclusive, or functionally identical.

I've tried to break it into things that people must do (here's the basic minimum) and things they should consider if they want to make it more secure. I've also tried to steer people towards `cloud` where it'll work for them, and if they just want remote access to the UI to use a VPN or Tor (I moved the SSH tunnel option last because it's the most complicated to set up of the three).


stephenmg12:
Does using Tor really keep home-assistant secure? If someone knew or found out the onion address wouldn't they be able to access home-assistant as well? I would move up the Debian guides with raspberry pi being the most popular hardware which is Debian based, maybe even with a note pointing that out. I would also add fail2ban in the mix to limit brute force attacks. In the proxy section, maybe mention Nginx as well, it tends to be less of a resource hog. Disagree on the SSH tunnel being most complicated, ssh is already installed on most Linux systems.

I do like the organization, separating what you must do from what you can do.

DubhAd:
Thanks.

Tor - looks like [you need a cookie](https://www.home-assistant.io/docs/ecosystem/tor/) for access. 

Debian guide was moved up, and references Raspbian.

NGINX linked along with Apache (I run it myself), fail2ban linked.

fabaff:
> Does using Tor really keep home-assistant secure?

[Without cookie](https://www.home-assistant.io/blog/2017/11/12/tor/#cranking-up-security) only partial as it's possible to brute-force the address.

carlchan:
Could add link to existing fail2ban howto: https://www.home-assistant.io/cookbook/fail2ban/

frenck:
This branch has now merge conflicts and this might get worse.
Any plans on moving this PR forward @DubhAd? Or can we close this one up for now?

DubhAd:
Well, I was hoping for some more feedback, but 🤷‍♂️ 

I'll rebase and update then maybe somebody will merge it

DubhAd:
@frenck I'm now just waiting on _somebody_ reviewing it ;)

frenck:
@DubhAd Maybe you could look at the conflict?
Ping me afterward, so I can take a look immediately after.

DubhAd:
@frenck Updated. I've also removed the piece about setting the API password, as that's no longer relevant

frenck (approved):

frenck:
I do think there is still room for improvement, nevertheless, I consider this a major step forwarding compared to the page at its current state.

✅ Approved.

PR #5464, opened 2018-05-31
Moved the power supply note

DubhAd:
People keep missing the power warning because it's long after the install notes


skalavala (approved):
good update!

PR #5466, opened 2018-05-31
Wether Underground clarification

kloggy:
Added a note to clarify that the WU sensor is added to the entity_registry  and that numbers are appended to subsequent weather stations after the first one. Nowhere as far as I know is this documented or made clear.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/sensor.wunderground.markdown
+++ b/source/_components/sensor.wunderground.markdown
@@ -195,4 +195,8 @@ group:
 Note: While the platform is called “wunderground” the sensors will show up in Home Assistant as “PWS” (eg: sensor.pws_weather).
 </p>
 
+Note that the Weather Underground sensor is added to the entity_registry. Note also that second and subsequent Personal Weather Station ID (pws_id) will have their monitored conditions suffixed with an index number e.g.
+
+      - sensor.pws_weather_1d_metric_2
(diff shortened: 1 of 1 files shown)


source/_components/sensor.wunderground.markdown
```diff
@@ -195,4 +195,8 @@ group:
 Note: While the platform is called “wunderground” the sensors will show up in Home Assistant as “PWS” (eg: sensor.pws_weather).
 </p>
 
+Note that the Weather Underground sensor is added to the entity_registry. Note also that second and subsequent Personal Weather Station ID (pws_id) will have their monitored conditions suffixed with an index number e.g.
```
DubhAd: Starting both sentences with `Note` feels repetitive. How about *The second and subsequent...*?

kloggy: Yes but I think it is important to highlight that it appears in entity_registry. How about:

The Weather Underground sensor is added to the entity_registry and any second and subsequent Personal Weather Station ID....

DubhAd: Sounds good, it's at least not repetitive :)

arsaboo (approved):
👍

PR #5475, opened 2018-06-01
No more free keys

DubhAd:
https://apicommunity.wunderground.com/weatherapi/topics/weather-underground-api-changes

They're no longer issuing free keys, and have stated that _for the time being, free keys will continue to work_.


fabaff (approved):
Thanks :bird:

PR #5477, opened 2018-06-02
NSW Fuel Station Price Sensor

nickw444:
**Description:**
Add documentation for `sensor.nsw_fuel_station` component

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#14757

## Checklist:

### DIFF
--- a/source/_components/sensor.nsw_fuel_station.markdown
+++ b/source/_components/sensor.nsw_fuel_station.markdown
@@ -30,9 +30,6 @@ sensor:
   - platform: nsw_fuel_station
     station_id: 291
     station_name: Coles Express Cremorne
-    fuel_types:
-      - E10
-      - P95
 ``` 
 
 {% configuration %}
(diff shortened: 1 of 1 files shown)


fabaff:
The docs should contain a hint there is a limit of [2500 requests per months](https://api.nsw.gov.au/price). If one starts to use multiple sensors and overwrites the update interval then she/he will hit the limit.

nickw444:
> The docs should contain a hint there is a limit of 2500 requests per months. If one starts to use multiple sensors and overwrites the update interval then she/he will hit the limit.

Where did you find this information about 2500 requests? The client does not actually use the [official API](https://api.nsw.gov.au/fuel-price-check/apis), since this requires each home-assistant user creating API credentials, waiting to be approved, and subsequently configuring them in home-assistant.

Instead, the client library `nsw-fuel-api-client` piggybacks on the API exposed for the [FuelCheck App](https://www.fuelcheck.nsw.gov.au/app), which does not require any authentication and doesn't appear to have any limits (however usage of this API is possibly questionable, but I don't think it differs from many of the other unofficial 3rd party platforms available on home-assistant.)

fabaff:
> Where did you find this information about 2500 requests?

At https://api.nsw.gov.au/price

nickw444:
I'm aware of the limit imposed by the official API, but as I mentioned in the previous comment, the underlying library uses an unofficial API exposed by the FuelCheck App, which I expect has higher limits since it is the underlying API for a [web](https://www.fuelcheck.nsw.gov.au/app) and mobile application used to show fuel information.

The main reason for using the unofficial API is that it does not require any authentication - using the official API requires consumers to apply for an API key, a process that took a week for me (which is why I resorted to the unofficial API).

fabaff (approved):
Thanks :bird:

PR #5486, opened 2018-06-03
Correct commands for 'homeassistant' user

trisk:
**Description:**
Correct commands for `homeassistant` user
* Update commands to start shell as `homeassistant` user
* Fix references to `hass` user

## Checklist:

### DIFF
--- a/source/_components/recorder.markdown
+++ b/source/_components/recorder.markdown
@@ -2,7 +2,7 @@
 layout: page
 title: "Recorder"
 description: "Instructions on how to configure the data recorder for Home Assistant."
-date: 2017-09-24 09:00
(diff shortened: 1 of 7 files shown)


DubhAd:
Only one problem, by default Hassbian doesn't give the `homeassistant` user a shell, hence the instructions - I've just gone through the source to verify that, and found only:

    useradd -u 1001 -g 1001 -rm homeassistant

The other edits are good however.

trisk:
@DubhAd `sudo -s` will actually start the invoking user's shell instead of the target user's shell, so this should be fine for that (my instance has `/bin/false` as `homeassistant`'s shell).

DubhAd (approved):
👍

PR #5489, opened 2018-06-03
Adding warning about backups

DubhAd:
Many people have failed to set up backups, and lost their config. Adding some warnings to hopefully reduce the numbers.


fabaff (approved):
Thanks :bird:

PR #5500, opened 2018-06-06
Update vacuum docs to be more generic

dshokouhi:
**Description:**

Docs kept mentioned `botvac` instead of `vacuum` so they are updated to be more generic.  They still tell the user that the services depend on their platform.

## Checklist:


DubhAd (approved):
👍

PR #5513, opened 2018-06-09
Required / recommended generic Linux packages

dobrynski:
As per Frenck's advice this is the current list of packages that Hass.io would like to have / require:
  apparmor-utils
  apt-transport-https
  avahi-daemon
  ca-certificates
  curl
  dbus
  jq
  network-manager
  socat
  software-properties-common

It would be helpful if this was mentioned in the generic Linux server install section.

**Description:**

## Checklist:

### DIFF
--- a/source/hassio/installation.markdown
+++ b/source/hassio/installation.markdown
@@ -54,7 +54,22 @@ If you copy over your existing Home Assistant configuration, make sure to enable
 
 ## {% linkable_title Alternative: install on generic Linux server %}
 
-For advanced users, it is also possible to try Hass.io on your [Linux server or inside a virtual machine][linux]. To do so, run the following command as root:
+For advanced users, it is also possible to try Hass.io on your [Linux server or inside a virtual machine][linux].
+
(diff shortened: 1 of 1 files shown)


source/hassio/installation.markdown
```diff
@@ -64,6 +64,20 @@ curl -sL https://raw.githubusercontent.com/home-assistant/hassio-build/master/in
...
+    apt-transport-https
+    avahi-daemon
+    ca-certificates
+    curl
+    dbus
+    jq
+    network-manager
+    socat
+    software-properties-common
+</p>
```
fabaff: I was thinking about adding them when I wrote the blog post. I didn't do it because it would assume that you want to use Ubuntu/Debian. I guess that now nobody will try it with Fedora/Alpine Linux anymore because Apparmor is used.

dobrynski:
Perhaps that can be clarified in an additional sentence and/or the list split into required and optional?

fabaff (approved):
Thanks :bird:

fabaff:
Let's go from here and see what the feedback is.

PR #5552, opened 2018-06-16
Promoting timezone issue to warning

DubhAd:
Given the long standing bug with the alert component - where the timezone in HA differs from the timezone of the OS when starting HA it either fires rapidly, or very slowly - promoting this to a warning.

See [this bug](https://github.com/home-assistant/home-assistant/issues/13609) and [this thread](https://community.home-assistant.io/t/error-alert-component/56242) which reminded me about it.


fabaff (approved):
:bird:

PR #5605, opened 2018-06-26
UVC updated configuration example

gottsman:
Updated configuration example and description.

## Checklist:

### DIFF
--- a/source/_components/camera.uvc.markdown
+++ b/source/_components/camera.uvc.markdown
@@ -35,6 +35,7 @@ camera:
   - platform: uvc
     nvr: IP_ADDRESS
     key: API_KEY
+    password: PASSWORD
 ```
 
 {% configuration %}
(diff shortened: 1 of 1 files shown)


frenck:
Thanks for opening a PR @gottsman!

Nevertheless, I cannot match this to the actual code.
The code shows that the password is currently optional, see:

https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/camera/uvc.py#L32

So why this change?

gottsman:
Agree that the password should be optional since it defaults to "ubnt" as part of [https://github.com/home-assistant/home-assistant/pull/7499](url).

The other part of the patch is simply reordering the arguments and changing the errant "name" wording to "password" in the config documentation.

frenck:
Since the password is optional; I've corrected the PR.

frenck (approved):

PR #5636, opened 2018-06-30
Update tradfri.markdown

linuxlurak:
**Description:**
Added hint on integrating multiple tradfri gateways

## Checklist:

### DIFF
--- a/source/_components/tradfri.markdown
+++ b/source/_components/tradfri.markdown
@@ -32,11 +32,12 @@ You can add the following to your `configuration.yaml` file if you are not using
 # Example configuration.yaml entry
 tradfri:
   host: IP_ADDRESS
+  host: IP_ADDRESS_SECOND_GATEWAY_IF_THERE_IS_ONE
 ```
 
 Configuration variables:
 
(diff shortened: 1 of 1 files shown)


source/_components/tradfri.markdown
```diff
@@ -32,11 +32,12 @@ You can add the following to your `configuration.yaml` file if you are not using
 # Example configuration.yaml entry
 tradfri:
   host: IP_ADDRESS
+  host: IP_ADDRESS_SECOND_GATEWAY_IF_THERE_IS_ONE
```
fabaff, line 35: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps users to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

linuxlurak: honestly I am not sure if it's correct at all... Is this the way to add multiple hosts? An your input is good about keeping the default example as simple as possible. I'll close this pull request. Perhaps some one can enlighten me with multiple gateway settings? Thanks.

PR #5721, opened 2018-07-09
Fix configuration descriptions

fabaff:
**Description:**
Fix configuration descriptions.

## Checklist:

### DIFF
--- a/source/_lovelace/entities.markdown
+++ b/source/_lovelace/entities.markdown
@@ -12,8 +12,6 @@ footer: true
 
 Entities will be the most common type of card that will also be the most familiar to people using the standard interface. It groups items together very close to how groups used to do.
 
-## {% linkable_title Options %}
-
 {% configuration %}
 type:
   required: true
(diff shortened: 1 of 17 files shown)


source/_lovelace/entities.markdown
```diff
@@ -12,41 +12,35 @@ footer: true
...
+  description: "A list of entity IDs or an `entity` object."
   type: list
+  keys:
+    entity:
+      required: true
+      description: "An `entity_id` to use in the frontend."
+      type: string
+    name:
+      required: true
+      description: "A new name for the `entity_id`."
```
c727: Not sure if you want to fix this in this PR, but keys like `name` or `tap_action` are optional.

Also the description text for `name` is inaccurate (It overwrites friendly_name)

source/_lovelace/entity-filter.markdown
```diff
@@ -19,20 +19,27 @@ This type of card can also be used together with rest of cards that allow multip
...
+      description: The name to use in the front for the entity.
+      type: string
+    entity:
+      required: false
+      description: "The `entity_id` to show in the frontend."
+      type: string
 state_filter:
   required: true
-  description: Array of strings representing states
+  description: Array of strings representing states.
```
c727, line 42: we should only use one word `array` or `list`. afaik `list` is more common in the docs

fabaff: Agree

PR #5756, opened 2018-07-11
Lovelace: correct vertical stack

c727:
**Description:**

correct config, improve text

## Checklist:

### DIFF
--- a/source/_lovelace/vertical-stack.markdown
+++ b/source/_lovelace/vertical-stack.markdown
@@ -10,9 +10,7 @@ sharing: true
 footer: true
 ---
 
-Stack card will allow you to stack together multiple cards so they always sit together in the same column one on top of the other. Keep in mind this can be used with any cards and even used alongside a [horizontal-stack](/lovelace/horizontal-stack/).
-
-## {% linkable_title Options %}
+Vertical stack allows you to group multiple cards to use the same column.
 
 {% configuration %}
 type:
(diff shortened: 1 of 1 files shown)


source/_lovelace/vertical-stack.markdown
```diff
@@ -10,9 +10,7 @@ sharing: true
 footer: true
 ---
 
-Stack card will allow you to stack together multiple cards so they always sit together in the same column one on top of the other. Keep in mind this can be used with any cards and even used alongside a [horizontal-stack](/lovelace/horizontal-stack/).
-
-## {% linkable_title Options %}
+Vertical stack allows you to group multiple cards to use the same column.
```
DubhAd: Reads a little stilted, Maybe *Vertical stack allows you to group multiple cards so they always sit in the same column*?

source/_lovelace/vertical-stack.markdown
```diff
@@ -21,24 +19,11 @@ type:
...
-      description: The type of the card to stack.
-      type: string
-    entity:
-      required: true
-      description: "An `entity_id` to use in the frontend."
-      type: string
-    image:
-      required: true
-      description: The URL to an image.
-      type: string
```
fabaff, line 38: Those keys are still present in the example below. Not documenting them will at some point lead to confusion as they seems to be required.

fabaff (changes requested):

source/_lovelace/vertical-stack.markdown
c727: These keys don't belong to this card. Instead they are part of the sub-cards

fabaff: I guess then that I was the only one that go confused.

fabaff (approved):

PR #5758, opened 2018-07-11
Lovelace: correct glance config

c727:
**Description:**

correct glance config

## Checklist:

### DIFF
--- a/source/_lovelace/glance.markdown
+++ b/source/_lovelace/glance.markdown
@@ -24,11 +24,26 @@ type:
   type: string
 entities:
   required: true
-  description: "A list of entity IDs or a `entity` objects."
+  description: "A list of entity IDs or `entity` objects, see below."
   type: list
 title:
   required: false
   description: Card title
+  type: string
+{% endconfiguration %}
+
+## {% linkable_title Options For Entities %}
+
+If you define entities as objects instead of strings, you can add more customization and configuration:
+
+{% configuration %}
+entity:
+  required: true
+  description: Home Assistant entity ID.
+  type: string
+name:
+  required: false
+  description: Overwrites friendly name.
   type: string
 {% endconfiguration %}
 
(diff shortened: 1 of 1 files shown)


source/_lovelace/glance.markdown
```diff
@@ -24,22 +24,12 @@ type:
   type: string
 entities:
   required: true
-  description: "A list of entity IDs or an `entity` object."
+  description: "A list of entity IDs or a `entity` objects."
```
DubhAd: *a `entity` objects* doesn't make sense, maybe *`entity` objects*

source/_lovelace/glance.markdown
```diff
@@ -24,22 +24,12 @@ type:
...
   type: list
-  keys:
-    entity:
-      required: true
-      description: "The `entity_id` to show."
-      type: string
-    name:
-      required: true
-      description: "A name for `the entity_id`."
-      type: string
```
fabaff, line 37: According to sample below those are valid keys. From my point of view those needs to be documented.

fabaff (dismissed):

source/_lovelace/glance.markdown
c727: The keys there are optional. Yep, I will add a new setion for this

source/_lovelace/glance.markdown
```diff
@@ -24,22 +24,27 @@ type:
...
+
+{% configuration %}
+entity:
+  required: true
+  description: Home Assistant entity ID.
+  type: string
+name:
+  required: false
+  description: Overwrites friendly name.
+  type: string
```
DubhAd, line 47: In the entities card docs it says that `type` is an entity option, but it's not listed here. Is an entity different in different cards?

c727: Yes, it's different

DubhAd (approved):
👍

PR #5785, opened 2018-07-13
Updating formatting and layout

DubhAd:
The enable section blends into the ignore section, so moving to the new formatting config. Also turned the mass of notes at the end into a troubleshooting section with linkable titles.


fabaff (approved):
:bird:

PR #5812, opened 2018-07-17
Fix Xiaomi vacuum token documentation

Dmurph24:
**Description:**

Adds more documentation for the Mi Robot (Roborock) vacuum on how to get the token. It is not clear that the iPhone app still works and the new documentation makes that clear and provides the specific version.


fabaff (approved):
Thanks :bird:

PR #5821, opened 2018-07-18
Update prometheus documentation

siretart:
Clarify that the "namespace" option is available in version 0.73 and later. Earlier version will error out.

**Description:**

This is a small clarification in the docs.


fabaff (approved):
Thanks :bird:

PR #5850, opened 2018-07-22
Added doc for icons

DubhAd:
As it now supports icons, adding that to the docs

Icon description lifted from customizing page

### DIFF
--- a/source/_lovelace/glance.markdown
+++ b/source/_lovelace/glance.markdown
@@ -60,6 +60,10 @@ name:
   required: false
   description: Overwrites friendly name.
   type: string
+icon:
+  required: false
+  description: Overwrites the entity's icon.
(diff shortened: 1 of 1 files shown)


Vasiley (approved):
:+1: as always

c727:
not correct, revert it

#5839

c727:
also, do reviewers once check if the syntax is correct and the page renders as expected?

DubhAd:
Rather than "not correct, revert it" - how about some **useful detail** on why it's not correct or basically anything to help know what's gone wrong?

Also, within the GitHub UI, config blocks never render as expected.

balloob:
Netlify build links to a copy of the live site with your changes

c727:
@DubhAd are you kidding? you submit configs that are not implemented (so you havn't even tested it) and ask what's wrong?

c727:
... and then people ask why the docs review process is a joke cc: @frenck

PR #5857, opened 2018-07-23
Enable missing octopress-filters gem

### LINKED ISSUE #5856: Include full URLs in atom xml feeds
The current xml feed uses relative URLs:

```html
<a href="/components/#version/0.74">
<img src="/images/blog/2018-07-0.74/components.png" style="border: 0;box-shadow: none;" />
</a>
```

By the time the feed is displayed to a human, the human's feed downloader/reader pipeline might have lost track of what the base url should be. The feed might be displayed in a web browser on another base url entirely.

Would it please be possible to include full URLs in the atom xml feed? Like so:

```html
<a href="https://www.home-assistant.io/components/#version/0.74">
<img src="https://www.home-assistant.io/images/blog/2018-07-0.74/components.png" style="border: 0;box-shadow: none;" />
</a>
```

https://github.com/home-assistant/home-assistant.io/blob/next/source/atom.xml already pipes through [octopress-filters' `full_urls`](https://github.com/octopress/filters) but the right gem isn't included in the Gemfile.

grischard:
**Description:**

Fixes #5856 maybe?

The atom.xml template uses the `full_urls` filter which is part of the `octopress-filters` gem, which is missing from the Gemfile.

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ group :jekyll_plugins do
   gem 'jekyll-time-to-read'
   gem 'octopress', '~> 3.0'
   gem 'octopress-include-tag'
+  gem 'octopress-filters'
 end
 
 gem 'sinatra', '~> 1.4.2'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -57,6 +57,11 @@ GEM
       colorator
     octopress-escape-code (2.1.1)
       jekyll (~> 3.0)
+    octopress-filters (1.4.0)
+      jekyll
+      octopress-hooks (~> 2.0)
+      rubypants-unicode
+      titlecase
     octopress-hooks (2.6.2)
       jekyll (>= 2.0)
     octopress-include-tag (1.1.3)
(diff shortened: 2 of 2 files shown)


grischard:
I'm not sure why the CI doesn't find the gem on rubygems, [it's right there](https://rubygems.org/gems/octopress-filters)...

frenck:
Sorry for the long wait.

The issue is that you've simply added it to the Gemfile (guess by the branch name, using the GitHub editor).

This, however, does not update the lock file. Run `bundle install` to update it and add the lock file.

frenck (changes requested):
Please update the lock file.

grischard:
I've updated the lock file.

grischard:
And now it's failing and I have no idea why - I couldn't really find something with `empty` or `nil:NilClass` so I imagine it's something like a missing argument?

grischard:
@frenck Help! Any ideas?

frenck:
Assigned myself an reminder to take a look.

frenck:
Looking into this one right now. I wonder if the gem is needed to achieve this. Will test this locally.

frenck:
@grischard I've been able to reproduce the issue locally.
It seems to be related to a plugin this website uses:

https://github.com/home-assistant/home-assistant.io/blob/fef1e594d5d2089f1987cc48f0fce08da8edb2cd/plugins/category_generator.rb#L98

Nevertheless, I was unable to get the wished output with your offered solution.

frenck:
Closing this issue because of the above reasons.
Feel free to open a PR with another solution.

PR #5888, opened 2018-07-25
Added details of how to use update

DubhAd:
Added details of how to use the update service.


turbokongen:
Maybe also mention that everything including and between the start tag `<Node id="?">` and `</Node>` is what need to be deleted. Maybe also note that a savecopy May be a good idea of the zwcfg file before editing.

frenck (approved):

frenck:
@turbokongen Maybe add your suggestion into a separate PR? I'll go ahead and merge this one in for now.

PR #5909, opened 2018-07-28
Added SD card size guide

DubhAd:


Vasiley (changes requested):
Please Capitalize We reccomend

DubhAd:
That would result in two short sentences, rather than a continuation of the statement, let me reword it then.

Vasiley (approved):
Looks Good :+1:

PR #5911, opened 2018-07-28
Warning that these are invalid now

DubhAd:
Apparently the add-on has been updated, but the docs haven't. Until that's addressed by somebody who knows it, I've added this warning ;)


frenck:
I personally do not agree on adding these type of PR, instead, we should fix it.
Nevertheless, I've seen a couple of questions being raised over the last period, so this lets add it for now.

frenck (approved):

PR #5944, opened 2018-08-03
specify security concerns and actions to mitigate them.

finity69x2:
With the focus on security concerns in the forums recently I think it would benefit users to be more aggressive in recommending actions to take to secure home assistant when exposing it to the internet. Vague warnings don't always catch the eye of the casual user. I was one of those users and the forum threads on security caused me to examine my own security and I then realized how potentially lacking that it was.

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/configuration/securing.markdown
+++ b/source/_docs/configuration/securing.markdown
@@ -20,7 +20,7 @@ One major advantage of Home Assistant is that it's not dependent on cloud servic
 - Don't run Home Assistant as root – consider the Principle of Least Privilege.
 - Keep your [secrets](/topics/secrets/) safe.
 
-If you want to allow remote access, consider taking the additional steps as listed below. They are listed from the most secure to the least secure. <b>It is highly recommended that if you are going to be opening your Home Assistant (hence, the control of your home) to the outside world that you should secure it with a VPN at the minimum</b>:
+If you want to allow remote access, consider taking the additional steps as listed below. They are generally listed with the more secure options first. <b>It is highly recommended that if you are going to be opening your Home Assistant (hence, the control of your home) to the outside world and you don't need remote access to the API (for example, for a device tracker) that you should secure it using one of the first three options</b>:
 
 - Protect your communication with [Tor](/cookbook/tor_configuration/).
 - Set up a VPN
--- a/source/_docs/configuration/securing.markdown
+++ b/source/_docs/configuration/securing.markdown
@@ -20,16 +20,16 @@ One major advantage of Home Assistant is that it's not dependent on cloud servic
 - Don't run Home Assistant as root – consider the Principle of Least Privilege.
 - Keep your [secrets](/topics/secrets/) safe.
 
-If you want to allow remote access, consider these additional points:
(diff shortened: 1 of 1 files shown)


source/_docs/configuration/securing.markdown
```diff
@@ -20,16 +20,16 @@ One major advantage of Home Assistant is that it's not dependent on cloud servic
 - Don't run Home Assistant as root – consider the Principle of Least Privilege.
 - Keep your [secrets](/topics/secrets/) safe.
 
-If you want to allow remote access, consider these additional points:
+If you want to allow remote access, consider taking the additional steps as listed below. They are listed from the most secure to the least secure. IT IS HIGHLY RECOMMENDED THAT IF YOU ARE GOING TO BE OPENING YOUR HOME ASSISTANT (HENCE THE CONTROL OF YOUR HOME) TO THE OUTSIDE WORLD THAT YOU SHOULD SECURE IT WITH A VPN AT THE MINIMUM:
```
frenck: We don't have to shout.
Secondly, VPN is not the golden ticket, there are more possibilities.

frenck (changes requested):
We should not shout in our docs...

EVER 😉 

So I'm not approving this in the current state.

finity69x2:
Of course. I always forget the "all caps is shouting" standard. I've corrected it with bolding. I hope that's acceptable.

As for the VPN recommendation - I realize there may be other more secure routes to take (which is why I made it #3 on the list) but it should be the minimum to feel fairly secure. And if there are other better options not on the list they should probably be added. I don't have the experience to know what they are or I would add them here with my edit.

source/_docs/configuration/securing.markdown
```diff
@@ -20,16 +20,16 @@ One major advantage of Home Assistant is that it's not dependent on cloud servic
 - Don't run Home Assistant as root – consider the Principle of Least Privilege.
 - Keep your [secrets](/topics/secrets/) safe.
 
-If you want to allow remote access, consider these additional points:
+If you want to allow remote access, consider taking the additional steps as listed below. They are listed from the most secure to the least secure. <b>It is highly recommended that if you are going to be opening your Home Assistant (hence, the control of your home) to the outside world that you should secure it with a VPN at the minimum</b>:
```
DubhAd: Hmmm, I'm having difficulty with the "Tor is more secure than a VPN or an SSH tunnel" statement. If we're going to have statements like that lets have some evidence ;)

finity69x2: I'm no internet security expert but in my investigating the different options it just seemed as the Tor option was a little more secure. I may be (and probably am...) wrong. 

I was just simply trying to impress upon new users the importance of securing their installation with something more substantial than a simple password if they want to access it from the outside world . And imparting the understanding that using Letsencrypt or a reverse proxy isn't much better since they are, ultimately, still just protected by a simple password.

If you have a recommendation on how to get that point across for new people who aren't internet security experts who want to be able to safely access their HA from the outside (and I think most people actually want to do that since that's the impression given from looking at the very first page on the HA website...), please feel free to share.

I'm pretty confident in saying that Tor, SSH tunnel or a VPN are all likely way more secure than a password, letsencrypt or a simple reverse proxy. We can quibble about the details but in the end  letting people know of more safe options is what's important.

DubhAd: I'd agree that Tor/SSH/VPN are likely more secure, but ranking those...

Maybe more a *we would suggest that if you don't need remote access to the API (for example, for a device tracker) you should use one of the first three options*

finity69x2:
@DubhAd 

I made a change as recommended. Hopefully it works for you.

DubhAd (approved):
Thanks

frenck:
The documentation has been revamped. This PR is now a use merge conflict.
Some elements of this PR are integrated as well.

Closing this one.

PR #5947, opened 2018-08-03
Add documentation for new RMV public transport sensor

cgtobi:
**Description:**
This adds documentation for a new public transport sensor for the Rhein-Main area. It is based upon PyRMVtransport which uses public data from opendata.rmv.de.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#15814

## Checklist:


source/_components/sensor.rmvtransport.markdown
```diff
@@ -0,0 +1,73 @@
...
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: rmvtransport
+    nextdeparture:
+     - station: STATION_OR_STOP_ID
+```
+
+Configuration variables:
```
MartinHjelmare: Please use our configuration tag syntax instead to describe the config options.

https://developers.home-assistant.io/docs/en/documentation_create_page.html#configuration

MartinHjelmare:
Looks like the build failed on an unrelated page.

source/_components/sensor.rmvtransport.markdown
cgtobi: Ok. Thanks for the info. I updated the config options.

source/_components/sensor.rmvtransport.markdown
```diff
@@ -0,0 +1,98 @@
...
+
+## {% linkable_title Configuration %}
+
+To enable this sensor, add the following lines to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: rmvtransport
+    nextdeparture:
```
DavidMStraub: Isn't this called  `next_departure` in the merged version? (Which by the way is not consistent with the MVG sensor, where we have `nextdeparture` and `timeoffset`. Not sure if this should be handled consistently.)

cgtobi: Thank's for pointing it out. I agree with your remark about consistency, but it actually is PEP 8 compliant. So it would rather make sense to change the names of the MVG sensor, no?

frenck:
This needs to be rebased in order to fix the build. Will take care of that.

frenck:
Rebased, awaiting build results before starting a review.

frenck (approved):

frenck:
Looks good to go! 🎉

PR #5970, opened 2018-08-07
Adding user details

DubhAd:
Existing instructions are generic `user:group` which causes confusion. Adding a line to create a `homeassistant` user, and then adjusting the following line to use that user.


fabaff (approved):
:bird:

PR #5971, opened 2018-08-07
Fixing formatting goof

DubhAd:
The two spaces after the list caused the description of the calendars to appear to be linked to the `ignore_availability` setting


fabaff (approved):
Thanks :bird:

PR #5789, opened 2018-07-14
Add Facebox auth

robmarkcole:
**Description:**
Adds Facebox auth and removes teach events

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant/pull/15439) (if applicable):** home-assistant/home-assistant#15439

## Checklist:

- [x] Branch: Fixes, changes and adjustments should be created against `current`. New documentation for platforms/components and features should go to `next`.
- [x] The documentation follow the [standards][standards].

[standards]: https://home-assistant.io/developers/documentation/standards/

PR #5975, opened 2018-08-08
Add info on teaching automations

robmarkcole:
**Description:**
Add info on teaching automations, since the teaching events were removed

## Checklist:

### DIFF
--- a/source/_components/image_processing.facebox.markdown
+++ b/source/_components/image_processing.facebox.markdown
@@ -118,6 +118,56 @@ A valid service data example:
 ```
 {% endraw %}
 
+You can use an automation to receive a notification when you train a face:
+
+{% raw %}
+```yaml
+- id: '1533703568569'
+  alias: Face taught
+  trigger:
+  - event_data:
+      service: facebox_teach_face
+    event_type: call_service
+    platform: event
+  condition: []
+  action:
+  - service: notify.pushbullet
+    data_template:
+      message: '{{ trigger.event.data.service_data.name }} taught 
+      with file {{ trigger.event.data.service_data.file_path }}'
+      title: Face taught notification
+```
+{% endraw %}
+
+Any errors on teaching will be reported in the logs. If you enable [system_log](https://www.home-assistant.io/components/system_log/) events:
+
+{% raw %}
+```yaml
+system_log:
+  fire_event: true
+```
+{% endraw %}
+
+you can create an automation to receive notifications on Facebox errors:
+
+{% raw %}
+```yaml
+- id: '1533703568577'
+  alias: Facebox error
+  trigger:
+    platform: event
+    event_type: system_log_event
+  condition:
+    condition: template
+    value_template: '{{ "facebox" in trigger.event.data.message }}'
+  action:
+  - service: notify.pushbullet
+    data_template:
+      message: '{{ trigger.event.data.message }}'
+      title: Facebox error
+```
+{% endraw %}
+
 ## {% linkable_title Optimising resources %}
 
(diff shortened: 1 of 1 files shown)


source/_components/image_processing.facebox.markdown
```diff
@@ -118,6 +118,56 @@ A valid service data example:
...
+    data_template:
+      message: '{{ trigger.event.data.service_data.name }} taught 
+      with file {{ trigger.event.data.service_data.file_path }}'
+      title: Face taught notification
+```
+{% endraw %}
+
+Any errors on teaching will be reported in the logs. If you enable [system_log](https://www.home-assistant.io/components/system_log/) events:
+
+{% raw %}
```
MartinHjelmare: Only templates require the raw tags.

tjorim: Could you please expand on that a bit more? When to use, when not etc.
This is pretty unclear in the ['documentation documentation'](https://developers.home-assistant.io/docs/en/documentation_create_page.html) in my opinion.

e.g.

> ## Configuration
> The Configuration Variables section must use the {% raw %}{% configuration %} ... {% endconfiguration %}{% endraw %} tag.

The '... tag', I see multiple tags? Should we use both those tags or is it in this case just poorly rendered and do we only use the {% configuration %}-tag?

> ## Embedding Code
> You can use the default markdown syntax to generate syntax highlighted code. For inline code wrap your code in {% raw %}`{% endraw %}.

Are those `{% raw %}`-tags something we should include? Is this again a case of bad rendering and does the documentation actually just mean to use ` for inline code?

robmarkcole: @tjorim my reading of it is that only the contents of the yaml (i.e. configuration) files should be enclosed in the `raw`, so my example service data should NOT be enclosed. Is this correct @MartinHjelmare ?

MartinHjelmare: We don't want jekyll to try to process the curly braces in jinja templates. So we use the raw tag to temporarily disable tag processing. Read more here eg:
http://ozzieliu.com/2016/04/26/writing-liquid-template-in-markdown-with-jekyll/

So to answer the question, only literal yaml that contains templates, ie curly braces, needs to be contained within raw tags. But of course there are other cases where we need to disable tag processing. Trying to list that is out of scope for this review.

MartinHjelmare (approved):

frenck:
Merged a little too quickly.. there was a spelling error.

source/_components/image_processing.facebox.markdown
```diff
@@ -118,6 +118,54 @@ A valid service data example:
...
+{% endraw %}
+
+Any errors on teaching will be reported in the logs. If you enable [system_log](https://www.home-assistant.io/components/system_log/) events:
+
+```yaml
+system_log:
+  fire_event: true
+```
+
+you can create an automation to receive notifications on Facebox errors:
```
frenck, line 149: `you` -> `You`

frenck:
Furthermore, this is not related to a parent PR, so it targets an incorrect branch. It should have been the `current` branch.

MartinHjelmare:
No it is related to a parent PR. It was just not linked.

frenck:
@MartinHjelmare In that case, it should have been linked before merging.

Could @robmarkcole or you add the link please?

MartinHjelmare:
This was a follow up of https://github.com/home-assistant/home-assistant.io/pull/5789.

frenck:
Thanks, @MartinHjelmare! 👍 

Parent PR: home-assistant/home-assistant#15439

PR #5990, opened 2018-08-11
Minor changes

DubhAd:
Taking out the "new and experimental" note - we're a dozen releases on, it's hardly new and seems to be (from a feature sense) stable.

Re-written para 2 to reflect that HA needs shut down, and that 0.75 introduced the "future" UI. Re-written the supported components list because it's horribly out of date, and without trawling the code it's impossible to know what's supported.


fabaff (approved):
Thanks :bird:

PR #5994, opened 2018-08-11
Making the warning clearer, and linkable

DubhAd:
Making the warning clearer (I'll add something similar to the template trigger docs), and putting a title in so that we can link to it.


fabaff (approved):
:bird:

PR #5995, opened 2018-08-11
Adding warning here about the use of now()

DubhAd:
Adding the warning here, rather than just lurking in the templating docs.


fabaff (approved):
:bird:

PR #5996, opened 2018-08-12
Making it clearer what column_width does

DubhAd:
Added more words to make it clearer what it does, and how to use a value to get the number of entities you want in a row


swetoast (approved):

fabaff (approved):
:bird:

PR #5997, opened 2018-08-12
Added link to the FAQ at the top

DubhAd:
Seems to be a common problem with people assuming that random browser X will "just work", and the FAQ is hidden away.


fabaff (approved):
:bird:

PR #6000, opened 2018-08-12
Adding details of merging trackers

DubhAd:
Since the merging of trackers works for more than just OwnTracks (at least GPS Logger, probably more), adding the details here at the top level.


fabaff (approved):
Thanks :bird:

PR #6002, opened 2018-08-12
Added undocument possibility for scan_interval

endor-force:
**Description:**
scan_interval is available to override in existing code. Only documentation update.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant - N/A

## Checklist:


endor-force:
It is actually documented in general for entities, but i think it will save some people a bit of time to be aware of the possibility directly when reading up on the component settings..

fabaff:
It's indeed already documented at [Entity component platform options](https://www.home-assistant.io/docs/configuration/platform_options/#scan-interval). We decided not to add it because it's an advanced feature and in 95 % of the use cases not needed.

PR #6008, opened 2018-08-13
Reordering config options

DubhAd:
A few folks have missed the `trusted_proxies` setting, and wondered why access without a password is permitted if the proxy is in the trusted networks. Moved it before the XFF setting, added a few words and some emphasis, and took the time to rework it into the new config layout.

### DIFF
--- a/source/_components/http.markdown
+++ b/source/_components/http.markdown
@@ -29,19 +29,65 @@ http:
 
 Configuration variables:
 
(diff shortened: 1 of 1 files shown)


frenck:
@DubhAd The build is failing, could you please take a look? 👍

frenck (approved):

PR #6032, opened 2018-08-17
List service homematic.put_paramset in documentation

klada:
**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#16024

## Checklist:


danielperna84:
It would be nice if you could also add an example [here](https://www.home-assistant.io/components/homematic/#advanced-examples).

klada:
@danielperna84 Should this really go on the main page? I could also add a sample [in the climate documentation](https://www.home-assistant.io/components/climate.homematic/) and call it "special note about wall thermostat (HM-TC-IT-WM-W-EU)".

danielperna84:
Yes. All the other examples use some kind of device too, and none of the platform-documentation have content besides linking to the component-documentation. Besides that, the service itself isn't specific to thermostats. Thermostats are just one of possibly many other device types where the service makes sense. And if you have an actor you want to use the service on it's unintuitive to search for a service-example in the thermostat-documentation just because the example happens to be thermostat-specific.

klada:
OK, docs are updated with a complete example.

frenck (approved):

frenck:
Thanks, @klada! 👍

PR #6042, opened 2018-08-18
Updated template condition

DubhAd:
Updated the template condition to use `state_attr` instead of the more error prone long form.


fabaff (approved):
:bird:

PR #6046, opened 2018-08-19
Update sensor.mold_indicator.markdown

tomlut:
Added optional 'name' configuration variable. I have confirmed this is supported but undocumented.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/sensor.mold_indicator.markdown
+++ b/source/_components/sensor.mold_indicator.markdown
@@ -21,6 +21,7 @@ The sensor data may be used e.g., to signal bad air quality (too high air humidi
 # Example configuration.yaml entry
 sensor:
   - platform: mold_indicator
+    name: Mould Sensor
     indoor_temp_sensor: sensor.temp
     indoor_humidity_sensor: sensor.humidity
     outdoor_temp_sensor: sensor.weather_temperature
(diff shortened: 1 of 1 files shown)


source/_components/sensor.mold_indicator.markdown
```diff
@@ -21,6 +21,7 @@ The sensor data may be used e.g., to signal bad air quality (too high air humidi
 # Example configuration.yaml entry
 sensor:
   - platform: mold_indicator
+    name: Mould Sensor
```
tjorim: Please remove this line from the example, as per [the documentation](https://developers.home-assistant.io/docs/en/documentation_create_page.html#configuration):

> This sample must contain only the required variables to make it easy to copy and paste it for users into their configuration.yaml file.

tjorim (changes requested):
Thanks for your contribution! Please consider removing the `name:`-key from the example configuration.yaml entry as explained in the comment below.

While updating this documentation anyway, would you mind changing the `Configuration variables:`-section to use the new {% configuration %} tags as explained [here](https://developers.home-assistant.io/docs/en/documentation_create_page.html#configuration)? Thanks for helping out.

tomlut:
Done. Though looking back I think I made a mistake. The calibration_factor variable type is a float not an int. Though I don't see this as an option in the Create a new page documentation. Should it be a string instead of an int?

tjorim:
Not all options are shown in the documentation (like boolean), as far as I know it can actually be anything. Feel free to use float instead of int, if it's wrong another dev will probably chime in.
Or ask in the Discord Development #documentation channel if you want a quick answer.

tomlut:
Thanks, I changed it to a float.

fabaff (approved):
Thanks :bird:

PR #6054, opened 2018-08-21
Fix misleading numeric_state trigger documentation

pnbruckner:
**Description:**

Existing documentation says numeric_state trigger will fire once when the numeric value of an entity's state "crosses a threshold." However, that's not how the code actually works. It will fire for any state change (from any of the specified entities) if the numeric value of the new state satisfies the above and/or below conditions.

## Checklist:

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -59,7 +59,7 @@ automation:
 
 ### {% linkable_title Numeric state trigger %}
 
-Triggers when numeric value of an entity's state crosses a given threshold. On state change of a specified entity, attempts to parse the state as a number and triggers once if value is changing from above to below or from below to above the given threshold.
+Triggers when numeric value of an entity's state satisfies given condition(s.) On state change of a specified entity, attempts to parse the state as a number and triggers if value is above and/or below given threshold(s.)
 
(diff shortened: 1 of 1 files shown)


pnbruckner:
I seem to be wrong. I'll close this PR until I'm sure. If I'm right I'll reopen. Otherwise I'll leave it closed.

pnbruckner:
FWIW, I did look at the code again, and indeed I was wrong. I had originally misunderstood how it worked. I also did some more extensive experimenting to convince myself. But, having said that, there is some behavior that isn't well documented. First, after a restart, if the entity's state satisfies the above and/or below conditions, it doesn't fire. This seems to trip some people up, even though it shouldn't, because with only one state value it can't have "crossed a threshold." But, there is another behavior which is probably unexpected to most. That is, if the first entity state change after restart satisfies the above and/or below conditions, the trigger _will_ fire, even if it _doesn't cross a threshold!_ To be clear, if the state satisfies the conditions after restart, it won't immediately fire, but if it changes to a new state value that _also_ satisfies the conditions, the automation _will_ trigger (again, even though it didn't cross a threshold.) It might be worth describing this behavior. Actually, that behavior should probably get "fixed", but that might break existing automations.

PR #6060, opened 2018-08-21
Remove RPI variants

pvizeli:
**Description:**

First, we support now all RPi's and second we support also more as only RPi.


fabaff (approved):
:bird:

PR #6068, opened 2018-08-23
Add auth docs

balloob:
**Description:**
Add documentation for the auth in Home Assistant. Generally speaking, people should not configure this, the defaults are good and secure. 

## Checklist:


source/_docs/authentication/providers.markdown
```diff
@@ -0,0 +1,86 @@
...
+   - type: legacy_api_password
+```
+
+## {% linkable_title Available auth providers %}
+
+Below is a list of currently available auth providers.
+
+### {% linkable_title Home Assistant auth provider %}
+
+This is the default auth provider which stores the users in your configuration directory. All passwords are stored hashed and with a salt, making it impossible for an attacker to figure out the password from the storage.
```
frenck: From a security perspective, I would say: `impossible for an attacker` is kinda fraud upon.
Are you willing to make that promise? I personally would advise being more careful.

Impossible? Not sure, "improbable" or "unlikely" might be a better fit.

frenck (dismissed):

PR #6072, opened 2018-08-24
Added some details

DubhAd:
Making it a little clearer that `40` is a numeric, also making it clearer that it's an *entity attribute* that's being checked


fabaff (approved):
:bird:

PR #6080, opened 2018-08-25
Rewording

DubhAd:
As *Rename Node* no longer exists, some of this makes no sense. Reworked to make more sense.


fabaff (approved):
Thanks :bird:

PR #6116, opened 2018-08-30
Removed automation: line

DubhAd:
One of the examples has two `automation:` lines, which clearly doesn't work. Removed after somebody had issues with it


fabaff (approved):
Thanks :bird:

PR #6140, opened 2018-09-01
Added details about console access

DubhAd:
Since there's nothing in the docs about the console access, and it's a common question, adding it here.


fabaff (approved):
Thanks :bird:

PR #6143, opened 2018-09-02
Update mailgun.markdown

plyblu:
Fix inconsistency where "domain" is required but "sandbox" can also be used to specify the domain.  It appears that the package https://github.com/pschmitt/pymailgunner in  def guess_domain allows multiple domains and "sandbox" can be used to choose a sandbox domain over a normal domain.  As far as I can tell, this feature isn't used by the Mailgun component.  There are also updates to the Mailgun Notify doc.

The "domain" also should not talk about a default if specifying one is required.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/mailgun.markdown
+++ b/source/_components/mailgun.markdown
@@ -12,7 +12,7 @@ ha_category: Notifications
 ha_release: 0.38
 ---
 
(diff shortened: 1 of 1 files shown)


fabaff (approved):
Thanks :bird:

PR #6154, opened 2018-09-02
Update tesla.markdown

ottersen:
updated to new category Car

**Description:**

## Checklist:

### DIFF
--- a/source/_components/tesla.markdown
+++ b/source/_components/tesla.markdown
@@ -8,7 +8,7 @@ comments: false
 sharing: true
 footer: true
 logo: tesla.png
-ha_category: Hub
+ha_category: Car
 ha_release: 0.53
 ha_iot_class: "Cloud Polling"
 ---


source/_components/tesla.markdown
```diff
@@ -8,7 +8,7 @@ comments: false
 sharing: true
 footer: true
 logo: tesla.png
-ha_category: Hub
+ha_category: Car
```
fabaff, line 11: Instead of the component should the platforms be moved to the categories.

ottersen: Please help me understand hw this should be done correctly. How I understand it, the idea is that it should be easy to find a component based on category. 

Categorizing everything as sensor or hub might be correct as per implementation, but this will make it more difficult to find an item in a certain category. Ideally it should be possible to have several tags, but that might be a longer step to implement.

ottersen: Added car category as per comment in discord #documentation

fabaff: Ok, let's do it the same way as for `DIY`.

fabaff (approved):

PR #6161, opened 2018-09-02
Updated to reflect changes to the internal MQTT broker in 0.76

brubaked:
See mqtt changes specified here - https://www.home-assistant.io/blog/2018/08/17/release-76/

**Description:**

## Checklist:

### DIFF
--- a/source/_components/mqtt.markdown
+++ b/source/_components/mqtt.markdown
@@ -23,6 +23,7 @@ To integrate MQTT into Home Assistant, add the following section to your `config
 ```yaml
 # Example configuration.yaml entry
 mqtt:
+  password: mypass
 ```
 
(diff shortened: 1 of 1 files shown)


source/_components/mqtt.markdown
```diff
@@ -23,6 +23,7 @@ To integrate MQTT into Home Assistant, add the following section to your `config
 ```yaml
 mqtt:
+  password: mypass
```
fabaff, line 26: This is already [covered](https://www.home-assistant.io/docs/mqtt/broker#embedded-broker). We keep the configuration sample minimal and using a password is optional.

brubaked: It doesn't work if you don't add a password though. It throws an error without the "password: mypass" entry.

PR #6163, opened 2018-09-02
Update Miflora documentation

PaulAnnekov:
**Description:**

Updated according change in PR below.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#16085

## Checklist:

### DIFF
--- a/source/_components/sensor.miflora.markdown
+++ b/source/_components/sensor.miflora.markdown
@@ -71,10 +71,9 @@ sensor:
 - **force_update** (*Optional*): Sends update events even if the value hasn't changed.
 - **median** (*Optional*): Sometimes the sensor measurements show spikes. Using this parameter, the poller will report the median of the last 3 (you can also use larger values) measurements. This filters out single spikes. Median: 5 will also filter double spikes. If you never have problems with spikes, `median: 1` will work fine.
 - **adapter** (*Optional*): Define the Bluetooth adapter to use (defaults to hci0). Run `hciconfig` to get a list of available adapters.
-- **scan_interval** (*Optional*): Define the update interval of the sensor in seconds (defaults to 1200 if not defined).
 
 <p class='note warning'>
-By default the sensor is only polled once every 20 minutes. On a Home Assistant restart sensor will report initial value. If you set `median: 3`, it will take _at least_ 40 minutes before the sensor will report an average value. Keep in mind though that reducing polling intervals will have a negative effect on the battery life.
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
Please change the base branch to `next`.

PaulAnnekov:
@MartinHjelmare changed. Why do you need it? I see no contradiction with checklist: "Branch: Fixes, changes and adjustments should be created against current. New documentation for platforms/components and features should go to next.". If `current` means "changes related to current version" and `next` means "changes that should go into next HA release", then checklist explanation is misleading and should be rewritten.

MartinHjelmare:
PR for template improvement is welcome.

source/_components/sensor.miflora.markdown
```diff
@@ -70,13 +70,11 @@ sensor:
 - **name** (*Optional*): The name displayed in the frontend.
 - **force_update** (*Optional*): Sends update events even if the value hasn't changed.
 - **median** (*Optional*): Sometimes the sensor measurements show spikes. Using this parameter, the poller will report the median of the last 3 (you can also use larger values) measurements. This filters out single spikes. Median: 5 will also filter double spikes. If you never have problems with spikes, `median: 1` will work fine.
-- **timeout** (*Optional*): Define the timeout value in seconds when polling (defaults to 10 if not defined)
-- **retries** (*Optional*): Define the number of retries when polling (defaults to 2 if not defined)
-- **cache_value** (*Optional*): Define cache expiration value in seconds (defaults to 1200 if not defined)
 - **adapter** (*Optional*): Define the Bluetooth adapter to use (defaults to hci0). Run `hciconfig` to get a list of available adapters.
+- **scan_interval** (*Optional*): Define the update interval of the sensor in seconds (defaults to 1200 if not defined).
```
MartinHjelmare: We usually don't document this for platforms as it is a general option.

PaulAnnekov: But we have custom interval (20 minutes, not 10 seconds) for this component and I think there will be more guys that will want to change it, than regular.

MartinHjelmare: @fabaff ?

fabaff: `scan_interval` is an advance option which should be set to a reasonable value in the platform already and only be changed by the users if they run into trouble. Using a different interval than the default 30 s is quite common. If we add this to all platforms that are doing polling people start messing with the intervals and then run into performance issues, are hammering third-party APIs or wondering why their 5 MiFlora sensors are not working when the `scan_interval` is 10 s.

The interval is already mentioned in the warning. To make it more clear replace "polling intervals" with `scan_interval` there.

PaulAnnekov: > The interval is already mentioned in the warning. To make it more clear replace "polling intervals" with scan_interval there.

Agree. Added.

MartinHjelmare (approved):

PR #6169, opened 2018-09-02
Improve warning for yi-hack-v3 versions

jacobtomlinson:
Today I followed the Yi Home Camera instructions without fully reading the warning about supported versions and accidentally flashed a version which is higher than the currently supported version. 

Rather than downgrading my firmware I found [a workaround](https://github.com/shadow-1/yi-hack-v3/issues/129#issuecomment-361723075) in the `yi-hack-v3` issues. Thanks to this workaround I have successfully set up my camera using the latest firmware.

I thought I would update the docs here to let people know they can also successfully follow this workaround, but I've left it stating that the older version is the highest supported without additional work.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #6188, opened 2018-09-03
For the updates to the customize docs

DubhAd:
For the updates to the customize docs, to explain how to change an entity ID


fabaff (approved):
:bird:

PR #6189, opened 2018-09-03
Reworking the customize docs

DubhAd:
Reworking the customize docs:

1. Add information on using the UI to change the entity ID and name (where supported)
2. Document the use of the UI to customize
3. Rewrite the configuration block to use the new configuration block standard
4. Move the manual editing section towards the end

[This change](https://github.com/home-assistant/home-assistant.io/pull/6188) provides the image referenced


fabaff (approved):
Nice. Thanks :bird:

PR #6196, opened 2018-09-04
Updated troubleshooting with authentication

iguana007:
I was able to setup new account only after I cleared my cookies in browser, otherwise normal login prompt has been displayed.

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/authentication.markdown
+++ b/source/_docs/authentication.markdown
@@ -48,4 +48,4 @@ While you should hopefully be storing your passwords in a password manager, if y
 * `auth_provider.homeassistant`
 * `onboarding`
 
-When you start Home Assistant next you'll be required to set up authentication again.
+When you start Home Assistant next you'll be required to set up authentication again (clear your cookies if not).


awarecan:
This is not right, we are not using cookie. We use Local Storage

iguana007:
But until I deleted cookies I was not able to setup new account after deletion of files in .storage folder, whatever I did, I saw login form instead of new account setup form. When I deleted cookies, new account form showed up.

awarecan:
You need understand different between local storage and cookie, see: https://stackoverflow.com/a/19869560

Some browser may simplified the term on UI, provided functions such as "Clear cache" or "Clear storage", under the hood it will clean both cookies and local storage. But in our document, we should use appropriate terms.

frenck (changes requested):
@iguana007 Please make the changes requested by @awarecan. Thanks already! 👍

iguana007:
@frenck I would love to, but I do not know how, eg. what exactly should be written there to make this pull request pass.

frenck:
@awarecan Could you provide an alternative description that fits this better?

frenck:
Since there is a lack of response on this PR, and because the documentation has been updated in the meanwhile, I'm going to close this PR.

PR #6197, opened 2018-09-04
Trying to direct people away from this

DubhAd:
Given the pain that uninformed people run into with a non-venv install, trying to push people towards the venv installs.


fabaff (approved):
Thanks :bird:

PR #6210, opened 2018-09-06
Clarity updates

DubhAd:
Given that turn_on and off don't work on some, would be handy to mention that ;)


fabaff (approved):
Thanks :bird:

PR #6215, opened 2018-09-07
Note about bearer tokens

DubhAd:
Given that there's zero documentation about those warnings, and it comes up many times a day, time to document *something* about them.


awarecan:
Can you double check? This message suppose to be INFO level if you use api_password access and enabled `http.api_password` at same time. It will show as WARNING, and it is indeed need to be WARNING when you used api_password to access but did not configured `http.api_password`.

frenck (approved):

PR #6219, opened 2018-09-07
Add the documentation for the new base_url option for tts.

lddubeau:
**Description:**

Adds the documentation for the new `base_url` option for the `tts` component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#16478

## Checklist:

### DIFF
--- a/source/_components/tts.markdown
+++ b/source/_components/tts.markdown
@@ -23,7 +23,7 @@ tts:
 ```
 
 <p class='note'>
-Depending on your setup, you might need to set a base URL (`base_url`) inside the [http component](/components/http/).
+Depending on your setup, you might need to set a base URL (`base_url`) inside the [http component](/components/http/) or in the parameters of this component.
 </p>
 
 The following optional parameters can be used with any platform. However, the TTS component will only look for global settings under the configuration of the first configured platform:
--- a/source/_components/tts.markdown
+++ b/source/_components/tts.markdown
@@ -28,13 +28,28 @@ Depending on your setup, you might need to set a base URL (`base_url`) inside th
 
 The following optional parameters can be used with any platform. However, the TTS component will only look for global settings under the configuration of the first configured platform:
 
-| Parameter           | Default | Description                                                                                                                                                                                                                                                                                                                                                                               |
-|---------------------|---------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `cache` | True    | Allow TTS to cache voice file to local storage. |
-| `cache_dir`  | tts      | Folder name or path to a folder for caching files. |
-| `time_memory`     | 300     | Time to hold the voice data inside memory for fast play on a media player. Minimum is 60 s and the maximum 57600 s (16 hours). |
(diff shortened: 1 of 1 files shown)


source/_components/tts.markdown
```diff
@@ -33,6 +33,8 @@ The following optional parameters can be used with any platform. However, the TT
 | `cache` | True    | Allow TTS to cache voice file to local storage. |
 | `cache_dir`  | tts      | Folder name or path to a folder for caching files. |
 | `time_memory`     | 300     | Time to hold the voice data inside memory for fast play on a media player. Minimum is 60 s and the maximum 57600 s (16 hours). |
+| `base_url` | '' | A base URL to use *instead* of the one set in the [http component](/components/http/). It is used as-is by the `tts` component. In particular, you need to include the protocol scheme `http://` or `https://` and the correct port number. They will not be automatically added for you. An empty string (the default) means that `tts` will use the one set in the [http component](/components/http/). |
```
DubhAd: Rather than `''` maybe either leave it blank or specify that it's the one from HTTP?

That said, this would also be a good time to update the configuration block to use the [one in the standards](https://www.home-assistant.io/developers/documentation/standards/)

DubhAd (dismissed):

source/_components/tts.markdown
lddubeau: Yes, given the code changes requested to [code PR](https://github.com/home-assistant/home-assistant/pull/16478), modifying this column is part of the plan.

I'll take a look at the configuration block template. Making the change should be simple enough.

lddubeau:
I've performed the switch to using a the configuration template and that took care of the issue with documenting a default value for `base_url`.

DubhAd (dismissed):
👍

frenck (approved):

frenck:
Thanks, @lddubeau.
I've done some small tweaks as well.

✅ Approved.

source/_components/tts.markdown
```diff
@@ -23,28 +23,66 @@ tts:
...
+    base_url: http://192.168.0.10:8123
 ```
 
+## {% linkable_title When do you need to set `base_url` here? %}
+
+The general answer is "whenever the global `base_url` set in [http component](/components/http/) is not adequate to allow the `say` service to run". The `say` service operates by generating a media file that contains the speech corresponding to the text passed to the service. Then the `say` service sends a message to the media device with a URL pointing to the file. The device fetches the media file at the URL and plays the media. Some combinations of a media device, network configuration and Home Assistant configuration can make it so that the device cannot fetch the media file.
+
+The following sections describe some of the problems encountered with media devices.
+
+### {% linkable_title Self-signed certificates %}
```
pvizeli, line 72: I'm not sure if we should maintain self signed certificate on that part. We prefer LE if the user don't have a certificate. The problem with self-signed certificates is, that is extrem complex to make sure that there is no mitma. So user they work with own certificate need to know what they do

lddubeau: The section is not advocating the use of self-signed certificates. It is telling users that if they do use a self-signed certificate, that may cause problems when they try to use the `say` service with their media device. If anything, I'd expect it would make some folks switch from using self-signed certificates in favor of using Let's Encrypt.

frenck: I do agree with you, @pvizeli, nevertheless, this piece of documentation help users who DO.

`This problem occurs when your Home Assistant instance is configured to be accessed through SSL, and you are using a self-signed certificate.`

There is no recommendation in this.

PR #6226, opened 2018-09-08
Adding WiFi note

DubhAd:
Adding a note that modern smart phones may turn off their WiFi when idle, so this tracker may not be reliable when used by itself.


ludeeus (approved):

fabaff (approved):
:bird:

PR #6228, opened 2018-09-08
Adding some titles and details

DubhAd:
Added:

* titles to make it possible to link to sections
* that you can only reload those sections if you're not using packages
* information on how to migrate your config, and that key files and folders are hidden by default


fabaff (approved):
Nice work, thanks :bird:

PR #6235, opened 2018-09-09
Added the list of supported domains

DubhAd:
Given how often we have to direct people to the Google Assistant component for this list, seems overdue to add it here


fabaff (approved):
Thanks :bird:

PR #6048, opened 2018-08-20
Add platform param

huangyupeng:
Add platform to distinguish accounts from different apps
The value could be 'tuya' or 'smart_life'. Default is 'tuya'.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#16058

## Checklist:

- [x] Branch: Fixes, changes and adjustments should be created against `current`. New documentation for platforms/components and features should go to `next`.
- [x] The documentation follow the [standards][standards].

[standards]: https://home-assistant.io/developers/documentation/standards/

PR #6250, opened 2018-09-10
Update tuya.markdown

DonnyBahama:
Values should be enclosed in single quotes so I updated the example accordingly. Platform required was set to false; it is required so I changed it to true and added it to the example. Only two platform options were given- tuya and smart_life. My devices required the platform to be set to 'jinvoo_smart' so I updated the platforms to include that option. If there are other known options, I suggest they be listed as well.

**Description:**

## Checklist:


fabaff:
The docs update was already merged: https://github.com/home-assistant/home-assistant.io/pull/6048

PR #6258, opened 2018-09-11
Update konnected.markdown

themanieldaniel:
**Description:**
Changed the instructions from use 6 character MAC address to using 12 character since the 6 character abbreviation does not work. Also added that the letters must be lower case since the configuration fails if using upper case letters.

## Checklist:


DubhAd (approved):
👍

PR #6269, opened 2018-09-12
Highlighting the package names

DubhAd:
Highlighting the package names because apparently we forgot to do so before


fabaff (approved):
:bird:

PR #6277, opened 2018-09-13
Update link to Google TTS Supported Languages

vikramgorla:
**Description:**
Looking at the implementation, it uses Google Translate as backend and not Google Cloud Speech engine, updating the link where supported languages can be seen.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #6322, opened 2018-09-20
Adding warning about long standing issue

DubhAd:
Adding warning about long standing issue


cgarwood:
I just did some tests, and state restoration works on my system with mysql (atleast for the components I have: automation, device_tracker, input_boolean, and input_select).

DubhAd:
Will go and run my test system up with Postgres and stuff then, and re-open if needed

PR #6332, opened 2018-09-21
More titles

DubhAd:
Added a couple of titles to make linking into the page easier. Also expanded on why capabilities are useful.


fabaff (approved):
Thanks :bird:

PR #6339, opened 2018-09-23
Clarity about the wakeup interval

DubhAd:
Lots of people seem to mistake the wakeup interval with how often the device reports changes


fabaff (approved):
Thanks :bird:

PR #6351, opened 2018-09-24
Remove stale version warning

emlove:
**Description:**

This feature is now released, so this warning isn't necessary any longer.


fabaff (approved):
Thanks :bird:

PR #6359, opened 2018-09-26
Adding magic

DubhAd:
Turns out you can trick OZW into thinking a device has failed, and allowing you to remove a failed node.


fabaff (approved):
:bird:

PR #6360, opened 2018-09-26
Adding how to use the Remove Failed node service

DubhAd:
Since it's been confirmed to work, adding the details here on how to trick the system


fabaff (approved):
:bird:

PR #6362, opened 2018-09-26
Bigger warning

DubhAd:
At least a couple of times a week somebody turns up with issues, who's using a 0.5A mobile charger, or USB port on some device. Clearly the warning needs to be bigger.


fabaff (approved):
:bird:

Issue #6385, opened 2018-09-30
Use the new style for configuration variables description


fabaff:
A while back an [advance way to describe the configuration variables](https://developers.home-assistant.io/docs/en/documentation_create_page.html#configuration) was introduced. 

```yaml
{% configuration %}
api_key:
  description: The API key to access the service.
  required: true
  type: string
name:
  description: Name to use in the frontend.
  required: false
  default: The default name to use in the frontend.
  type: string
{% endconfiguration %}
```

At the moment not all platform and component documentation is ported to the new style.

This the tracking issue. Please don't close but add as a reference to your Pull request.

WofWca:
Should we open individual PRs for each page changed?

fabaff:
Yes, please or at least don't add 50 platforms to one Pull Request. If you want you can combine the changes for a component and its platforms.

If the Pull Requests are small it easier to spot for others which docs are in progress or already done.

WofWca:
I think that the documentation, in accordance to which these changes must be done, is a bit outdated.
home-assistant/developers.home-assistant#104

fabaff:
@WofWca, Doc were updated. Please feel free to improve them if you think that something is not clear.. Thanks for the feedback, BTW.

WofWca:
Opened a new issue home-assistant/developers.home-assistant#107

datafx:
How would we handle this for something like modbus which has different required configuration variables depending whether you are using tcp or serial communication?  Right now, the component actually has two different configuration variable sections.

tjorim:
I'm currently checking all components. Here's a working list while I'm going through them.

Components I'm currently updating:

- [ ] alarmdecoder.markdown
- [x] binary_sensor.bayesian.markdown
- [x] binary_sensor.gc100.markdown
- [x] binary_sensor.rpi_gpio.markdown
- [x] binary_sensor.rpi_pfio.markdown
- [ ] calendar.todoist.markdown
- [ ] climate.eq3btsmart.markdown
- [ ] climate.generic_thermostat.markdown
- [ ] climate.knx.markdown
- [ ] climate.proliphix.markdown
- [ ] cover.command_line.markdown

Components that needs to be updated:

- [ ] cover.rflink.markdown
- [ ] cover.rpi_gpio.markdown
- [ ] demo.markdown
- [ ] switch.snmp
- [ ] switch.pilight

Components that have not been checked yet.
[...]

PR #6429, opened 2018-10-01
Update emulated_hue configuration variable style

fredrikbaberg:
**Description:**
Update style of emulated_hue documentation to follow new configuration variables description.

Related to #6385.

## Checklist:

### DIFF
--- a/source/_components/emulated_hue.markdown
+++ b/source/_components/emulated_hue.markdown
@@ -49,37 +49,54 @@ emulated_hue:
 emulated_hue:
 ```
 
-Configuration variables:
-
-- **type** (*Optional*): The type of assistant which we are emulating. Either `alexa` or `google_home`, defaults to `google_home`. **This configuration option is deprecated and will be removed in a future release. It is no longer necessary to define type.**
-- **host_ip** (*Optional*): The IP address that your Home Assistant installation is running on. If you do not specify this option, the component will attempt to determine the IP address on its own.
-- **listen_port** (*Optional*): The port the Hue bridge API web server will run on. If not specified, this defaults to 8300. This can be any free port on your system.
-
-- **advertise_ip** (*Optional*): If you need to override the IP address used for UPnP discovery. (For example, using network isolation in Docker)
-- **advertise_port** (*Optional*): If you need to specifically override the advertised UPnP port.
-
-- **upnp_bind_multicast** (*Optional*): Whether or not to bind the UPnP (SSDP) listener to the multicast address (239.255.255.250) or instead to the (unicast) host_ip address specified above (or automatically determined). The default is true, which will work for most situations.  In special circumstances, like running in a FreeBSD or FreeNAS jail, you may need to disable this.
-
-- **off_maps_to_on_domains** (*Optional*): The domains that maps an "off" command to an "on" command.
-
-  For example, if `script` is included in the list, and you ask Alexa to "turn off the *water plants* script," the command will be handled as if you asked her to turn on the script.
-
-  If not specified, this defaults to the following list:
-
-  - `script`
-  - `scene`
-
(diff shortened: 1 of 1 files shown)


source/_components/emulated_hue.markdown
```diff
@@ -49,37 +49,54 @@ emulated_hue:
...
+  required: false
+  type: boolean
+  default: true
+off_maps_to_on_domains:
+  description: The domains that maps an "off" command to an "on" command. For example, if `script` is included in the list, and you ask Alexa to "turn off the *water plants* script," the command will be handled as if you asked her to turn on the script.
+  required: false
+  type: list
+  default: [script, scene]
+expose_by_default:
+  description: "Whether or not entities should be exposed via the bridge by default instead of explicitly (see the ‘emulated_hue’ customization below). Warning: If you have a lot of devices (more than 49 total across all exposed domains), you should be careful with this option. Exposing more devices than Alexa supports can result in it not seeing any of them. If you are having trouble getting any devices to show up, try disabling this, and explicitly exposing just a few devices at a time to see if that fixes it."
```
fredrikbaberg, line 86: Had to quote this due to special character in text ("Warning:")

WofWca: My apologies, @fredrikbaberg . I thought the quotes were going to be shown at the resulting page. Sorry you had to do unnecessary work.

fredrikbaberg: I understood the documentation as both being possible, but the quotes are required when there are control characters (https://developers.home-assistant.io/docs/en/documentation_standards.html). The example I started to look at used quotes for everything (and of course I can't find that now, so either I misread or it's been updated).
After your comment regarding quotes I'm only using them where required, and this is my interpretation of "add a note" (see above link to documentation standards).

WofWca: IMO the paragraph you provided only covers the immediate user's YAML config files, not the descriptions for internal use (like this). "Add a note" implies informing a user that if he does not quote particular values in his config, it may be invalid.

fredrikbaberg: Learning lot of new things this week, then I'll use quotes when needed in documentation.
You think the PR is OK as it is? Anything I should correct?

WofWca (commented):
TBH I still find documentation a little fuzzy.
Maybe content starting from line 123 may be embedded in the main part as it's done [here](https://www.home-assistant.io/components/sensor.template). On the other hand, there it says that variable `sensors` is a map with keys such as `friendly_name`, `friendly_name_template`, etc, which, in YAML representation, would look like this (or am I missing something?):
```yaml
sensors:
  friendly_name: "Sun angle"
  friendly_name_template: "{{ states.sun.sun.attributes.elevation }}"
```
But it doesn't:
```yaml
sensors:
  solar_angle:
    friendly_name: "Sun angle"
    unit_of_measurement: 'degrees'
    value_template: "{{ states.sun.sun.attributes.elevation }}"
```
Also [the standard says](https://github.com/home-assistant/developers.home-assistant/blob/master/docs/documentation_create_page.md) (since home-assistant/developers.home-assistant#105) that

> If you use `map`/`list` then you need to define `keys:`

But what if we use just a list of strings, like [here](https://www.home-assistant.io/components/emulated_hue) (off_maps_to_on_domains variable)?

```yaml
emulated_hue:
  host_ip: 192.168.1.186
  listen_port: 8300
  advertise_ip: 10.0.0.10
  advertise_port: 8080
  off_maps_to_on_domains:
    - script
    - scene
```

We don't need to specify any keys in that case.

frenck:
@WofWca We have updated the documentation a little, to make it a little more consistent.
Would you be so kind to create an issue or PR with suggestions on the home-assitant/developers.home-assistant repo with your suggestions?

This allows us to keep the PR's flowing. Thx 👍

frenck (approved):

frenck:
@fredrikbaberg Thank you for this 👍 

While there are some discussions on specific formatting (as discussed above), this is now "in-line" with the current documentation and therefore a step forward. 

Whatever the discussion changes to the "standards", we would have to update a large part of our documentation anyways.

Happy Hacktoberfest 🎉 

../Frenck

PR #6433, opened 2018-10-02
Ebusd integration

CrazYoshi:
**Description:**
Integration between ebusd, daemon for communication with eBUS heating systems, and homeassistant using sensor component.

https://github.com/home-assistant/home-assistant/pull/16899


frenck:
Thanks, @CrazYoshi.

Still not sure why (again) removed/replaced the PR template...

Marked ready for review.

CrazYoshi:
> Thanks, @CrazYoshi.
> 
> Still not sure why (again) removed/replaced the PR template...
> 
> Marked ready for review.

Sorry, I was not able to get the template again so I create a new one. My mistake

source/_components/sensor.ebusd.markdown
```diff
@@ -0,0 +1,90 @@
...
+description: "The ebusd sensor allows the integration between eBUS heating system and Home Assistant."
+date: 2018-10-01 15:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: ebusd.png
+ha_category: Sensor
+ha_iot_class: "Local Push"
+ha_release: 0.79
```
frenck: 0.79 has already been released. Please up this to the next release.

source/_components/sensor.ebusd.markdown
```diff
@@ -0,0 +1,90 @@
...
+      - 'Zone1TimerWednesday'
+      - 'Zone1TimerThursday'
+      - 'Zone1TimerFriday'
+      - 'Zone1TimerSaturday'
+      - 'Zone1TimerSunday'
+      - 'Zone1OperativeMode'
+      - 'PowerEnergyConsumptionThisMonth'
+```
+
+Configuration variables:
```
frenck: Please use the configuration tags, for more information please see:
https://home-assistant.io/developers/documentation/create_page/#configuration

frenck (changes requested):

frenck:
@CrazYoshi I've reviewed your PR and left some comments. Could you please take a look? 👍

source/_components/sensor.ebusd.markdown
```diff
@@ -0,0 +1,146 @@
...
+comments: false
+sharing: true
+footer: true
+logo: ebusd.png
+ha_category: Sensor
+ha_iot_class: "Local Push"
+ha_release: "0.80"
+---
+
+Integration between ebusd (https://github.com/john30/ebusd), daemon for communication with eBUS heating systems, and homeassistant using sensor component.
```
fredrikbaberg: This is probably the line causing the error in netlify. Could you try to format the line as a markdown link? Should be something like [ebusd](https://github.com/john30/ebusd)

CrazYoshi:
> @CrazYoshi I've reviewed your PR and left some comments. Could you please take a look? 👍

Hi, I'm trying to find out which is the root cause of the failed validation after I change the documentation using configuration tags as you suggested but without success. Could you please give me an hint?

source/_components/sensor.ebusd.markdown
```diff
@@ -0,0 +1,147 @@
...
+  description: The port your ebus daemon uses, defaults to 8888.
+  type: integer
+  required: false
+  default: 8888
+name: 
+  description: The name to use when displaying this ebusd instance.
+  type: string
+  required: false
+monitored_conditions:
+  description: array: List of condition to monitor.
```
fredrikbaberg: Try to add a quote to this description (since it contains ":").
I'm checking out the code now to build locally and see if it is enough.

fredrikbaberg: It should be enough to correct this line (77, "description: array: List of condition to monitor."), it removes the error when I try it locally. If you use any special characters (such as :, [] etc.), you need to place the text in quotes.
In this case it might be easier to just remove "array: ", since I think it's clear from "map" that it's a list(?)

source/_components/ebus.ebusd.markdown
```diff
@@ -0,0 +1,147 @@
...
+circuit: 
+  description: The heating circuit name to monitor, eg. 700.
+  required: true
+  type: string
+port:
+  description: The port your ebus daemon uses, defaults to 8888.
+  type: integer
+  required: false
+  default: 8888
+name:
```
tjorim, line 48: Name has a default value as well: ebusd

source/_components/ebus.ebusd.markdown
```diff
@@ -0,0 +1,147 @@
...
+  type: integer
+  required: false
+  default: 8888
+name: 
+  description: The name to use when displaying this ebusd instance.
+  type: string
+  required: false
+monitored_conditions:
+  description: List of condition to monitor.
+  type: map
```
tjorim: Should be list?

source/_components/ebus.ebusd.markdown
```diff
@@ -0,0 +1,147 @@
...
+host:
+  description: This is the IP address of your ebus daemon, eg. 127.0.0.1.
+  required: true
+  type: string
+circuit: 
+  description: The heating circuit name to monitor, eg. 700.
+  required: true
+  type: string
+port:
+  description: The port your ebus daemon uses, defaults to 8888.
```
tjorim: No need to repeat that 8888 is the default in the description, we have the `default:` for that.

source/_components/ebus.ebusd.markdown
```diff
@@ -0,0 +1,147 @@
...
+ha_release: "0.80"
+---
+
+Integration between [ebusd](https://github.com/john30/ebusd/) daemon for communication with eBUS heating systems, and homeassistant using sensor component.
+
+## {% linkable_title Configuration %}
+
+Enable the sensor by adding the following to your `configuration.yaml` file:
+
+```yaml
```
tjorim, line 22: Get rid of `port` and `name`, this example must only contain the required variables (and they were set at the default anyway). Limit `monitored_conditions:` to a few like `ActualFlowTemperatureDesired`, `PumpStatus` and `WaterPressure`.

source/_components/ebus.ebusd.markdown
```diff
@@ -8,7 +8,7 @@ comments: false
 sharing: true
 footer: true
 logo: ebusd.png
-ha_category: Sensor
+ha_category: Ebus
```
tjorim, line 11: This is not a valid component category (yet).

tjorim (changes requested):
@CrazYoshi, I left some feedback for you 👍 

⚠️ The `ha_release` will need an update before merging this one in.

source/_components/ebus.ebusd.markdown
CrazYoshi: which is the right category?

frenck:
Closing this PR, since the parent PR has been closed as well.

PR #6440, opened 2018-10-02
Add comment about alarm_type versions

eresonance:
**Description:**
Added info about the alarm_type from the openzwave code. I believe there is some mixed info based on older V1 alarms vs new V2 alarms. V2 have fixed alarm_types so I've separated the two to make it clear which one a user should look at.

## Checklist:

### DIFF
--- a/source/_docs/z-wave/entities.markdown
+++ b/source/_docs/z-wave/entities.markdown
@@ -25,16 +25,32 @@ This is for a single purpose sensor, multi sensors are explained under Multi Sen
 
 Devices (usually sensors) that support the Alarm command class will create entities starting with `sensor`, and with some generic suffixes, and a suffix that relates to the supported alarm class. For example, the smoke detector `lounge` will have an entity `sensor.lounge_smoke`, and possibly also `sensor.lounge_alarm_type` and `sensor.lounge_alarm_level`. If the device creates a `binary_sensor` entity, it is recommended to use that rather then the `sensor` entity.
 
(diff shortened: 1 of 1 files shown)


frenck (approved):

frenck:
Thank you for this @eresonance!

Nice first-time contribution to our documentation! Feel free to submit some more! 🎖 

Happy Hacktoberfest 🎉

eresonance:
Actually I need to revise these changes :(

I thought more about this at work and I realized where I was getting confused. Openzwave and homeassistant already handle the alarm class v2 enum properly by creating a new sensor called {entity}_{alarm v2 type}. They do this via a zwave command to get the alarm types supported by the device. So for instance for my garage door sensor I have a {entity}_burglar sensor that represents the tamper switch in the case.

In the alarm report command, the older style alarm_type and alarm_level values are present in the data:
image
I think the intention by the zwave std is for the older alarm_type and alarm_level to be 0 when a report is sent, but on my device the "alarm type" is set to the same value for which the newer "zwave alarm type" is also set (burglar so type=7). Hass will make two other sensors to handle these, a {entity}_alarm_type and {entity}_alarm_level, which handle the older style v1 alarms.

Really the docs should say that if you have a v2 device, you should ignore the alarm_type and alarm_level sensors as the v2 alarm type sensors are also created and supersede the older ones.

Also it would be nice if hass interpreted/translated the alarm event as per the v2 spec:
image

Maybe I'll submit a change to do that :)

PR #6459, opened 2018-10-03
Update homematic.markdown config variables style

rdbahm:
Per #6385 updating the configuration description.

**Description:**
I've needed to make a few guesses based on the documentation - Homematic's configuration examples are divided into three parts because it uses heavily nested configuration. It is not immediately obvious to me if the system can handle having multiple {% configuration %} blocks in a single page - let me know and I can always adjust how it's set up.

## Checklist:

### DIFF
--- a/source/_components/homematic.markdown
+++ b/source/_components/homematic.markdown
@@ -50,7 +50,7 @@ local_ip:
 local_port:
   description: Port for connection with Home Assistant. By default it is randomly assigned.
   required: false
-  type: int
+  type: integer
 {% endconfiguration %}
 
 Configuration variables (interface):
@@ -61,9 +61,9 @@ host:
   required: true
   type: string
 port:
-  description: Port of CCU/Homegear XML-RPC Server. Wireless: 2001, wired: 2000, IP: 2010, thermostatgroups: 9292.
+  description: "Port of CCU/Homegear XML-RPC Server. Wireless: 2001, wired: 2000, IP: 2010, thermostatgroups: 9292."
   required: false
-  type: int
+  type: integer
 callback_ip:
   description: Set this, if Home Assistant is reachable under a different IP from the CCU (NAT, Docker etc.).
   required: false
@@ -71,23 +71,16 @@ callback_ip:
 callback_port:
   description: Set this, if Home Assistant is reachable under a different IP from the CCU (NAT, Docker etc.).
   required: false
-  type: int
+  type: integer
 resolvenames:
   description: Try to fetch device names. Defaults to `false` if not specified.
   required: false
-  type: map
+  type: string
   default: false
-  keys:
-    metadata:
-      description: Homegear provides device-names through the metadata devices have internally. When using an HM-CFG-LAN interface, you typically use a configuration software ("HomeMatic-Komponenten konfigurieren" is the name of the shortcut on your desktop by default) to pair and configure your devices. If you have paired devices, you'll see them listed in a table. The leftmost column (Name) is prefilled with default names. You can click such a name and enter whatever you like.
-    json:
-      description: The CCU allows to fetch details of the paired devices via JSON-RPC. For this to work you need to add valid credentials to your component-configuration. Guest-access is sufficient to query for device names.
-    xml:
-      description: If you use a CCU, there is an add-on called the "XML-API". With it installed, you are able to fetch all kinds of information from you CCU using XML-RPC. We can leverage this and fetch the names of devices set within the CCU. We don't support authentication with this method.
 jsonport:
   description: Port of CCU JSON-RPC Server. The default is 80, but it may be different when running CCU virtually via Docker.
   required: false
-  type: int
+  type: integer
 username:
   description: When fetching names via JSON-RPC, you need to specify a user with guest-access to the CCU.
   required: false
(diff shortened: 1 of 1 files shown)


frenck:
Hi there @rdbahm,

It seems like the build is failing, can you please take a look? 👍 

../Frenck

source/_components/homematic.markdown
```diff
@@ -34,28 +34,90 @@ homematic:
...
-- **username** (*Optional*): When fetching names via JSON-RPC, you need to specify a user with guest-access to the CCU.
-- **password** (*Optional*): When fetching names via JSON-RPC, you need to specify the password of the user you have configured above.
-- **path** (*Optional*): Set to `/groups` when using port 9292.
+{% configuration %}
+host:
+  description: IP address or Hostname of CCU/Homegear device or Hass.io add-on.
+  required: true
+  type: string
+port:
+  description: Port of CCU/Homegear XML-RPC Server. Wireless: 2001, wired: 2000, IP: 2010, thermostatgroups: 9292.
```
WofWca: There's a `:`. Quotation needed.

source/_components/homematic.markdown
```diff
@@ -34,28 +34,90 @@ homematic:
...
+  required: false
+  type: string
+callback_port:
+  description: Set this, if Home Assistant is reachable under a different IP from the CCU (NAT, Docker etc.).
+  required: false
+  type: int
+resolvenames:
+  description: Try to fetch device names. Defaults to `false` if not specified.
+  required: false
+  type: map
```
WofWca: I think it is incorrect to say that the type is `map` here. It is string, but with only a few valid values.

WofWca (changes requested):
The conversion is incorrect.

source/_components/homematic.markdown
```diff
@@ -34,28 +34,90 @@ homematic:
...
+  required: false
+  type: list
+local_ip:
+  description: IP of device running Home Assistant. Override auto-detected value for exotic network setups.
+  required: false
+  type: string
+local_port:
+  description: Port for connection with Home Assistant. By default it is randomly assigned.
+  required: false
+  type: int
```
frenck: `integer`

source/_components/homematic.markdown
frenck: Indeed. @rdbahm this should be a `string`.

source/_components/homematic.markdown
```diff
@@ -34,28 +34,90 @@ homematic:
...
+callback_ip:
+  description: Set this, if Home Assistant is reachable under a different IP from the CCU (NAT, Docker etc.).
+  required: false
+  type: string
+callback_port:
+  description: Set this, if Home Assistant is reachable under a different IP from the CCU (NAT, Docker etc.).
+  required: false
+  type: int
```
frenck: `integer`

source/_components/homematic.markdown
```diff
@@ -34,28 +34,90 @@ homematic:
...
-- **path** (*Optional*): Set to `/groups` when using port 9292.
+{% configuration %}
+host:
+  description: IP address or Hostname of CCU/Homegear device or Hass.io add-on.
+  required: true
+  type: string
+port:
+  description: Port of CCU/Homegear XML-RPC Server. Wireless: 2001, wired: 2000, IP: 2010, thermostatgroups: 9292.
+  required: false
+  type: int
```
frenck: `integer`

source/_components/homematic.markdown
```diff
@@ -34,28 +34,90 @@ homematic:
...
+callback_port:
+  description: Set this, if Home Assistant is reachable under a different IP from the CCU (NAT, Docker etc.).
+  required: false
+  type: int
+resolvenames:
+  description: Try to fetch device names. Defaults to `false` if not specified.
+  required: false
+  type: map
+  default: false
+  keys:
```
frenck: I suggest restoring the part where those where explained (since this is not a map)

frenck (changes requested):

source/_components/homematic.markdown
rdbahm: Makes sense. Is there a better way than "keys" to specify what the options are? (Does "keys"work without it being set to "map" for type?")

source/_components/homematic.markdown
rdbahm: Documentation says "int" for the type - is it incorrect? https://developers.home-assistant.io/docs/en/documentation_create_page.html

source/_components/homematic.markdown
rdbahm: Found the comment below where you suggest restoring the section - agreed, and I'll do that in the next commit.

rdbahm:
Okay, I've made the requested changes. An outstanding question I had left (outside the scope of this one request) is "int" vs "integer." The documentation says to use "int" (https://developers.home-assistant.io/docs/en/documentation_create_page.html#configuration) while it looks like "integer" is what you're looking for. Should the documentation be updated?

WofWca:
[It is updated](https://github.com/home-assistant/developers.home-assistant/blob/master/docs/documentation_create_page.md),  but the updates are not live yet.

frenck (approved):

frenck:
Thanks for updating the PR @rdbahm! 👍 

Happy Hacktoberfest 🎉 

../Frenck

PR #6475, opened 2018-10-03
Adding detail about initial state

DubhAd:
It's always been a bit of mystery why automations sometimes get turned off. This should shed some light on it.

I appreciate the intent is to fix this, but until it's fixed, it makes sense to document it.


fabaff (approved):
:bird:

PR #6486, opened 2018-10-03
Added mdi icons to ipanel example

jladefoged:
**Description:**
Added mdi icons to ipanel example

## Checklist:


fabaff:
`icon` is an optional configuration setting.

We try to keep all [configuration samples minimal](https://developers.home-assistant.io/docs/en/documentation_create_page.html#configuration). Thus, no optional requirement in the default sample. This helps users to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

ladefoged81:
So the opposite of #4861 ?

PR #6537, opened 2018-10-05
Made true/false lowercase

frenck:
**Description:**

Changed True/False -> true/false. 
Old values where truty...

## Checklist:


fabaff:
According to [Boolean Language-Independent Type for YAML™ Version 1.1](http://yaml.org/type/bool.html) you can use various entries for a Boolean. Also, our validator is not complaining.

Are there known issues which are forcing us to stick to `true/false`?

frenck:
@fabaff https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy

It is not wrong, but the documentation currently uses a mix.

DubhAd (approved):
I'm in favour of consistency, it causes less confusion in the less experienced folks.

PR #6575, opened 2018-10-07
Little note that triggers are or

DubhAd:
We don't really call out that triggers are `or`, so adding a little about that here.


frenck (approved):

frenck:
Thanks, @DubhAd.
That is a nice addition. I bet first-timers will be thankful for that little addition.

Happy Hacktoberfest 🎉 

../Frenck

PR #6606, opened 2018-10-08
Add device registry to MQTT cover

OttoWinter:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#17245

## Checklist:

### DIFF
--- a/source/_components/cover.mqtt.markdown
+++ b/source/_components/cover.mqtt.markdown
@@ -169,7 +169,7 @@ device:
       required: false
       type: list, string
     connections:
-      description: 'A list of connections of the device to the outside world as a list of tuples `[connection_type, connection_identifier]`. For example the MAC address of a network interface: `[["mac", "02:5b:26:a8:dc:12"]]`.'
+      description: 'A list of connections of the device to the outside world as a list of tuples `[connection_type, connection_identifier]`. For example the MAC address of a network interface: `"connections": [["mac", "02:5b:26:a8:dc:12"]]`.'
       required: false
       type: list, tuple
     manufacturer:
--- a/source/_components/cover.mqtt.markdown
+++ b/source/_components/cover.mqtt.markdown
@@ -171,7 +171,7 @@ device:
     connections:
       description: 'A list of connections of the device to the outside world as a list of tuples `[connection_type, connection_identifier]`. For example the MAC address of a network interface: `"connections": [["mac", "02:5b:26:a8:dc:12"]]`.'
       required: false
-      type: list, tuple
+      type: list
     manufacturer:
       description: 'The manufacturer of the device.'
       required: false
--- a/source/_components/cover.mqtt.markdown
+++ b/source/_components/cover.mqtt.markdown
@@ -159,6 +159,35 @@ unique_id:
   description: An ID that uniquely identifies this cover. If two covers have the same unique ID, Home Assistant will raise an exception.
   required: false
   type: string
+device:
+  description: 'Information about the device this cover is a part of to tie it into the [device registry](https://developers.home-assistant.io/docs/en/device_registry_index.html). Only works through [MQTT discovery](/docs/mqtt/discovery/) and when [`unique_id`](#unique_id) is set.'
+  required: false
+  type: map
+  keys:
+    identifiers:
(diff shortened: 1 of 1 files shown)


source/_components/cover.mqtt.markdown
```diff
@@ -159,6 +159,35 @@ unique_id:
...
+  description: 'Information about the device this cover is a part of to tie it into the [device registry](https://developers.home-assistant.io/docs/en/device_registry_index.html). Only works through [MQTT discovery](/docs/mqtt/discovery/) and when [`unique_id`](#unique_id) is set.'
+  required: false
+  type: map
+  keys:
+    identifiers:
+      description: 'A list of IDs that uniquely identify the device. For example a serial number.'
+      required: false
+      type: list, string
+    connections:
+      description: 'A list of connections of the device to the outside world as a list of tuples `[connection_type, connection_identifier]`. For example the MAC address of a network interface: `[["mac", "02:5b:26:a8:dc:12"]]`.'
```
DubhAd: Can you make the same clarity update here please.

DubhAd (dismissed):

source/_components/cover.mqtt.markdown
```diff
@@ -159,6 +159,35 @@ unique_id:
...
+  type: map
+  keys:
+    identifiers:
+      description: 'A list of IDs that uniquely identify the device. For example a serial number.'
+      required: false
+      type: list, string
+    connections:
+      description: 'A list of connections of the device to the outside world as a list of tuples `[connection_type, connection_identifier]`. For example the MAC address of a network interface: `"connections": [["mac", "02:5b:26:a8:dc:12"]]`.'
+      required: false
+      type: list, tuple
```
frenck: `tuple` is not a valid type for the documentation.
Seem like this needs to be a `map` with possible `keys`?

source/_components/cover.mqtt.markdown
```diff
@@ -159,6 +159,35 @@ unique_id:
   description: An ID that uniquely identifies this cover. If two covers have the same unique ID, Home Assistant will raise an exception.
   required: false
   type: string
+device:
+  description: 'Information about the device this cover is a part of to tie it into the [device registry](https://developers.home-assistant.io/docs/en/device_registry_index.html). Only works through [MQTT discovery](/docs/mqtt/discovery/) and when [`unique_id`](#unique_id) is set.'
```
frenck, line 163: `Information about the device this cover` sounds weird.

frenck (changes requested):
@OttoWinter Left you some comments, could you please take a look? 👍

source/_components/cover.mqtt.markdown
OttoWinter: No, it's not a map. It's a list of 2-tuples, for example `"[["mac", "02:5b:26:a8:dc:12"], ["mac", "02:5b:26:a8:dc:13"]]` which in context would look like:

```json
{
  "state_topic": "...",
  # ...
  "device": {
    "connections": [["mac", "02:5b:26:a8:dc:12"]]
  }
}
```

So in JSON (this configuration option only works through MQTT discovery, so only JSON) it would *actually* be a list of list, but that would look even weirder in the documentation. And a term to describe a list with fixed size is usually `tuple` - hence a list of tuples.

I can just remove `tuple` from above and keep it as `type: list`

source/_components/cover.mqtt.markdown
OttoWinter: Ok, I don't really know how to put it else. I think one option would be to just remove the second part of that: `Information about the device to tie it into the device registry.`

frenck: I've been thinking about another wording...
I was just saying it sounds weird / doesn't flow. But it is correct.
So let's leave it like it is.

source/_components/cover.mqtt.markdown
frenck: Aah. I would keep it a `type: list` in that case. Simply because we don't have tuples defined at this point in time... and well... it, in fact, is a list.

frenck (approved):

PR #6614, opened 2018-10-08
Removed API password warning

DubhAd:
Given that as of 0.78 auth is unavoidable, removed the warning of enabling the api password.

Also updated the description of the API password, and nudged towards using a long lasting access token instead.


frenck (approved):

frenck:
Thanks, @DubhAd 👍

PR #6622, opened 2018-10-08
Updated harmony configuration to use configuration variables

luukdobber:
**Description:**
Updated the Harmony Hub documentation to use the new configuration variables (#6385).

## Checklist:

### DIFF
--- a/source/_components/remote.harmony.markdown
+++ b/source/_components/remote.harmony.markdown
@@ -43,15 +43,31 @@ You can override some default configuration values on a discovered hub (e.g., th
     activity: Watch TV
 ```
 
-Configuration variables:
+{% configuration %}
+name:
+  description: The hub's name to display in the frontend. This name must match the name you have set on the Hub.
+  required: true
+  type: string
+host:
+  description: The Harmony device's IP address. Leave empty for the IP to be discovered automatically.
+  required: false
+  type: string
+port:
+  description: The Harmony device's port.
+  required: false
+  type: integer
+  default: 5222
+activity:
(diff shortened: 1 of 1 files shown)


source/_components/remote.harmony.markdown
```diff
@@ -43,15 +43,31 @@ You can override some default configuration values on a discovered hub (e.g., th
...
+  type: integer
+  default: 5222
+activity:
+  description: Activity to use when `turn_on` service is called without any data. Overrides the `activity` setting for this discovered hub.
+  required: false
+  type: string
+delay_secs:
+  description: Default duration in seconds between sending commands to a device.
+  required: false
+  type: float
```
klaasnicolaas, line 67: possible that this can also be an integer

arsaboo: Yes, let us use integer here.

luukdobber: Are you sure? It's a float in the [harmony component's source](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/remote/harmony.py#L36). Also, one of the existing examples uses a float.

Shouldn't we specify float or both types (the input_number component also supports both, but only mentions float on the documentation page)?

klaasnicolaas: You are right, usually an integer is always used because whole seconds are used. But I now see in an example script that the 0.6 sec is used so it should be a float indeed

klaasnicolaas (approved):

arsaboo (approved):
Thanks :100:

PR #6641, opened 2018-10-09
influxdb component config var formatting

nwesterhausen:
Updating the configuration variable section.

I wasn't sure that the customize variables should be included, since they apply to all components from my understanding.
```
- **component_config**, **component_config_domain**, **component_config_glob** (*Optional*): These attributes contains component-specific override values. See [Customizing devices and services](/getting-started/customizing-devices/) for format.
  - **override_measurement** (*Optional*): Measurement name to use for this component, takes precedence over the global 'override_measurement' and component-specific 'unit_of_measurement' attribute.
```

Related to #6385.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/influxdb.markdown
+++ b/source/_components/influxdb.markdown
@@ -40,17 +40,15 @@ username:
   type: string
   description: The username of the database user. The user needs read/write privileges on the database
   required: false
-  default: home-assistant
 password:
   type: string
   description: The password for the database user account.
   required: false
-  default: password
 database:
   type: string
   description: Name of the database to use. The database must already exist.
   required: false
-  default: home_assistant
+  default: will use the default database of your influxdb instance
 ssl:
   type: boolean
   description: Use https instead of http to connect.
@@ -60,7 +58,7 @@ verify_ssl:
   type: boolean
   description: Verify SSL certificate for https request.
   required: false
-  default: false
+  default: true
 max_retries:
   type: int
   description: Set this to allow the component to retry if there was a network error when transmitting data.
--- a/source/_components/influxdb.markdown
+++ b/source/_components/influxdb.markdown
@@ -109,6 +109,36 @@ tags_attributes:
   description: The list of attribute names which should be reported as tags and not fields to InfluxDB. For example, if set to `friendly_name`, it will be possible to group by entities' friendly names as well, in addition to their ids.
   required: false
   default: 0
+component_config:
+  type: string
+  required: false
+  description: These attributes contains component-specific override values. See [Customizing devices and services](/getting-started/customizing-devices/) for format.
+  keys:
+    override_measurement:
+      type: boolean
+      description: Measurement name to use for this component, takes precedence over the global ‘override_measurement’ and component-specific ‘unit_of_measurement’ attribute.
+      required: false
+      default: false
+component_config_domain:
+  type: string
+  required: false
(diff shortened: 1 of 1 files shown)


source/_components/influxdb.markdown
```diff
@@ -25,28 +25,91 @@ influxdb:
...
-- **tags** (*Optional*): Tags to mark the data.
-- **tags_attributes** (*Optional*): The list of attribute names which should be reported as tags and not fields to InfluxDB. For example, if set to `friendly_name`, it will be possible to group by entities' friendly names as well, in addition to their ids.
+{% configuration %}
+host:
+  type: string
+  description: IP address of your database host, e.g., 192.168.1.10
+  required: false
+  default: localhost
+port:
+  type: int
```
frenck: `int` -> `integer`

source/_components/influxdb.markdown
```diff
@@ -25,28 +25,91 @@ influxdb:
...
+  required: false
+password:
+  type: string
+  description: The password for the database user account.
+  required: false
+database:
+  type: string
+  description: Name of the database to use. The database must already exist.
+  required: false
+  default: will use the default database of your influxdb instance
```
frenck: According to the code (maybe I misread it), the default is `home_assistant`.

https://github.com/home-assistant/home-assistant/blob/e903f7ffdafc7f5e0797f654c9b8314529412775/homeassistant/components/influxdb.py#L40

nwesterhausen:
Right now, this component page links to the blog post on influxdb (from 2015) which outlines getting influxdb set up and using it with home assistant. The process in the blog post will still work, it uses an old version of influxdb. The current version of influxdb no longer has [built-in administration](https://docs.influxdata.com/influxdb/v1.6/about_the_project/releasenotes-changelog/#deprecations-1). 

Do we want to have updated instructions or possibly link to [this forum post](https://community.home-assistant.io/t/complete-guide-on-setting-up-grafana-influxdb-with-home-assistant-using-official-docker-images/42860) as an up to date alternative for the blog post?

frenck (changes requested):
@nwesterhausen Thank you for doing this! 👍 

About the customize variables, I would opt to add them, each individually (Rather be explicit on the options).

I'm not for updating/rewriting blog posts. Those are a snapshot of history.
Maybe move the update of the contents to a secondary PR?

I also left some comments on the PR itself. Could you please take a look?

source/_components/influxdb.markdown
nwesterhausen: Yes, I second guessed myself and misread the python. Thanks for catching that!

nwesterhausen:
I agree we shouldn't change the blog post. I should have been more explicit in my suggestion: either putting the instructions into this page, a secondary page with more instructions on influx db, or link to the forum post which has more up to date instructions.

I think based off some other components which involve extra steps, it makes sense to put the instructions into this component page.

frenck:
@nwesterhausen Agreed.

nwesterhausen:
I initially misunderstood the custom_* variables and what they would be used for. It turns out it actually is really simple: if there is a set of entities or a domain that you want to specify a special measurement value for the database (e.g. "% Humidity" instead of just "%" for humidity values) you could add a configuration value here that would accomplish that for you.

nwesterhausen:
I believe this is ready to be merged.

frenck (approved):

frenck:
Thank, @nwesterhausen! 👍 

Happy Hacktoberfest 🎉 

../Frenck

PR #6644, opened 2018-10-09
Adding relink section

DubhAd:
A few folks have commented that if they hit *Account linking failed* then clearing browser history and cache can solve that.


fabaff (approved):
:bird:

frenck (approved):

PR #6647, opened 2018-10-09
Simplification of the supported controllers

DubhAd:
Stripped it down to the key fact - *static controllers* are all that'll work with OZW.

Noted that Z-Wave Plus controllers are recommended, to take advantage of all the features of Z-Wave Plus.

Noted that S2 and Smart Start aren't supported by OZW

### DIFF
--- a/source/_docs/z-wave/controllers.markdown
+++ b/source/_docs/z-wave/controllers.markdown
@@ -11,7 +11,15 @@ footer: true
 
 ## {% linkable_title Supported Z-Wave USB Sticks & Hardware Modules %}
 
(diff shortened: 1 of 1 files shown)


source/_docs/z-wave/controllers.markdown
```diff
@@ -11,22 +11,9 @@ footer: true
 
 ## {% linkable_title Supported Z-Wave USB Sticks & Hardware Modules %}
 
-You need to have a [supported Z-Wave USB stick or module](https://github.com/OpenZWave/open-zwave/wiki/Controller-Compatibility-List) installed.
-
-| Device                    | Works on Linux | Works on Windows | Works on OSX |
-|---------------------------|----------------|------------------|--------------|
-| Aeotec Z-Stick            |   &#10003;     |                  |              |
```
balloob, line 18: Let's keep a list of controllers that work. People want to see an easy list and we already have that info.

frenck (approved):

frenck:
Thanks @DubhAd 👍

PR #6655, opened 2018-10-10
Clarity update

DubhAd:
The _user_ docs imply that setting the device class sets the unit of measurement. This clearly isn't the case, so doing a set of clarity edits.


frenck (approved):

frenck:
Thanks 👍

PR #6657, opened 2018-10-10
Clarity update

DubhAd:
As somebody pointed out on the forum, using different examples adds confusion. Updated both styles to use the same contents, so it's clearer what's going on.


frenck (approved):

frenck:
Thanks 👍

PR #6682, opened 2018-10-11
Update Kankun switch component configuration variable

klaasnicolaas:
Update style of Kankun switch component documentation to follow new configuration variables description.
Related to #6385.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/switch.kankun.markdown
+++ b/source/_components/switch.kankun.markdown
@@ -30,14 +30,40 @@ switch:
       host: hostname_or_ipaddr
 ```
 
-Configuration variables:
-
-- **switches** (*Required*): The array that contains all Kankun switches.
-  - **identifier** (*Required*): Name of the Kankun switch as slug. Multiple entries are possible.
-    - **host** (*Required*): Hostname or IP address of the switch on the local network.
-    - **name** (*Optional*): Friendly name of the switch.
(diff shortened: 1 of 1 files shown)


source/_components/switch.kankun.markdown
```diff
@@ -30,14 +30,40 @@ switch:
...
+switches:
+  description: The array that contains all Kankun switches.
+  required: true
+  type: map
+  keys:
+    identifier:
+      description: Name of the Kankun switch as slug. Multiple entries are possible.
+      required: true
+      type: string
+    host:
```
frenck: this is incorrect... `host` (and all the others below) are options of a unique device and are therefore nested deeper.

frenck (changes requested):
@klaasnicolaas Thanks for the PR. I've left you a comment, could you please take a look? 👍

source/_components/switch.kankun.markdown
klaasnicolaas: so the type of identifier has to be a map and then put everything under it in those keys?

frenck: This is the confusing part of the documentation at this point. We are still uncertain about a permanent solution. Until then, we keep it this way.

frenck (approved):

frenck:
Thanks, @klaasnicolaas 👍 

Happy Hacktoberfest 🎉

PR #6708, opened 2018-10-12
Corrected the Aruba OID

tyler-8:
**Description:**
The previous Aruba OID was the MAC address of the access point itself. The new OID is client mac addresses.

Tested and verified on an Aruba Instant 325 AP.


fabaff (approved):
Thanks :bird:

PR #6714, opened 2018-10-12
Adding some details

DubhAd:
The Getting Started doesn't mention how long the install typically takes, or what to do if `hassio.local` doesn't work.


frenck (approved):

frenck:
Thanks, @DubhAd 👍

PR #6722, opened 2018-10-12
Update tibber.markdown

Danielhiversen:
**Description:**

Is hub correct?
Or how should a platform be categorized?

### DIFF
--- a/source/_components/tibber.markdown
+++ b/source/_components/tibber.markdown
@@ -8,7 +8,7 @@ comments: false
 sharing: true
 footer: true
 logo: tibber.png
-ha_category: Energy
+ha_category: Hub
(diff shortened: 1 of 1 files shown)


fabaff:
If something matches a group then it should be in that group. The idea is that `hub` and `sensor` have less entries. I hope that with the config flow we can get rid of platform docs because they often only say set up the component and if you are looking for something you should find it in the appropriate category (like `weather` stuff in "weather" category even if it's a hub (meaning a component) and not need to browse through the "hub" category).

Danielhiversen:
Ok, thanks.
It is not possible to know where the platform docs are here: https://www.home-assistant.io/components/#search/tibber

PR #6770, opened 2018-10-13
Added URL example internal

noxhirsch:
Added an example for weblink with which you not only can link to external URLs - you also can link to internal pages or panels


fabaff (approved):
Thanks :bird:

PR #6791, opened 2018-10-14
Correcting device_class link

DubhAd:
Yes, I know it isn't a link in the source, but it is in the live docs 🤷‍♂️

### DIFF
--- a/source/_docs/configuration/customizing-devices.markdown
+++ b/source/_docs/configuration/customizing-devices.markdown
@@ -69,7 +69,7 @@ assumed_state:
 device_class:
   description: Sets the class of the device, changing the device state and icon that is displayed on the UI (see below). It does not set the `unit_of_measurement`.
   required: false
-  type: device_class
+  type: [device_class](/docs/configuration/customizing-devices/#device-class)
   default: None
 unit_of_measurement:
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
This should not be modified in this location.

https://github.com/home-assistant/home-assistant.io/blob/2dbbafef628c691de84a705dee6547a802795dd8/plugins/configuration.rb#L5

fabaff:
The content is coming from https://github.com/home-assistant/home-assistant.io/blob/next/plugins/configuration.rb#L3 as @frenck already mentioned.

If `device_class` is included in a sensor the link is pointing to https://www.home-assistant.io/components/sensor/#device-class and for a binary sensor to https://www.home-assistant.io/components/sensor/#device-class.

For the Customizing entities page this is not working. Thus I suggest that we just display the string because below is a section that contains the links to the component pages.

fabaff:
Was not working for covers: https://github.com/home-assistant/home-assistant.io/pull/6793

frenck:
So is there anything to do now on this PR @DubhAd 🤷‍♀️

frenck:
Closing this PR, since it is running stale.

PR #6825, opened 2018-10-15
Specified model/OS version that Aruba OID was verified on

tyler-8:
**Description:**
In https://github.com/home-assistant/home-assistant.io/pull/6708 I updated the OID for Aruba client MAC addresses. This PR adds the platform that I used for verification to the appropriate column. I should have done this in the original PR but forgot to.


fabaff (approved):
Thanks :bird:

PR #6872, opened 2018-10-16
Fixing service names

DubhAd:
Some of the service names were still in the old JSON format for some reason


fabaff (approved):
Thanks :bird:

PR #6873, opened 2018-10-16
Patch 1

arraylabs:
**Description:**
add info for scan_interval

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#17535

## Checklist:

- [X ] The documentation follow the [standards][standards].


fabaff:
You need to `rebase` your branch.

ludeeus:
The `scan_interval` are covered here <https://www.home-assistant.io/docs/configuration/platform_options/#scan-interval>

tjorim:
`scan_interval` is a general entity option and shouldn't be added to this page, if I'm not mistaken. https://www.home-assistant.io/docs/configuration/platform_options/

Edit: why doesn't GitHub show it when someone else is already commenting 🙈

arraylabs:
@ludeeus @tjorim no problem, I thought I would add it as I'm not sure users are aware that it's available for most components. I honestly didn't realize it was available until I really starting digging around in the docs. I can close the request if that's best. Thanks!

tjorim:
I do realize that and agree the `scan_interval` option isn't clear to most users.
It took me some time to figure out it existed as well.

In my opinion it doesn't belong in the 'Advanced Configuration' section,
all tucked away down there.

fabaff:
The reason that `scan_interval` is in the advance section is that the developers of the platform or integration were taking care that API limits are not exceeded in a standard use case and resources not wasted. There are cases were we need to have that options but the regular users should not have any need to change it. Especially as most of the time the interval is decreased between the updates and not increased.

PR #6876, opened 2018-10-17
Updating unifi documentation to be clearer about the site_id value

keirans:
**Description:**
I encountered some initial challenges with getting the unifi tracker working for my environment because I was not using the default site.

I'm adding some more information to the documentation to ensure that those who have similar configurations know how to find the value for the site_id parameter as it is not as clear that is not the name of the site, but a controller generated value.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #6889, opened 2018-10-17
Clarity updates

DubhAd:
Added note about testing templates in the template tool

Added section about how `|` has priority over non-bracket operators

Some minor changes to steer people towards `states.('...')` rather than `states...state`.

### DIFF
--- a/source/_docs/configuration/templating.markdown
+++ b/source/_docs/configuration/templating.markdown
@@ -107,6 +107,18 @@ If your template uses an `entity_id` that begins with a number (example: `states
 Rendering templates with time (`now()`) is dangerous as updates only trigger templates in sensors based on entity state changes.
 </p>
 
+## {% linkable_title Priority of operators %}
+
+The default priority of operators is that the filter (`|`) has priority over everything except brackets. This means that:
+
+{% raw %}
+```yaml
+{{ states('sensor.temperature') | float / 10 | round(2) }}
+```
+{% endraw %}
+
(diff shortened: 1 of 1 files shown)


source/_docs/configuration/templating.markdown
```diff
@@ -107,6 +107,18 @@ If your template uses an `entity_id` that begins with a number (example: `states
 Rendering templates with time (`now()`) is dangerous as updates only trigger templates in sensors based on entity state changes.
 </p>
 
+## {% linkable_title Priority of operators %}
+
+The default priority of operators is that the filter (`|`) has priority over everything except brackets. This means that:
```
zsarnett: I would make it `The default priority of operators is the filter (`|`) has priority over everything except brackets. This means that:` The `that` just didn't flow well

fabaff (approved):
Thanks :bird:

source/_docs/configuration/templating.markdown
```diff
@@ -107,6 +107,18 @@ If your template uses an `entity_id` that begins with a number (example: `states
...
+
+{% raw %}
+```yaml
+{{ states('sensor.temperature') | float / 10 | round(2) }}
+```
+{% endraw %}
+
+Would round `10` to 2 decimal places, then divide `states('sensor.temperature')` by that.
```
zsarnett: `then divide `states('sensor.temperature')` by the result.`

PR #6907, opened 2018-10-17
RFC: Put default value between code quotes

tjorim:
**Description:**

I think it would be better if the default value is shown as code.
This is already done by some people for some docs.
What are other people's opinion on this?

## Checklist:

### DIFF
--- a/plugins/configuration.rb
+++ b/plugins/configuration.rb
@@ -69,7 +69,7 @@ module Jekyll
         end
         markup << "</p>"
         if attr.key? 'default'
-          markup << "<p class='default'>Default value: #{attr['default']}</p>"
+          markup << "<p class='default'>Default value: `#{attr['default']}`</p>"
         end
         markup << "</dd>"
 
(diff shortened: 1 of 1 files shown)


tjorim:
Only issue I can think about is when the default is more complicated than just a single value.

Edit: Luckily Jekyll has a liquid filter to count the number of words.

nwesterhausen:
I like the way the values look as code, but it then becomes very busy since the default line is already italicized.  

Putting the value in a code block highlights it literally is the value that will be used, and when I see the page without the italics on .default, it's less busy.

Besides just my own preference though, I can't present an argument that this really helps users parse the information any clearer, since it already is a separate line under the description and italicized.

frenck:
I personally do not agree on this. There are many situations where this is not a good solution (e.g., where the default is described instead of a hard value). 

I'm wondering how @fabaff things of this suggestion.

tjorim:
@frenck As it is now it won't affect cases where it's described with a longer text, only single word values (which things like boolean, integers... are most of the time) will show as code block.

frenck:
@fabaff What do you think?\
@klaasnicolaas You've done a bit quite a bit of documentation lately, what is your opinion on this?

frenck:
No response from anyone for a long time. Closing this PR because of the lack of support and thus being stale.

PR #6952, opened 2018-10-19
fix get_state syntax examples

afex:
**Description:**

when called as given, errors with `TypeError: get_state() takes from 1 to 2 positional arguments but 3 were given`

## Checklist:


fabaff (approved):
Thanks :bird:

PR #6983, opened 2018-10-20
Updated to include new devices configuration

kbickar:
**Description:**
Updated to include new devices configuration

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#17645

## Checklist:


source/_components/sensor.sense.markdown
```diff
@@ -10,7 +10,7 @@ footer: true
 logo: sense.png
 ha_category: Energy
 ha_iot_class: "Cloud Polling"
-ha_release: 0.65
+ha_release: 0.80
```
frenck: Please do not update the `ha_release` for existing documentation.

source/_components/sensor.sense.markdown
```diff
@@ -40,6 +43,8 @@ Two types of sensors can be monitored and will be created with the following nam
 
 Weekly, Monthly and Yearly variants are also available.
 
+Additionally, each device detected by Sense can be used as a binary sensor.  The device name must match exactly what is shown in the Sense app and will by named as such as a binary sensor.
```
frenck:
```suggestion
Additionally, each device detected by Sense can be used as a binary sensor.  The device name must match exactly what is shown in the Sense app and will be named as such as a binary sensor.
```

source/_components/sensor.sense.markdown
```diff
@@ -51,7 +56,7 @@ password:
   type: string
 monitored_conditions:
   description: List of sensors to display in the front end.
-  required: true
+  required: false
```
frenck: It his becomes optional, the option should be removed from the initial example on top of the page.

source/_components/sensor.sense.markdown
```diff
@@ -31,6 +31,9 @@ sensor:
     - active_production
     - daily_usage
     - daily_production
+  devices:
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

frenck (dismissed):
Thanks for the PR @kbickar!
I've left you some comments, could you take a look? 👍

fabaff (approved):
Thanks :bird:

PR #6813, opened 2018-10-15
Update Caddy Server Docs

ikifar2012:
**Description:**
Updated Caddy Server Documentation to be more in-depth

PR #6996, opened 2018-10-21
Update Caddy Docs

ikifar2012:
Update Caddy Docs - add much needed improvements to the docs
I have rebased and fixed the issues with my previous pull request

### DIFF
--- a/source/_docs/ecosystem/caddy.markdown
+++ b/source/_docs/ecosystem/caddy.markdown
@@ -10,12 +10,46 @@ footer: true
 ---
 
 Configure [Caddy Server](https://caddyserver.com/) for use as a reverse proxy to Home Assistant.
+### {% linkable_title Install %}
 
+<p class='note warning'>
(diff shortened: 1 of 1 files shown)


tjorim:
Previous PR here: https://github.com/home-assistant/home-assistant.io/pull/6813

frenck:
So basically this is for users running manual type of installations? (`venv`)?

How about Hass.io or Docker users?

ikifar2012:
> So basically this is for users running manual type of installations? (`venv`)?
> 
> How about Hass.io or Docker users?

havent looked into docker or hassio

ikifar2012:
> So basically this is for users running manual type of installations? (`venv`)?
> 
> How about Hass.io or Docker users?

What should I add?

frenck:
At this point, I'm actually unsure if we should add installation instructions for Caddy.
The problem with adding installation instructions for third-party programs and the many environments that Home Assistant provides/works with is a recipe for documentation hell.

Imaging all the components that involve third-party software providing installation instructions in the HA manual for each method... where does it end?

frenck:
Closing this PR because of the above reasoning. Adding installation instructions for third-party application will be a beast to maintain.

PR #7002, opened 2018-10-22
Clarity on how to use LLAT

DubhAd:
LLAT isn't obviously a long lived access token, and the intro leaves the reader hanging as if things are handled automatically. Adding some clarity.


fabaff (approved):
:bird:

PR #7013, opened 2018-10-22
change plex activity monitor category from media_player to sensor

quazzie:
**Description:**
Changes "Plex activity monitor" to Sensor category because it is a sensor.

### DIFF
--- a/source/_components/sensor.plex.markdown
+++ b/source/_components/sensor.plex.markdown
@@ -8,7 +8,7 @@ comments: false
 sharing: true
 footer: true
 logo: plex.png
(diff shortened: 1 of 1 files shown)


fabaff:
The category should match the topic the integration is covering and it's functionality.

quazzie:
@fabaff what about for example the zoneminder plugins? Shouldnt all of them be in the same category?
Or the amcrest IP camera sensor?
The plex monitor is not a media player, its a sensor.

quazzie:
It even sais "The plex sensor platform will monitor activity ..." in the description of the activity monitor

PR #7023, opened 2018-10-22
Update USPS configuration

tjorim:
**Description:**
Using new config vars
Add linkable titles
Remove trailing spaces
Related to #6385

## Checklist:


Vasiley:
This should not be merged ....USPS has been known to cause users to get banned from site

tjorim:
What does that issue have to do with updating the documentation according to the standards or not?

Rather than just not merging this PR this should be addressed elsewhere in my opinion.
Maybe someone could open up a PR which adds a warning for users.

Vasiley:
This componet does not work..what good does it do to update the docs

arsaboo:
This PR only corrects the formatting, which is fine by me. 

If the component is not working and needs to be taken down, that is a separate issue.

arsaboo (approved):
Merging it, as it fixes the current docs. We may decide to take down this component entirely at some point.

PR #7039, opened 2018-10-23
add note for hassOS boot folder

sholzmayer:
**Description:**
add note where to find boot/config file in hassOS, because it is pretty confusing...


frenck (changes requested):
I think I understand what you are trying to say.
This sentence is correct, in case the user has direct development/debug access to the host OS of HassOS. This is something we do not want to encourage. This file is editable directly on the SD as well, and that makes the file path specified in your addition incorrect and misleading.

sholzmayer:
ok yeah it should be just a DEV-note... so i guess we don't need it in the ZWave addon configuration.

PR #7083, opened 2018-10-25
TensorFlow image_processing component

hunterjm:
**Description:**

Adding a [TensorFlow Object Detection](https://github.com/tensorflow/models/blob/master/research/object_detection/README.md) image_processing component to run either pre-trained or custom built machine learning object detection models through TensorFlow and expose the results to Home Assistant.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#17795

## Checklist:


source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+footer: true
+logo: tensorflow.png
+ha_category: Image Processing
+featured: false
+ha_release: 0.82
+---
+
+The `tensorflow` image processing platform allows you to detect and recognize objects in a camera image using [TensorFlow](https://www.tensorflow.org/). The state of the entity is the number of objects detected, and recognized objects are listed in the `matches` attribute. The attribute provides the confidence `score` for recognition and the bounding `box` of the object for each detection category.
+
+## {% linkable_title Setup %}
```
frenck, line 23: Missing newline.
```suggestion
## {% linkable_title Setup %}

```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+TensorFlow is pre-configured in the [official home assistant docker image](https://hub.docker.com/r/homeassistant/home-assistant/).  If you are using this image, you can skip this section.
```
frenck:
```suggestion
TensorFlow is pre-configured in the [official Home Assistant Docker image](https://hub.docker.com/r/homeassistant/home-assistant/).  If you are using this image, you can skip this section.
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+
+TensorFlow is special in that it can run on the CPU, or an Nvidia GPU using CUDA.  Because of this, tensorflow is not listed in `REQUIREMENTS`.  In your instance, you will need to pre-install either `tensorflow` or `tensorflow-gpu` via pip:
```
frenck:
```suggestion
TensorFlow is special in that it can run on the CPU, or a Nvidia GPU using CUDA.  Because of this, TensorFlow is not listed in `REQUIREMENTS`.  In your instance, you will need to pre-install either `tensorflow` or `tensorflow-gpu` via pip:
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+
+```bash
+# CPU
+pip3 install tensorflow
+# GPU
+pip3 install tensorflow-gpu
+```
+
+Object detection also requires some dependancies outside of the core tensorflow package.  These live in the [tensorflow/models](https://github.com/tensorflow/models/tree/master/research/object_detection) GitHub repository.  The one we are interested in for this use case lives in `research/object_detection`.  Since the repository is large, we will be cloning in `/tmp` and only moving the files this component needs.  The rest can safely be deleted.  Run the following, making sure to replace `/home/homeassistant/.homeassistant/` with the path to your config directory:
```
frenck:
```suggestion
Object detection also requires some dependencies outside of the core TensorFlow package.  These live in the [tensorflow/models](https://github.com/tensorflow/models/tree/master/research/object_detection) GitHub repository.  The one we are interested in for this use case lives in `research/object_detection`.  Since the repository is large, we will be cloning in `/tmp` and only moving the files this component needs.  The rest can safely be deleted.  Run the following, making sure to replace `/home/homeassistant/.homeassistant/` with the path to your config directory:
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+
+# Cleanup
+rm -rf /tmp/*
+```
+
+Finally, you are ready to pick a model.  Researchers have already built multiple TensorFlow object detection models using many open data sets like [COCO](http://cocodataset.org), [Kitti](http://www.cvlibs.net/datasets/kitti/), and [Open Images](https://github.com/openimages/dataset).
+
+This component can support all of them, and if you are feeling adventurous, you can even [roll your own](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/defining_your_own_model.md).
+
+For now, start with one of the COCO models available in the [Model Detection Zoo](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/detection_model_zoo.md).  The trade-off here is accuracy vs speed.  The `mAP` is an indication of the model's accuracy while the `ms` is inference speed (on an Nvidia Titan X).  Most users won't see detection speeds close to what is listed.
```
frenck:
```suggestion
For now, start with one of the COCO models available in the [Model Detection Zoo](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/detection_model_zoo.md).  The trade-off here is accuracy vs speed.  The `mAP` is an indication of the model's accuracy while the `ms` is inference speed (on a Nvidia Titan X).  Most users won't see detection speeds close to what is listed.
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+
+It is recommended to start with the `faster_rcnn_inception_v2_coco` model for users with a dedicated server with a decent CPU.  If you are running on an ARM device like a Rasberry Pi, stick with a SSD MobileNet model like the `ssd_mobilenet_v2_coco` which was built for mobile devices.
```
frenck:
```suggestion
It is recommended to start with the `faster_rcnn_inception_v2_coco` model for users with a dedicated server with a decent CPU.  If you are running on an ARM device like a Rasberry Pi, stick with an SSD MobileNet model like the `ssd_mobilenet_v2_coco` which was built for mobile devices.
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+    type: list
+model:
+    description: Information about the TensorFlow model.
+    required: true
+    type: map
+    keys:
+        graph:
+            description: Full path to `frozen_inference_graph.pb`.
+            required: true
+            type: file
```
frenck: `file` is not a valid documentation type.
```suggestion
            type: string
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+        labels:
+            description: Full path to a `*label_map.pbtext`.
+            required: false
+            type: file
```
frenck: `file` is not a valid documentation type.
```suggestion
            type: string
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+            default: /usr/src/app/tensorflow/object_detection/data/mscoco_label_map.pbtxt
+        model_dir:
+            description: Full path to tensorflow models directory.
+            required: false
+            type: directory
```
frenck: `directory` is not a valid documentation type.
```suggestion
            type: string
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+            default: /usr/src/app/tensorflow
+        area:
+            description: Custom detection area.  Only objects fully in this box will be reported. Top of image is 0, bottom is 1.  Same left to right.
+            required: false
+            type: map
+            keys:
+                top:
+                    description: Top line defined as % from top of image.
+                    required: false
+                    type: small_float
```
frenck: `small_float` is not a valid documentation type.
```suggestion
                    type: float
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+                    default: 0
+                left:
+                    description: Left line defined as % from left of image.
+                    required: false
+                    type: small_float
```
frenck: `small_float` is not a valid documentation type.
```suggestion
                    type: float
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+                    default: 0
+                bottom:
+                    description: Bottom line defined as % from top of image.
+                    required: false
+                    type: small_float
```
frenck: `small_float` is not a valid documentation type.
```suggestion
                    type: float
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+                    default: 1
+                right:
+                    description: Right line defined as % from left of image.
+                    required: false
+                    type: small_float
```
frenck: `small_float` is not a valid documentation type.
```suggestion
                    type: float
```

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,180 @@
...
+                    default: 1
+        categories:
+            description: List of categories to include in object detection.  Can be seen in the file provided to `labels`.
```
frenck:
```suggestion
            description: List of categories to include in object detection. Can be seen in the file provided to `labels`.
```

frenck (dismissed):
@hunterjm I've reviewed your PR and left some initial comments.

I'm also kinda against the installation instructions for TensorFlow.
e.g.:
- Some people use `venv`?
- Some people Hass.io?
- How about none 64 bits installations? ARM64? 
- What if versions need upgrading? (e.g., on `protobuf`)? Who maintains the docs?

I think in general, we should avoid being an instruction manual for third-party software.

If you could take a look? 👍

hunterjm:
@frenck - Thank you for the feedback.  The detailed instructions are for those using hassbian or venv.  @arsaboo is running a venv and they were initially put together to help him set it up.  If approved, I would eventually plan to have it pre-configured in the Hass.io image as well.

Currently I am running Hass.io on a NUC8 and have just replaced the home-assistant image with one that was built from my modifications to the Dockerfile.  Google does not release a TensorFlow wheel for alpine linux, so that installation becomes more complicated, but I believe it can be tackled as a separate PR.  They do have ARM images compatible with the Raspberry Pi.

The install script is specific to x86_64.  Maybe instead of providing a script I can outline the dependencies?  I felt that leaving it out entirely would create too much confusion, and it really isn't that difficult to setup, but you would need to sort through multiple documents in the tensorflow/models repo to figure it out yourself.

As far as upgrades/versioning, protobuf is just used to build the python models from the protos since TensorFlow is a multi-language library.  We could pin the commit to download, the Detection Model Zoo version, and the TensorFlow wheel version in the documentation to have a working version as long as those resources are still available, even if some get updated.

Thoughts?

arsaboo:
I think for this PR, we can include brief instructions. We only need instructions for Tensorflow installation (based on the system) and then list the files/dependencies that are required. The detailed instructions (e.g., `mv` commands) can be provided in a gist, like we do for [OpenCV](https://www.home-assistant.io/components/image_processing.opencv/).

hunterjm:
What about simply linking to [this](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/installation.md)? Then providing basic instructions on what to do after?

The problem is only half of that is relevant for this use case.

robmarkcole:
I agree that HA specific instructions are preferable to linking to an external site with extraneous info which will cause confusion. However these instructions could be hosted on a gist as @arsaboo suggested. For opencv I usually `pip install opencv-python`, wouldn't hurt to add?

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,160 @@
...
+{% endconfiguration %}
+
+`categories` can also be defined as dictionary providing an `area` for each category as seen in the advanced configuration below:
+
+```yaml
+# Example advanced configuration.yaml entry
+image_processing:
+  - platform: tensorflow
+    source:
+      - entity_id: camera.local_file
```
arsaboo: Since this is an advanced config example, let us add more more camera here to show users how to use this with multiple cameras.

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,160 @@
...
+
+A sample script can be found at [this gist](https://gist.github.com/hunterjm/6f9332f92b60c3d5e448ad936d7353c3).
+
+## {% linkable_title Model Selection %}
+
+Finally, you are ready to pick a model.  Researchers have already built multiple TensorFlow object detection models using many open data sets like [COCO](http://cocodataset.org), [Kitti](http://www.cvlibs.net/datasets/kitti/), and [Open Images](https://github.com/openimages/dataset).
+
+This component can support all of them, and if you are feeling adventurous, you can even [roll your own](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/defining_your_own_model.md).
+
+For now, start with one of the COCO models available in the [Model Detection Zoo](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/detection_model_zoo.md).  The trade-off here is accuracy vs speed.  The `mAP` is an indication of the model's accuracy while the `ms` is inference speed (on a Nvidia Titan X).  Most users won't see detection speeds close to what is listed.
```
arsaboo: Is this too much detail for users? Can we just refer them to the Model Detection Zoo and recommend the `faster_rcnn_inception_v2_coco` or `ssd_mobilenet_v2_coco` models to get started. Then, they can experiment.

hunterjm: I guess that depends on if we want to give people the resources to research how it works, or just a dumbed down version to get it working in Home Assistant.

arsaboo: The Model detection zoo website has more information than we can possibly provide users in our documentation. So, I feel that giving them a good starting point should be the objective of this docs.

hunterjm:
Comments addressed.  I also added a warning to let Hass.io users know that this component will not work out of the box with their setup, and linked to the system requirements of TensorFlow.

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,161 @@
...
+
+A sample script can be found at [this gist](https://gist.github.com/hunterjm/6f9332f92b60c3d5e448ad936d7353c3).
+
+## {% linkable_title Model Selection %}
+
+Lastly, it is time to pick a model.  It is recommended to start with one of the COCO models available in the [Model Detection Zoo](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/detection_model_zoo.md).
+
+The trade-off between the different models is accuracy vs speed.  Users with a decent CPU should start with the `faster_rcnn_inception_v2_coco` model.  If you are running on an ARM device like a Rasberry Pi, start with the `ssd_mobilenet_v2_coco` model.
+
+Whichever model you choose, download it and place the `frozen_inference_graph.pb` file somewhere in your config directory.
```
robmarkcole: Perhaps its worth recommending somewhere specific rather than `somewhere`, perhaps in the `tensorflow` dir we created earlier?

hunterjm:
Finished testing on Hassbian - updated warning to reflect additional package requirements.

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,161 @@
...
+
+The `tensorflow` image processing platform allows you to detect and recognize objects in a camera image using [TensorFlow](https://www.tensorflow.org/). The state of the entity is the number of objects detected, and recognized objects are listed in the `summary` attribute along with quantity. The `matches` attribute provides the confidence `score` for recognition and the bounding `box` of the object for each detection category.
+
+<p class='note warning'>
+  This package will not yet work on Hass.io installations, but should work on Hassbian.  [See here](https://www.tensorflow.org/install/pip) for system requirements.
+</p>
+
+## {% linkable_title Setup %}
+
+TensorFlow object detection requires some dependencies outside of the core tensorflow package.  These live in the [tensorflow/models](https://github.com/tensorflow/models/tree/master/research/object_detection) GitHub repository.  The one we are interested in for this use case lives in `research/object_detection`.  A general outline of the installation process is as follows:
```
robmarkcole: I find this instruction a bit confusing myself. Perhaps for clarity something like: 

`
This component requires files to be downloaded, compiled on your computer, and added to the Home-Assistant configuration directory. These steps are can be performed using the sample script at this gist. Alternatively, if you wish to perform the process manually, the process is as follows:
`

robmarkcole:
@hunterjm I added a couple of minor comments. I wouldn't worry too much about getting these docs perfect, as we will get a lot of useful feedback when the component is released to the community. 

One final comment, we are here using `score` rather than `confidence` (a configurable for the face recognition components, also the metric reported by Facebox). I think we should compile a list of these terms and their meanings in the [platform page](https://www.home-assistant.io/components/image_processing/). I see `score` as being the output from the model, whilst `confidence` being a threshold we define as being meaningful. @arsaboo what do you think of this suggestion?

robmarkcole:
LGTM @frenck @arsaboo

arsaboo:
Looks good to me.

arsaboo (dismissed):

source/_components/image_processing.tensorflow.markdown
```diff
@@ -0,0 +1,163 @@
...
+The `tensorflow` image processing platform allows you to detect and recognize objects in a camera image using [TensorFlow](https://www.tensorflow.org/). The state of the entity is the number of objects detected, and recognized objects are listed in the `summary` attribute along with quantity. The `matches` attribute provides the confidence `score` for recognition and the bounding `box` of the object for each detection category.
+
+<p class='note warning'>
+  To work with Hassbian after following the setup and installing these packages:
+  `$ sudo apt-get install libatlas-base-dev libopenjp2-7 libtiff5`
+</p>
+
+## {% linkable_title Setup %}
+
+You need install `tensorflow` Python packages with: `$ pip3 install tensorflow`. The wheel are not available for all platforms, it could be that you need compile it from source. Hass.io have this already preinstalled.
```
arsaboo: Since we are now providing TF install instructions, would it make sense to also include the command for Nvidia one as well. 

You need install `tensorflow` Python packages with: `pip3 install tensorflow` or `pip3 install tensorflow-gpu` (for those who use a Nvidia GPU).

hunterjm: Resolved with a link to the official install page on tensorflow.org

arsaboo:
Looks good to me. 

@frenck comments are addressed as well (as most of those are install related). I will merge this tomorrow, unless there are some residual concerns.

arsaboo (approved):

PR #7120, opened 2018-10-26
Update sensor.influxdb.markdown configuration

hmmbob:
**Description:**
Update style of Influxdb documentation to follow new configuration variables description.
Related to #6385

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** n/a

## Checklist:

### DIFF
--- a/source/_components/sensor.influxdb.markdown
+++ b/source/_components/sensor.influxdb.markdown
@@ -27,24 +27,77 @@ sensor:
         measurement: '"°C"'
 ```
 
+{% configuration %}
+host:
+  description: IP address of your database host, eg. 192.168.1.10.
+  required: false
+  default: `localhost`
+  type: string
+port:
+  description: Port to use. 
+  required: false
+  default: 8086.
+  type: string
+username:
+  description: The username of the database user.
+  required: false
+  type: string
+password:
+  description: The password for the database user account.
+  required: false
+  type: string
+ssl:
+  description: Use `https` instead of `http` to connect. 
+  required: false
+  default: `false`.
+  type: boolean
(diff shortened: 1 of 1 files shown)


source/_components/sensor.influxdb.markdown
```diff
@@ -27,24 +27,77 @@ sensor:
         measurement: '"°C"'
 ```
 
+{% configuration %}
+host:
+  description: IP address of your database host, eg. 192.168.1.10.
```
klaasnicolaas: Must be: `e.g.,`

source/_components/sensor.influxdb.markdown
```diff
@@ -27,24 +27,77 @@ sensor:
...
+  required: false
+  default: `localhost`
```
klaasnicolaas: the error arises in my opinion because you start a value with ` do this in between "" such as: "`localhost`" 😉

klaasnicolaas (changes requested):

source/_components/sensor.influxdb.markdown
hmmbob: Big thanks, I was looking for tabs etc etc.....

hmmbob: Or should I just drop the '-symbols at all?

klaasnicolaas: Maybe even better 😉 (but only with the defaults)

frenck (approved):

frenck:
Thanks, @hmmbob!

Happy Hacktoberfest 🎉

PR #7128, opened 2018-10-26
Fix Thermostat card config

arsaboo:


thomasloven (approved):

balloob (approved):

HadiHassaoui:
Hi, have a small glitch in the thermostat card.When  the background of the card has transparency (ex: rgba(0, 0, 0, 0.6)), since the "diamond" that mask the bottom part of the gauge use the same color "paper-card-background-color", the result is not good. 

See the PNG attached for what I'm saying....
capture d ecran 2018-11-13 00 07 12

thomasloven:
Please don't comment on closed or merged pull requests. Open an issue instead.

PR #7140, opened 2018-10-27
Fix MQTT topic and ON/OFF displaying as true/false

rwagoner:
Fix MQTT documentation so ON and OFF are correctly displayed. These currently show as true and false. Also fix MQTT topic that had a leading forward slash.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7166, opened 2018-10-27
Remove incorrect truth table

emlove:
**Description:**
It turns out this truth table wasn't correct for the actual implementation of sun conditions, and was causing confusion. I don't think having columns with a mix of arrows and "+1s" and ambiguous headers like "night/at sunrise" provides additional clarity anyway.


DubhAd (approved):
👍

PR #7197, opened 2018-10-28
Adding configurator note

DubhAd:
Now it can be set up with the integration, we probably should mention that ;)


fabaff (approved):
Thanks :bird:

PR #7214, opened 2018-10-28
Update weather.yweather.markdown

bonanitech:
Include a note warning users about the Yahoo Weather API retirement.


fabaff (approved):
Thanks :bird:

PR #7216, opened 2018-10-28
Update NO-IP docs to clarify what it does and doesn't do

lindsaymarkward:
**Description:**
When I first installed this component, I hoped it would automatically renew/confirm my free hostname, but it doesn't. I've added that detail to the docs to make it clearer what it does.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7249, opened 2018-10-29
Re-add truth table for sun conditions

emontnemery:
**Description:**
Re-add truth table for sun conditions which was removed in 44ec6aeaca2d3c87f168adf344e9d783cde61e96, but align with the implementation

## Checklist:

### DIFF
--- a/source/_docs/scripts/conditions.markdown
+++ b/source/_docs/scripts/conditions.markdown
@@ -149,7 +149,7 @@ condition:
         before: sunrise
 ```
 
-Here is a truth table to clarify the parameters with and without offset:
+Here is a truth table to clarify the conditions with and without offset:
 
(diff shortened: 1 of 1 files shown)


emlove:
I'd rather not re-add the table myself. The reason I got rid of it is because a table just adds confusion rather than making the description easier to understand.

I'm planning to create a timeline graphic in the next few days that illustrate when the condition will be true.

emontnemery:
@armills The table you removed in 44ec6ae was not aligned with the implementation. The table in this PR is correct. Please also note the additional column to make it clearer that the conditions are reset at midnight.

I agree some graphic would be nice though.

emlove:
OK, I opened a PR that adds a graphic alternative to the info in this table: https://github.com/home-assistant/home-assistant.io/pull/7262

emontnemery:
Closing this, the graphic is clearly much better!

PR #7262, opened 2018-10-30
Add sun conditions timeline graphic

emlove:
## Description:
This PR adds a chart which shows an example of when the sun conditions will be true.

I've also added the document I used to generate this image, in case we need to change this in the future. I'm not sure if we have somewhere specific we commit these files, so for the moment it's adjacent to the actual image file.

**Screenshot:**

image

## Checklist:


emontnemery:
This is great! What tool did you use to generate the graph?

balloob (approved):

emlove:
I used Libreoffice Calc, and included the source document in this PR. There's a function in Calc to export graphs as an image which I used to generate the SVG.

PR #7267, opened 2018-10-30
Add example sun condition which is true during daytime

emontnemery:
**Description:**
Add example sun condition which is true during daytime

## Checklist:

### DIFF
--- a/source/_docs/scripts/conditions.markdown
+++ b/source/_docs/scripts/conditions.markdown
@@ -139,6 +139,16 @@ condition:
         before: sunrise
 ```
 
+```yaml
+condition:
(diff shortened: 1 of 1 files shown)


source/_docs/scripts/conditions.markdown
```diff
@@ -139,6 +139,16 @@ condition:
         before: sunrise
 ```
 
+```yaml
+condition:
+    condition: and  # 'when light' condition: before sunset and after sunrise
```
thomasloven, line 144: This is not right.
The condition says After sunset and Before sunrise.

thomasloven (changes requested):

thomasloven (approved):

PR #7271, opened 2018-10-30
Update Discovery component configuration

klaasnicolaas:
**Description:**
Update style of Discovery component documentation to follow new configuration variables description.
Related to #6385.

## Checklist:

### DIFF
--- a/source/_docs/mqtt/discovery.markdown
+++ b/source/_docs/mqtt/discovery.markdown
@@ -33,10 +33,19 @@ mqtt:
   discovery: true
   discovery_prefix: homeassistant
 ```
-Configuration variables:
 
-- **discovery** (*Optional*): If the MQTT discovery should be enabled or not. Defaults to `False`.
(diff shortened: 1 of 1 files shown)


nwesterhausen:
You should go ahead modify the example configuration while you're at it. The example should be as generic as possible (so it should not include optional configuration). Typically you can just re-label the existing one as a configured example or advanced configuration example and provide a simpler _# Example configuration.yaml entry_

klaasnicolaas:
I will first let @fabaff or @frenck look to it, to decide if this is really necessary in this case

ludeeus:
For most cases, I would agree with what @nwesterhausen said since that is the standard.
But in this case, it makes sense to have them there since this is a subpage for `MQTT` with "advanced" configuration.

_Note: Feel free to disregard my views :D_

fabaff (approved):
Thanks :bird:

fabaff:
Let's keep this PR to the configuration variable update and update the examples later if needed.

PR #7288, opened 2018-10-30
Hint on lxml installation

akloeckner:
**Description:**
Manual lxml installation s required for some reason. I added this to the docs.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** none

## Checklist:


cgtobi:
Since `lxml` is a dependency of `fritzconnection` it should be installed automatically. No need to install it by hand. If that would be required it should rather be fixed upstream and only be mentioned as a intermediate workaround

source/_components/sensor.fritzbox_callmonitor.markdown
```diff
@@ -27,6 +27,13 @@ $ sudo apt-get install libxml2-dev libxslt-dev \
   python3-setuptools zlib1g-dev build-essential
 ```
 
+If you installed Home Assistant in a virtualenv, also run the following command inside it. 
+Be patient this will take a while.
+```bash
+pip3 install lxml
+```
+
+
```
frenck: One empty line to much...

source/_components/sensor.fritzbox_callmonitor.markdown
```diff
@@ -27,6 +27,13 @@ $ sudo apt-get install libxml2-dev libxslt-dev \
   python3-setuptools zlib1g-dev build-essential
 ```
 
+If you installed Home Assistant in a virtualenv, also run the following command inside it. 
+Be patient this will take a while.
+```bash
+pip3 install lxml
```
frenck, line 33: Why not simply add this to the instructions above?

frenck (changes requested):
@akloeckner Thanks for the PR! I've left some comments, could you take a look?

Secondly, the PR should go against the `current` branch, since this changes the documentation of our current release. So please change the target and rebase the PR if needed.

Third, just like @cgtobi mentioned... this should not be needed?

source/_components/sensor.fritzbox_callmonitor.markdown
akloeckner: fixed by new commit

source/_components/sensor.fritzbox_callmonitor.markdown
akloeckner: A user might be tempted to run all the commands in a sequence. But this is not possible, because he would have to bin/activate the VirtualEnv in between first. Also, this additional hint apparently is valid only for VirtualEnv installations. I cannot explain why. But it's stated in the links I'll post in my general answer below. 

I suggest to keep it separate unless we (i.e. not me) know better.

PR #7320, opened 2018-11-01
hint on lxml installation

akloeckner:
**Description:** I messed up my branches in trying to move this to the current branch... So, well... here goes the replacement PR for #7288... Please see there also for a few previous discussion posts...

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** none

## Checklist:

### DIFF
--- a/source/_components/sensor.fritzbox_callmonitor.markdown
+++ b/source/_components/sensor.fritzbox_callmonitor.markdown
@@ -27,6 +27,12 @@ $ sudo apt-get install libxml2-dev libxslt-dev \
   python3-setuptools zlib1g-dev build-essential
 ```
 
+If you installed Home Assistant in a virtualenv, also run the following command inside it. 
+Be patient this will take a while.
+```bash
+pip3 install lxml
+```
+
 ## {% linkable_title Setup%}
 
 To activate the call monitor on your Fritz!Box, dial #96\*5\* from any phone connected to it.


akloeckner:
**First:** @frenck, I addressed your requests in the dialogs, then failed to change the branch. The remarks are adressed in this new PR and below. As to the merging of the two code blocks, I repeat my comment here to have it all in one PR: 

A user might be tempted to run all the commands in a sequence. But this is not possible, because he would have to bin/activate the VirtualEnv in between first. Also, this additional hint apparently is valid only for VirtualEnv installations. I cannot explain why. But it's stated in the links I'll post in my general answer below.

I suggest to keep it separate unless we (i.e. not me) know better.

**Second:** Ok... Here's the result of this process...

**Third:**
@cgtobi you *should* be right. But unfortunately, for some reason, this dependency failed to install. Yes, the dependency install apparently started, as I could see in my `top` processes, which showed a `ccl` being executed by homeassistant for a long time. (The exact same thing that happens when installing manually.)  But I waited until it finished and then had errors about the failed dependency in my log. (Unfortunately, I rebooted, so the logs are gone.) I don't know, maybe there is a timeout on the dependency installation?

However, I am not the only one with this problem. I did find (at least) the following sources for the procedure I propose to document here:
1. The device tracker doc contains the exact same procedure: https://www.home-assistant.io/components/device_tracker.fritz/
2. The netmonitor also points this out: https://www.home-assistant.io/components/sensor.fritzbox_netmonitor/
3. People are being pointed to this documentation: https://github.com/home-assistant/home-assistant/issues/1323

In summary: This procedure helped me set up the component. The procedure described in the docs did not. A working procedure should be described in the docs. Hence the PR.

frenck (approved):

frenck:
Thanks, @akloeckner!

Let's merge it in.

PR #7329, opened 2018-11-02
import appdaemon.appapi outdated

CrazYoshi:
appdaemon.appapi as appapi changed to import appdaemon.plugins.hass.hassapi as hass

**Description:**
import in AppDaemon example changed to appdaemon.plugins.hass.hassapi as hass


fabaff (approved):
Thanks :bird:

PR #7335, opened 2018-11-02
Changed on/off or yes/no to true/false

klaasnicolaas:
**Description:**
Title says more then enough.

## Checklist:

### DIFF
--- a/source/_components/group.markdown
+++ b/source/_components/group.markdown
@@ -12,7 +12,7 @@ ha_category: Organization
 ha_qa_scale: internal
 ---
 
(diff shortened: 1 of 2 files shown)


fabaff (approved):
Thanks :bird: 

From a YAML point of view it was [ok](http://yaml.org/type/bool.html), for consistency in the documentation it's better.

klaasnicolaas:
What do you mean exactly? @fabaff

frenck:
@fabaff From a YAML point of view, it is no longer OK.
YAML 1.2 spec does not longer support `yes`/`no`, for example, our YAML parser used for Lovelace is already using that spec.

http://yaml.org/spec/1.2/spec.html#id2803629

PR #7354, opened 2018-11-04
add note to configure port if deCONZ is not running on port 80

chbndrhnns:
**Description:**
On the deCONZ docs, a note is missing how to configure it when using a non-default port (for example when using the [docker container](https://github.com/marthoc/docker-deconz).

This note has been added.

## Checklist:


Kane610:
Thanks for PR!

You're changing stuff in this PR that is not relevant, and also wrong.

There is a PR when discovery fails will allow manual config through the frontend. The components documentation shall not mention the configuration.yaml.

Also; if you have configured the container correctly deconz will get the correct port through discovery.

Closing this PR.

PR #7356, opened 2018-11-04
Adding some documentation for the services

DubhAd:
As nowhere do we document `homeassistant.update_entity`, or the other `homeassistant` services, it seems overdue that we actually document them.


fabaff (approved):
:bird:

PR #7376, opened 2018-11-05
additional warning for broken dependency

wattsra:
**Description:**
The dependency "locationsharinglib" is currently broken and does not pick up shared locations. 

A fix in the works, but not verified or merged yet. This bug has been present for over a month so wiki should be updated to reflect the issue so as not to confuse new users.

Only one file changed: source/_components/device_tracker.google_maps.markdown.

Added in single line: 
This component is currently broken (November 2018, v0.81.6) due to an upstream dependency bug, see here: https://github.com/costastf/locationsharinglib/issues/42 and here: https://github.com/home-assistant/home-assistant/issues/17410 

## Checklist:

### DIFF
--- a/source/_components/device_tracker.google_maps.markdown
+++ b/source/_components/device_tracker.google_maps.markdown
@@ -12,6 +12,10 @@ ha_release: 0.67
 ha_category: Presence Detection
 ha_iot_class: "Cloud Polling"
 ---
+<p class='note warning'>
+This component is currently broken (November 2018, v0.81.6) due to an upstream dependency bug, see here: https://github.com/costastf/locationsharinglib/issues/42 and here: https://github.com/home-assistant/home-assistant/issues/17410 
+</p>
+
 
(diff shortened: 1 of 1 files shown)


fabaff:
Sorry, we are not going to track issues in the documentation as long as somebody is working on the problem.

Like with the changelog if it's not automated it's not going to happen.

wattsra:
Ok. If that's the agreed approach. It's been going on quite a while by the looks of it. So I was hoping to ease the path of newbies but I'm ok.

WedHumpDay:
So, Hass.io users are getting no love. This is still broken under 0.82.0.

wattsra:
@WedHumpDay it was mentioned above by fabaff that documentation is not updated to show that a component is currently broken. 

As I said above , I don't agree with that strategy, but I am willing to accept that if it is the agreed precedent.

frenck:
@WedHumpDay "Hass.io users are getting no love"? Seriously?

I'm locking down further discussion on this PR. If you want to discuss the issue, do it in the issue created for it. Thanks.

PR #7452, opened 2018-11-10
Alexa: proactive api config

abmantis:
**Description:**
Add config docs for enabling Alexa's proactive API.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#18114

## Checklist:

### DIFF
--- a/source/_components/alexa.markdown
+++ b/source/_components/alexa.markdown
@@ -376,13 +376,17 @@ alexa:
       switch.stairs:
         display_categories: LIGHT
 ```
-The `endpoint`, `client_id` and `client_secret` are optional, and are only required if  you want to enable Alexa's proactive mode.
 This exposes an HTTP POST endpoint at `http://your_hass_ip/api/alexa/smart_home`
 which accepts and returns messages conforming to the
 [Smart Home v3 payload](https://developer.amazon.com/docs/smarthome/smart-home-skill-api-message-reference.html).
 You must then create an Amazon developer account with an Alexa skill and Lambda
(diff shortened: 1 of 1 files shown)


pc-coholic:
Hi @abmantis,

thank you for integrating the proactive notifications - a feature I was hoping for :-)

I just tested it and it works pretty well - however I would like to propose to add a little bit more information to the documentation regarding enabling the proactive ChangeReports:
- Document that there are different endpoints available for different regions. Either have them directly in the documentation or link to https://developer.amazon.com/docs/smarthome/send-events-to-the-alexa-event-gateway.html#endpoints
- Explain, that the client_id and client_secret are *not* the ones used by the skill that have been set up using "Login with Amazon" (in the Alexa Developer Console: Build > Account Linking), but rather from the "Alexa Skill Messaging" (in the Alexa Developer Console: Build > Permissions > Alexa Skill Messaging).
- In order to get mentioned client ID and secret, the "Send Alexa Events"-permission must be checked.
- If  the "Send Alexa Events" was not enabled before, the user needs to unlink and relink the skill/account using the Alexa-App - else it won't work (`Token invalid and no refresh token available.`)

Best regards,
Martin

abmantis:
@pc-coholic Thank you for the suggestions, Marting, I've updated the docs with them.

balloob:
It seems like you have some commits in this branch that are not yours, can you please it up.

abmantis:
It is fixed now.

balloob (approved):

PR #7456, opened 2018-11-10
Note about needed doorbird user's permission

bayi:
**Description:**

This adds a note about a user permission which is needed to be enabled on the doorbird before home assistant can use the device. This only applies to 0.82+, everything worked without this permission in older versions.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7465, opened 2018-11-11
Fix typo in Lametric notify setup example

deftdawg:
* fix typo in Lametric notify setup example

### DIFF
--- a/source/_components/notify.lametric.markdown
+++ b/source/_components/notify.lametric.markdown
@@ -59,7 +59,7 @@ Check out the list of all icons at [https://developer.lametric.com/icons](https:
 ```yaml
 # Example configuration.yaml entry
 notify:
-  name: NOTIFIER_NAME
+  name: lametric
   platform: lametric
   lifetime: 20
   icon: a7956


frenck:
@jumpkick Thank you for opening a PR. First, please make changes to current documentation to the `current` branch. The `next` branch is for the upcoming releases, which is used for "new" documentation. I've changed the target for you already.

Secondly, this is not a typo. Capitalized text like `API_TOKEN` indicates something a user should replace. In this case, a user should define/give the name of their likings.

Therefore, I'm going to decline and close the PR.

deftdawg:
@frenck Sorry I misfiled the pull, the pull request checklist says to use "next" for documentation changes, but perhaps that's only for new features.

Regarding the typo... I found the page a bit confusing because if you look at the [entire page](https://www.home-assistant.io/components/notify.lametric/), it shows an configuration example at the top of the page with NOTIFICATION_NAME... 

Then towards the bottom of the page there is a second "full" configuration example in which the only field not replaced with sample data is NOTIFICATION_NAME.   This is what I changed, because it seemed inconsistent and it allowed the sample implementations below it(all of which replaced NOTIFICATION_NAME with "lametric") to work properly.

If you think it's best left the way it is, that's okay.  🥂

frenck:
The PR template states:

> Branch: next is for changes and new documentation that will go public with the next home-assistant release. Fixes, changes and adjustments for the current release should be created against current.

Which is pretty clear to me, if you have a suggestion on improving it, feel free to open a PR.

The initial examples are always bare minimal examples (no optional fields, unless the optional field is almost always needed), with placeholders (migration to this format is still in progress). 

Additional examples are allowed to be larger and may contain sample data. This specific page follow that format perfectly fine. So lets leave it like so.

PR #7490, opened 2018-11-13
Add deconz entry manually

kirbby:
A few days ago I tried for a few hours to get deconz to work with the auto discovery and didn't succeed.
I found a ticket for a bug in the deconz github, that the discovery request always returns an empty array. Which will probably be resolved in the future and is most likely the cause of it not working for me.
But I think there would be no harm in adding the manual way of including the deconz component in the configuration.yaml

**Description:**

## Checklist:

### DIFF
--- a/source/_components/deconz.markdown
+++ b/source/_components/deconz.markdown
@@ -34,6 +34,15 @@ If you don't have the API key, you can generate an API key for deCONZ by using t
 
 You can manually add deCONZ by going to the integrations page.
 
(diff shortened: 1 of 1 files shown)


Kane610:
Thanks for PR. There will be a fallback from the front end which will allow you to fill host and port directly if the discovery fails. Coming next release so there is no reason to add this.

kirbby:
Ok, sounds good. Thanks 

Am 13.11.2018 20:18, schrieb Robert Svensson:

> Thanks for PR. There will be a fallback from the front end which will allow you to fill host and port directly if the discovery fails. Coming next release so there is no reason to add this. 
> 
> --
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].

  

Links:
------
[1]
https://github.com/home-assistant/home-assistant.io/pull/7490#issuecomment-438401130
[2]

PR #7510, opened 2018-11-15
Update device_tracker.aruba.markdown

JDW2018:
**Description:**

This is an update to the Aruba device tracker to allow it to connect and interface with both IAP's and AP's. This allows devices to be tracked around the home depending on which access point they are connected

In the configuration the type option is optional to make sure this is not a breaking change. by selection AP it changes from the current IAP mode to AP mode

Home assistant PR Number https://github.com/home-assistant/home-assistant/pull/18490

## Checklist:

### DIFF
--- a/source/_components/device_tracker.aruba.markdown
+++ b/source/_components/device_tracker.aruba.markdown
@@ -1,7 +1,7 @@
 ---
 layout: page
 title: "Aruba"
-description: "Instructions on how to integrate Aruba routers into Home Assistant."
+description: "Instructions on how to integrate Aruba access points into Home Assistant."
 date: 2015-08-31 08:45
 sidebar: true
 comments: false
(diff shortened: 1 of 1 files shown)


frenck:
Not stale, parent PR is still open.

source/_components/device_tracker.aruba.markdown
```diff
@@ -47,6 +60,10 @@ password:
   description: The password for your given admin account.
   required: true
   type: string
+type:
+  description: type of device that you want to connect to.
```
DubhAd, line 64: _The type of device that you want to connect to_

However, what are the options? From the example above we can guess that `AP` is one

DubhAd (changes requested):

frenck:
Closing this PR, since the parent PR has been closed.

PR #7517, opened 2018-11-15
hassio instalation: Addition to required packages

efratomer:
**Description:**
Add docker.io to required packages list

## Checklist:


frenck:
This is actually incorrect. You should NOT use that package (system provided = outdated and different).

Install docker according to the instructions provided by Docker.

PR #7520, opened 2018-11-15
Change category to Sensor for WaterFurnace

mdallaire:
**Description:**
Changed the category to "Sensor" since a geothermal system is not related at all with irrigation and, since this component does not support the thermostat function, it should probably not be put in the "Climate" category.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7524, opened 2018-11-16
Update alert.markdown

point-4ward:
Added raw tags for template in last example as the templates weren't displaying correctly.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7561, opened 2018-11-17
Fix then/than typo in climate.generic_thermostat

ekeih:
**Description:** The documentation of climate.generic_thermostat used `then` to compare temperatures. I replaced both occurrences with the correct term `than`.
Reference for the usage of `then vs. than`: https://writingexplained.org/then-vs-than-difference

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7563, opened 2018-11-17
add example for climate device

dhzl84:
**Description:**
based on HA 0.81.6
current_temperature_template has no abbreviation according to the docs

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** not applicable

## Checklist:

### DIFF
--- a/source/_docs/mqtt/discovery.markdown
+++ b/source/_docs/mqtt/discovery.markdown
@@ -244,3 +244,41 @@ Setting up a switch using topic prefix and abbreviated configuration variable na
 - Command topic: `homeassistant/switch/irrigation/set`
 - State topic: `homeassistant/switch/irrigation/state`
 - Payload:  `{"~": "homeassistant/switch/irrigation", "name": "garden", "cmd_t": "~/set", , "stat_t": "~/state"}`
+
+Setting up a climate component (heat only) with abbreviated configuration variable names to reduce payload length.
+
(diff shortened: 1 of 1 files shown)


frenck (approved):

frenck:
Thanks, @dhzl84!
Nice first-time contribution to our documentation 👍

dhzl84:
Somehow the JSON templates for the state topic are missing in the [current version](https://www.home-assistant.io/docs/mqtt/discovery/) of the docs, intentional?

frenck:
@dhzl84 State topic = optional.

dhzl84:
Shouldn't we then also remove the state topic payload. In the current state I'd consider the example as incomplete and misleading for people facing discovery for the first time. Personnaly I'd prefer to have a complete real world example.

frenck:
@dhzl84 I'm not sure where you are heading, all of those are valid real-world examples.
It is not wrong to set optional fields. Some people prefer to be explicit by defining all fields, other like to configure only the minimal needed sets to keep stuff clean.

So that is just a matter of opinion I guess.

dhzl84:
Personnaly I just want to keep things as simple and obvious as possible. If one would just copy & paste the example, it would not work cause the mentioned state topic is of JSON format but the templates do not "decode" it.
If this is not the goal of the docs then I'm fine with it as well.

PR #7579, opened 2018-11-19
Add documentation for the Awair sensor platform

ahayworth:
**Description:**

This PR adds documentation for the Awair sensor platform.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#18570

## Checklist:


source/_components/sensor.awair.markdown
```diff
@@ -0,0 +1,68 @@
...
+title: "Awair Air Quality Monitor"
+description: "Instructions on how to setup Awair devices in Home Assistant."
+date: 2018-11-18 20:41
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: awair.jpg
+ha_category: Health
+ha_release: 0.83
```
MartinHjelmare: 0.84

source/_components/sensor.awair.markdown
```diff
@@ -0,0 +1,68 @@
...
+      - uuid: UUID
+```
+
+{% configuration %}
+access_token:
+  description: The access token for the Awair API.
+  required: true
+  type: string
+devices:
+  description: An optional list of devices to manually configure rather than relying upon API discovery.
```
MartinHjelmare: ... to manually configure devices rather than...

source/_components/sensor.awair.markdown
ahayworth: :+1: 

Aside - is there any documentation around what the release cadence is? I haven't found any yet. It's not a problem or anything, just seeking to understand - for example, we're on 0.82.1 publicly, but we're starting to tag already for 0.84.

source/_components/sensor.awair.markdown
ahayworth: I changed it a bit - I wasn't exactly sure what you were trying to express, though.

source/_components/sensor.awair.markdown
MartinHjelmare: We have a two week release cycle where the beta goes out one week before release. So it's two weeks between releases and two weeks between betas. The beta release for 0.83 was cut Wednesday, and we don't add new features (normally) after the beta cut, only fixes. So that's why this feature will go into 0.84.

This blog post kinda describes it, but it's a bit inaccurate since we still have a two week release cycle:
https://www.home-assistant.io/blog/2018/03/24/new-release-schedule/

ahayworth: Thanks!

MartinHjelmare (approved):

PR #7595, opened 2018-11-21
Added scan_interval to documentation

shafr:
If scan_interval is not specified - it causes issues with system. Specifying this parameter seems to fix this.

**Description:**

## Checklist:


fabaff:
`scan_interval` is covered in the documentation. This change would overwrite the default of the sensor of 120 seconds. If there is an issue, with the interval then this should be fixed in the code.

PR #7642, opened 2018-11-25
The current text assumes your user is always "homeassistant"

dimitripb:
If you follow the advanced guide to setup a virtual environment a user named "homeassistant" or whatever you want will be created. See:
https://www.home-assistant.io/docs/installation/raspberry-pi/
If you follow this guide the "homeassistant" user won't be created:
https://www.home-assistant.io/docs/installation/virtualenv
This page actually refers to the Autostart page on which is assumed that you created a user named "homeassistant". It's better to refer to the user which runs HA in general.

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/autostart/systemd.markdown
+++ b/source/_docs/autostart/systemd.markdown
@@ -21,8 +21,8 @@ If the preceding command returns the string `systemd`, continue with the instruc
 A service file is needed to control Home Assistant with `systemd`. The template below should be created using a text editor. Note, root permissions via `sudo` will likely be needed. The following should be noted to modify the template:
 
 - `ExecStart` contains the path to `hass` and this may vary. Check with `whereis hass` for the location.
-- For most systems, the file is `/etc/systemd/system/home-assistant@[your user].service` with [your user] replaced by the user account that Home Assistant will run as (normally `homeassistant`).  In particular, this is the case for Ubuntu 16.04.
-- If unfamiliar with command-line text editors, `sudo nano -w [filename]` can be used with `[filename]` replaced with the full path to the file.  Ex. `sudo nano -w /etc/systemd/system/home-assistant@[your user].service`.  After text entered, press CTRL-X then press Y to save and exit.
+- For most systems, the file is `/etc/systemd/system/home-assistant@YOUR_USER.service` with YOUR_USER replaced by the user account that Home Assistant will run as (normally `homeassistant`).  In particular, this is the case for Ubuntu 16.04.
+- If unfamiliar with command-line text editors, `sudo nano -w [filename]` can be used with `[filename]` replaced with the full path to the file.  Ex. `sudo nano -w /etc/systemd/system/home-assistant@YOUR_USER.service`.  After text entered, press CTRL-X then press Y to save and exit.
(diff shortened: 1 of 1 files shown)


source/_docs/autostart/systemd.markdown
```diff
@@ -51,7 +51,7 @@ After=network-online.target
 [Service]
 Type=simple
 User=%i
-ExecStart=/srv/homeassistant/bin/hass -c "/home/homeassistant/.homeassistant"
+ExecStart=/srv/homeassistant/bin/hass -c "/home/[your user]/.homeassistant"
```
frenck: We generally use an all-caps single string for replacement values in our documentation. e.g., `YOUR_USER`.

```suggestion
ExecStart=/srv/homeassistant/bin/hass -c "/home/YOUR_USER/.homeassistant"
```

source/_docs/autostart/systemd.markdown
```diff
@@ -41,7 +41,7 @@ WantedBy=multi-user.target
 
 ### {% linkable_title Python virtual environment %}
 
-If you've setup Home Assistant in `virtualenv` following our [Python installation guide](/getting-started/installation-virtualenv/) or [manual installation guide for Raspberry Pi](/getting-started/installation-raspberry-pi/), the following template should work for you. If Home Assistant install is not located at `/srv/homeassistant`, please modify the `ExecStart=` line appropriately.
+If you've setup Home Assistant in `virtualenv` following our [Python installation guide](/getting-started/installation-virtualenv/) or [manual installation guide for Raspberry Pi](/getting-started/installation-raspberry-pi/), the following template should work for you. If Home Assistant install is not located at `/srv/homeassistant`, please modify the `ExecStart=` line appropriately. [your user] should replaced by the user account that Home Assistant will run as (e.g `homeassistant`).
```
frenck: We generally use an all-caps single string for replacement values in our documentation. e.g., `YOUR_USER`.

```suggestion
If you've setup Home Assistant in `virtualenv` following our [Python installation guide](/getting-started/installation-virtualenv/) or [manual installation guide for Raspberry Pi](/getting-started/installation-raspberry-pi/), the following template should work for you. If Home Assistant install is not located at `/srv/homeassistant`, please modify the `ExecStart=` line appropriately. `YOUR_USER` should be replaced by the user account that Home Assistant will run as (e.g `homeassistant`).
```

frenck (changes requested):

source/_docs/autostart/systemd.markdown
dimitripb: Sounds plausible but on the same page [your user] is used everywhere else. Maybe all occurences of [your user] should be changed then.

frenck: In the past, those things weren't standardized. For new changes, we try to keep sticking to some form of standard.

Updating the current docs is quite a big task, but slowly happening. If you'd like to update the other occurrences as well... go ahead! It is greatly appreciated! 👍

dimitripb: Done for this page!

frenck (approved):

frenck:
Thanks, @dimitripb! 👍

PR #7671, opened 2018-11-28
Update ifttt.markdown

olskar:
**Description:**
I had a hard time understanding that the automation-part was needed. Before this change only the body in ifttt was needed. I updated the doc a bit to make clear that also the automation bit is needed for ifttt integration to work.


fabaff (approved):
Thanks :bird:

PR #7672, opened 2018-11-28
Update list of dependencies

magnusoverli:
Add docker to the list of dependencies

**Description:**

## Checklist:

### DIFF
--- a/source/hassio/installation.markdown
+++ b/source/hassio/installation.markdown
@@ -72,6 +72,7 @@ This is the list of packages you need to have available on your system that will
  - ca-certificates
  - curl
  - dbus
+ - docker
  - jq
  - network-manager
  - socat


fabaff (approved):
Thanks :bird:

frenck:
@magnusoverli This is confusing and a risky change.
Reasoning, please might install the Docker that came with their OS distribution, which is a really bad idea...

magnusoverli:
Suggested fix in [#7678](https://github.com/home-assistant/home-assistant.io/pull/7678).

fabaff:
It's just a list of deps. We don't say how to install it. It's an advance method and really depends on the underlying OS.

magnusoverli:
In that case, specifying Docker-CE should be enough?

PR #7688, opened 2018-11-29
Internal DNS notes

mateuszdrab:
Add a neat trick to force google casts to use internal DNS, useful when TTS doesn't work because Google cast ignores internal DNS.

Worth adding to the docs?


frenck (approved):

frenck:
Thanks for this addition @mateuszdrab! 👍 

Made a little tweak, can be merged as soon as the build passes.

mateuszdrab:
Thanks!

PR #7701, opened 2018-12-01
Making owntracks breaking change clearer

DubhAd:
Many folks have suffered because they didn't know to remove the old device tracker before upgrading


fabaff (approved):
:bird:

MartinHjelmare:
Shouldn't this go into current?

ludeeus:
@MartinHjelmare yes, next would be 0.84, when that happen few will even look at this :/

PR #7702, opened 2018-12-01
monitored_variables required now?

hdsheena:
When I did my install, the TILEs weren't tracked until I added the monitored_variables with Phone and Tile specified.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/device_tracker.tile.markdown
+++ b/source/_components/device_tracker.tile.markdown
@@ -25,6 +25,10 @@ device_tracker:
   - platform: tile
     username: email@address.com
     password: MY_PASSWORD_123
+    monitored_variables:
+      - TILE
+      - PHONE
+  
 ```
 
 {% configuration %}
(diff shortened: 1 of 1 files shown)


source/_components/device_tracker.tile.markdown
```diff
@@ -25,6 +25,10 @@ device_tracker:
   - platform: tile
     username: email@address.com
     password: MY_PASSWORD_123
+    monitored_variables:
```
fabaff, line 28: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps users to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_components/device_tracker.tile.markdown
```diff
@@ -37,7 +41,7 @@ device_tracker:
     required: true
     type: string
   monitored_variables:
-    description: the Tile types to monitor; valid values are `TILE` and `PHONE` (default is for all types to be included)
+    description: the Tile types to monitor; valid values are `TILE` and `PHONE` (default is for only 'PHONE' to be monitored)
```
fabaff, line 44: See https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/device_tracker/tile.py#L44

hdsheena: Then I think the issue isn't the documentation, but a bug in the way the code runs. I'll see if I can find time to test it again, adding the monitored_variables line to my config resolved the issue, though.

PR #7709, opened 2018-12-02
Add note about not being on same network as Home Assistant during relink

Xiol:
**Description:**
I had trouble relinking GA to Hass due to being on the same network, I suspect this is down to how I'm overriding DNS locally for my hass instance, but it solved the issue for me when I was having trouble relinking after moving to 0.80+

## Checklist:


frenck (approved):

frenck:
Thanks, @Xiol! Good tip!

Issue #7368, opened 2018-11-05
The water_heater documentation about away mode is not correct


kennedyshead:
https://www.home-assistant.io/components/water_heater/#water-heater-control-services

It shoud not be `turn_away_mode_on` / `turn_away_mode_off` but rather `set_away_mode`

klaasnicolaas:
If this has to be adjusted, the [python code](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/water_heater/__init__.pyt) must also be looked at, because I see functions with: `turn_away_mode_on` / `turn_away_mode_off`.

So not only the documentation should then be adjusted.

nwesterhausen:
We may need someone more familiar with how the components are supposed to work, but when I read through the water heater __init__ some things stand out to me:

The only mentions that match turn_away_mode_off/on are functions which get called by a handler for the "away_mode" attribute [here](https://github.com/home-assistant/home-assistant/blob/master/homeassistant/components/water_heater/__init__.py#L240)

It also has the exposed service for away mode being "set_away_mode" with it wanting an entity id and a boolean for away mode on/off [here](https://github.com/home-assistant/home-assistant/blob/master/homeassistant/components/water_heater/__init__.py#L63)

I will go ahead and do a PR to change the documentation. If I'm wrong at understanding the python, we can just not accept the PR.

frenck:
PR's have been merged. Crisis diverted 😆

PR #7742, opened 2018-12-05
Fix service call documentation

nwesterhausen:
**Description:**
See #7368 
The component doesn't seem to register "turn_away_mode_off" and "turn_away_mode_on" as services and instead uses "set_away_mode".
Fix the Water Heater documentation to reflect the registered services.

## Checklist:

### DIFF
--- a/source/_components/water_heater.markdown
+++ b/source/_components/water_heater.markdown
@@ -77,13 +77,14 @@ automation:
         operation_mode: eco
 ```
 
-### {% linkable_title Service `water_heater.turn_away_mode_on` %}
+### {% linkable_title Service `water_heater.set_away_mode` %}
 
-Turn away mode on for water heater device
+Turn away mode on or off for water heater device
 
 | Service data attribute | Optional | Description |
 | ---------------------- | -------- | ----------- |
 | `entity_id` | yes | String or list of strings that point at `entity_id`'s of water heater devices to control. Else targets all.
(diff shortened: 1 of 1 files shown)


klaasnicolaas:
Pherhaps then also change the function names in the python code. Then there is no doubt at all about the use of `turn` or `set` 😋

nwesterhausen:
I think it's just a case of copy/paste gone wrong when creating documentation? I mean there does exist a python function for the away_mode service. I've used the climate device before, and it uses `set_away_mode`, are there some examples which use `turn` instead?

klaasnicolaas:
[Here](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/water_heater/__init__.py#L32) it is clear that it is using `set` for calling a service. 
Only [here](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/water_heater/__init__.py#L210) is says `turn` 😋 

But the changes you have made to the documentation are good in any case.

nwesterhausen:
I think I found out where to look, there's only one spot where it registers services with Home Assistant: [here](https://github.com/home-assistant/home-assistant/blob/master/homeassistant/components/water_heater/__init__.py#L80) and there is only one registered for `set_away_mode` and none for `turn_away_mode_x`. 😄

frenck (approved):

frenck:
Thanks, @nwesterhausen! 👍

PR #7748, opened 2018-12-05
Fix issue with proximity configuration

ludeeus:
**Description:**

Fix issue with proximity configuration

## Checklist:

### DIFF
--- a/source/_components/proximity.markdown
+++ b/source/_components/proximity.markdown
@@ -49,29 +49,28 @@ proximity:
 ```
 
 {% configuration %}
-proximity:
-  zone:
-    description: The zone to which this component is measuring the distance to. Default is the home zone.
-    required: false
-    type: map
-    keys:
-      ignored_zones:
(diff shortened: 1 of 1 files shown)


source/_components/proximity.markdown
```diff
@@ -49,29 +49,28 @@ proximity:
 ```
 
 {% configuration %}
-proximity:
```
klaasnicolaas, line 52: Where is the proximity variable? 😆 

Think it is important to let it in and the rest as sub-items below it, since you also have to place the proximity variable in your config file.

klaasnicolaas (changes requested):

source/_components/proximity.markdown
ludeeus: It's gone :D
If you look at other components this key is not present 🤷‍♂️

cogneato (approved):
Missing info now visible. Looks good to me.

fabaff (approved):

PR #7771, opened 2018-12-08
XS1 component

markusressel:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#19115

## Checklist:


DubhAd (dismissed):
👍

source/_components/xs1.markdown
```diff
@@ -0,0 +1,103 @@
...
+title: "EZcontrol XS1"
+description: "Instructions on how to integrate an XS1 Gateway within Home Assistant."
+date: 2018-12-04 20:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: xs1.jpg
+ha_category: Hub
+featured: true
```
MartinHjelmare: Set to false.

source/_components/xs1.markdown
```diff
@@ -0,0 +1,103 @@
...
+ha_iot_class: "Local Polling"
+---
```
MartinHjelmare, line 19: Please add `ha_release`.

MartinHjelmare: If there's not much info in the platform pages, we now prefer to not create those pages but just create artifical entries in the component index by adding categories and redirects here in the component page front matter.

See eg:
https://github.com/home-assistant/home-assistant.io/pull/8427

source/_components/xs1.markdown
```diff
@@ -0,0 +1,103 @@
...
+```yaml
+# Example configuration.yaml entry
+xs1:
+  host: "192.168.2.100"
+```
+
+The component will automatically detect the configuration of the XS1 Gateway only **on initialization** which currently means when Home Assistant boots. When you change the configuration of the XS1 you (currently) have to restart Home Assistant to see the effects.
+
+{% configuration %}
+xs1:
```
MartinHjelmare: We don't include this key.

source/_components/xs1.markdown
markusressel: done

source/_components/xs1.markdown
markusressel: removed

source/_components/xs1.markdown
markusressel: ok

source/_components/xs1.markdown
```diff
@@ -0,0 +1,112 @@
...
+sharing: true
+footer: true
+logo: xs1.jpg
+ha_category: 
+  - Hub
+  - Climate
+  - Sensor
+  - Switch
+redirect_from:
+  - /components/climate.xs1/
```
MartinHjelmare: I guess we only need the redirects if there were already pages. I think we can remove them. Hopefully the index will still populate.

markusressel: Seems like it doesn't? I don't understand what the build error message is trying to tell me :disappointed:

MartinHjelmare: My guess was that it failed due to the redirects looking for pages that didn't exist. But I guess I'm wrong.

@klaasnicolaas any idea?

klaasnicolaas: `redirect_from` is only for pages that already exist.

markusressel: @klaasnicolaas I removed that already but it still doesn't compile :cry:

klaasnicolaas:
Pherhaps that @balloob can take a look at the deploy error

frenck:
Rebased, awaiting build results.

frenck:
Thanks for the PR @markusressel! 🎉 🥇

frenck (approved):

PR #7790, opened 2018-12-10
Added information about attributes + note fix  …

dagobert:
Added detailed information about attributes in a separate section. Moved warning level information into attribute section.

Fixed problem with note section. It was not showing correctly on hass.io page.


fabaff (approved):
Thanks :bird:

dagobert:
Is it possible, that some of the markdown language is not "translated" correctly within the `<p>` section? The list is not correctly shown on the website but the code inserts.
How should I do the list in the `<p>` section? Should I use HTML code?

PR #7799, opened 2018-12-10
Removed the venv link from the sidebar

DubhAd:
Removed the venv link, since that gets people the worst of all options, and helps them break their system when they install another Python program. Obviously the link (and others) are still available if they click _Installation_.


emontnemery:
What's wrong with Python venv and why is it dangerous causing system breakage?

fabaff:
Using a `venv` is the way to go for installation directly on a system. At least 15 % of all HA setup are done this way.

Removing it from the sidebar is a good idea, though.

fabaff (approved):
:bird:

PR #7805, opened 2018-12-12
Update device_tracker.gpslogger.markdown

w1ll1am23:
While it makes it nicer to view, most people are probably going to copy and paste the URL from the site via their Android device and not type it all out. Having the line breaks results in %20 (or spaces) in the URL when sending to Home Assistant which doesn't work. Removing those spaces will make it easier for users to copy/paste.

**Description:**

## Checklist:


fabaff (approved):
:bird:

PR #7810, opened 2018-12-12
Add KS8500 to supported SamsungTV models

DallasO:
**Description:**
Adding another supported model to the SamsungTV component. Multiple features tested, including:
* Power On/Off
* Volume Up/Down
* Play Pause/Resume
* Play Next/Previous

Did not test changing channel, but am assuming it will work like similar models. I can test if you would like before accepting this Pull Request.  

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7813, opened 2018-12-12
Update matrix.markdown

adebeun:
**Description:**

The example is incomplete (or maybe out of date). For it to work (in version 0.83.3) I had to make 2 changes.
Also, for whatever reason (perhaps because of nested layers of markup), the part with the double braces on the last line does not show up on the actual displayed page (it shows when editing the page, but not the normal page). So when copy-pasting the example, a crucial bit is missing. I'm not sure how to fix that aspect of it and since I can't test it either, I haven't proposed a change for this.

Arthur

## Checklist:

### DIFF
--- a/source/_components/matrix.markdown
+++ b/source/_components/matrix.markdown
@@ -96,6 +96,7 @@ matrix:
     - "#someothertest:matrix.org"
   commands:
     - word: testword
+      name: testword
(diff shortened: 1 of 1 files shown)


fabaff (approved):
Thanks :bird:

PR #7847, opened 2018-12-14
iFrame card -> Iframe card

AlecRust:
**Description:**

The [iframe element](https://www.w3.org/TR/2011/WD-html5-20110525/the-iframe-element.html) is part of the HTML spec, it's not written as "iFrame" which would imply it's an Apple product.

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7853, opened 2018-12-15
Fixing the numeric state examples

DubhAd:
There's a bunch of copy/paste numeric state conditions that have quotes around them, when they shouldn't


ludeeus (approved):
Thanks 👍 
Can be merged when build passes.

PR #7875, opened 2018-12-16
hash in passwords

sveip:
**Description:**

## Checklist:

### DIFF
--- a/source/_docs/configuration/secrets.markdown
+++ b/source/_docs/configuration/secrets.markdown
@@ -35,6 +35,7 @@ The `secrets.yaml` file contains the corresponding password assigned to the iden
 ```yaml
 http_password: YOUR_PASSWORD
 ```
+If password contains the hash character, you have to embed it in double quotes to avoid it being treated as a comment. `http_password: "YOUR_PASSWORD_WITH_#"
 
 ### {% linkable_title Debugging secrets %}
 


source/_docs/configuration/secrets.markdown
```diff
@@ -35,6 +35,7 @@ The `secrets.yaml` file contains the corresponding password assigned to the iden
 ```yaml
 http_password: YOUR_PASSWORD
 ```
+If password contains the hash character, you have to embed it in double quotes to avoid it being treated as a comment. `http_password: "YOUR_PASSWORD_WITH_#"
```
fabaff, line 38: That's YAML and already covered in the documentation.

sveip: Well, if that logic is applied generally, most of the doc would be superfluous ;) It's common mistake, and not mentioned anywhere in the docs that are usually read y inexperienced users. Maybe it should rather be mentined as a "Tip"?

fabaff: Well, moving the secrets out the configuration is an advanced topic and placed in that section while the YAML intro is the very first topic when it comes to configuration.

sveip: Ah, sorry, then I misunderstood your first comment. If it's covered in the config then it's probably OK. Even though I don't think it's bad to ad hints and tricks, it's hard to "overdocument."

PR #7888, opened 2018-12-17
Lazy image load logo

balloob:
**Description:**
- Fix total component count after Jekyll upgrade. Caused by https://github.com/jekyll/jekyll/issues/6038
- Add lazy load to the logos in the cards.


balloob:
Loading the all components page

image

PR #7889, opened 2018-12-17
Added example to notify-group platform

hAmpzter:
**Description:**

## Checklist:

### DIFF
--- a/source/_components/notify.group.markdown
+++ b/source/_components/notify.group.markdown
@@ -15,12 +15,12 @@ ha_qa_scale: internal
 
 The `group` notification platform allows you to combine multiple `notify` platforms into a single service.
 
-To use this notification platform in your installation, add the following to your `configuration.yaml` file:
+## {% linkable_title Configuration %}
 
 ```yaml
 # Example configuration.yaml entry
 notify:
-  - name: NOTIFIER_NAME
+  - name: NAME_OF_NOTIFIER_GROUP
     platform: group
     services:
       - service: html5
@@ -48,3 +48,14 @@ services:
       required: false
       type: string
 {% endconfiguration %}
+
+## {% linkable_title Example %}
+
+An example on how to use inside of an automation:
+
+```yaml
+action:
+  service: notify.NAME_OF_NOTIFIER_GROUP
+  data:
+    message: "The sun is {% raw %}{% if is_state('sun.sun', 'above_horizon') %}up{% else %}down{% endif %}{% endraw %}!"
+```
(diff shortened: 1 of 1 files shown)


source/_components/notify.group.markdown
```diff
@@ -15,12 +15,12 @@ ha_qa_scale: internal
 
 The `group` notification platform allows you to combine multiple `notify` platforms into a single service.
 
-To use this notification platform in your installation, add the following to your `configuration.yaml` file:
```
frenck, line 18: Not sure why this became a title. This is kinda not matching our other documentation.

source/_components/notify.group.markdown
```diff
@@ -48,3 +48,14 @@ services:
...
+
+## {% linkable_title Example %}
+
+An example on how to use inside of an automation:
+
+```yaml
+action:
+  service: notify.NAME_OF_NOTIFIER_GROUP
+  data:
+    message: "The sun is {% raw %}{% if is_state('sun.sun', 'above_horizon') %}up{% else %}down{% endif %}{% endraw %}!"
```
frenck: `{% raw %}` should be wrapped outside the code block right (outside of the opening of the yaml code block)?
Since the user is not supposed to copy that part I assume.

hAmpzter: Why should they not be able to copy that, the code should be working inside of an automation-configuration?
Or what do you mean?

hAmpzter: Ohh, now I think I saw what you mean.
I think I copied that part from the notify-example. But hold on and I'll change it!

hAmpzter: No, the code is the same as the example on the other page: [Edit notify-markdown](https://github.com/home-assistant/home-assistant.io/edit/current/source/_components/notify.markdown).
I assume it should be the same as there?!

frenck: Well, that is incorrect as well in that case...
The raw now causes the parts between it not to be executed in the user's example...

frenck (changes requested):
Changes requested

source/_components/notify.group.markdown
hAmpzter: Sorry, forgot this! :)
Sounds more like I should remove the raw-tags as the example should be copy&paste-enabled

hAmpzter:
@frenck Changes committed! :)

frenck (approved):

frenck:
Thanks, @hAmpzter! 👍

PR #7895, opened 2018-12-18
Make device_class type a string

nwesterhausen:
**Description:**
As mentioned in #6791, we should have "device_class" as a string in the documentation.

That pull was closed and there was not change for "device_class" on the template documentation page. This should correct it, having it still say "device class" but will not link to a non-existent page.

## Checklist:

### DIFF
--- a/source/_docs/configuration/customizing-devices.markdown
+++ b/source/_docs/configuration/customizing-devices.markdown
@@ -69,7 +69,7 @@ assumed_state:
 device_class:
   description: Sets the class of the device, changing the device state and icon that is displayed on the UI (see below). It does not set the `unit_of_measurement`.
   required: false
-  type: device_class
+  type: device class
   default: None
 unit_of_measurement:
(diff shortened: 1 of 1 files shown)


nwesterhausen:
I think it would be best to link "device class" in the configuration block to the anchor on the same page for device class. But having no link is definitely better than having a broken link (as it was, and [is currently](https://www.home-assistant.io/docs/configuration/customizing-devices/#device_class)).

frenck:
IMHO this is a workaround for the actual problem?

nwesterhausen:
It is a workaround -- with this patch I'm not fixing the underlying issue. With this patch I'm getting rid of the link to a non-existent page on the documentation site, until the work around is figured out.

I don't know much about Jeckyll or what other options we have to fix the link on this one page, but based off of #6791 which was closed because people stopped replying it's not an easy fix. To me, I don't know how much sense putting a single page exception into the configuration formatting logic than it is to just modify that single page so it doesn't trigger it.

frenck:
@nwesterhausen Created #8179 as a more generic, solid and side-wide solution.

Closing this PR in favor of #8179.

PR #7897, opened 2018-12-18
Update systemd.markdown

a-tom-s:
The quotation marks must be left off for the container name. Otherwise you will get the message
```/usr/bin/docker: Error response from daemon: Invalid container name ("home-assistant-pi"), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed.```
After I removed the quotation marks from my systemd service file, everything worked fine.

**Description:**

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7899, opened 2018-12-18
Updated docs of climate.knx

marvin-w:
**Description:**

Updated the docs of the climate entity for the KNX platform.

## Checklist:

### DIFF
--- a/source/_components/climate.knx.markdown
+++ b/source/_components/climate.knx.markdown
@@ -47,6 +47,25 @@ climate:
     operation_mode_comfort_address: '5/1/7'
 ```
 
+If you want to overwrite the supported operation modes use:
+
+```yaml
+# Example configuration.yaml entry
+climate:
+  - platform: knx
+    name: HASS-Kitchen.Temperature
+    temperature_address: '5/1/1'
+    setpoint_shift_address: '5/1/2'
+    setpoint_shift_state_address: '5/1/3'
+    target_temperature_address: '5/1/4'
+    operation_mode_frost_protection_address: '5/1/5'
+    operation_mode_comfort_address: '5/1/7'
+    override_supported_operation_modes:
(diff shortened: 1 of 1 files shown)


fabaff (approved):
Thanks :bird:

marvin-w:
@fabaff: Can you revert this please? I was looking at https://github.com/XKNX/xknx/blob/master/home-assistant-plugin/custom_components/climate/xknx.py#L18-L37 

instead of https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/climate/knx.py#L18-L32

Maybe @Julius2342 can use it when he releases the next version. Sorry :/.

Julius2342:
I will port the changes of the xknx-custom-component to the main hass components asap. 

(KNX climate devices are a terrible mess, all manufacturers do it slightly different. And many users don't have an ETS so they have to use the configuration their electricians left a long time ago.)

marvin-w:
Thanks @Julius2342, yes I know this problem. I still don't really know why they charge 1k for a single license.

One could also argue with the quality of those configurations made by at least some electricians.

Julius2342:
@marvin-w : What do you think about the idea of splitting up the different climate devices into separate climate devices with different logic behaviour? My hope would be that this would simplify things a lot - especially we would avoid this mess of dozens of parameters.

marvin-w:
@Julius2342: Sounds good to me but we'd have to define those climate devices upfront and also add proper documentation. If you're up to it I could help you doing that and also help during the implementation but, as I mentioned, my knowledge of python is limited but it didn't look to hard.

Maybe we could discuss those changes in some kind of instant messaging like slack for instance. Would you be up for that?

Julius2342:
We have a chat here: https://discord.gg/5XARFNT

PR #7900, opened 2018-12-19
Update Radius

tradiuz:
Clarified the radius parameter, and added it to the example.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/geo_location.usgs_earthquakes_feed.markdown
+++ b/source/_components/geo_location.usgs_earthquakes_feed.markdown
@@ -36,6 +36,7 @@ To integrate the U.S. Geological Survey Earthquake Hazards Program feed, add the
 geo_location:
   - platform: usgs_earthquakes_feed
     feed_type: 'past_day_all_earthquakes'
+    radius: 50
 ```
 
 {% configuration %}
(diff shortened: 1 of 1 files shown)


source/_components/geo_location.usgs_earthquakes_feed.markdown
```diff
@@ -36,6 +36,7 @@ To integrate the U.S. Geological Survey Earthquake Hazards Program feed, add the
 geo_location:
   - platform: usgs_earthquakes_feed
     feed_type: 'past_day_all_earthquakes'
+    radius: 50
```
fabaff: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps users to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

tradiuz: OK, I'll add a full config with defaults to show expected input.

tradiuz (commented):
Updated the default to be more clear (50km is not a valid float).

Provided a full example at the bottom.

fabaff (approved):
Thanks :bird:

PR #7911, opened 2018-12-20
Fix IHC config schema

mopolus:
**Description:**
Documentation updates for fix to issue with config schema.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):**
https://github.com/home-assistant/home-assistant/pull/19415

**Related issue:**
home-assistant/home-assistant#19353

## Checklist:


MartinHjelmare:
Since the source fix is already released I think we should target `current` branch instead of `next` branch. Please rebase on latest current and switch target branch.

mopolus:
> Since the source fix is already released I think we should target `current` branch instead of `next` branch. Please rebase on latest current and switch target branch.

Branch updated. Not sure how I can do the rebasing from here (currently only have web access)

MartinHjelmare:
Probably then easiest to make a new pull request from the correct base and apply your changes again. From the command line it would be easier.

mopolus:
@MartinHjelmare I have added PR #7913 from current
Thank you for advice and assistance :-)

PR #7914, opened 2018-12-20
scan_interval mandatory

vg8020:
As per @fronzbot message
https://community.home-assistant.io/t/fail2ban-sensor-error/32948/3
and own testing, it seems configuration variable scan_interval should be mandatory.


frenck:
This has been fixed upstream, so should not go into the documentation.
Closing PR.

PR #7926, opened 2018-12-22
Patch 1

ceejii:
**Description:**
I discovered that Discovery indeed supports ignoring octoprint as well. This is in reference to https://github.com/home-assistant/home-assistant/issues/18627 but does not necessarily close that issue since this commit is just about missing documentation.

## Checklist:

### DIFF
--- a/source/_components/discovery.markdown
+++ b/source/_components/discovery.markdown
@@ -85,6 +85,7 @@ Valid values for ignore are:
  * `ikea_tradfri`: IKEA Trådfri (Tradfri)
  * `logitech_mediaserver`: Logitech media server (Squeezebox)
  * `netgear_router`: Netgear routers
+ * `octoprint`: Octoprint
  * `openhome`: Linn / Openhome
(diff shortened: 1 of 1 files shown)


ludeeus (changes requested):
Thanks for this @ceejii , can you also add it to the list of supported components?
As a new line after `* [Netgear routers](/components/device_tracker.netgear/)`

ceejii:
I could easily and did think about it, but do you know if it is **actually** supported and works? I don't know that. 

My reason for this doc update is because the Discovery component falsely triggers an error of incorrect config when I haven't even added the config to my home assistant config. So all I've tested is adding it to discovery´s ignore to confirm that the warnings went away. And since ignore for octoprint wasn't documented I added what documentation I was sure of without having developed any part of the component.

ludeeus:
It is part of the `discovery` component, it looks like it was just forgotten to add it here when the component was added :)

ludeeus (approved):

PR #7928, opened 2018-12-22
Update homematic.markdown

danielperna84:
**Description:**
Recently there was confusion in the forum why the usage of variables didn't work. It requires a user with Admin-privileges, which the documentation does not mention yet. Hence this change.

## Checklist:


ludeeus (approved):
Thanks 👍

PR #7937, opened 2018-12-23
Updated docs of knx climate device

marvin-w:
**Description:**

Updated docs for KNX climate devices.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#19546

## Checklist:


source/_components/climate.knx.markdown
```diff
@@ -112,6 +135,24 @@ operation_mode_comfort_address:
   description: KNX address for switching on/off comfort mode.
   required: false
   type: string
-{% endconfiguration %}
-
-`operation_mode_frost_protection_address` / `operation_mode_night_address` / `operation_mode_comfort_address` are not necessary if `operation_mode_address` is specified.
+operation_modes:
+  description: Overrides the supported operation modes.
+  required: false
```
MartinHjelmare, line 139: What happens if the user doesn't specify this? What operation modes will be activated?

marvin-w: @MartinHjelmare some users don't have explicit knx addresses for setting the frost night and comfort mode directly. If they don't set this property here the available operation modes will be made up of the information we have. If their device doesn't support operation modes at all we will also not set the support flags for it. So I think this is fine.

source/_components/climate.knx.markdown
```diff
@@ -47,6 +47,21 @@ climate:
...
 ```
 
+`operation_mode_frost_protection_address` / `operation_mode_night_address` / `operation_mode_comfort_address` are not necessary if `operation_mode_address` is specified.
+
+If your device doesn't support setpoint_shift calculations (i.e. if you don't provide a `setpoint_shift_address` value) please set the `min_temp` and `max_temp`
+attributes of the climate device to avoid issues with increasing the temperature in the frontend.
+
+The following values are valid for the `operation_modes` attribute:
+
+- Comfort
```
MartinHjelmare: We should document somewhere how these map to the home assistant modes.

marvin-w: I agree. But due to the fact that we currently only support one type of climate devices and we need to work out a way to support multiple types without mixing of the mapping I would adjust the docs in the next PR and leave it as it is for now.

An example for this problem is as follows.

One type of devices has the operation mode `Comfort`. This maps to the HEAT mode in HA but a second device type has the operation mode `Heat` which would also map to HEAT within HA thus making it impossible to properly distinguish between the two if we would support all of them as of now.

We need to add this in the library and we will do that in a future release, but for now, I assume it's good if we support at least some of those modes.

Are there existing docs on the currently available operation modes within HA other than in the code itself and in the Lovelace climate card where I can refer users to?

MartinHjelmare: I haven't seen those docs.

marvin-w: @MartinHjelmare  Alright, so you want me to update the docs now or are you fine that I'll do it once we did the refactoring I was talking about? I'll fix the conflicts when I'm at home.

MartinHjelmare: I think we should document the current behavior.

marvin-w: I added the mapping.

MartinHjelmare (approved):

PR #7942, opened 2018-12-24
GitHub Sensor

timmo001:
**Description:**

Adds docs for GitHub sensor component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#19561

## Checklist:

### DIFF
--- a/source/_components/sensor.github.markdown
+++ b/source/_components/sensor.github.markdown
@@ -62,7 +62,7 @@ access_token:
   description: Your GitHub Access Token
   required: false
   type: string
-server_url:
+url:
   description: If you are using a GitHub Enterprise server, add it's URL here. For example, `https://mygithubserver.com`
   required: false
   type: string
--- a/source/_components/sensor.github.markdown
+++ b/source/_components/sensor.github.markdown
@@ -9,7 +9,7 @@ sharing: true
 footer: true
 logo: github.png
 ha_category: Sensor
-ha_release: 0.87
+ha_release: 0.88
 ha_iot_class: "Cloud Polling"
 ---
 
--- a/source/_components/sensor.github.markdown
+++ b/source/_components/sensor.github.markdown
@@ -58,7 +58,7 @@ repositories:
       required: true
       type: string
     name:
-      description: Name of repository Gives the sensor a custom name in Home Assistant. Defaults to the repository name from the path.
+      description: Name of repository Gives the sensor a custom name in Home Assistant. Defaults to the repository name from GitHub.
       required: false
       type: string
 {% endconfiguration %}
--- a/source/_components/sensor.github.markdown
+++ b/source/_components/sensor.github.markdown
@@ -17,7 +17,7 @@ The GitHub sensor integrates data from [GitHub](https://github.com/) to monitor 
 
 ## {% linkable_title Setup %}
 
-You will need your GitHub username and password. You can also create an access token [here](https://github.com/settings/tokens). An access token is required if you use two factor authentication with your account.
+You will need your GitHub username and password. You can also create an access token [here](https://github.com/settings/tokens). An access token is required if you use two-factor authentication with your account.
 
 ## {% linkable_title Configuration %}
 
--- a/source/_components/sensor.github.markdown
+++ b/source/_components/sensor.github.markdown
@@ -9,7 +9,7 @@ sharing: true
 footer: true
 logo: github.png
 ha_category: Sensor
-ha_release: 0.86
+ha_release: 0.87
 ha_iot_class: "Cloud Polling"
 ---
 
--- a/source/_components/sensor.github.markdown
+++ b/source/_components/sensor.github.markdown
@@ -23,6 +23,23 @@ You will need your GitHub username and password. You can also create an access t
 
 To enable this platform, please add the following to your `configuration.yaml` file:
 
+### {% linkable_title Username / Password %}
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: github
+    username: 'githubusername'
+    password: !secret github_password
+    repositories:
+      - path: 'home-assistant/home-assistant'
+        name: 'Home Assistant'
+      - path: 'hassio-addons/addon-ide'
+        name: 'IDE add-on'
+```
+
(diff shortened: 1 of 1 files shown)


source/_components/sensor.github.markdown
```diff
@@ -0,0 +1,64 @@
...
+```
+
+{% configuration %}
+access_token:
+  description: Your GitHub Access Token
+  required: true
+  type: string
+username:
+  description: Your GitHub Username
+  required: true
```
ludeeus: The code states that this is optional.

source/_components/sensor.github.markdown
```diff
@@ -0,0 +1,64 @@
...
+      - path: 'home-assistant/home-assistant'
+        name: 'Home Assistant'
+      - path: 'hassio-addons/addon-ide'
+        name: 'IDE add-on'
+```
+
+{% configuration %}
+access_token:
+  description: Your GitHub Access Token
+  required: true
```
ludeeus, line 38: The code states that this is optional.

source/_components/sensor.github.markdown
```diff
@@ -0,0 +1,64 @@
...
+  type: string
+username:
+  description: Your GitHub Username
+  required: true
+  type: string
+password:
+  description: Your GitHub Password
+  required: true
```
ludeeus: The code states that this is optional.

source/_components/sensor.github.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+## {% linkable_title Configuration %}
+
+To enable this platform, please add the following to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: github
+    access_token: !secret github_access_token
```
ludeeus, line 30: Please remove all non required options from the example.

ludeeus (commented):
There are some mismatches here and in the code on what is required.

source/_components/sensor.github.markdown
timmo001: Technically, either username/password or an access_token is required. Since either/or is required, what should I state here? Examples for both methods?

ludeeus: Good question... preferably the user will use `access_token` for communication, so maybe this part should be as is.

source/_components/sensor.github.markdown
```diff
@@ -0,0 +1,81 @@
...
+ha_category: Sensor
+ha_release: 0.85
+ha_iot_class: "Cloud Polling"
+---
+
+The GitHub sensor integrates data from [GitHub](https://github.com/) to monitor your favorite repositories.
+
+## {% linkable_title Setup %}
+
+You will need your GitHub username and password. You can also create an access token [here](https://github.com/settings/tokens). An access token is required if you use two factor authentication with your account.
```
frenck:
```suggestion
You will need your GitHub username and password. You can also create an access token [here](https://github.com/settings/tokens). An access token is required if you use two-factor authentication with your account.
```

source/_components/sensor.github.markdown
```diff
@@ -0,0 +1,81 @@
...
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: github
+    username: 'githubusername'
+    password: !secret github_password
+    repositories:
+      - path: 'home-assistant/home-assistant'
+        name: 'Home Assistant'
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

frenck (changes requested):
Thanks for this @timmo001, left 2 comments for you to look at. Thanks 👍

source/_components/sensor.github.markdown
timmo001: See new commit. Removed any optional properties from examples

source/_components/sensor.github.markdown
```diff
@@ -0,0 +1,75 @@
...
+ha_category: Sensor
+ha_release: 0.85
+ha_iot_class: "Cloud Polling"
+---
+
+The GitHub sensor integrates data from [GitHub](https://github.com/) to monitor your favorite repositories.
+
+## {% linkable_title Setup %}
+
+You will need your GitHub username and password. You can also create an access token [here](https://github.com/settings/tokens). An access token is required if you use two-factor authentication with your account.
```
ludeeus:
```suggestion
To set up this sensor you will need your GitHub username and password, or an [access token][accesstoken]. 

If you have two-factor authentication enabled on your account an [access token][accesstoken] is required.
[accesstoken]: https://github.com/settings/tokens
```
A litte sugestion for rewording, this is my personal preference, disregard this if you don't agree :)

timmo001: Looks good :+1: I've added it

source/_components/sensor.github.markdown
```diff
@@ -62,6 +62,10 @@ access_token:
   description: Your GitHub Access Token
   required: false
   type: string
+server_url:
+  description: If you are using a GitHub Enterprise server, add it's URL here. For example, `https://mygithubserver.com`
+  required: false
```
frenck, line 42: If it isn't required, it has a default.. right?

frenck (changes requested):

source/_components/sensor.github.markdown
timmo001: Yes, it defaults to `https://github.com` as the default via the library

ludeeus: Something like this then?
```
description: URL to GitHub server (only change this if you use GitHub Enterprise).
required: false
default: https://github.com
```

timmo001: This URL should be blank / not added still since the library handles the defaults. If someone set `https://github.com` as the `server_url` it may cause issues or would have no effect

frenck (dismissed):

frenck (dismissed):

klaasnicolaas:
Time to merge this PR?

PR #7961, opened 2018-12-27
Ebusd integration

CrazYoshi:
**Description:**
Integration between ebusd, daemon for communication with eBUS heating systems, and homeassistant using sensor component.

https://github.com/home-assistant/home-assistant/pull/19607

### DIFF
--- a/source/_components/ebusd.markdown
+++ b/source/_components/ebusd.markdown
@@ -24,19 +24,16 @@ Enable the sensor by adding the following to your `configuration.yaml` file:
 ebusd:
   host: 127.0.0.1
   circuit: '700'
-  monitored_conditions:
-    - 'MaxFlowTemperatureDesired'
-    - 'WaterPressure'
-    - 'PumpStatus'
+
 ```
 
 {% configuration %}
 host:
-  description: This is the IP address of your ebus daemon, eg. 127.0.0.1.
+  description: This is the IP address of your ebus daemon, e.g., `127.0.0.1`.
   required: true
   type: string
 circuit: 
-  description: The heating circuit name to monitor, eg. 700.
+  description: The heating circuit name to monitor, e.g., 700.
   required: true
   type: string
 port:
--- a/source/_components/ebusd.markdown
+++ b/source/_components/ebusd.markdown
@@ -13,7 +13,7 @@ ha_iot_class: "Local Polling"
 ha_release: "0.85"
 ---
 
-Integration between [ebusd](https://github.com/john30/ebusd/) daemon for communication with eBUS heating systems, and homeassistant using sensor component.
+Integration between [ebusd](https://github.com/john30/ebusd/) daemon for communication with eBUS heating systems, and Home Assistant using sensor component.
 
 ## {% linkable_title Configuration %}
 
--- a/source/_components/ebusd.markdown
+++ b/source/_components/ebusd.markdown
@@ -1,0 +1,123 @@
+---
+layout: page
+title: "ebusd"
+description: "The ebusd component allows the integration between eBUS heating system and Home Assistant."
+date: 2018-12-27 15:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: ebusd.png
+ha_category: Ebusd
+ha_iot_class: "Local Polling"
+ha_release: "0.85"
+---
+
+Integration between [ebusd](https://github.com/john30/ebusd/) daemon for communication with eBUS heating systems, and homeassistant using sensor component.
+
+## {% linkable_title Configuration %}
+
+Enable the sensor by adding the following to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+ebusd:
+  host: 127.0.0.1
+  circuit: '700'
+  monitored_conditions:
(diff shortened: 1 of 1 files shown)


source/_components/ebusd.markdown
```diff
@@ -0,0 +1,123 @@
...
+## {% linkable_title Configuration %}
+
+Enable the sensor by adding the following to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+ebusd:
+  host: 127.0.0.1
+  circuit: '700'
+  monitored_conditions:
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_components/ebusd.markdown
```diff
@@ -0,0 +1,123 @@
...
+comments: false
+sharing: true
+footer: true
+logo: ebusd.png
+ha_category: Ebusd
+ha_iot_class: "Local Polling"
+ha_release: "0.85"
+---
+
+Integration between [ebusd](https://github.com/john30/ebusd/) daemon for communication with eBUS heating systems, and homeassistant using sensor component.
```
frenck:
```suggestion
Integration between [ebusd](https://github.com/john30/ebusd/) daemon for communication with eBUS heating systems, and Home Assistant using sensor component.
```

source/_components/ebusd.markdown
```diff
@@ -0,0 +1,123 @@
...
+  circuit: '700'
+  monitored_conditions:
+    - 'MaxFlowTemperatureDesired'
+    - 'WaterPressure'
+    - 'PumpStatus'
+```
+
+{% configuration %}
+host:
+  description: This is the IP address of your ebus daemon, eg. 127.0.0.1.
```
frenck:
```suggestion
  description: This is the IP address of your ebus daemon, e.g., `127.0.0.1`.
```

source/_components/ebusd.markdown
```diff
@@ -0,0 +1,123 @@
...
+  required: true
+  type: string
+circuit: 
+  description: The heating circuit name to monitor, eg. 700.
```
frenck:
```suggestion
  description: The heating circuit name to monitor, e.g., 700.
```

frenck (changes requested):
Thanks for the PR @CrazYoshi! 👍 
I've reviewed it and left a couple of small comments. Could you take a look? 👍

source/_components/ebusd.markdown
```diff
@@ -0,0 +1,119 @@
...
+layout: page
+title: "ebusd"
+description: "The ebusd component allows the integration between eBUS heating system and Home Assistant."
+date: 2018-12-27 15:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: ebusd.png
+ha_category: Ebusd
```
frenck: This is a non-existing category. Please check our current documentation for a list.

frenck (changes requested):
Thanks for updating the PR @CrazYoshi! 👍 
Nevertheless, found one more issue. Could you take a look? Thanks!

source/_components/ebusd.markdown
CrazYoshi: This is a new one, I have created a new component inside component/ebusd folder. Which is the correct category otherwise?

frenck: I would say: Sensor
Since it creates sensors for the ebusd heating systems.

CrazYoshi: yes, but it also have a service so it is not only a sensor. If you confirm I will make this correction anyway

frenck: As per home-assistant/architecture#131, sensors should not have services.

CrazYoshi: That's why I have created a new component named ebusd. That's also a category, correct?

frenck: No, it is not. Go to our website, hit the components page, categories are on your left menu. It would be weird to see a `ebusd` there.

source/_components/ebusd.markdown
```diff
@@ -0,0 +1,119 @@
...
+layout: page
+title: "ebusd"
+description: "The ebusd component allows the integration between eBUS heating system and Home Assistant."
+date: 2018-12-27 15:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: ebusd.png
+ha_category: Hub
```
frenck: We've had some discussion internally about the "Hub" category.
Conclusion: What "Hub" should be is when another Home Automation system/hub integrates with Home Assistant (e.g., Samsung SmartThings). While there are many things in Hub that should be moved out of there (which we started working on already), we should not add to it.

This component provides a Sensor, so that is what it should be categorized in.

```suggestion
ha_category: Sensor
```

frenck (changes requested):

frenck:
Thanks, @CrazYoshi!

frenck (dismissed):

rohankapoorcom (approved):

PR #7974, opened 2018-12-28
Update device_tracker.markdown

Jonher937:
**Description:**
hide_if_away does not take "no" as a valid argument according to the rest of the docs.
Let's change it to false where it was previously no

## Checklist:


fabaff (approved):
Thanks :bird:

PR #7982, opened 2018-12-29
Update harmony remote for new aioharmony library

ehendrix23:
**Description:**

-) Updated example as configuration file written is now in JSON format.
-) Added that Harmony configuration file is now automatically updated anytime the Harmony HUB configuration is updated.
-) Improve send_command example as it now also allows for the name of the device and not just device ID.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#19595

## Checklist:

### DIFF
--- a/source/_components/remote.harmony.markdown
+++ b/source/_components/remote.harmony.markdown
@@ -75,6 +75,8 @@ Upon startup one file will be written to your Home Assistant configuration direc
 - List of all programmed device names and ID numbers
 - List of all available commands per programmed device
 
+This file will be overwritten whenever the Harmony HUB has a new configuration, there is no need to restart HASS.
+
 ### {% linkable_title Service `remote.turn_off` %}
 
 Turn off all devices that were switched on from the start of the current activity.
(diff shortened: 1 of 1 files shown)


source/_components/remote.harmony.markdown
```diff
@@ -97,8 +99,14 @@ Start an activity. Will start the default `activity` from configuration.yaml if
 In the file 'harmony_REMOTENAME.conf' you can find the available activities, for example:
 
 ```text
-Activities
-  22054907 - Watch TV
+{
+    "Activities": {
```
MartinHjelmare, line 103: Were the docs stale from before this change? The merged PR didn't change the config file format, right?

ehendrix23: The config file that is written is in JSON format now. 
The file itself is not used by HASS but is rather just informative so people can reference to write services etc. to know when to write services etc. what different devices and commands are supported.


That was in the PR for the Harmony fix

MartinHjelmare: But it already saved the file as json? If we did change this, it's a breaking change, so it's important that we are clear about this.

ehendrix23: It did as from PR #19440 as at that time it was faster for me to get the websocket fixes in based on how pyharmony was done and other working in-progress code I had.

MartinHjelmare: Ah ok, so this is just correcting the docs from the earlier PR change.

Is this file meant for manual editing too?

ehendrix23: Correct. 

No, this file is informational so people know what type of commands can be sent etc.
The file is overwritten anytime HASS starts or configuration change on the Harmony itself.

MartinHjelmare (approved):

PR #7992, opened 2018-12-30
Adds instructions for adding BME680 to hass.io

ctborg:
**Description:**
Adds instructions for adding BME680 to hass.io

## Checklist:

### DIFF
--- a/source/_components/sensor.bme680.markdown
+++ b/source/_components/sensor.bme680.markdown
@@ -180,6 +180,33 @@ group:
       - sensor.bme680_sensor_air_quality
 ```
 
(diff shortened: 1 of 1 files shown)


frenck:
This feels more like a hack/workaround than an actual solution.
Secondly, while the debug command line is available, we do not want the user to use it for general configuration purposes.

IMHO it would make more sense on getting those kernel modules enabled in HassOS?

frenck:
Looked at it a little more. This can be done using a configuration USB stick and there is no need to go into the HostOS for this.
Therefore, I'm going to close this PR.

ctborg:
Sounds good.   Can you point to the documentation where I can enable I2C using the USB configuration.

PR #7996, opened 2018-12-31
Add Hass.io on generic Linux server as installation method

ajfriesen:
Also renamed Hass.io to Hass.io on HassOS because Hass.io can be installed in different ways.

**Description:**

## Checklist:


source/_docs/installation.markdown
```diff
@@ -27,7 +27,13 @@ Home Assistant provides multiple ways to be installed. A requirement is that you
     <div class='img-container'>
       <img src='/images/supported_brands/home-assistant.png' />
     </div>
-    <div class='title'>Hass.io<br>(Beginner friendly)</div>
+    <div class='title'>Hass.io on HassOS<br>(Beginner friendly)</div>
```
fabaff: No need to add HassOS as beginner shouldn't need to care about the underlying operating system.

source/_docs/installation.markdown
```diff
@@ -27,7 +27,13 @@ Home Assistant provides multiple ways to be installed. A requirement is that you
...
+  </a>
+  <a class='option-card' href='/hassio/installation/#alternative-install-on-generic-linux-server'>
```
fabaff: This is something which should go into the "Alternative installs" because that definitely not something that we recommend.

fabaff (changes requested):

source/_docs/installation.markdown
ajfriesen: That is something the home assistant team decides and that is totally okay.
When I want to add this to the page should I revert my commit and create a new one with this change but in the "Alternative installs" section?

Just do not know the workflow within a big project like this.

frenck: It is OK to update this PR @ajfriesen, that way this discussion and the end result will stick together.

DubhAd:
The install page has changed since you submitted this. You'll need to resubmit this for the new layout, in the community section.

fabaff (approved):

PR #8005, opened 2018-12-31
Extend disable Auto Start - HomeKit

cdce8p:
**Description:**
Include option with wait_template.
In response to: https://community.home-assistant.io/t/homekit-persistent-storage/84299/7

## Checklist:

### DIFF
--- a/source/_components/homekit.markdown
+++ b/source/_components/homekit.markdown
@@ -227,6 +227,41 @@ automation:
 ```
 {% endraw %}
 
+In some cases it might be desirable to check that all entities are available before starting `HomeKit`. This can be accomplished by adding and additional `binary_sensor` as follows:
+
+{% raw %}
+```yaml
+# Example checking specific entities to be available before start
+homekit:
+  auto_start: False
+
+binary_sensor:
+  - platform: template
+    sensors:
+      homekit_entity_status:
+        value_template: >-
+          {% if not states.light.kitchen_lights %}
+            false
+          {% elif not states.sensor.outside_temperature %}
+            false
+          # Repeat for every entity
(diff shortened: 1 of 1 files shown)


frenck:
Just out of curiosity, why do you create an extra binary sensor instead of just adding the Jinja template into the `wait_template`?

cdce8p:
Didn't though about it earlier. Although personally I do perfer the clean seperation of automation and sensor here. Especially with more then two entities to check for.

frenck:
While that might feel like so, there are 2 issues with this approach.

1. The template engine has to figure out which entities are used in the template. So if you take this approach, the best practice would be to specify the list of entities used in the binary sensor.
2. The binary sensor needs updating in the state engine, each time one of his members updates (even way after HomeKit as been started already). This is rather inefficient for something that could be just a wait template.

cdce8p:
Regarding your comment:
1. Although advised in general, in this case it isn't necessary to specify the entities since they would get extracted without any issues. At the moment it seems the given list would even be ignored. [binary_sensor/template.py](https://github.com/home-assistant/home-assistant/blob/ead38f60054162cd73e19ecf453e8b7933e82f14/homeassistant/components/binary_sensor/template.py#L75-L81)
 2. You are right here. I didn't though about this earlier.

I've updated the PR to use only the `wait_template`. Thanks for noticing this.

cdce8p:
@frenck Any update on this PR? I would really like for this solution to be available on the components page as well, so other user could find it easier.

frenck (approved):
Thanks for updating the PR @cdce8p!
And sorry for me not responding, somehow, I've missed the notification for this 😢

PR #8012, opened 2019-01-02
Adds Integration Sensor

dgomes:
**Description:**
The integration platform provides the Riemann sum (integration) of a numeric sensor.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#19703

## Checklist:


source/_components/sensor.energy.markdown
```diff
@@ -0,0 +1,49 @@
...
+## {% linkable_title Configuration %}
+
+To enable Energy Sensor in your installation, add the following to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: energy 
+    source: sensor.current_power
+    name: "energy consumed"
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_components/sensor.energy.markdown
```diff
@@ -0,0 +1,49 @@
...
+    round: 2
```
frenck: Same as above.

source/_components/sensor.energy.markdown
```diff
@@ -0,0 +1,49 @@
...
+```
+
+{% configuration %}
+source:
+  description: The entity ID of the sensor providing power readings (must provide values in W or kW).
+  required: true
+  type: string
+name:
+  description: Name to use in the frontend.
+  required: false
```
frenck: Missing `default`

source/_components/sensor.energy.markdown
```diff
@@ -0,0 +1,49 @@
...
+  type: string
+round:
+  description: Round the calculated energy value to at most N decimal places.
+  required: false
```
frenck: Missing `default`

frenck (changes requested):
Thanks for the PR @dgomes, I've reviewed it and left some small comments. Could you please take a look? Thanks 👍

source/_components/sensor.energy.markdown
dgomes: done 👍

source/_components/sensor.energy.markdown
dgomes: done 👍

source/_components/sensor.energy.markdown
dgomes: Right!

dgomes:
@frenck this sensor has changed scope after discussion in the original PR and therefore I've updated all the documentation

frenck (dismissed):
Thanks for updating the PR @dgomes! 👍 

frenck (approved):

frenck:
Thanks, @dgomes! 👍

PR #8019, opened 2019-01-02
Documenting impact of state restore change

DubhAd:
As it turns off all the automations without an `initial_state`, and it's hitting lots of folks, documenting it here ;)


ludeeus (approved):
Thanks 👍

PR #8065, opened 2019-01-05
Add redirects for esphome Platform

tjorim:
**Description:**
We only have docs for the esphome component, Balloob said there doesn't need to be pages for the other platforms, but at least redirects. *new platform in 0.85, so should be rc maybe?*

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#<home-assistant PR number goes here> NA

## Checklist:


MartinHjelmare:
Just a note that this PR included a diff to the credits when merged.

Issue #8069, opened 2019-01-05
Quotes on binary_sensor.command_line


Konstigt:
In the examples and guide text regarding payload_on/off on https://www.home-assistant.io/components/binary_sensor.command_line/ there are no quotes, single quotes, double quotes + a note in the bottom that says that you should use single quotes.

Confusing yes?

emontnemery:
Then please submit a PR to improve the examples and guide text. This can be done by just clicking the "edit this page on GitHub" link.

Konstigt:
Sorry, can't. Because I don't know what is the correct one of these three options. That's why I visited the page in the first place :)

emontnemery:
Aha, I didn't realize it was a question!

yaml supports both no quotes, single quotes and double quotes for strings.

With double quotes, escaping is possible, maybe this is why there is a suggestion to use single quotes to not have unintended escaping of commands.

@bphermansson you added the note about using single quotes, can you comment?

bphermansson:
My experience is that single quotes is what's working with system commands.

PR #8119, opened 2019-01-10
Update binary_sensor.command_line.markdown

emontnemery:
**Description:**
Clarify why it's preferred to enclose command string in single quotes and align the examples.
Related: #8069 

## Checklist:

### DIFF
--- a/source/_components/binary_sensor.command_line.markdown
+++ b/source/_components/binary_sensor.command_line.markdown
@@ -25,7 +25,9 @@ binary_sensor:
   - platform: command_line
     command: 'cat /proc/sys/net/ipv4/ip_forward'
 ```
-
+<p class='note'>
+It's highly recommended to enclose the command in single quotes `'` as it ensures all characters can be used in the command and reduces the risk of unintentional escaping. To include a single quote in a command enclosed in single quotes, double it: `''`.
+</p>
 {% configuration %}
 command:
   description: The action to take to get the value. It's highly recommended to enclose the command in single quotes `'` as it ensures all characters can be used in the command and reduces the risk of unintentional escaping. To include a single quote in a command enclosed in single quotes, double it: `''`.
--- a/source/_components/binary_sensor.command_line.markdown
+++ b/source/_components/binary_sensor.command_line.markdown
@@ -30,7 +30,7 @@ It's highly recommended to enclose the command in single quotes `'` as it ensure
 </p>
 {% configuration %}
 command:
-  description: The action to take to get the value. It's highly recommended to enclose the command in single quotes `'` as it ensures all characters can be used in the command and reduces the risk of unintentional escaping. To include a single quote in a command enclosed in single quotes, double it: `''`.
+  description: The action to take to get the value.
   required: true
   type: string
 name:
--- a/source/_components/binary_sensor.command_line.markdown
+++ b/source/_components/binary_sensor.command_line.markdown
@@ -23,12 +23,12 @@ To use your Command binary sensor in your installation, add the following to you
 # Example configuration.yaml entry
 binary_sensor:
   - platform: command_line
-    command: cat /proc/sys/net/ipv4/ip_forward
+    command: 'cat /proc/sys/net/ipv4/ip_forward'
 ```
 
 {% configuration %}
 command:
-  description: The action to take to get the value.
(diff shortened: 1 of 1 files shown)


emontnemery:
@bphermansson, @Konstigt: Do you think this is clearer?

source/_components/binary_sensor.command_line.markdown
```diff
@@ -23,12 +23,12 @@ To use your Command binary sensor in your installation, add the following to you
...
 binary_sensor:
   - platform: command_line
-    command: cat /proc/sys/net/ipv4/ip_forward
+    command: 'cat /proc/sys/net/ipv4/ip_forward'
 ```
 
 {% configuration %}
 command:
-  description: The action to take to get the value.
+  description: The action to take to get the value. It's highly recommended to enclose the command in single quotes `'` as it ensures all characters can be used in the command and reduces the risk of unintentional escaping. To include a single quote in a command enclosed in single quotes, double it: `''`.
```
klaasnicolaas:
```suggestion
  description: The action to take to get the value. It's highly recommended to enclose the command in single quotes `'` as it ensures all characters can be used in the command and reduces the risk of unintentional escaping. To include a single quote in a command enclosed in single quotes, double it: `''`.
```

Will solve the build error or remove the `:` after "double it"

source/_components/binary_sensor.command_line.markdown
```diff
@@ -23,12 +23,12 @@ To use your Command binary sensor in your installation, add the following to you
 # Example configuration.yaml entry
 binary_sensor:
   - platform: command_line
-    command: cat /proc/sys/net/ipv4/ip_forward
+    command: 'cat /proc/sys/net/ipv4/ip_forward'
 ```
```
klaasnicolaas: Why do not you put the warning / note here? With this: `<p class='note'> </p>?`

Like:
```suggestion
<p class='note'>
It's highly recommended to enclose the command in single quotes `'` as it ensures all characters can be used in the command and reduces the risk of unintentional escaping. To include a single quote in a command enclosed in single quotes, double it: `''`.
</p>
```
I wonder if you should put the (above) text in the description of the configuration, I think this is not really the intention.

source/_components/binary_sensor.command_line.markdown
emontnemery: @klaasnicolaas Your suggested change has no change, can you try again?

source/_components/binary_sensor.command_line.markdown
emontnemery: You mean move the warning / note from the configuration variable description to the basic example?

source/_components/binary_sensor.command_line.markdown
klaasnicolaas: What do you mean with try again?

source/_components/binary_sensor.command_line.markdown
klaasnicolaas: Change it to my suggestion, the warning is not something to put in the description of a variable.

source/_components/binary_sensor.command_line.markdown
emontnemery: I mean, your suggested change is same as what is already there. Maybe github is being buggy.. Can you explain how I should change to solve the build error again?

source/_components/binary_sensor.command_line.markdown
emontnemery: OK

source/_components/binary_sensor.command_line.markdown
emontnemery: nvm, moved the warning as you suggest below

klaasnicolaas (approved):
Think it is now better, [see also here](https://deploy-preview-8119--home-assistant-docs.netlify.com/components/binary_sensor.command_line/)

emontnemery:
Yeah, agree.
Do you think the language is OK btw, it does sound a bit off:
> It’s highly recommended to enclose the command in single quotes ' as it ensures all characters can be used in the command and reduces the risk of unintentional escaping. To include a single quote in a command enclosed in single quotes, double it: ''.

The point is that yaml has many ways of specifying strings and:
- Unquoted strings are tricky as they may be interpreted as other type than string
- Double quoted strings allow escaping, which is probably not desired

frenck (approved):
Thanks, @emontnemery! 👍

PR #8151, opened 2019-01-13
Minut point link fix 

ikifar2012:
**Description:**
The link to Minut Point was not correct. I switched it to minut.com


fabaff (approved):
Thanks :bird:

PR #8167, opened 2019-01-14
Fix wrong update interval for SolarEdge sensor

LouisMT:
**Description:**
While working on another PR I noticed that the documentation says that the SolarEdge sensor updates every 5 minutes, but I think this should be 10 minutes [according to these lines](https://github.com/home-assistant/home-assistant/blob/master/homeassistant/components/sensor/solaredge.py#L25-L26).

## Checklist:


fabaff (approved):
Thanks :bird:

PR #8174, opened 2019-01-15
guide for asuswrt to use the rt-n56u project, Padavan custom firmware

orrpan:
**Description:** Guide for making it possible to use asuswrt device tracker on the rt-n56u project, as the only difference is where the `dnsmasq.leases` is stored. 
`asuswrt`: /var/lib/misc/dnsmasq.leases
`the rt-n56u project` (currently): /tmp/dnsmasq.leases

## Checklist:

### DIFF
--- a/source/_components/device_tracker.asuswrt.markdown
+++ b/source/_components/device_tracker.asuswrt.markdown
@@ -28,15 +28,13 @@ Follow these steps to setup the link.
 ```
 $ find / -name "dnsmasq.leases" 
 ```
-3. Copy the full path of, example: 
-```
-$ /tmp/dnsmasq.leases
-```
+3. Copy or remember the full path of, example: 
+`/tmp/dnsmasq.leases`
 4. Create the folder if it does not exist: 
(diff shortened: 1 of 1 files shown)


frenck:
I'm not sure if we should add such specific documentation (which essentially is a "hack" to make it work). @fabaff, what do you think?

DubhAd:
IMO a separate component, or an update to the existing component with a configuration option, would be the right way to do this. That said, I personally have no objection to simple "hacks" like this.

fabaff:
I agree with @DubhAd that the proper way would be to extent the `asuswrt` tracker. Till then we should add those details.

fabaff:
Hmm, the branch is gone. I will fix it manually.

klaasnicolaas:
Did a merge so the file is now gone 😉

fabaff:
Was added with https://github.com/home-assistant/home-assistant.io/commit/d3160f40ad0838b007b05c9cefe96aebd8959506

PR #8180, opened 2019-01-15
Add documentation for Danfoss Air HRV.

JonasPed1:
**Description:**
Add documentation for Danfoss Air HRV component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#20138

## Checklist:


source/_components/danfoss_air.markdown
```diff
@@ -0,0 +1,32 @@
+---
+layout: page
+title: "Danfoss Air"
+description: "How to integrate Danfoss Air HRV in Home Assistant."
+date: 2019-01-14 20:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
```
rohankapoorcom, line 9: Please add a logo. You can add it to `source/images/supported_brands`

source/_components/danfoss_air.markdown
```diff
@@ -0,0 +1,32 @@
...
+ha_category: Hub
```
rohankapoorcom: This category doesn't seem right. I think this would be better to be a in `climate` or `fan`.

rohankapoorcom (changes requested):
Please also add pages for the platforms included with this component.

rohankapoorcom:
Looking like this is almost set - just needs pages for the platforms that are created as well.

JonasPed1:
I was actually a bit unsure if documentation was needed for the individual platforms as no configuration for these is possible. I have updated PR with documentation for sensor and binary sensor platform.

rohankapoorcom:
I think the new platforms documentation should indicate what things it provides sensors for.

JonasPed1:
I think making a list with the sensors will make sense. Should this go in to the platform pages or the component page? Looks like it is a bit of a mix in the existing documentation?

rohankapoorcom:
Platform pages makes more sense since there each platform exposes different sensors.

JonasPed1:
Updated the platform pages with a short description of each sensor.

rohankapoorcom (approved):
This looks good to me. Thanks for working through it.

rohankapoorcom:
@cgarwood I think this needs a version change to 0.87

cgarwood:
Good catch. New PR to update version: #8249

PR #8204, opened 2019-01-17
Update calendar.google.markdown

Chaotic:
Add in the expected format for the offset to help prevent bugs being entered that it only supports 2 digits (it does but it supports hours)
https://github.com/home-assistant/home-assistant/issues/20168
https://github.com/home-assistant/home-assistant/issues/11886
https://github.com/home-assistant/home-assistant/issues/5791

**Description:**

The doc for offset appears to be missing the expected format.  This has led to several bugs being filed when people attempt to use an offset over 2 digits since it is looking for MM or HH:MM format and not MMM format

## Checklist:

### DIFF
--- a/source/_components/calendar.google.markdown
+++ b/source/_components/calendar.google.markdown
@@ -129,7 +129,8 @@ entities:
     offset:
       description: >
         A set of characters that precede a number in the event title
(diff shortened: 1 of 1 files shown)


fabaff (approved):
Thanks :bird:

PR #8215, opened 2019-01-19
Partial reload IS supported with packages

tjorim:
**Description:**

The docs says partial reloading (of automations, customize, groups and scripts) is not supported when using packages. However, it's supported since 0.84 (and should have been documented when merging that PR).

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** https://github.com/home-assistant/home-assistant/pull/18884

## Checklist:

### DIFF
--- a/source/_docs/configuration.markdown
+++ b/source/_docs/configuration.markdown
@@ -35,7 +35,8 @@ If you run into trouble while configuring Home Assistant, have a look at the [co
 
 ## {% linkable_title Reloading changes %}
 
-You will have to restart Home Assistant for most changes to `configuration.yaml` to take effect. You can load changes to [automations](/docs/automation/), [customize](/docs/configuration/customizing-devices/), [groups](/components/group/), and [scripts](/components/script/) without restarting if you're not using [packages](/docs/configuration/packages/).
+You will have to restart Home Assistant for most changes to `configuration.yaml` to take effect.
+You can load changes to [automations](/docs/automation/), [customize](/docs/configuration/customizing-devices/), [groups](/components/group/), and [scripts](/components/script/) without restarting.
 
 <p class='note warning'>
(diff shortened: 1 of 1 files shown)


source/_docs/configuration.markdown
```diff
@@ -35,7 +35,8 @@ If you run into trouble while configuring Home Assistant, have a look at the [co
 
 ## {% linkable_title Reloading changes %}
 
-You will have to restart Home Assistant for most changes to `configuration.yaml` to take effect. You can load changes to [automations](/docs/automation/), [customize](/docs/configuration/customizing-devices/), [groups](/components/group/), and [scripts](/components/script/) without restarting if you're not using [packages](/docs/configuration/packages/).
+You will have to restart Home Assistant for most changes to `configuration.yaml` to take effect.
+You can load changes to [automations](/docs/automation/), [customize](/docs/configuration/customizing-devices/), [groups](/components/group/), and [scripts](/components/script/) without restarting.
```
rohankapoorcom, line 39: Worth mentioning that it works with packages as well?

ludeeus: I think we should only specify elements that do not work (like packages in the past).
Otherwise, we _should_ add that this works with config option 1 and 2, split configurations using `!include` and packages 🤷‍♂️

tjorim: I agree with Ludeeus things like this should be implied, and only things that are not compatible are worth mentioning.

ludeeus (approved):

ludeeus:
LGTM, @rohankapoorcom do you have an objection to merging this?

rohankapoorcom:
LGTM

rohankapoorcom (approved):

PR #8216, opened 2019-01-19
Make Liquid more strict

tjorim:
**Description:**

Instead of warnings (by default) it should output error messages on the console for each error and stop the build. Therefore this changes the error_mode to `strict`, it happens far too often that people forget to use raw-tags.

This option is documented here: https://jekyllrb.com/docs/configuration/liquid/

**Note:**
strict_variables and strict_filters are still false by default, should we enable this as well? https://github.com/Shopify/liquid#undefined-variables-and-filters

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** NA

## Checklist:

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -32,8 +32,12 @@ destination: public/
 plugins_dir: plugins
 code_dir: downloads/code
 category_dir: blog/categories
+
 markdown: kramdown
+highlighter: rouge
 timezone: UTC
+liquid:
+  error_mode: strict
 
 kramdown:
   input: GFM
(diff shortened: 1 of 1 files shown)


tjorim:
Looks to be working as intended since PR https://github.com/home-assistant/home-assistant.io/pull/8210 is not included yet. Once that gets merged I'll upstream this PR to check if the new build is successful.

Edit: okay, here we gooo!!

Edit 2: LGTM, it builds now.

frenck:
Tested locally and works like a charm!
Really nice addition @tjorim! 👍

frenck (approved):

PR #8232, opened 2019-01-21
Add command line argument --init

jurgenhaas:
See also https://github.com/home-assistant/home-assistant/issues/20260, the --init argument makes sure that the python service does not run with PID 1 which on some host systems can not be killed, i.e. restarting the service from within the UI is unreliable without that argument.

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/installation/docker.markdown
+++ b/source/_docs/installation/docker.markdown
@@ -15,19 +15,19 @@ Installation with Docker is straightforward. Adjust the following command so tha
 ### {% linkable_title Linux %}
 
 ```bash
-$ docker run -d --name="home-assistant" -v /PATH_TO_YOUR_CONFIG:/config -v /etc/localtime:/etc/localtime:ro --net=host homeassistant/home-assistant
+$ docker run --init -d --name="home-assistant" -v /PATH_TO_YOUR_CONFIG:/config -v /etc/localtime:/etc/localtime:ro --net=host homeassistant/home-assistant
 ```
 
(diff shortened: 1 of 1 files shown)


kellerza (approved):
Thanks @jurgenhaas! From the docs I read it seems valid, not sure why its currently working fine in my environment.

@michaelarnauts might also have a better view on this and would appreciate his review.

michaelarnauts:
I haven't used `--init` before with docker, but after a quick look at the docs, it makes sense to use something like this in since we are starting python directly as pid 1.

However, it seem to be an unsupported option in docker-compose.yml files, so I'm not sure if most installations have a real benefit for this, since docker-compose is a more convenient way to start hass. 

Maybe it makes more sense to create a wrapper script (something like tini does for the `--init` option) and run that instead. A really quick look at hass.io seems that they do this by starting a hassio python module instead of hass directly, and then the hassio module can restart hass when it is restarted trough the UI.

I think @pvizeli has a more clear view on this.

michaelarnauts:
If we solve this issue by editing the Dockerfile to use a startup script, we don't need to change the documentation (since then we will do this twice).

frenck:
IMHO, we should use a small init system like dumb init or tiny to solve the general issue around this subject.

pvizeli (approved):

frenck:
Moving this one forward. Thanks for the PR @jurgenhaas! 👍

jurgenhaas:
Thanks @frenck glad I was able to help, even if with a tiny piece. But you have to get started somewhere, I guess.

PR #8313, opened 2019-01-27
Fixing the example

DubhAd:
The example isn't valid 🤦‍♂️


klaasnicolaas (approved):

klaasnicolaas:
Thnx for the PR! 👍

havardelnan:
This breaks most configs as ssl isnt on by default on unc. A optional setting should not be listed in the main example with the non default value. Better to remove ssl as it isnt mandatory.

klaasnicolaas:
@havardelnan I have opened a new PR to solve this, you are right. It is optional so it should not be in the example config. #8317

PR #8315, opened 2019-01-27
ADS light: Add "friendly_name" config option

carstenschroeder:
**Description:**
Add config option friendly_name to ADS light component. When it's used, option "name" will only be used for entity id definition. This way we can have stable entity ids even if the friendly names change.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#20511

## Checklist:


carstenschroeder:
Not needed anymore as unique id will be used...

PR #8329, opened 2019-01-28
Update recorder.markdown

quentinsf:
**Description:**

On my up-to-date Hassio installation, the 'mysql+pymysql' syntax doesn't work for MariaDB.  Using just 'mysql:' and specifying the port works fine.

## Checklist:

### DIFF
--- a/source/_components/recorder.markdown
+++ b/source/_components/recorder.markdown
@@ -142,6 +142,7 @@ Purging does not necessarily remove all entries before a given date. For example
 | SQLite          | `sqlite:////PATH/TO/DB_NAME`                             |
 | MariaDB         | `mysql+pymysql://SERVER_IP/DB_NAME?charset=utf8`                 |
 | MariaDB         | `mysql+pymysql://user:password@SERVER_IP/DB_NAME?charset=utf8`   |
+| MariaDB         | `mysql://user:password@SERVER_IP:3307/DB_NAME?charset=utf8`   |
 | MySQL           | `mysql://SERVER_IP/DB_NAME?charset=utf8`         |
 | MySQL           | `mysql://user:password@SERVER_IP/DB_NAME?charset=utf8` |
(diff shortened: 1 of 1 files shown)


frenck:
I guess both MariaDB examples can be replaced since it is exactly the same as MySQL (which has been modified recently).

source/_components/recorder.markdown
```diff
@@ -142,6 +142,7 @@ Purging does not necessarily remove all entries before a given date. For example
 | SQLite          | `sqlite:////PATH/TO/DB_NAME`                             |
 | MariaDB         | `mysql+pymysql://SERVER_IP/DB_NAME?charset=utf8`                 |
 | MariaDB         | `mysql+pymysql://user:password@SERVER_IP/DB_NAME?charset=utf8`   |
+| MariaDB         | `mysql://user:password@SERVER_IP:3307/DB_NAME?charset=utf8`   |
```
DubhAd, line 145: Should mention the version number of MariaDB here, and remove the note under the table

DubhAd (changes requested):

source/_components/recorder.markdown
quentinsf: Thanks - some thoughts:

* If these lines are specifications rather than examples, then I guess the 3307 should be replaced by PORT.  The key thing is that the default port is different from MySQL, though.  So perhaps it's worth leaving the 3307 and changing the message under the table to point out that you may need to change this if you aren't using the default.

* The most important thing, though, is that the two mysql+pymysql MariaDB lines above *do not* work by default on typical current MariaDB installations (mine is v10), so I'd suggest they are removed or at least demoted to a lower place, and my guess is that, if the +pymysql variant works at all, it would also need :PORT to be added.

Many thanks!

frenck:

quentinsf:
Apologies, Franck - I hadn't noticed that the ball was back in my court on this one.   I'll create a new PR with the discussed changes.

Quentin

PR #8346, opened 2019-01-30
Change documentation for googlehome

ludeeus:
**Description:**

Changes to the documentation for googlehome.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#19971

## Checklist:


klaasnicolaas (dismissed):

source/_components/googlehome.markdown
```diff
@@ -0,0 +1,69 @@
...
+layout: page
+title: "Google Home"
+description: "Instructions on how to connect to your Google Home device."
+date: 2018-11-04 00:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: google_home.png
+ha_release: 0.83
```
MartinHjelmare: 0.88

MartinHjelmare: The component is new but not the platform. 🤷‍♂️

source/_components/googlehome.markdown
```diff
@@ -0,0 +1,69 @@
...
+    rssi_threshold:
+      description: The threshold for the RSSI signal of the device.
+      required: false
+      default: -70
+      type: integer
+    device_types:
+      description: Device types that will be tracked [see device types](#device_types), by default all types are tracked.
+      required: false
+      type: list
+{% endconfiguration %}
```
MartinHjelmare, line 54: We should add the possible items of the list here in `keys` I think.

source/_components/googlehome.markdown
ludeeus: Was unsure about this, updated to 88

source/_components/googlehome.markdown
```diff
@@ -0,0 +1,68 @@
...
+      description: The threshold for the RSSI signal of the device.
+      required: false
+      default: -70
+      type: integer
+    device_types:
+      description: Device types that will be tracked [see device types](#device_types), by default all types are tracked.
+      required: false
+      type: list
+      keys:
+        1:
```
MartinHjelmare: Maybe this just looks messy? If we want to keep it we have to quote the numbers. But I think we can revert to the previous way it was. Sorry.

ludeeus: This way is more correct according to the standard, but I agree that this looks messier 🤷‍♂️

ludeeus: Added quotes now, so we can see when it has rendered :)

MartinHjelmare: The previous way was better I think. I don't understand what 1, 2, 3 means.

MartinHjelmare (approved):

PR #8380, opened 2019-02-02
Update sensor.yr.markdown

Danielhiversen:
**Description:**
Add location parameter to config

## Checklist:

### DIFF
--- a/source/_components/sensor.yr.markdown
+++ b/source/_components/sensor.yr.markdown
@@ -71,18 +71,18 @@ monitored_conditions:
 latitude:
   description: Manually specify latitude.
   required: false
-  type: number
-  default: Provided by Home Assistant configuration
+  type: float
+  default: Provided by Home Assistant configuration.
 longitude:
   description: Manually specify longitude.
   required: false
-  type: number
-  default: Provided by Home Assistant configuration
+  type: float
+  default: Provided by Home Assistant configuration.
 altitude:
   description: Manually specify altitude.
   required: false
-  type: number
-  default: Provided by Home Assistant configuration
+  type: float
+  default: Provided by Home Assistant configuration.
 {% endconfiguration %}
 
 A full configuration example can be found below:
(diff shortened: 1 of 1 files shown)


source/_components/sensor.yr.markdown
```diff
@@ -68,6 +68,21 @@ monitored_conditions:
       description: Medium cloud level.
     highClouds:
       description: High cloud level.
+latitude:
+  description: Manually specify latitude. By default the value will be taken from the Home Assistant configuration.
```
DubhAd: By default the value will be taken from the Home Assistant configuration.

Seems redundant when you mention the default immediately below ;)

source/_components/sensor.yr.markdown
```diff
@@ -68,6 +68,21 @@ monitored_conditions:
...
+  required: false
+  type: number
+  default: Provided by Home Assistant configuration
+longitude:
+  description: Manually specify longitude. By default the value will be taken from the Home Assistant configuration.
```
DubhAd: Same

source/_components/sensor.yr.markdown
```diff
@@ -68,6 +68,21 @@ monitored_conditions:
...
+  required: false
+  type: number
+  default: Provided by Home Assistant configuration
+altitude:
+  description: Manually specify altitude. By default the value will be taken from the Home Assistant configuration.
```
DubhAd: Same

DubhAd (dismissed):

source/_components/sensor.yr.markdown
```diff
@@ -68,6 +68,21 @@ monitored_conditions:
       description: Medium cloud level.
     highClouds:
       description: High cloud level.
+latitude:
+  description: Manually specify latitude.
+  required: false
+  type: number
```
klaasnicolaas: Is number a legit type? Or should it be a float?

Danielhiversen: Is not a float a sub type of number?

klaasnicolaas: Well I see [here](https://developers.home-assistant.io/docs/en/documentation_create_page.html) number not as variable type.

frenck: In the docs that would be translated into a `float`.
`number` is not valid in the docs.

frenck (approved):

PR #8389, opened 2019-02-02
Fix SmartThings Switch Platform

andrewsayre:
**Description:**
Reverts the changes of #8357 so that the switch platform is consistent with the other platforms that have integration details (#8388, #8376, and #8365).

While switch doesn't have details today, it will in the future, once it supports other capabilities, like power meter.  To keep the component page lean (instructions for setting up the integration), the SmartThings-device-to-platform mapping details are within each platform page.

## Checklist:

### DIFF
--- a/source/_components/smartthings.switch.markdown
+++ b/source/_components/smartthings.switch.markdown
@@ -1,0 +1,19 @@
+---
+layout: page
+title: "SmartThings Switch"
+description: "Instructions on setting up Samsung SmartThings switches within Home Assistant."
+date: 2018-01-14 00:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: samsung_smartthings.png
+ha_category: Switch
+ha_release: "0.87"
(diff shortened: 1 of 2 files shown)


source/_components/smartthings.switch.markdown
```diff
@@ -0,0 +1,19 @@
+---
```
fabaff, line 1: The SmartThings documentation should follow the new approach for the documentation for integrations. We decided to keep the things in the documentation together thus we don't want platform documentation for integration which are powered by a component. The platform details should be part of the component documentation. There are already examples present and this work much better than providing multiple times the same standard phrases.

fabaff (changes requested):

source/_components/smartthings.switch.markdown
andrewsayre: OK, I'll close this PR and update the other in-flight items.  A new PR will be needed to pull in lights.

PR #8415, opened 2019-02-05
Clarify MDI version currently used by HA

akasma74:
HA is currently using [mdi 3.0.39](https://cdn.materialdesignicons.com/3.0.39/), which is not the current one.
As per https://github.com/home-assistant/home-assistant-polymer/issues/2684

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/configuration/customizing-devices.markdown
+++ b/source/_docs/configuration/customizing-devices.markdown
@@ -58,7 +58,7 @@ entity_picture:
   required: false
   type: string
 icon:
-  description: Any icon from [MaterialDesignIcons.com](http://MaterialDesignIcons.com) ([Cheatsheet](https://materialdesignicons.com/cheatsheet)). Prefix name with `mdi:`, ie `mdi:home`.
+  description: Any icon from [MaterialDesignIcons.com](http://MaterialDesignIcons.com)'s v.3.0.39 ([Cheatsheet](https://cdn.materialdesignicons.com/3.0.39/)). Prefix name with `mdi:`, ie `mdi:home`.
   required: false
   type: string
 assumed_state:


klaasnicolaas:
Tomorrow it will be upgraded to 3.3.92 😉 
https://rc--home-assistant-docs.netlify.com/blog/2019/01/30/release-87/

fabaff (changes requested):
We shouldn't add releases numbers because this increases the amount of maintenance of the page significantly.

fabaff:
Link to https://github.com/home-assistant/home-assistant-polymer/blob/dev/package.json#L20 could be an option.

akasma74:
> We shouldn't add releases numbers because this increases the amount of maintenance of the page significantly.

Then you need to come up with an alternative to it, otherwise a lot of people will spend hours trying to find out why that particular icon does not show up.

akasma74:
> Link to https://github.com/home-assistant/home-assistant-polymer/blob/dev/package.json#L20 could be an option.

I can see only "@mdi/svg": "^3.0.39".
The problem is it does not reflect the current release state - for example, HA 0.87.0 uses a different version of MDI.
So the issue stands - we need a way to know the MDI version supported.
Any thoughts?

frenck:
@akasma74 Use the master branch instead of dev?
<https://github.com/home-assistant/home-assistant-polymer/blob/master/package.json#L20>

akasma74:
I can see the same "@mdi/svg": "^3.0.39" when follow your link

frenck:
@akasma74 Yep, but at least it is not the dev branch, so more accurate... (usually).

akasma74:
I still don't know the right link to use.
HA devs, any ideas?

akasma74:
bump

frenck:
@akasma74 There is no way to link the code base (truth) to the documentation right now.
So whatever we list in here, will be outdated, probably, every release.

We could, however, say something like: `Please note, that new icons might not be available in Home Assistant yet.`

IMHO it is simple, honest and explanatory.

frenck:
Superseded by #9416.
Closing this PR.

akasma74:
> @akasma74 There is no way to link the code base (truth) to the documentation right now.

got it

> We could, however, say something like: `Please note, that new icons might not be available in Home Assistant yet.`

that will do to start with.
but I'm curious how exactly does the process of adding a new MDI icon to HA look?
what is involved?

maybe by thinking about it I can come up with more detailed sentence as there is no such thing as too much details, is it? ;)

frenck:
Well, for starters, yes, there is a thing called too much detail. There is a big difference between user and developer documentation.

The file linked before shows the version scope the MDI icon package is in at the Home Assistant repo right now. On release, the build script will fetch the latest version matching those version requirements. Sometimes the version is pinned, sometimes it isn't.

PR #8419, opened 2019-02-06
Clarification about Lost Connection Detection

dagobert:
**Description:**

For people who are new to the CCU and specially writing a program it is helpful to mention that the script needs to reside in the "Then" clause even there is no condition. Even programming aware people might run into this problem.


frenck (approved):
Thanks, @dagobert 👍

PR #8420, opened 2019-02-06
Updated automation to switch tariffs

ardeus-ua:
Great sensor, automation is checking the condition after HA start (to avoid wrong tariff) and then triggers on the time of tariff change.
It would be great if tariff time would be possible to configure within a sensor setup

**Description:**

## Checklist:

### DIFF
--- a/source/_components/utility_meter.markdown
+++ b/source/_components/utility_meter.markdown
@@ -117,25 +117,18 @@ a time based automation can be used:
 
 ```yaml
 automation:  
-  - alias: switch_tariff
-    initial_state: true
-    trigger:
-      - platform: homeassistant
-        event: start
-      - platform: time
-        at: '09:00:00'
-      - platform: time
-        at: '21:00:00'
-      - platform: state
-        entity_id: sensor.tariff
-    action:
-      - service: utility_meter.select_tariff
-        data_template:
-          entity_id: 'utility_meter.daily_energy'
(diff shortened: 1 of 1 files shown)


klaasnicolaas:
`rc` is not the right branch to merge in it, please rebase to `current`

frenck:
Dual tariff is not used in a large part of the world. Maybe add it as a second example instead of replacing the current?

ardeus-ua:
As this component is actually made to measure dual tariff, or triple, you can easily expand to as many tariffs you want. My add is to switch to exact tariff, not to the next one. It may occur that you restart your HA right in a moment of tariff switch or your server's power source is not secured with UPS and you had power outage. This way you will know for sure. But to be honest, both examples will give users more possibilities to apply in their exact situation.

frenck:
> But to be honest, both examples will give users more possibilities to apply in their exact situation.

Hence my suggestion on adding it as a second example, you've currently replaced it.

ardeus-ua:
> Hence my suggestion on adding it as a second example, you've currently replaced it.
Agree. added instead of replacing

frenck:
@ardeus-ua Your PR still has a build error. Are you planning on resolving that?

Furthermore, a merge conflict has appeared. Please resolve, so we can go ahead with the merge.

frenck:

PR #8421, opened 2019-02-06
Neato supports custom cleaning

RomRider:
**Description:**
Adding custom cleaning to Neato

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#20779
## Checklist:

### DIFF
--- a/source/_components/neato.markdown
+++ b/source/_components/neato.markdown
@@ -67,11 +67,11 @@ Currently supported services are:
 
 And a specific Platform Service:
 
-- `neato.neato_custom_cleaning`
+- `vacuum.neato_custom_cleaning`
 
 ### {% linkable_title Platform Services %}
 
-#### {% linkable_title Service `neato.neato_custom_cleaning` %}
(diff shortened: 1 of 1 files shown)


source/_components/neato.markdown
```diff
@@ -64,6 +64,7 @@ Currently supported services are:
 - `return_to_base`
 - `locate`
 - `spot_clean`
+- `neato_custom_cleaning` (The `zone` parameter is only supported for the Botvac D7)
```
dshokouhi: I wonder if we should provide a table of what all the values mean and which values are supported by which botvac models.  Maybe for the time being we just link to this page so users can see a better description?  The only issue is that actual botvac models are not listed here.

https://developers.neatorobotics.com/api/robot-remote-protocol/housecleaning

RomRider: Done in the latest commit

frenck (approved):
Look good @RomRider! Thanks! 👍

PR #8431, opened 2019-02-07
Adding example for doing a search

DavidFW1960:
Not intuitive and it took me a while but I worked it out so I'm documenting here.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/sensor.imap.markdown
+++ b/source/_components/sensor.imap.markdown
@@ -28,6 +28,7 @@ sensor:
     port: 993
     username: YOUR_USERNAME
     password: YOUR_PASSWORD
+    search: FROM <sender@email.com>, SUBJECT <subject here>
 ```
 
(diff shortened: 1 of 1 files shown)


klaasnicolaas:
If you are doing it anyway, you could also add the other [configuration variables](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/sensor/imap.py#L33) that are still missing.

You have to delete the `port` and `search` variable from the yaml example, only the variables that are required is allowed on the first example (so that beginners can easily copy it to there own config).

DavidFW1960:
The only thing I have done is add a search example. Nothing else has been changed.

klaasnicolaas:
I know that, but with my additional suggestions, the page will be in line with the guidelines we use.

fabaff (approved):

Issue #6402, opened 2018-10-01
Add "AntiFeatures" categories to component listings


arthurlutz:
First off : kudos to home-assistant and it's brilliant ecosystem. I'm a big fan. 

Second : I know this might be a controversial feature request for the web site, I am hoping that this will be a constructive discussion on how we can provide useful information to existing and potential users of home-assistant. 

My feature request is the following : it would be awesome to have some way of categorising and listing components according to the following question "does the component use an external non-free service?" "does the component rely on closed source hardware?" "does the component download and use non-free software to work?"

The closest type of approach I can think of (thus the title of the issue) is the categories provided by fdroid https://f-droid.org/ (a FLOSS alternative to Google Play for android)  https://f-droid.org/wiki/page/Antifeatures

I think the following "flags" can get us starting the discussion : 

* Non-free Network Services 
* Non free Dependencies
* Tracking 
* FLOSS web services needed (possibility of self-hosting) 

As a side note : maybe there could be a "flag" such as "Requires registration of a web service". 

One example : network speed probing : 

* fast.com : Non-free Network Services 
* speedtest.net : Non-free Network Services 
* iperf3 : FLOSS web service

I can think of a bunch more examples, but I'll stop there for conciseness.

timmo001:
I guess this could be done with labels similar to how GitHub works.

Not sure if this is a definite, but I know there is a PR for migrating the site to Hugo. #5744 

From that, Hugo has a feature known as [taxonomies](https://gohugo.io/content-management/taxonomies/) which are essentially tags. We could probably throw in a bunch of tags for each component like `non-free`, `3rd-party`, `self-hosted`, `diy` etc.

fabaff:
This topic was coming up a few times in the past but was never implemented. The reasons are manifold. The biggest blocker is that there are no resources to perform reviews to categorized existing  integrations and check it for new ones. Even if we limit it to everything that is communicating at the moment with a third-party web service (sensors and notifications) are we talking about approx. 100 integrations (I'm guessing). There will be only value for the users if the coverage over the integrations is large. I doubt that a lot of users care about a "Non-free Network Services" as long as the usage is free. And those who cares will probably be capable to figure out an alternative.

Also, for contributors it must be super easy to determine if the integrations is a match for the particular flag. This means that "Non free Dependencies" and "Tracking" are not really practicable because it would require that a license review is performed and the network traffic analysed in detail.

[...]

PR #8498, opened 2019-02-12
Add tag information for "antifeatures"

arthurlutz:
so far : self-hosting, requires-registration, proprietary,
opensource, openhardware

Related to #6402

## Checklist:

### DIFF
--- a/source/_components/arduino.markdown
+++ b/source/_components/arduino.markdown
@@ -12,6 +12,7 @@ ha_category: DIY
 featured: true
 ha_release: pre 0.7
 ha_iot_class: "Local Polling"
+ha_tags: opensource, openhardware
 ---
 
 The [Arduino](https://www.arduino.cc/) device family are microcontroller boards that are often based on the ATmega328 chip. They come with digital input/output pins (some can be used as PWM outputs), analog inputs, and a USB connection.
--- a/source/_components/binary_sensor.nest.markdown
+++ b/source/_components/binary_sensor.nest.markdown
@@ -11,6 +11,7 @@ logo: nest.png
 ha_category: Binary Sensor
 ha_release: pre 0.7
 ha_iot_class: "Cloud Push"
+ha_tags: proprietary, requires-registration
 ---
 
 The `nest` binary sensor platform lets you monitor various states of your [Nest](https://nest.com) devices.
--- a/source/_components/binary_sensor.octoprint.markdown
+++ b/source/_components/binary_sensor.octoprint.markdown
@@ -11,6 +11,7 @@ logo: octoprint.png
 ha_category: Binary Sensor
 ha_release: 0.19
 ha_iot_class: "Local Polling"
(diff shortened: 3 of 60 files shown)


frenck:
@arthurlutz IMHO you are to soon, the discussion in the related issue was not resolved. Nevertheless, you are opening a PR.

We are currently reviewing the way the documentation is structured. One of the recent changes made it possible to reduce the number of pages and allows a page to be in multiple categories.

While I do appreciate the effort, I do not agree on the chosen route. Just adding a `tags` parameter and simply displaying them? The name tags are misused in this case, it is not a tag, but a collection of properties. 

Secondly, as @fabaff pointed out in #6402:

> Also, for contributors it must be super easy to determine if the integrations is a match for the particular flag. This means that "Non free Dependencies" and "Tracking" are not really practicable because it would require that a license review is performed and the network traffic analysed in detail.

Other thoughtfull mentions: self-hosted is not always the only option. Some offert both self-hosted and a hosted version. So how would that reflect in this change?

I think we should close this PR, until the discussion is resolved, before doing this.

frenck (approved):
Blocking PR until the discussion is resolved.

frenck (changes requested):
Blocking PR until the discussion is resolved.

arthurlutz:
@frenck thanks for taking the time to review my code and answer in detail. 

I am trying to figure out how to start on this subject, and I think that discussing over some code (and the awesome netlify previews) is easier than discussing theoretically in the issue so I am glad to get your feed back. 

No, displaying is not enough, I am trying out changes to components/index.html to have more facets and displaying information. 

As for the name of "property", lets discuss it, "Naming things is difficult", that's ok. 

For the self-hosted and hosted also available. I would recommend having a  "self-hosting" (first) and "hosted-sass-also-available" to encourage user to choose self-hosting which is better in terms of privacy.

balloob:
I don't want to merge this.

With Home Assistant we're aiming to reach a wide audience, wider than just "techies". The non-techies will not understand this and for them it will just be extra words on the page that they don't understand, scaring them away.

I also agree with others that the maintenance burden is too big. You're new to this community and want us to make changes that have to be supported forever, even when you are long gone.

So no, we're not doing this.

arthurlutz:
I'm sorry to see this closed this way and the way it is being closed. 

> You're new to the community

How would you define "being established in the community" ? I've been a user of home-assistant for a while now, have presented home-assistant in python conferences in France (https://youtu.be/Eu6umBJ51I4?t=180) ... I am starting to regret promoting the plateform if we can't find a user friendly way of encouraging **privacy first** services with home-assistant. 

> even when you are gone.

I'm familiar with open source projects and accept that you can refuse something you're not at ease maintaining, but please don't presume of my implication in the project. 

I'm trying to stay open to this discussion because I firmly believe in open-source, privacy oriented, user-oriented informed choice for home automation, but you're making it a bit hard.

Issue #8502, opened 2019-02-12
Display and filter by ha_iot_class on components search page


arthurlutz:
On https://www.home-assistant.io/components/ it would be nice to have the `ha_iot_class` displayed and part of the facets.

frenck:
Already covered in #6387

arthurlutz:
Sorry for the duplicate, will update my pull request.

Issue #6387, opened 2018-09-30
Check the IoT classes


fabaff:
The [`ha_iot_class`](https://developers.home-assistant.io/docs/en/documentation_create_page.html#configuration) entries could be outdated. Check if the class still matches the current implementation.

This the tracking issue. Please don't close but add as a reference to your Pull request.

fabaff:
#6408

klaasnicolaas:
will look if I can help with this 😉

fabaff:
That would be awesome. Adding a feature to filter the integration according the `ha_iot_class` is still on the ToDo list somewhere.

klaasnicolaas:
not `stale` still a work in progress

PR #8506, opened 2019-02-12
ha_iot_class in facets

arthurlutz:
Related to #8502 and #6387

## Checklist:

### DIFF
--- a/source/_components/binary_sensor.egardia.markdown
+++ b/source/_components/binary_sensor.egardia.markdown
@@ -10,7 +10,7 @@ footer: true
 logo: egardia.png
 ha_release: 0.65
 ha_category: Binary Sensor
-ha_iot_class: "Polling"
+ha_iot_class: "Cloud Polling"
 ---
 
 The `egardia` platform allows you to get data from your [Egardia](http://www.egardia.com)/[Woonveilig](http://www.woonveilig.nl) binary sensors from within Home Assistant.
--- a/source/_components/binary_sensor.mercedesme.markdown
+++ b/source/_components/binary_sensor.mercedesme.markdown
@@ -9,7 +9,7 @@ sharing: true
 footer: true
 logo: mercedesme.png
 ha_category: Car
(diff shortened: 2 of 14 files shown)


frenck:
#8502 has been closed and not merged, how does this PR related to a closed one?
Could you please clarify?

- Addition of iot_classes has not been completed as well, I'm not sure if this is the right timing.
- iot_class is misused in some places by listing multiple, you've updated that to be `,` separated? Why not turn it into a list?
- PR should target the `current` branch.

Secondly, you start opening PR on structural changes like a train...
Could you please stop that and start discussing things in the appropriate locations, please?

Thanks.

balloob:
Breaks the component page + we cannot put IoT class in the same list as the categories.

image

arthurlutz:
@balloob indeed, thanks for taking a look. Some CSS is probably required to distinguish categories and the IOT_class part. (:heart: netlify). Am also having difficulties setting up my local jekyll setup (more of a python person than ruby, so am a bit lost, but finding my way).  

@frenck here are a few answers : 

* it feels like the coverage of iot_class is pretty good to start using it, but I can understand that this change could have to wait for the coverage to be greater (having this view also helps check the consistency and might help in going through the components which are missing this information or where it is not applicable)
* you're right about the list, I'll fix to avoid parsing strings
* can you explain why the PR should target main branch ? the PR template asks to target next
* I'd rather try out code and discuss over netlify and screenshots instead of over-engineering in issues, this helps me get into jekyll and the documentation, and I fully understand that these PR will not be accepted as is, and that they will require improvements or not fit the needs and be closed

frenck:
@arthurlutz As per PR template: 
> `next` is for changes and new documentation that will go public with the next [home-assistant](https://github.com/home-assistant/home-assistant) release.

Since this change has no upstream change in our code base and thus does not depend on the next Home Assistant release; it should go into `current`.

klaasnicolaas:
Will close it for now, feel free to open it when the time is there 😉

PR #8515, opened 2019-02-12
Remove section which talks about loggin in with google/facebook

arthurlutz:
"Information We Collect from Other Sources" probably comes from the Automattic privacy policy since you can login to wordpress with Google and Facebook. From what I understand this does not apply to home-assistant's site or documentation.

## Checklist:


fabaff:
The forum allows logging in with the help of third-party services.

Changes to policies and alike needs to be discussed in the arch repo.

arthurlutz:
@fabaff thanks for taking the time to take a look at this. 

Sorry for the confusion, since the forum has its own privacy policy at https://community.home-assistant.io/privacy I though this privacy policy was only for the website. Maybe the forum is missing a link to https://www.home-assistant.io/privacy/ to say that it also applies to the community forum ?

PR #8519, opened 2019-02-12
Added refresh_rate + scan_interval co2signal

passie:
Added refresh_rate + scan_interval to avoid problems with rate limit

**Description:**
Co2signal component is hitting a rate limit.
This will fix the rate limit and avoid being blocked.


frenck:
This is an update to the current documentation, and therefore needs to go into the `current` branch. Please change the target and rebase if needed.

Actually, why are we adding this? Shouldn't we update the component to have sane defaults?

passie:
> This is an update to the current documentation, and therefore needs to go into the `current` branch. Please change the target and rebase if needed.
> 
> Actually, why are we adding this? Shouldn't we update the component to have sane defaults?

I agree on both parts and have changed the branch to current. 
Since I don't have the time nor the knowledge to change these settings in the current code , this is a quick work around to not hit the limit which is set by the co2signal owners.

klaasnicolaas:
You must rebase this PR

PR #8557, opened 2019-02-13
Add method option to sensor.integration

dgomes:
**Description:**

Adds 'method' option to sensor.integration

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#21050

## Checklist:

### DIFF
--- a/source/_components/sensor.integration.markdown
+++ b/source/_components/sensor.integration.markdown
@@ -7,14 +7,14 @@ sidebar: true
 comments: false
 sharing: true
 footer: true
-ha_category: Utility 
+ha_category: Utility, Energy 
 ha_release: 0.87
 ha_iot_class: "Local Push"
 logo: integral.png
 ha_qa_scale: internal
 ---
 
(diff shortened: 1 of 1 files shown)


source/_components/sensor.integration.markdown
```diff
@@ -7,14 +7,14 @@ sidebar: true
 comments: false
 sharing: true
 footer: true
-ha_category: Utility 
+ha_category: Utility, Energy
```
klaasnicolaas:
```suggestion
ha_category: 
  - Utility 
  - Energy 
```

klaasnicolaas (dismissed):

source/_components/sensor.integration.markdown
dgomes: Thank you soo much yet again :)

tim-devel:
I have just started using this sensor and was a bit blinded by some of the mathematical terminology. Can you consider adding some more description for the laymen please.  For example what is the difference between trapezoidal and left right and what are the use cases for each

dgomes:
The best explanation is IMHO the wikipedia article which is already linked (explaining Reimann sum)

The use case for each of them depend too much on the user, and it's specific source :(

balloob (approved):

PR #8593, opened 2019-02-15
🚜 Merges/Redirect Mysensor component pages

klaasnicolaas:
**Description:**
Merges components pages for Mysensor and add redirects for other component platform pages.

## Checklist:


MartinHjelmare:
I don't like this change. The component page is now huge and even harder than before to get a grasp on.

I think merging platform pages that don't have details with the component page was a good move. Merging detailed platform pages into the component isn't.

klaasnicolaas:
I agree with you, but you can easily click through to the correct paragraph which makes navigation a little easier.

If the rest also does not like it, then we just close it

klaasnicolaas:
Other (and mayby better) change is to adjust only the `ha_category` of all 9 pages. So not only `DIY` but more like:
```
ha_category:
  - DIY
  - Binary Sensor
```

frenck:
I do agree this one is becoming a large page, nevertheless, it will be one of the few / one of the exceptions, since most components do not contain such amount of information.

@klaasnicolaas I don't think @MartinHjelmare is referring to the component/platform index page here... but to the content length of this specific component.

MartinHjelmare:
Why can't we let platforms that need details have their own pages?

frenck:
In this case, we maybe should leave it @MartinHjelmare until we have more structure.

The main reasoning for this @MartinHjelmare is maintainability and usability. Originally, this change was induced from the hundreds of pages that refer to the component page. Others have a small section on the platform page, which would just as nice fit onto the component page.

Not only is the example above a burden on maintaining, but is also not very user-friendly. 

Having that many pages induce other issues as well. Jekyll has issue handling that many pages (especially with all the stuff going in plugins and templates in our case) and the flexibility to change/move our website has been reduced by this as well (for example, standardize and improve the general look and feel of the docs, or maybe even move away from Jekyll).

The question here is, how large may a page become?
Maybe we just hit the answer to that question...

[edit]
Build it locally; That indeed is a long page, mainly caused by the number of sketches inside the documentation. This is truly an exception in our documentation base (MQTT might be the other exception).

@klaasnicolaas I kinda agree with @MartinHjelmare to not merge this.

MartinHjelmare:
Yes, as I said above I think it's a good move to merge platform pages that don't have details. It could also be ok to move minor notes from platforms to component pages to be able to merge. But if platform pages need details we should let them have their own pages.

Where the line is drawn decision should be made between the documentation team and each respective component contributor or code owner.

I propose that we stop merging platforms that already have considerable details for now.

frenck:
> propose that we stop merging platforms that already have considerable details for now.

That is not a workable proposal since "considerable" is not quantifiable.

MartinHjelmare:
If we're not sure what considerable details is for a case, then we don't merge that until we've had a discussion with the code owner etc. That discussion can happen when the next PR for that page happens.

klaasnicolaas:
We can close this, I will make a better adjustment for this case.

PR #8594, opened 2019-02-15
Add additional config parameters for HomematicIP

SukramJ:
**Description:**

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#21076

## Checklist:

### DIFF
--- a/source/_components/homematicip_cloud.markdown
+++ b/source/_components/homematicip_cloud.markdown
@@ -65,6 +65,8 @@ Add the information to your `configuration.yaml` file:
 homematicip_cloud:
   - accesspoint: IDENTIFIER
     authtoken: AUTHTOKEN
+    enable_group_switches: true
+    enable_group_sec_sensors : false
   - name: Location2
     accesspoint: IDENTIFIER2
     authtoken: AUTHTOKEN2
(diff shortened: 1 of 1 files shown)


source/_components/homematicip_cloud.markdown
```diff
@@ -65,6 +65,8 @@ Add the information to your `configuration.yaml` file:
 homematicip_cloud:
   - accesspoint: IDENTIFIER
     authtoken: AUTHTOKEN
+    enable_group_switches: true
```
klaasnicolaas, line 68: This is an optional variable and should not be included in the first example yaml config.

source/_components/homematicip_cloud.markdown
```diff
@@ -65,6 +65,8 @@ Add the information to your `configuration.yaml` file:
...
+    enable_group_sec_sensors : false
```
klaasnicolaas, line 69: Same

klaasnicolaas (changes requested):

worm-ee:
Thanks for your contribution! 
I would just add the group devices, there is no need to add the config options. For hassio users it might be even confusing - the have only the options to remove and add the platform again.

SukramJ:
I removed the config options, so there is no need for the PR any more

PR #8608, opened 2019-02-16
Owlet Component Documentation

oblogic7:
**Description:**
Replaces #8217 since PyCharm keeps blowing up my branches during rebase or merge.

Add documentation for Owlet component

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#21108

## Checklist:


source/_components/owlet.markdown
```diff
@@ -0,0 +1,50 @@
...
+layout: page
+title: "Owlet"
+description: "Instructions on how to integrate Owlet baby monitor into Home Assistant."
+date: 2019-01-19
+sidebar: true
+comments: false
+sharing: true
+footer: true
+logo: owlet.svg
+ha_category: Health
```
MartinHjelmare: @klaasnicolaas what's the standard now for categories? Is it enough with one category also for components that have platforms or should we always add the categories that correspond to the platform modules, eg binary_sensor etc?

MartinHjelmare (commented):
We usually document somewhere in the page body what platforms are supported.

MartinHjelmare:
Here's the merged parent PR:
https://github.com/home-assistant/home-assistant/pull/21108

source/_components/owlet.markdown
fabaff: I think that platforms should be mentioned too.

source/_components/owlet.markdown
```diff
@@ -0,0 +1,59 @@
...
+ha_release: "0.89"
+ha_iot_class: "Cloud Polling"
+---
+
```
fabaff, line 18: It would be nice to have a sentence with some details about the integration.

klaasnicolaas (dismissed):

source/_components/owlet.markdown
klaasnicolaas: You will now mention both, the category and the platforms.

It is possible that if we have cleaned up a lot of nonsense pages this will still change, but then we can also better categorize and show everything to the users.

fabaff (approved):

PR #8618, opened 2019-02-18
Update traccar documentation with new parameters

lapy:
**Description:**
Added scan_interval and monitored_conditions

**Pull requests in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#21149
home-assistant/home-assistant#21079

## Checklist:

### DIFF
--- a/source/_components/device_tracker.traccar.markdown
+++ b/source/_components/device_tracker.traccar.markdown
@@ -70,3 +70,12 @@ monitored_conditions:
   required: false
   type: list
 {% endconfiguration %}
+
+The parameter `monitored_conditions` allows you to track non standard attributes from the traccar platform and use them in your Home Assistant. For example if you need to monitor the state of the non standard attribute `alarm` and a custom computed attribute `mycomputedattribute` just fill the configuration with:
+
+```yaml
+device_tracker:
+  - platform: traccar
+    ...
+    monitored_conditions: ['alarm', 'mycomputedattribute']
+```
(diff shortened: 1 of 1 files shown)


source/_components/device_tracker.traccar.markdown
```diff
@@ -60,4 +60,13 @@ verify_ssl:
...
\ No newline at end of file
+scan_interval:
+  description: Defines the polling interval.
+  required: false
+  type: time
+  default: 00:00:30
+monitored_conditions:
+  description: Additional traccar computed attributes or device-related attributes to include in the scan.
+  required: false
+  type: list
```
klaasnicolaas, line 66: Where are the keys?

lapy: I can't define keys, the attributes are custom defined in the traccar platform, they can be any name the user wants.

klaasnicolaas: I think we should include this information, normally you have a list that you can choose from. This can otherwise create confusion.

lapy: Where do you think this can be clarified? In the parameter description?

klaasnicolaas: My advice is to make a separate paragraph for it, with how the `monitored_conditions` are proportionate to what you set up in traccar.

klaasnicolaas (dismissed):
Much better with the clarification 😉

frenck:
I removed `scan_interval`. It is already documented at [Entity component platform options](https://www.home-assistant.io/docs/configuration/platform_options/#scan-interval). We decided not to add it because it's an advanced feature and in 95 % of the use cases not needed.

frenck (approved):

Issue #8619, opened 2019-02-18
Links broken in components pages.


eliseomartelli:
The links are broken on the components pages.

Some examples:
https://www.home-assistant.io/components/media_player.emby/
points at:
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/emby/media_player.py
but it should be:
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/media_player/emby.py

fabaff:
I think that @frenck knows the current state of the changes to make the links work again.

klaasnicolaas:
This is something that is being worked on, more people have opened that same type of issue.

klaasnicolaas:
let's use this issue for discussion, I notice that many people are going to open issues that are about the same. You can better respond to this issue.

peternijssen:
Just noticed the same thing in the following pages:
https://www.home-assistant.io/components/sensor.buienradar/
https://www.home-assistant.io/components/weather.buienradar/

hmmbob:
> Just noticed the same thing in the following pages:
> https://www.home-assistant.io/components/sensor.buienradar/
> https://www.home-assistant.io/components/weather.buienradar/

Lol, that's how I came into this issue - noticed the same while I was fixing a key error for the buienradar sensor

nikolak:
Note that also (all?) binary sensors have invalid links:

For example

https://www.home-assistant.io/components/binary_sensor.template/

https://www.home-assistant.io/components/binary_sensor.trend/

https://www.home-assistant.io/components/binary_sensor.bayesian/

https://www.home-assistant.io/components/binary_sensor.iss/

The docs structure points to e.g:

`/components/trend/binary_sensor.py`

While the real path is in [`/components/binary_sensor/`](https://github.com/home-assistant/home-assistant/tree/dev/homeassistant/components/binary_sensor)`filename.py`

In general there seems to be a lot of broken links, should I write a quick script to find them all, or is this something that's already known or otherwise unecessary?

Since this is being worked on I'm assuming that there is some work done, or are contribtions still welcome?

frenck:
## ⚠️ Please DO NOT add all the broken links you can find to this issue.

The links are generated and currently we know which links break and why.
The breakage is caused by [The Great Migration](https://developers.home-assistant.io/blog/2019/02/19/the-great-migration.html) in our code base.

I've added a couple of fixes to make them work in around 80% of the cases.

[...]

PR #8634, opened 2019-02-19
Update sensor.template.markdown

pdbogen:
I spend a bit of time troubleshooting why a template sensor wasn't updating, even though I had `sensor.time` listed in `entity_id`. It turns out that `sensor.time` didn't (yet) exist for me!

I've updated the documentation so that other folks don't get stuck on that.

As a side note, the source link on https://www.home-assistant.io/components/sensor.template/ is broken; it points to https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/template/sensor.py, which doesn't exist, but it should point to https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/sensor/template.py, which does; note `sensor/template.py` in the working URL. I'm not sure how this link is generated, since it doesn't appear in this documentation file.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/sensor.template.markdown
+++ b/source/_components/sensor.template.markdown
@@ -253,14 +253,11 @@ sensor:
 
 The `template` sensors are not limited to use attributes from other entities but can also work with [Home Assistant's template extensions](/docs/configuration/templating/#home-assistant-template-extensions).
 
-This template contains no entities that will trigger an update, so we add an `entity_id:` line with an entity that will force an update - here we're using a [date sensor](/components/sensor.time_date/) to get a daily update:
+This template contains no entities that will trigger an update, so we add an `entity_id:` line with an entity that will force an update - here we're using a [date sensor](/components/sensor.time_date/) (which may need to be registered) to get a daily update:
 
 {% raw %}
 ```yaml
 sensor:
-- platform: time_date
-  display_options:
-  - date
 - platform: template
   sensors:
     nonsmoker:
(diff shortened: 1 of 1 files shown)


fabaff:
> As a side note, the source link on https://www.home-assistant.io/components/sensor.template/ is broken

See #8619

source/_components/sensor.template.markdown
```diff
@@ -258,6 +258,9 @@ This template contains no entities that will trigger an update, so we add an `en
 {% raw %}
 ```yaml
 sensor:
+- platform: time_date
+  display_options:
+  - date
```
fabaff: There is no need to add the sensor as it's mentioned in the intro for the section that a `time_date` sensor is used. It it's not clear then the intro should be updated. 

None of the other sample is showing the source entity. Thus if one is not using the `sun` component the sample will not work.

pdbogen: Thanks for the feedback! I adjusted my proposal to be more in line with your commentary.

I do think _some_ mention is warranted. It would've at least saved me some time, and I'm inclined to believe that other folks would run into the same confusion.

frenck: I don't think that the latest change makes it clearer... "may need to"? So when in that case?
Also, I'm not sure if "registered" is the right term here.

frenck:
Since there was no response in a month, I'm declaring this PR stale and closing it.

PR #8665, opened 2019-02-21
Update history_graph.markdown

mrvanes:
refresh should be refresh_interval

**Description:**
I was bitten by this bug https://github.com/home-assistant/home-assistant-polymer/issues/1914 until I read this remark https://github.com/home-assistant/home-assistant-polymer/issues/1914#issuecomment-459232270

That helps a lot! Maybe next time HASS developers shouldn't be so harsh on users that try to comment/reopen unresolved issues?

### DIFF
--- a/source/_components/history_graph.markdown
+++ b/source/_components/history_graph.markdown
@@ -45,7 +45,7 @@ hours_to_show:
   required: false
   default: 24
   type: integer
-refresh:
+refresh_interval:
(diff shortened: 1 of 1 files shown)


frenck:
This change is incorrect. You are confusing Lovelace with the actual component here.

Component source reference: https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/history_graph/__init__.py#L18

As you can see in the above code link, it is actually `refresh` for the component side of things.

PR #8666, opened 2019-02-21
Default port for MariaDB 10 is 3306

jeffh0821:
**Description:**

## Checklist:

### DIFF
--- a/source/_components/recorder.markdown
+++ b/source/_components/recorder.markdown
@@ -150,7 +150,7 @@ Purging does not necessarily remove all entries before a given date. For example
 | MS SQL Server   | `mssql+pymssql://user:password@SERVER_IP/DB_NAME?charset=utf8` |
 
 <p class='note'>
-If you use MariaDB 10 you need to add port 3307 (or another port depending on which port is used by, for example: your hosting provider.) to the SERVER_IP, e.g., `mysql://user:password@SERVER_IP:3307/DB_NAME?charset=utf8`.
+If you use MariaDB 10 you need to add port 3306 (or another port depending on which port is used by, for example: your hosting provider.) to the SERVER_IP, e.g., `mysql://user:password@SERVER_IP:3306/DB_NAME?charset=utf8`.
 </p>
 
 <p class='note'>


klaasnicolaas:
Well this change is not true.

MariaDB 10 = 3307 and MariaDB 5 = 3306

jeffh0821:
The documentation says otherwise and the default port on the MariaDB 10.x official docker containers is 3306.
https://mariadb.com/kb/en/library/connecting-to-mariadb/

fabaff:
I don't think that they changed the port between the releases. The default port is 3306 as far as I know.

klaasnicolaas:
https://forum.synology.com/enu/viewtopic.php?t=90966&start=30

Today I have set the recorder component in my config and I used 3307 as port for my MariaDB on the synology NAS. I added that piece of text in the past because I noticed that a port number had to be added.

Edit:
Just tested port 3306 and I get the message that Home Assistant can no longer connect to the database, in my case I have to use port 3307 for connecting to my MariaDB 10 database.

jeffh0821:
Your reference article is specific to a parallel installation of MariaDB 5 and MariaDB 10 which would require independent ports. However, MariaDB documentation and docker images are all adhering to the standard MariaDB port of 3306. The HA page that I am attempting to correct is wrong and likely specific to your implementation only. It causes unneeded confusion and issues. Unless there is MariaDB official documentation stating that the default port in MariaDB changed from 3306 to 3307, the PR should be approved.

jeffh0821:
Or the PR should be rejected and the documentation should be corrected to clarify the default port MAY be overridden in certain-use cases. In either case, we SHOULD NOT be documenting that the MariaDB 10.x default port is 3307 as it clearly is not.

klaasnicolaas:
That is a better option, I think the 3306 is the default, but it may be that you have to use 3307 if it turns out that you still get a `can not connect to mysql database` error.

In any case, I had to approach my Synology NAS and my Hosting provider via 3307, 3306 results in an error.

I'll see what Synology does at MariaDB 5

klaasnicolaas:
MariaDB 5 and 10 side by side on the Synology
photo

jeffh0821:
I still don't think you understand that YOUR implementation is not typical and it is YOUR implementation of MariaDB that requires port 3307.

All MariaDB versions use port 3306 as default. Other ports MAY be used based on the requirements of the implementation. It is not relevant how it is setup on your Synology NAS.

I will close the PR and do a new one.

klaasnicolaas:
YES I know what you mean 😉 

And YES also the setup on a synology or third parties is important. It's fine that the port is changed to 3306 if that is the default. There should only be a note that the port can also be different, Before people will open an issue because the port for them also was not correct 😉

PR #8674, opened 2019-02-22
Fix example automation trigger

partytimeexcellent:
I requested a change earlier today that turned out to be incorrect. It solved the problem of the trigger not firing, but it caused the 'event_data' field to be ignored. I have updated this so not only does the trigger fire, but the event data works correctly as well.

**Description:**

## Checklist:


frenck:
This has been resolved in #8726 already.

Nevertheless, thanks for the PR! 👍

PR #8716, opened 2019-02-24
Update tellstick to match with new sensor configuration

endor-force:
**Description:**
Sensor configuration will now use static identifiers instead of ID number to set the name of named sensors.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#21402

## Checklist:

### DIFF
--- a/source/_components/tellstick.markdown
+++ b/source/_components/tellstick.markdown
@@ -108,7 +108,7 @@ sensor:
   required: false
   type: string
 only_named:
-  description: Only show the named sensors. Add sensors to only include specified sensors. If this is not specified the all sensors will be imported and the names will be based on each sensors ID number.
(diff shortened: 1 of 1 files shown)


source/_components/tellstick.markdown
```diff
@@ -108,10 +108,19 @@ sensor:
   required: false
   type: string
 only_named:
-  description: Only show the named sensors. Set to `true` to hide sensors.
+  description: Only show the named sensors. Add sensors to only include specified sensors. If this is not specified the all sensors will be imported and the names will be based on each sensors ID number.
```
MartinHjelmare:
```suggestion
  description: Only add and include specified sensors. If this is not specified all sensors will be imported and the names will be based on each sensor's ID number.
```

MartinHjelmare (changes requested):

MartinHjelmare (approved):

PR #8750, opened 2019-02-26
Update docs for the new iperf3 component

rohankapoorcom:
**Description:**
Iperf3 was split into a component and a sensor.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#21138

## Checklist:

### DIFF
--- a/source/_components/iperf3.markdown
+++ b/source/_components/iperf3.markdown
@@ -8,17 +8,23 @@ comments: false
 sharing: true
 footer: true
 logo: iperf3.png
-ha_category: System Monitor
+ha_category:
+  - System Monitor
+  - Sensor
 featured: false
 ha_release: 0.71
 ha_iot_class: "Local Polling"
+redirect_from:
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
Please do not add/document `scan_interval`. It is already documented at [Entity component platform options](https://www.home-assistant.io/docs/configuration/platform_options/#scan-interval). We decided not to add it to every component, because it's an advanced feature and in 95 % of the use cases not needed.

rohankapoorcom:
@frenck, this is needed on the component level, not the entity level.

Setting it in an entity will do nothing since the entity is not polled. Setting it in the component will actually change the frequency at which speed tests are run. Since most components do not support scan_interval I think it's important to mention here.

frenck (approved):
In that case, this will do!

Thanks, @rohankapoorcom! 👍

PR #8773, opened 2019-02-28
Added description for source priority.

gorynychzmey:
**Description:**

Added description for source priority to Person (with examlpe).

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#21479

## Checklist:

### DIFF
--- a/source/_components/person.markdown
+++ b/source/_components/person.markdown
@@ -13,7 +13,11 @@ ha_qa_scale: internal
 ha_release: 0.88
 ---
 
-The person component allows to connect device tracker entities to one or more person entities. The last state update of a connected device tracker will set the state of the person. For example if you connect your router and your OwnTracks device as trackers to your person, the last state update from either the router or your OwnTracks device will set the state of your person.
+The person component allows to connect device tracker entities to one or more person entities. The state updates of a connected device trackers will set the state of the person. When multiple device tracers used, the state of berson will be determined next way:
+
+1. If there are sources presenting status 'home', than latest of this sources will be taken.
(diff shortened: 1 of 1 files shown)


fabaff (approved):

source/_components/person.markdown
```diff
@@ -13,7 +13,11 @@ ha_qa_scale: internal
 ha_release: 0.88
 ---
 
-The person component allows to connect device tracker entities to one or more person entities. The last state update of a connected device tracker will set the state of the person. For example if you connect your router and your OwnTracks device as trackers to your person, the last state update from either the router or your OwnTracks device will set the state of your person.
+The person component allows to connect device tracker entities to one or more person entities. The state updates of a connected device trackers will set the state of the person. When multiple device tracers used, the state of berson will be determined next way:
+
+1. If there are sources presenting status 'home', than latest of this sources will be taken.
```
MartinHjelmare, line 18: This isn't correct totally correct, since gps trackers are excluded from this group. This needs clarification. Please also add examples.

gorynychzmey: @MartinHjelmare, I've corrected description in documentation, examples will be later. But I still haven't decided, should I really exclude GPS sources from 'home' sources or not. Seems to be that nothing changes for status, only difference is, which device will be presented as source.

MartinHjelmare: Let's update the docs so it's correct according to the current source code first.

Then please open an architecture issue where we discuss the details of the implementation.

Thanks for responding quickly!

PR #8780, opened 2019-02-28
Add Home Assistant API example to RESTful command description

ronschaeffer:
I cannot find this type of example documented anywhere else in the manual, and the few examples on the forum are incorrect.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/rest_command.markdown
+++ b/source/_components/rest_command.markdown
@@ -120,3 +120,15 @@ automation:
         status: "At Work"
         emoji: ":calendar:"
 ```
+You can also use REST commands to access the Home Assistant API without using shell commands. This example uses a long-lived access token to access the Home Assistant API and set the state of a light to on.
+
+```yaml
+set_study_light_status_to_on:
+  url: http://localhost:8123/api/states/light.study_light
+  method: POST
+  headers: 
+    authorization: 'Bearer YOUR_LONG_LIVED_ACCESS_TOKEN'
+    content-type: 'application/json'
+  payload: '{"state":"on"}'
+```
+Replace YOUR_LONG_LIVED_ACCESS_TOKEN with your token. You can obtain a token by logging into the frontend using a web browser and going to your profile http://IP_ADDRESS:8123/profile.


frenck:
Hi @ronschaeffer, 

Thanks for the PR. Nevertheless, I'm wondering why one would ever do this / use this method?
This is in Home Assistant right? You'd just change states of things directly?

../Frenck

ronschaeffer:
Hi @frenck -
Have I overlooked a simpler way of achieving my goal? I want to set the state of an entity, for example to on from off, without also executing the associated action.

The use case is to set a light with an assumed state to on or off when it is controlled from the wall switch instead of through HA. I use other methods to track the state of the light through a binary sensor, and use the binary sensor state change as a trigger for an automation to change the state of the light in the frontend to match its actual state. (Hope that sentence makes sense!)

According to manual https://www.home-assistant.io/docs/configuration/state_object/, "if you overwrite a state via the states dev tool or the API, it will not impact the actual device." As far as I can find, the manual does not mention an alternative method within HA of changing just the state without executing the associated action.

This use of RESTful commands could be applied to any object with an assumed state, especially those for which you can use other methods to know the actual state.

arsaboo:
@ronschaeffer Isn't using the REST command the same as executing the action? I agree with @frenck that this may not be the right use case.  

Also, other than using the long_lived_token, the REST command is like any other REST command. Thus, I am not sure if we need it.

frenck:
Thanks for the PR and detailed explanation @ronschaeffer, nevertheless, we are going to friendly decline it for the earlier stated reasons.

Again, thanks, since opening a PR is really appreciated 👍

ronschaeffer:
@arsaboo In this case, it's not the same as performing the action. It's about setting the state without performing the action. Unless I've missed it, a rest command or a shell command with curl are the only ways to automate/script that within HA. There are a few threads on the forum where people are looking for ways to do that. The main point here, though, was to show how to use the long-lived access token with the rest command. I couldn't find it documented anywhere else. Maybe there is a better location in the manual?

fabaff:
The [REST API](https://developers.home-assistant.io/docs/en/external_api_rest.html) documentation should describe the usage.

ronschaeffer:
OK @fabaff . Raised a PR over there https://github.com/home-assistant/developers.home-assistant/pull/194. Can close that one if you guys think this is too much of an edge use case, but it would have saved me a world of trial and error pain following incorrect or outdated advice in the forum.

PR #8806, opened 2019-03-03
Add 'apps' config entry to Fire TV

JeffLIrion:
**Description:**

Allow users to provide friendly names for apps.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** https://github.com/home-assistant/home-assistant/pull/21601

## Checklist:

### DIFF
--- a/source/_components/media_player.firetv.markdown
+++ b/source/_components/media_player.firetv.markdown
@@ -80,7 +80,7 @@ get_sources:
   default: true
   type: boolean
 apps:
-  description: A dictionary where the keys are app IDs and the values are app names.
+  description: A dictionary where the keys are app IDs and the values are app names that will be displayed in the UI; see example above.
   required: false
   default: {}
   type: dict
--- a/source/_components/media_player.firetv.markdown
+++ b/source/_components/media_player.firetv.markdown
@@ -26,23 +26,27 @@ Steps to configure your Amazon Fire TV stick with Home Assistant:
   - From the main (Launcher) screen, select Settings.
   - Select System > About > Network.
 
-To add Fire TV to your installation, add the following to your `configuration.yaml` file:
+
+## {% linkable_title Configuration %}
 
 ```yaml
 # Example configuration.yaml entry
 media_player:
-  # use the Python ADB implementation
+  # Use the Python ADB implementation without authentication
   - platform: firetv
     name: Fire TV 1
     host: 192.168.0.111
-    adbkey: "/config/android/adbkey"
-    apps:
-      com.amazon.tv.launcher: "Fire TV"
 
-  # use an ADB server for sending ADB commands
+  # Use the Python ADB implementation with authentication
   - platform: firetv
     name: Fire TV 2
     host: 192.168.0.222
+    adbkey: "/config/android/adbkey"
+
+  # Use an ADB server for sending ADB commands
+  - platform: firetv
+    name: Fire TV 3
+    host: 192.168.0.123
     adb_server_ip: 127.0.0.1
 ```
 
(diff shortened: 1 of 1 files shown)


source/_components/media_player.firetv.markdown
```diff
@@ -26,26 +26,23 @@ Steps to configure your Amazon Fire TV stick with Home Assistant:
...
 ```yaml
 media_player:
-  # a device that does not require ADB authentication
+  # use the Python ADB implementation
   - platform: firetv
     name: Fire TV 1
     host: 192.168.0.111
+    adbkey: "/config/android/adbkey"
+    apps:
```
rohankapoorcom: For simplicity, we don't show the optional parameters in the configuration example.

source/_components/media_player.firetv.markdown
```diff
@@ -82,6 +79,11 @@ get_sources:
   required: false
   default: true
   type: boolean
+apps:
+  description: A dictionary where the keys are app IDs and the values are app names.
```
rohankapoorcom: It's unclear what the difference between an app id and name is.

rohankapoorcom (changes requested):

source/_components/media_player.firetv.markdown
JeffLIrion: Should I change it back to having 3 example configurations? I removed the first configuration with only required parameters because all Fire TV devices require authentication now, so that configuration wouldn't work. Either `adbkey` (use Python ADB implementation) or `adb_server_ip` (use an ADB server) needs to be provided.

source/_components/media_player.firetv.markdown
JeffLIrion: I agree. Is it sufficient to reference the example above? I think `com.amazon.tv.launcher: "Fire TV"` makes it clear. 

```
description: A dictionary where the keys are app IDs and the values are app names that will be displayed in the UI; see example above.
```

JeffLIrion: I changed this to what I posted above. If that's not what you had in mind, let me know and I'll fix it.

source/_components/media_player.firetv.markdown
rohankapoorcom: Has the component been updated to reflect the fact that either `adbkey`/`adb_server_ip` is required? If it has not, we should make the schema change there.

Regardless, we shouldn't have `apps` in the first example. After the configuration variables section, you can add a full configuration. Here's an [example page](https://www.home-assistant.io/components/zoneminder/) that's structured like that.

JeffLIrion: I suppose I made too strong of a statement above: I can't say for sure that every single Fire TV device out there requires authentication. It's possible that someone out there has blocked their Fire TV stick from accessing the internet and so it's running an old version of the OS, and requiring one of those two config entries would break their setup. 

What do you think? Either way, I can provide a basic configuration first and then one or two full configurations.

rohankapoorcom: In that case, I would say that we should then show all three ways of connecting (unauthenticated, plus the two different ways to authenticate) and then have a full example at the bottom with the `apps` section.

JeffLIrion: Ok, I'll make those changes. 

In the meantime, could you please merge this pull request and I'll rebase this one on top of it: 
https://github.com/home-assistant/home-assistant.io/pull/8722

rohankapoorcom: Done.

rohankapoorcom (approved):

rohankapoorcom:
This looks good! Thanks for the quick turnaround!

JeffLIrion:
Thanks @rohankapoorcom!

PR #8831, opened 2019-03-05
Add ClickSend "caller" option

fabtesta:
**Description:**
Added optional "caller" parameter to let ClickSend API call be more compliant with some cellular networks that do not allow incoming calls from the same recipient number.
If "caller" is not defined, then the recipient parameter value is used.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant):** home-assistant/home-assistant#20780

## Checklist:

### DIFF
--- a/source/_components/notify.clicksend_tts.markdown
+++ b/source/_components/notify.clicksend_tts.markdown
@@ -26,6 +26,7 @@ notify:
     username: CLICKSEND_USERNAME
     api_key: CLICKSEND_API_KEY
     recipient: PHONE_NO
+    caller: PHONE_NO
 ```
 
 {% configuration %}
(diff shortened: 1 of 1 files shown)


source/_components/notify.clicksend_tts.markdown
```diff
@@ -26,6 +26,7 @@ notify:
     username: CLICKSEND_USERNAME
     api_key: CLICKSEND_API_KEY
     recipient: PHONE_NO
+    caller: PHONE_NO
```
klaasnicolaas: You can remove this line, also the name variable. 
Only required items in the first example.

klaasnicolaas (changes requested):

klaasnicolaas (approved):
✅ Can be merged when deploy is ready

frenck (approved):
Thanks, @fabtesta! 👍

PR #8833, opened 2019-03-05
port number typo on sesame sample URL

mrsnyds:
The current version of the sample URL in the sesame section shows the port as 8123, like so ...

https://hassio.yourdomain.com:8123/somesecretnobodycanguess

..but I believe it should be port 3218, like so ...

https://hassio.yourdomain.com:3218/somesecretnobodycanguess

**Description:**

## Checklist:


frenck (approved):
Thanks, @mrsnyds! That is indeed correct 👍

PR #8880, opened 2019-03-09
Update detecting lost connections sample

mhaack:
**Description:**
Old detecting lost connections using sensors does not work with current HA versions any more.

## Checklist:


danielperna84 (changes requested):
Some spelling:
> The important part is the `sensor.time` entiy (from time_date component). This will update the binary sensor on every change of the sensor and every minute. If the Homematic senso does not send any updates any more the `sensor.time` will set the binary sensor to off 10 minutes after the last sensor update. This will trigger the automation.

1. entiy -> entity
2. senso ->sensor
3. any more -> anymore
4. Add a `,` after anymore
5. off -> `off`

Also the usage of template-code in your example automation seems to break the framework that processes the document. I don't know what the correct solution for this might be. Probably some escaping that allows to include raw code.

mhaack:
@danielperna84 you can check again, addressed your comment

danielperna84 (approved):

PR #8891, opened 2019-03-10
Components page improvements

robbiet480:
**Description:**

- Add this nice new dropdown to allow seeing specific version new components. Required a new filter: `group_components_by_release`, [the output of which looks like this.](https://gist.github.com/robbiet480/6c8ef4a3d06574f2a9286958e3aba508)

image

- Remove totally invalid or dead component pages (broke the selector otherwise)
- Ensure that _every_ component page has a `ha_release` set. I wrote a handy little script to get the earliest tag version out of Git (see below) and another little script to write the version into the YAML.
- Simplify the logic for the quick pick links using work done for the select. **BONUS**: Shaved off 10 seconds of compile time just by doing this 🎉!
- If Jekyll experiences an issue during preview, output the trace.

My script to gather earliest Git tags:

```bash
#!/bin/bash

declare -a files=("homeassistant/components/air_quality/__init__.py" "homeassistant/components/alarm_control_panel/__init__.py")

for i in "${files[@]}"
do
   echo "$i - $(git describe --always --contains $(git log --follow --format="%H" -- "$i" | tail -n 1) | sed 's/~.*//')"
done
```

My script to write the newly found versions into the front matter (`npm install js-yaml`):

```nodejs
const fs = require('fs');
const yaml = require('js-yaml');

var files = {"source/_components/zwave.markdown": "0.7"}

Object.keys(files).forEach(function(filePath) {
  var introduced = files[filePath];
  fs.readFile(filePath, 'utf8', function(err, contents) {
    if(err) { return console.log(err); }

    if(introduced !== 'pre 0.7' && introduced.split('.').length-1 == 1 && introduced.substr(-1) !== '0') {
      introduced = Number(introduced)
    }

    var chunks = contents.split('\n---\n');
    
    chunks[0] = chunks[0] + '\nha_release: '+introduced;
    
    var allContents = chunks.join('\n---\n')

    fs.writeFile(filePath, allContents, function(err) {
      if(err) { return console.log(err); }

      console.log(filePath, "saved");
    });
  });
});
```


frenck (dismissed):
👍 ❤️ 🏅 💯 🍪 💛 🏆

frenck (approved):

PR #8897, opened 2019-03-10
Fix some careless frontmatter mistakes

robbiet480:
Fix a bunch of careless mistakes in front matter, including by not limiting to:

- Unifying `ha_iot_class` into the following:
```
Assumed State
Cloud Polling
Cloud Polling and Cloud Push
Cloud Push
Configurable
Local Polling
Local Push
Local Push and Cloud Polling
```
- Removing quotes where unnecessary
- Removing `featured: false`
- Removing duplicated keys
- Fix key misspellings (like `redirect_form` and `redirect_frome`)
- Fix incorrect keys (like `ha_version` instead of `ha_release`


source/_components/light.tikteck.markdown
```diff
@@ -8,8 +8,8 @@ comments: false
 sharing: true
 footer: true
 ha_category: Light
-ha_iot_class: "Local Polling"
-ha_logo: tikteck.png
```
frenck, line 12: Is this correct?

robbiet480: `ha_logo` is invalid, `ha_iot_class` is a duplicate key.

frenck (approved):

PR #8900, opened 2019-03-10
Additional clarification

arretx:
I was a bit confused about what the base url was supposed to be here and that setting it here was overriding the HTTP setting in configuration.yaml.  Often I see in the documentation that there are examples without explanation of why that was used as the example.  So for this one, I assumed that it was `asking` for the base_url that I had set, rather than allowing me to define a custom base url.

**Description:**
Helping other understand that the URL is being customized here, not requested.

## Checklist:

### DIFF
--- a/source/_components/tts.markdown
+++ b/source/_components/tts.markdown
@@ -62,6 +62,9 @@ tts:
     time_memory: 300
     base_url: http://192.168.0.10:8123
 ```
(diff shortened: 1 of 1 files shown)


frenck (approved):

frenck:
Thanks, @arretx 👍

PR #8945, opened 2019-03-15
Update PS4 Docs

ktnrg45:
**Description:**
Added additional config info.

## Checklist:

### DIFF
--- a/source/_components/ps4.markdown
+++ b/source/_components/ps4.markdown
@@ -63,7 +63,9 @@ Home Assistant installed on a Debian-type OS may require configuration. This sec
 
 In terminal run the following command:
 
-`sudo setcap 'cap_net_bind_service=+ep' <python>`
+```bash
+sudo setcap cap_net_bind_service=+ep <python>
+```
 
 Replace `<python>` with your **system path** to Python that is running Home Assistant and/or your virtual environment if used. The path **should not** be a **symlink** or be **inside of a virtual environment**.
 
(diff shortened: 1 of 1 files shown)


source/_components/ps4.markdown
```diff
@@ -2,7 +2,7 @@
 layout: page
 title: "Sony PlayStation 4"
 description: "Instructions on how to integrate a Sony PlayStation 4 into Home Assistant."
-date: 2019-02-12 01:08
+date: 2019-03-15 01:08
```
klaasnicolaas: Please do not change this

source/_components/ps4.markdown
```diff
@@ -53,9 +52,40 @@ There are varying methods to perform this, dependent on your OS that is running
...
+
+- Debian
+- Hassbian
+- Rassbian
+- Armbian
+- Ubuntu
+
+In terminal run the following command:
+
+`$ sudo setcap 'cap_net_bind_service=+ep' <python>`
```
klaasnicolaas: Please use [this kind](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) of format for code in markdown.

klaasnicolaas (changes requested):

source/_components/ps4.markdown
ktnrg45: Ok. Thanks. Reverted.

source/_components/ps4.markdown
ktnrg45: Removed $ prefix from code

klaasnicolaas: I think you must do more then that, I will change somthing as example

ktnrg45: How about now?

PR #8950, opened 2019-03-15
Telegram_bot: Add option to allow fetching data from unverified SSL e…

nbarrientos:
**Description:**

This patch documents a new option (`verify_ssl`) that can be passed as part of the payload when calling certain services of the Telegram platform of the Notify component.

The variable is allowed to be unset, which means that nothing will be passed to `python-requests` relying therefore on the default behaviour of the library (currently it's to verify). From the documentation's POV the most correct description of the default behaviour would be then "python-requests' default" but I'm not sure that's very appropriate for user documentation. Your call.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#22069

## Checklist:

### DIFF
--- a/source/_components/notify.telegram.markdown
+++ b/source/_components/notify.telegram.markdown
@@ -186,6 +186,11 @@ authentication:
   required: false
   default: basic
   type: string
+verify_ssl:
+  description: Set to false to skip the validation of the server's SSL certificate.
+  required: false
+  default: true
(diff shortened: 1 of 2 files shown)


cgtobi (approved):
LGTM

rohankapoorcom (approved):

PR #8968, opened 2019-03-17
WIP add an overview diagram to installation.md

yoflippo:
## Description:
This is a proposal to include some kind of overview diagram which illustrates the distinction between the main HA projects. When I started looking into the development side of the HA I was confused. This is an attempt to mitigate possible initial confusion. Of course, the [main GitHub project page](https://github.com/home-assistant) gives an overview. But IMO at first this is still confusing and does not provide clear associations.

Below I have summarized some suggestions and relevant questions for further development. Of course, if this does not help in any way, than remarks in that direction are welcome and this PR can be closed.

## Checklist:

## Image source:

- [This is the concept overview diagram](https://pasteboard.co/I5PldKr.png)
- [The source draw.io image](https://github.com/yoflippo/drawio_drawings/tree/master/IN4315_Software_Architecture/HA_Main_Project_Overview)

## Suggestions and questions:

- Browsing through the documentation the [installation](https://www.home-assistant.io/docs/installation/) looked most approriate. But is it? Should we introduce it in another place?
- In it's initial state this addition only covers the projects which are for a beginning developer possibly confusing. Maybe we should recreate an overview of all main GitHub projects (i.e. home-assistant, home-assistant.io, home-assistant-polymer, hassio, hassos)?
- Should hassbian be added to the overview in a more explicit manner?


fabaff:
> Browsing through the documentation the installation looked most approriate. But is it? Should we introduce it in another place?

There is the [user documentation](https://www.home-assistant.io/) and the [developer documentation](https://developers.home-assistant.io/en/). We shouldn't start to mix things and the developer documentation is already covering the core parts (incl. backend, frontent, HassIO and HassIO add-ons).

> In it's initial state this addition only covers the projects which are for a beginning developer possibly confusing. Maybe we should recreate an overview of all main GitHub projects (i.e. home-assistant, home-assistant.io, home-assistant-polymer, hassio, hassos)?

I don't see how an overview could help here as most parts are describing what they are doing. There are over 50 repos under the HA orga. Some are relevant for the available installers and some are just helpers.

> Should hassbian be added to the overview in a more explicit manner?

If one decides to start contributing to Hassbian then it the starting point should be the Hassbian repo.

yoflippo:
@fabaff thanks for your reply. You are right, those should not be mixed. I guess regular end-users will not look at the developer manual. So if some kind of overview figure should be added then it should be placed in the developers documentation.

The example image I made, would have helped me, when I started to make clear distinctions between the _main_ projects. If some kind of overview does not help in anyway, or I am the only one with the problem, then we should close the PR.

frenck:
IMHO, this graph does not distinguish anything.

Even worse, it can be pretty easly be interpreded. For example the HassOS part... it shows like this is not Hass.io? HassOS **runs** Hass.io...

I'm also not in favor of adding this.

Since I'm the second reviewer, I'll go ahead and close this PR.

PR #8974, opened 2019-03-18
Updated docs for Google Hangout hub component

teliov:
Added information on how to authenticate using the authorization code, and also how to obtain the authorization code

**Description:**
This PR updates the documentation with instructions on obtaining an authorization code from Google for authentication and how to use this code to complete the authentication flow for the hangout component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#22158

## Checklist:

### DIFF
--- a/source/_components/hangouts.markdown
+++ b/source/_components/hangouts.markdown
@@ -41,22 +41,16 @@ Once the code is obtained fill in the form with your email, password and the aut
 
 ### {% linkable_title Steps to obtain Authorization Code %}:
 
-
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
There is a lot of unneeded HTML in here (e.g., the `<br>` linebreaks are not needed in Markdown).
Please remove those.

Also, there is a merge conflict which needs to be resolved.

@teliov Could you please take a look? 👍

teliov:
@frenck, maybe I'm doing something wrong but this is how the preview looks without the <br> tags.
It was why I added the tags in the first place. My guess is that the fact the content is in the  `<p class='note'></p>` section is causing markdown to not behave properly.

Any pointers would be appreciated!

Screen Shot 2019-04-10 at 6 08 09 PM

klaasnicolaas:
I think it is better to put this not in a note 😉

teliov:
@klaasnicolaas , good point. Was trying to preserve the original format but yeah the instructions do not need to be in a note

PR #9000, opened 2019-03-21
Consolidate the netgear_lte configuration

amelchio:
**Description:**

This consolidates the configuration under the `netgear_lte` component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#22105

## Checklist:

### DIFF
--- a/source/_components/netgear_lte.markdown
+++ b/source/_components/netgear_lte.markdown
@@ -19,7 +19,7 @@ redirect_from:
   - /components/sensor.netgear_lte/
 ---
 
-The Netgear LTE integration for Home Assistant allows you to observe and control [Netgear LTE modems](https://www.netgear.com/home/products/mobile-broadband/lte-modems/default.aspx), currently only tested with LB2120.
+The Netgear LTE integration for Home Assistant allows you to observe and control [Netgear LTE modems](https://www.netgear.com/home/products/mobile-broadband/lte-modems/default.aspx).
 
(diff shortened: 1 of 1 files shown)


source/_components/netgear_lte.markdown
```diff
@@ -52,63 +60,44 @@ password:
...
+      type: string
+sensor:
+  description: Configuration options for sensors.
   required: false
-  type: string
+  type: map
+  keys:
+    monitored_conditions:
+      description: Sensor types to create.
+      required: true
```
MartinHjelmare: false

MartinHjelmare: Add `default: usage`.

MartinHjelmare (changes requested):

source/_components/netgear_lte.markdown
amelchio: Hmm ... I would like the default to be underspecified but maybe that is not something we do?

MartinHjelmare: Why? Nah, we should be accurate in the docs.

amelchio: To discourage users from depending on it. I wanted the default to just hint that sensors are available and then users could specify the ones they actually want.

I am fine with listing the default though.

MartinHjelmare (approved):

PR #9008, opened 2019-03-23
Clarifies expose_by_default and exposed_domains in google_assistant config

dthulke:
**Description:**

Clarifies the documentation on the `expose_by_default` and `exposed_domains` config fields in the Google Assistant config.
These have caused confusion among users in the past (see https://github.com/home-assistant/home-assistant/issues/21577).

Ideally, the behaviour would be fixed in the code but this would break existing configs at least in some edge cases (see https://github.com/home-assistant/home-assistant/pull/17745).

The proposed new description should explain the current behaviour of these options.

## Checklist:

### DIFF
--- a/source/_components/google_assistant.markdown
+++ b/source/_components/google_assistant.markdown
@@ -139,12 +139,12 @@ api_key:
   required: false
   type: string
 expose_by_default:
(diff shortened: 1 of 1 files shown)


frenck (approved):

frenck:
Thanks, @dthulke! 👍

PR #9029, opened 2019-03-25
Added example to mobile app component

havardelnan:
There should be an example even if its a one liner.


frenck (approved):
Thanks, @havardelnan 👍

PR #9078, opened 2019-03-30
Updated docs for support of multiple elk devices.

gjbadros:
**Description:**
I just submitted the PR for the multiple elk devices: change for support of multiple elk devices.\

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#22573

## Checklist:

### DIFF
--- a/source/_components/elkm1.markdown
+++ b/source/_components/elkm1.markdown
@@ -48,246 +48,256 @@ section to your `configuration.yaml` file:
 ```yaml
 # Example configuration.yaml entry
 elkm1:
-  host: elk://IP_ADDRESS
+  devices:
+  - host: elk://IP_ADDRESS1
+  ....
+  - host: elk://IP_ADDRESS2
 ```
 
 {% configuration %}
-host:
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
Thanks for your PR, @gjbadros!

Nevertheless, since this PR depends on a parent PR in our code base, it should target the `next` branch (since it depends on a next/future release).

gjbadros:
Got it; I clicked that branch "next" option -- I didn't want to assert this has to be accepted into the next version, but now I think I understand that that just reflects that there's a dependency on a PR before the doc change should go live.

gjbadros:
Docs, gurus: is there a way to structurally document that the configuration fields can be repeated for multiple hosts?  I inserted a devices: element so it could clearly contain a list of -host and other configuration fields, but that results in a breaking change to users who only have a single elk m1 controller.  I'd rather leave "devices" out, but then I don't see how to make clear that the subsequent configuration can be repeated multiple times. 

Thoughts?

klaasnicolaas:
I close this PR because the parent PR is also closed

PR #9150, opened 2019-04-04
improve Configuration for Credentials

mrsnyds:
Hi there.  I found this page only because of the log warning that the lambda notify component would be deprecated, so I suspect that means it is new.  I was able to get the new component working, but a edits would have helped a bit. 

1.)  The current doc lists `region_name`as required, so I configured it, but then received and invalid option error when checking the configuration.yaml in dev tools ... so I rather think it does not belong and am suggesting it be removed.  

2.) I was unclear how the credentials section and notify sections should work together, but I think the idea is to list credentials once, and then each notifier underneath it will refer to it ... yes?  I am not sure what edits to suggest, but I did notice that the comment in the example, `# use the first credential defined in aws component by default` did not hold true in my case.  I had to explicitly define credentials inside my notifier.  So maybe the current state of the component and the doc' are not in synch yet?

**Description:**

## Checklist:

### DIFF
--- a/source/_components/aws.markdown
+++ b/source/_components/aws.markdown
@@ -59,11 +59,6 @@ profile_name:
   description: A credentials profile name.
   required: false
   type: string
-region_name:
-  description: The region identifier to connect to.
-  required: true
-  default: "`us-east-1`"
-  type: string
 name:
   description: Setting the optional parameter `name` allows multiple notifiers to be created. The notifier will bind to the service `notify.NOTIFIER_NAME`.
   required: false


klaasnicolaas:
I think it is better to set `region_name` as optional, instead of deleting it. In fact, this variable does occur in the python code.

mrsnyds:
> I think it is better to set `region_name` as optional, instead of deleting it. In fact, this variable does occur in the python code.

I have not tried looking at the code, as I am not a developer.   But I did observe that the `region_name` was listed as _Required_ within both the "CONFIGURATION FOR CREDENTIALS"  and the "CONFIGURATION FOR NOTIFY"  sections of the doc'.   I believe it cannot be _Required_ in both places , because when I tried to add it to both sections in the configuration.yaml, the Check Config in dev tools gave an error stating that `region_name `was an invalid option.   When I removed it from the CREDENTIALS section, then the component and notifier worked as expected.

My apologies if I have misunderstood something, but I think there is something in the doc' that could be improved to help someone like me so that we don't have to do trial and error.   If it really is _Optional_, someone that understands the code could maybe suggest an edit to indicate where and how it is optional.

awarecan:
This is a simple copy/paste error, `region_name` should be removed.

awarecan (approved):

PR #9159, opened 2019-04-06
add homekit battery config

adrum:
**Description:**

Adds the ability to link an accessory to pull its battery level from a separate entity. I have many Z-Wave devices with template sensors for their main states and separate template sensors for their battery levels. This would allow users to define where to pull their battery level from, in case the `battery_level` attribute is absent on the main entity for the accessory.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#22788

## Checklist:

### DIFF
--- a/source/_components/homekit.markdown
+++ b/source/_components/homekit.markdown
@@ -114,8 +114,7 @@ homekit:
                 required: false
                 type: string
               battery:
-                description: The `entity_id` of the battery entity for the accessory. The entity must
-                have the device_class of `battery` for it to function properly.
+                description: The `entity_id` of the battery entity for the accessory. The entity must have the device_class of `battery` for it to function properly.
                 required: false
                 type: string
                 default: '`<No entity_id>`'
--- a/source/_components/homekit.markdown
+++ b/source/_components/homekit.markdown
@@ -31,6 +31,8 @@ homekit:
       - alarm_control_panel
       - light
       - media_player
+    include_entities:
+      - binary_sensor.living_room_motion
   entity_config:
     alarm_control_panel.home:
       code: 1234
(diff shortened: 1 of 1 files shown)


source/_components/homekit.markdown
```diff
@@ -113,6 +113,11 @@ homekit:
                 description: Name of the entity to show in HomeKit. HomeKit will cache the name on the first run so a device must be removed and then re-added for any change to take effect.
                 required: false
                 type: string
+              linked_battery_sensor:
+                description: The `entity_id` of the battery entity for the accessory.
```
cdce8p: This should mention that it must be a `sensor` entity.

cdce8p: As for your comment: https://github.com/home-assistant/home-assistant/pull/22788#issuecomment-481055199
I think the `description` of the `name` attribute above could fit quite well (slightly altered of course).

Nevertheless, what do you think about adding a new topic inside the troubleshooting section. Something like `The linked battery sensor isn't recognized`?

source/_components/homekit.markdown
```diff
@@ -113,6 +113,11 @@ homekit:
...
+                required: false
+                type: string
+                default: '`<No entity_id>`'
```
cdce8p: Just remove the `default`

cdce8p (changes requested):
As mentioned https://github.com/home-assistant/home-assistant/pull/22788#issuecomment-481442060
Maybe we should also integrate this into the example config at the top. What do you think?

source/_components/homekit.markdown
adrum: I put something together for this in the troubleshooting section, as well as mentioning it in the description. Let me know what you think.

cdce8p:
@adrum I like it. Just pushed some last changes, but I think this is ready. What do you think?

adrum:
@cdce8p Looks good to me 👍

klaasnicolaas (approved):

cdce8p (approved):

cdce8p:
@adrum Thanks for the good work you've done 👍 I'm looking forward to your next HomeKit PR.

PR #9176, opened 2019-04-09
Update docs to reflect multiple devices support in opentherm_gw

mvn23:
**Description:**
The opentherm_gw component is being rewritten to support more than one OpenTherm Gateway.
This PR contains updated docs for the rewrite of the component.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#22932

## Checklist:

### DIFF
--- a/source/_components/opentherm_gw.markdown
+++ b/source/_components/opentherm_gw.markdown
@@ -23,13 +23,7 @@ redirect_from:
 
 The `opentherm_gw` component is used to control the [OpenTherm Gateway](http://otgw.tclcode.com/) from Home Assistant.
 
-The following device types are currently supported within Home Assistant:
-
-- Binary Sensor
-- Climate
-- Sensor
-
-This component will automatically add a `climate` entity to Home Assistant for each configured gateway.
+This component will add a single `climate` entity and multiple `sensor` and `binary_sensor` entities to Home Assistant for each configured gateway.
 
 <p class='note'>
 The OpenTherm protocol is based on polling. The thermostat sends requests to the boiler at specific intervals. As a result, it may take some time for changes to propagate between Home Assistant and the thermostat.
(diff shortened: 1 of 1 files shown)


mvn23:
Still waiting for the parent PR to be reviewed...

source/_components/opentherm_gw.markdown
```diff
@@ -364,8 +163,319 @@ The value you provide here will be used with the GPIO `home` (5) and `away` (6)
...
 | `temperature` | no  | The setback temperature. Accepted values are `0.0` through `30.0`.
 
+## {% linkable_title Sensors %}
+
+The following `sensor` entities will be created for each configured gateway. The `entity_id` of every sensor will have a suffix containing the `gateway_id` of the gateway to which it belongs.
+<p class='note'>
+Not all boilers and thermostats properly support all OpenTherm features, so not all of the sensors will have useful values.
+</p>
+
+- <b>burner_hours</b><br />
```
MartinHjelmare: We can format text as bold with double stars around the target text in markdown.

```
- **burner_hours**
```

MartinHjelmare: `<br>`
br has no end tag I think.

Maybe replace those with simple line breaks in the markdown?

mvn23: The bold format will be changed to the stars notation.

`<br>` has no end tag, the slash is there to indicate a void element. Either `<br>` `<br/>` or `<br />` would do the same, with the latter being the most compatible as far as I know (not sure if that's relevant in this case though).
A line break alone would not translate to the additional info being on a new line. The markdown solution for this would be to replace the `<br />` tags with 2 trailing spaces, but personally I am not a big fan of trailing whitespaces as part of any syntax.
If, however, this is the preferred way to go within Home Assistant, I will change it.

MartinHjelmare: Sounds good! Either way works for me.

MartinHjelmare (approved):

source/_components/opentherm_gw.markdown
mvn23: Changed it to the trailing whitespaces in the last commit, it's a markdown file after all...

frenck (approved):

PR #9208, opened 2019-04-13
Update getting-started configurator section

lolli42:
Younger versions of the HASS Configurator have a
reduced set of config options, especially username and
password are gone. "Save" fails if trying to use the currently
documented set of options. The default config needs no change
at first install anymore. Reflect this in the getting started docs.


DubhAd (approved):
👍

PR #9247, opened 2019-04-18
remove reference to panel_iframe for configurator

cogneato:
adding panel_custom configuration for adding entry to nav panel

**Description:**

## Checklist:

- [x ] Branch: `next` is for changes and new documentation that will go public with the next [home-assistant](https://github.com/home-assistant/home-assistant) release. Fixes, changes and adjustments for the current release should be created against `current`.

### DIFF
--- a/source/getting-started/configuration.markdown
+++ b/source/getting-started/configuration.markdown
@@ -51,11 +51,16 @@ Time for the first practice with the configurator. We're going to add the Config
  - Copy and paste the following to the end of the `configuration.yaml` file:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


klaasnicolaas:
Perhaps it is better to put both ways there. You can continue to use the configurator without ingress as long as you provide a port number.

balloob:
We should stick to a single approach in the getting started. It should be aimed at getting people up to speed as fast as possible. Later they can learn about other things.

balloob (approved):

PR #9269, opened 2019-04-21
Use valid commands for open ports on xiaomi_aqara

savin-alexey:
From specification:
`psm-get <module> <variable>`
`psm-set <module> <variable> <value>`


frenck (approved):

frenck:
Thanks, @savin-alexey! 👍

PR #9281, opened 2019-04-23
new schema Mqtt State Vacuum 

pszafer:
**Description:**

New schema is provided for mqtt vacuum. This is documentation for it.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** https://github.com/home-assistant/home-assistant/pull/23171

## Checklist:


emontnemery:
We should maybe explain that the `state` schema is preferred because the vacuum will then be represented as a `StateVacuumDevice`. Also, it would be good to mention here https://developers.home-assistant.io/docs/en/entity_vacuum.html that the documentation applies for `StateVacuumDevice` and that `VacuumDevice` is deprecated.

pszafer:
I added a note about `StateVacuumDevice` but I'm not sure if it is not misleading for enduser (enduser shouldn't care about parent class for vacuum).

emontnemery:
MQTT is a bit more low level/hands on than most integrations, but maybe you're right.

klaasnicolaas (approved):

PR #9302, opened 2019-04-25
Add Life360 integration

pnbruckner:
**Description:**
Add Life360 integration.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#23405

## Checklist:


source/_components/life360.markdown
```diff
@@ -0,0 +1,234 @@
...
+You must first [create a Life360 account](https://www.life360.com/websignup). Then add the following section to your `configuration.yaml` file:
+
+{% raw %}
+```yaml
+# Example configuration.yaml entry
+device_tracker:
+  - platform: life360
+    username: LIFE360_USERNAME
+    password: LIFE360_PASSWORD
+    prefix: life360
```
klaasnicolaas: Only required items in the first example

klaasnicolaas (changes requested):

source/_components/life360.markdown
pnbruckner: Although prefix is not _technically_ required, I think it would be beneficial for most users to specify a unique prefix like this. (And since many, if not most, users simply copy the first example...)

I had thought about specifying a default, but then it wasn't clear how one would specify no prefix. Would any of the following be intuitive?
```text
prefix: ''
prefix: null
prefix:
```
I'm all ears if you have a better suggestion.

pnbruckner: Ok, decided to make the default 'life360', and any of the above would work to specify no prefix. I'll make a comment to that effect as well.

arsaboo:
@klaasnicolaas I don't see the parent-PR yet, was the `has-parent` label applied by mistake?

klaasnicolaas:
> @klaasnicolaas I don't see the parent-PR yet, was the `has-parent` label applied by mistake?

Goes almost automatically when I see that it is a new component and goes in the `next` 😆 Normally then there is always first a PR in the parent repo and then a PR for the docs.

pnbruckner:
Yep, still working on the actual code. Hope to submit that soon. Since I have to do both, and they can't be done exactly at the same time, and therefore reference one another, I had to decide which to do first. Decided to do this first, which ended up being a good idea, because I got good "requirements" feedback from @klaasnicolaas already! 😄

source/_components/life360.markdown
klaasnicolaas: Seems to be a good decision 😉

pnbruckner:
Requested changes completed. PR for code submitted (and initial comment above updated.)

source/_components/life360.markdown
```diff
@@ -0,0 +1,286 @@
...
+          Update for {{
+            state_attr(trigger.event.data.entity_id, 'friendly_name') or
+            trigger.event.data.entity_id
+          }} restored after {{ trigger.event.data.wait }}.
+```
+{% endraw %}
+
+## {% linkable_title Disclaimer %}
+
+It does not appear that Life360 officially supports its REST API for use with other than its own apps. This integration is based on reverse engineering that has been done by the open source community, and an API token that was somehow discovered by the same community. At any time Life360 could disable that token or otherwise change its REST API such that this custom component would no longer work.
```
klaasnicolaas:
```suggestion
It does not appear that Life360 officially supports its REST API for use with other than its own apps. This integration is based on reverse engineering that has been done by the open source community, and an API token that was somehow discovered by the same community. At any time Life360 could disable that token or otherwise change its REST API such that this component would no longer work.
```

klaasnicolaas (changes requested):

source/_components/life360.markdown
pnbruckner: Good catch! But actually, now it's an "integration", right?

klaasnicolaas: you are right, but a component or integration are the same. Even though we are pushing a bit to only talk about integrations 😄

pnbruckner:
Code PR not accepted.

PR #9314, opened 2019-04-26
Add visibility with a note

Yllelder:
Just as the top of the page is reported in the "note" format of a particular change in a specific version, the clarification at the bottom of the page should be in the "note" format as well.


klaasnicolaas:
Perhaps it is better to simply delete this entire line, the chance is very small that someone will still run a version under 0.35 🤣

fabaff (approved):

PR #9372, opened 2019-05-03
Add nws weather.

MatthewFlamm:
**Description:**. Adds documentation for NWS Weather.

~NWS requires permission to use their logo.  How should this be done?  Should I make this request?  See [NWS disclaimer page](https://www.weather.gov/disclaimer).  I have commented out the logo line and have not included it yet.~

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** https://github.com/home-assistant/home-assistant/pull/23647

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9372"></a>


klaasnicolaas:
> Permission is not required to display unaltered NWS products which include the NWS name or NWS/NOAA visual identifier as part of the original product. Neither the name nor visual identifier may be used, however, in a manner that implies an endorsement or affiliation with NOAA/NWS.

MatthewFlamm:
@klaasnicolaas I'm not sure that we are using "unaltered NWS products" or " as part of the original product.". I would say it is minimally altered, does this meet this requirement?

MatthewFlamm:
You are right @klaasnicolaas.  After reviewing this [page](https://www.weather.gov/logorequest-faq), and in particular this section,  I think our use is allowed.  This addresses my concern about the minimal amount of altering we have done, as long as we explicitly only describe NWS/NOAA as being the data source.

> 3.  Question – How can I attribute the original source of the data I use to NWS or NOAA even if the display is not considered to be “unaltered?”
> 
>     Answer – You may, without requesting a license, attribute the source of the original data to NWS or NOAA as along as it is clearly indicated that the data are from NOAA/NWS and not the display (e.g., “Data source:  NOAA”).  In addition, as required by 17 U.S.C. 403, you cannot copyright the government content within your data/products.

The documentation in this PR follows this guidance in my opinion.

Since we need to align on naming between NOAA and NWS with #9458, I'm going to hold off on adding the logo at this time.

source/_components/weather.nws.markdown
```diff
@@ -0,0 +1,78 @@
...
+description: "Instructions on how to integrate National Weather Service data within Home Assistant."
+date: 2019-05-02 12:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+featured: false
+#logo: nws.png
+ha_category: Weather
+ha_release: 0.93
```
klaasnicolaas:
```suggestion
ha_release: 0.96
```

source/_components/nws.markdown
```diff
@@ -0,0 +1,78 @@
...
+description: "Instructions on how to integrate National Weather Service data within Home Assistant."
+date: 2019-05-02 12:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+featured: false
+#logo: nws.png
+ha_category: Weather
+ha_release: 0.93
```
klaasnicolaas:
```suggestion
ha_release: 0.96
```

klaasnicolaas (changes requested):

source/_components/nws.markdown
```diff
@@ -0,0 +1,78 @@
...
+layout: page
+title: "NWS"
+description: "Instructions on how to integrate National Weather Service data within Home Assistant."
+date: 2019-05-02 12:00
+sidebar: true
+comments: false
+sharing: true
+footer: true
+featured: false
+#logo: nws.png
```
klaasnicolaas: Still missing a logo 😉

klaasnicolaas (dismissed):

source/_components/nws.markdown
MatthewFlamm: I was waiting to get aligned with the other PR with NOAA as the name, but I think that the author suggested that NWS could be used.  I've read through the NWS disclaimer page one more time now.  The section I quoted above only applies to attribution and not the logo (at least it doesn't mention the logo).

The quote you gave clearly says you need no permission only if it is unaltered and part of original product.  Are you comfortable with that assertion?  The data is altered, in particular, changing the condition to match a predefined set of HA conditions.  The rest of alterations are more trivial like rounding, and unit conversions.  This doesn't seem to meet this strict criteria in my interpretation.

MatthewFlamm:
@klaasnicolaas Now that the backend PR will hopefully get some movement with a simplified structure, can we revisit the logo issue?  I'm still not convinced that we fall under the specific requirements for no permission needed.  I think we are very close to that line, but the language is very restrictive.

frenck:
@MatthewFlamm Please leave it out. Thanks 👍

MatthewFlamm:
Thanks @frenck.  I'm okay with this path forward for now.  I do think we should consider asking permission if this integration gets included.  I think our use is clearly within their acceptable use category, just not within the category where no permission is needed to use.

I will note here for completeness that noaatides integration already uses an NOAA logo, which is the umbrella organization of NWS, and has the same requirements for logo usage.  I don't use this integration, so I cannot confirm, but it seems that this integration may be displaying the raw data without altering the data.  So, my reading is that this would fall under the no permission needed category.

One tricky thing here is that if anyone else adds a new NOAA application that does not conform to this requirement, you may have to remove the logo at that point.  Just pointing this out for future visibility.

source/_components/nws.markdown
```diff
@@ -8,7 +8,7 @@ comments: false
 sharing: true
 footer: true
 featured: false
-#logo: nws.png
+#logo: nws.png # permission needed
```
frenck: Commented out code, please just remove.

MatthewFlamm:
I assumed this would go in 0.99 since 0.98 is in beta.

klaasnicolaas:
Yes thats correct

klaasnicolaas:
Can you update this PR with the latest changes from our upstream repo? Thnx 😄

MatthewFlamm:
Thanks, I saw the build error, but didn't have time to investigate right away.

frenck (approved):
Thanks, @MatthewFlamm! 👍

MatthewFlamm:
Thanks @klaasnicolaas and @frenck!

PR #9408, opened 2019-05-07
Update google_assistant.markdown

ericellb:
This was changed in 0.84.0. If set to true Google will fail to link to your HA.

Did tons of troubleshooting to figure this out as it is not documented.
https://github.com/home-assistant/home-assistant/issues/23730

**Description:**

## Checklist:

### DIFF
--- a/source/_components/google_assistant.markdown
+++ b/source/_components/google_assistant.markdown
@@ -138,7 +138,7 @@ api_key:
   description: Your Homegraph API key (for the `google_assistant.request_sync` service)
   required: false
   type: string
-expose_by_default:
+expose_by_default: (Changed in 0.84. Will not sync devices if set to true)
   description: "Expose devices in all supported domains by default. If `exposed_domains` domains is set, only these domains are exposed by default. If `expose_by_default` is set to false, devices have to be manually exposed in `entity_config`."
   required: false
   default: true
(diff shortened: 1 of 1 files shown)


source/_components/google_assistant.markdown
```diff
@@ -138,7 +138,7 @@ api_key:
   description: Your Homegraph API key (for the `google_assistant.request_sync` service)
   required: false
   type: string
-expose_by_default:
+expose_by_default: (Changed in 0.84. Will not sync devices if set to true)
```
awarecan: ~There is no such change~

awarecan (changes requested):
How did you got this conclusion? I do not think you found right answer.

source/_components/google_assistant.markdown
awarecan: Alright, I withdraw my previous review. In 0.84, there was a change to fix a bug exists in 0.80-0.83.

This sentence is not right anyway, true/false is reversed and the actual situation is more complex. Moreover, since current document technically correct, I don't think we need add a note for a changed made in half year ago.

ericellb: If you set expose_by_default to *false*, and expose specific domains nothing will sync. It seems like you need to expose specific entities. This can be very confusing when setting stuff up, no reason for expose_by_default to even exist.

awarecan:
> Will not sync devices if set to false

Above statement is not correct

> If `expose_by_default` is set to false, devices have to be manually exposed in `entity_config`.

Above is correct

frenck:
As pointed out by awarecan, this PR contains an invalid change. Therefore, I'm going to close this PR.

Nevertheless, thank you for opening one, that is always greatly appreciated 👍

PR #9519, opened 2019-05-26
Initial simplification of doc

gerdesj:
Remove some euphemisms, simplify language and move note wrt venv for all Linux and not just RPi to top of doc.  Keep tone of original where possible.

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/installation/raspberry-pi.markdown
+++ b/source/_docs/installation/raspberry-pi.markdown
@@ -10,52 +10,27 @@ footer: true
 redirect_from: /getting-started/installation-raspberry-pi/
 ---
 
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
While I do like some of the changes, in general, you've changed the guide to a generic Linux guide. We already have guides for that. Also, it removed some valuable information in the process.

For example: How to connect & login to your Pi (the guide now assumes you know), but also the warning about the power supply. Especially that last one, is, unfortunately, often the case in issue reports.

gerdesj:
OK, I'll retract my merge request.  I was aiming to make this a generic Linux guide and then perhaps create a separate set of notes for specific cases - eg RPi (which might be better off with Hassio instead anyway)

PR #9535, opened 2019-05-30
Add esphome to discovery components

olskar:
**Description:**

Add esphome to discovery components

## Checklist:

- [ X] Branch: `next` is for changes and new documentation that will go public with the next [home-assistant](https://github.com/home-assistant/home-assistant) release. Fixes, changes and adjustments for the current release should be created against `current`.
- [ X] The documentation follows the [standards][standards].


Jc2k:
This is not strictly true for 0.94 onwards - the following components are now detected by `zeroconf` instead of `discovery`:

 * axis
 * esphome
 * ikea_tradfri
 * homekit_controller
 * philips_hue

balloob:
@jc2k is correct. We should update the other pages, will do so in the rc branch.

PR #9537, opened 2019-05-30
Remove esphome breaking change

OttoWinter:
**Description:**

See https://github.com/home-assistant/home-assistant/pull/24043#issuecomment-497360049 and https://github.com/home-assistant/home-assistant/pull/24043#issuecomment-497360925

Since ESPHome was only ever configurable through config flow. Plus it seems like the zeroconf change is not a breaking change yet anyway.

## Checklist:


balloob:
This is automatically generated based off the labels, which I will run a few times probably. Will merge this but in future, it's not necessary 👍

PR #9544, opened 2019-06-01
🔨 Change the device class description

klaasnicolaas:
**Description:**
This PR ensures that all links for the device classes now refer to the page on the developers website.

## Checklist:

### DIFF
--- a/source/_components/ads.markdown
+++ b/source/_components/ads.markdown
@@ -250,6 +250,6 @@ name:
   type: string
 device_class:
   required: false
-  description: Sets the class of the device, changing the device state and icon that is displayed on the UI (awning, blind, curtain, damper, door, garage, shade, shutter, window)
+  description: Sets the [class of the device](https://developers.home-assistant.io/docs/en/entity_cover.html#device-classes), changing the device state and icon that is displayed on the UI.
   type: device_class
 {% endconfiguration %}
--- a/source/_components/ads.markdown
+++ b/source/_components/ads.markdown
@@ -102,7 +102,7 @@ name:
   required: false
   type: string
 device_class:
-  description: The [type/class](/components/binary_sensor/) of the sensor to set the icon in the frontend.
+  description: The [type/class](https://developers.home-assistant.io/docs/en/entity_binary_sensor.html#available-device-classes) of the sensor to set the icon in the frontend.
   required: false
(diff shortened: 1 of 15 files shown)


source/_components/w800rf32.markdown
```diff
@@ -68,7 +68,7 @@ devices:
       required: false
       type: string
     device_class:
-      description: "The [type or class of the sensor](/components/binary_sensor/) to set the icon in the frontend."
+      description: "The [type or class of the sensor](https://developers.home-assistant.io/docs/en/entity_binary_sensor.html#available-device-classes) to set the icon in the frontend."
```
DubhAd: The device class also changes the way the state is referred to. For instance a `door` will be open or closed rather than on or off

source/_components/ads.markdown
```diff
@@ -102,7 +102,7 @@ name:
   required: false
   type: string
 device_class:
-  description: The [type/class](/components/binary_sensor/) of the sensor to set the icon in the frontend.
+  description: The [type/class](https://developers.home-assistant.io/docs/en/entity_binary_sensor.html#available-device-classes) of the sensor to set the icon in the frontend.
```
DubhAd: Linking to the developer docs, for a user configuration, just seems very wrong. This is either a user feature documented in the user docs, or a developer feature...

frenck: Or both userdocs and developer feature.

I agree links from user docs to dev docs should be forbidden unless the explicit need is to redirect developers to the right docs (which is absolutely not the case here).

Nevertheless, it solves an (apparently) direct need from users for now.

So I'm fine with it for now. Searching for "developers.home-assistant.io" in our documentation can help us solve / fix things better later on.

frenck (dismissed):

source/_components/ads.markdown
balloob: So I agree with both of you. I think we did jump the gun by removing the device classes, and there are more use cases than I anticipated, and linking to the dev docs is not good practice.

Let's merge this, and then let's do a new PR to put the device classes back

klaasnicolaas: I did a new commit where the old links are back, with a small change in the description everywhere because it not only changes the icon but also the state.

PR #9550, opened 2019-06-02
Fix ipv6 issue with Nginx not redirecting to ipv4 instance of Hass

Barrysv:
In the Nginx config rather than specifying localhost:8123 use 127.0.0.1:8123 so this will ensure incoming IPv6 requests get redirected to Hass on IPv4. "localhost:8123" means Nginx will try to connect to Hass using the same protocol that came in - and Hass doesn't do dual stack at present

**Description:**

## Checklist:


frenck (approved):
Thanks, @Barrysv! 👍

PR #9552, opened 2019-06-02
Link to integration WIP

MatthewFlamm:
**Description:**
Currently several weather sensor plarforms link to generic components/sensor rather than to integration/sensor.

## Checklist:

### DIFF
--- a/source/_components/sensor.buienradar.markdown
+++ b/source/_components/sensor.buienradar.markdown
@@ -12,6 +12,8 @@ ha_category:
   - Weather
 ha_release: 0.47
 ha_iot_class: Cloud Polling
+redirect_from:
+ - /components/sensor.buienradar/
 ---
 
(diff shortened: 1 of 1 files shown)


MatthewFlamm:
My attempted fix didn't work.  From my observation, pages with redirect_from have the correct link, but it apparently doesn't work so simply.

klaasnicolaas:
`redirect_from` is used for other reasons, this to link existing URLs to a new page. Now you redirect the url to the same page, what was your intention?

MatthewFlamm:
The source link points to /components/sensor/ rather than /components/integration/.  My observation was that files with a redirect_from had the correct link.  I  now think it is the file name, and this creates problems when that file already exists.  See buienradar and dark sky for examples where one of the two pages links to the generic component.  I'm not sure how to fix that without combining the pages together.  Another idea is to have a specifiable source link in the header.

klaasnicolaas:
This is because with both the buienradar component and darksky there has not been a single page migration yet. Now it consists of a component page and a separate platform page, the platform pages must eventually disappear. This has already happened with 90% of all integrations.

`redirect_from` has no influence on the source url.

See also #8619

MatthewFlamm:
I understand now, thanks!

PR #9561, opened 2019-06-03
Update yaml.markdown

Wuppie007:
included code reusability for YAML files

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/configuration/yaml.markdown
+++ b/source/_docs/configuration/yaml.markdown
@@ -93,6 +93,25 @@ lights: !include lights.yaml
 ```
 
 More information about this feature can also be found at [splitting configuration](/docs/configuration/splitting_configuration/).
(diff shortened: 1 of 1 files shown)


fabaff:
I'm leaning in the direction not to add this to the documentation. First we are moving to configure the system with the help of the frontend and second, it will make it very hard to debug people's configuration. 

Splitted configurations and anchors/aliases could become a massive source of configuration issues. Hardcore users will know how to deal with it already but the average users should not be handle their configuration in this way.

frenck:
I agree, although I would think adding a link or a reference to more extensive YAML documentation/spec, could be helpful maybe.

frenck:
This was discussed and we decided not to add this as reasoned by fabaff.

Nevertheless, thank you for your time creating this PR. That is greatly appreciated.

PR #9661, opened 2019-06-19
Add tolerance to dlib face identify

gonzalezcalleja:
**Description:**

Documentation for new tolerance parameter.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#24497

## Checklist:


source/_components/dlib_face_identify.markdown
```diff
@@ -50,6 +51,12 @@ faces:
   description: List of faces sources.
   required: true
   type: list
+confidence:
+  description: How much distance between faces to consider it a match. Using tolerance values lower than 0.6 will make the comparison more strict.
+  required: false
+  type: integer
```
frenck: According to the parent PR, this should be a float.
```suggestion
  type: float
```

source/_components/dlib_face_identify.markdown
```diff
@@ -30,6 +30,7 @@ image_processing:
    faces:
      Jon: /home/hass/jon.jpg
      Bob: /home/hass/bob.jpg
+   confidence: 0.6
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

frenck (changes requested):
Thanks for opening the PR @gonzalezcalleja! 👍 

I've reviewed it and left some comments. Could you please take a look? Awesome! 👍

source/_components/dlib_face_identify.markdown
gonzalezcalleja: Done! thanks for your comments.

source/_components/dlib_face_identify.markdown
gonzalezcalleja: Ok, I'll remove confidence from sample configuration.

frenck (approved):

PR #9669, opened 2019-06-21
Add note on conditions not working without forecast

KarimGeiger:
It tripped me off that some monitored conditions only work with the forecast attribute set, as they require a forecast to be made. This will add a note that explains this behaviour.

## Checklist:

### DIFF
--- a/source/_components/darksky.markdown
+++ b/source/_components/darksky.markdown
@@ -160,7 +160,7 @@ scan_interval:
 {% endconfiguration %}
 
 <p class='note'>
-Please note that some monitored conditions, such as `temperature_high` or `temperature_high`, may only work when setting the `forecast` attribute to at least `0` (current day).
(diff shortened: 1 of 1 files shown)


source/_components/darksky.markdown
```diff
@@ -159,6 +159,10 @@ scan_interval:
   type: time
 {% endconfiguration %}
 
+<p class='note'>
+Please note that some monitored conditions, such as `temperature_high` or `temperature_high`, may only work when setting the `forecast` attribute to at least `0` (current day).
```
klaasnicolaas: `temperature_high` or `temperature_high`?

KarimGeiger: Thanks. That obviously wasn't correct.

frenck (approved):
Thanks, @KarimGeiger! 👍

PR #9679, opened 2019-06-22
Additional instructions for Mariadb 10.3

ghost:
In Debian/Rasbian buster (MariaDB 10.3) the command mysql_config needed by the python package mysqlclient was moved to the package libmariadb-dev-compat.   The python package will not install if libmariadb-dev-compat is missing in Debian Buster and everyone who is upgrading from stretch is going to run into this error, so this additional step is important.


frenck (approved):
Thanks, @eyager1! 👍

PR #9687, opened 2019-06-24
notes on using states(xxx) vs states.xxx

akasma74:
as per this discussion - https://community.home-assistant.io/t/using-better-templates-with-state/122912/14

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/configuration/templating.markdown
+++ b/source/_docs/configuration/templating.markdown
@@ -59,7 +59,7 @@ script:
 ### {% linkable_title States %}
 - Iterating `states` will yield each state sorted alphabetically by entity ID.
 - Iterating `states.domain` will yield each state of that domain sorted alphabetically by entity ID.
-- `states.sensor.temperature` returns the state object for `sensor.temperature`.
+- `states.sensor.temperature` returns the state object for `sensor.temperature` if `sensor.temperature` exists, and `unknown` otherwise (for example, during Home Assistant startup for [template sensor](https://www.home-assistant.io/components/template/#considerations) *producing error messages in log file*. Use one of the functions below to avoid this:
 - `states('device_tracker.paulus')` will return the state string (not the object) of the given entity or `unknown` if it doesn't exist.
 - `is_state('device_tracker.paulus', 'home')` will test if the given entity is the specified state.
 - `state_attr('device_tracker.paulus', 'battery')` will return the value of the attribute or None if it doesn't exist.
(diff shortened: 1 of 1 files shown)


homeassistant:
This pull request has been mentioned on **Home Assistant Community**. There might be relevant details there:

https://community.home-assistant.io/t/using-better-templates-with-state/122912/21

frenck:
Create an issue for this in for a broader improvement #9711

akasma74:
Fair enough.
Do you want me to close this PR then?
And what help do you need?

frenck:
I'm looking into it. Please leave this PR open. I've assigned myself. Thanks.

frenck:
I made it more explicit by moving the message into a warning. Made the message a little broader and as a result, makes the original list as clean as before.

frenck (approved):

frenck:
Thanks for bringing this up @akasma74! 👍

akasma74:
@frenck Thanks for improving it!
One thought though that I lost at some point - this sentence is a bit confusing, especially for newcomers:
> Avoid using `states.sensor.temperature`, instead use `states('sensor.temperature')`

Mainly because they return different things - state object and state object's state respectively.
I don't know the reason why in HA world they introduced term State object that has state attribute, and it all exists within `states` container so if one wants to know state of an object (`open`/`low` etc) they need to use a bit over-the-top `states.sensor.temperature.state`, in which case it is safe to recommend switching to  `states('sensor.temperature')`.
And just a note that if one need access to `last_changed`/`last_updated`, there is no choice as it's impossible to get these using `state_attr` so it will take some effort to write a code that does not produce errors if `sensor.temperature` does not exist (I might be wrong though ;)

frenck:
@akasma74 I know the difference.

The reasoning behind this wording is that it is hard to explain the difference between an object and string to the average user. The function cover about 95% of the use cases out there.

Besides, the list above does mention the difference.

akasma74:
> @akasma74 I know the difference.

I bet you do ;)

> The reasoning behind this wording is that it is hard to explain the difference between an object and string to the average user. The function cover about 95% of the use cases out there.

Well, that's exactly why I said that warning could confuse users. Agree about 95%

How about something like "If you need to get state of your entity (i.e Open/Off etc), it is advisable to use `states('sensor.temperature')` and not `states.sensor.temperature`. Generally, it is recommended to use the access methods `states()`, `is_state()`, `state_attr()` and `is_state_attr()` as much as possible, to avoid errors messages when the entity is unknown or isn't ready yet (e.g., during Home Assistant startup)."

> Besides, the list above does mention the difference.

correct

frenck:
It is not just the state, also state attributes, that is what made it difficult. "If you need to get state of your entity (i.e Open/Off etc)" that might throw people off for attributes.

PR #9693, opened 2019-06-26
Added link to add-on configuration doc

esciara:
**Description:**
Added link to add-on configuration doc but I think this part of the doc needs to be reworked, with example, as there are a lot of concepts that are taken for granted... I certainly could not set this up without a lot of trial and error...

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** not applicable

## Checklist:

### DIFF
--- a/source/hassio/external_storage.markdown
+++ b/source/hassio/external_storage.markdown
@@ -11,7 +11,7 @@ footer: true
 
 ### {% linkable_title ResinOS / Generic %}
 
-Map the USB drive into add-on with `devices` options. If you need it on multiple add-ons, you can use the `/share` folder which is accessible from various add-ons.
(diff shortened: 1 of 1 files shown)


frenck:
I don't see how the added pieces have anything to do with external storage?

esciara:
Hi @frenck . Then I do not understand what the doc means to say... the title is "External storage"... the only `devices` option I could find was the one I pointed with the link. So I thought that to point add-ons to use an external storage, one needed to use this option.

If not, could you clarify what `devices` option is talked about and point to a link that explains where it is and how to use it ?

frenck:
This is a weird page, still dating from the early days of Hass.io.
I'm looking into it.

frenck:
The problem here is that the page should not be here at all.
I've created #9712 to remove it.

Closing up this PR, nevertheless, thank you for taking the time to open one! That is appreciated.

esciara:
Thanks @frenck

Issue #9637, opened 2019-06-15
Microsoft Text-to-Speech issues when add language code: vi-VN


nguyenanh87:
HI,
When i config Microsoft TTS service with language code is vi-VN, the system say:

Invalid config for [tts.microsoft]: value is not allowed for dictionary value @ data['language']. Got 'vi-VN'. (See ?, line ?). Please check the docs at https://home-assistant.io/components/tts.microsoft/

I've checked language code, type at the document: https://docs.microsoft.com/en-us/azure/cognitive-services/Speech/api-reference-rest/bingvoiceoutput

![Screen Shot 2019-06-15 at 22 32 20](https://user-images.githubusercontent.com/40736871/59552145-16c3ce00-8fbe-11e9-8e95-50fbfedab64a.png)

![Screen Shot 2019-06-15 at 22 32 33](https://user-images.githubusercontent.com/40736871/59552147-1deadc00-8fbe-11e9-96c1-5694932e3505.png)

Please suggest me the solutions or add more language support to source code.

klaasnicolaas:
What if you do it lowercase? like `vi-vn`

nguyenanh87:
I’ve tried with same issues

aidbish:
The only supported languages currently for the component are

    'ar-eg', 'ar-sa', 'ca-es', 'cs-cz', 'da-dk', 'de-at', 'de-ch', 'de-de',
    'el-gr', 'en-au', 'en-ca', 'en-gb', 'en-ie', 'en-in', 'en-us', 'es-es',
    'es-mx', 'fi-fi', 'fr-ca', 'fr-ch', 'fr-fr', 'he-il', 'hi-in', 'hu-hu',
    'id-id', 'it-it', 'ja-jp', 'ko-kr', 'nb-no', 'nl-nl', 'pl-pl', 'pt-br',
    'pt-pt', 'ro-ro', 'ru-ru', 'sk-sk', 'sv-se', 'th-th', 'tr-tr', 'zh-cn',
    'zh-hk', 'zh-tw',

frenck:
Let's add the list to the integration page.

frenck:
### 🆘 Marked as "Help wanted"

If you want to help out on this, go ahead, it is really appreciated!

Please leave a comment if you started working on this, so we can avoid multiple people working on the same things. Thanks! 👍

aidbish:
PR submitted.
[...]

PR #9721, opened 2019-06-29
Add list of supported languages.

aidbish:
The integration does not list the languages that it supports.
Some languages are missing from the integration itself compared to the Microsoft documentation, adding the languages it supports should prevent users trying to add an non-supported language

https://github.com/home-assistant/home-assistant.io/issues/9637

### DIFF
--- a/source/_components/microsoft.markdown
+++ b/source/_components/microsoft.markdown
@@ -35,7 +35,7 @@ api_key:
   required: true
   type: string
 language:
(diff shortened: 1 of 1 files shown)


source/_components/microsoft.markdown
```diff
@@ -70,6 +70,14 @@ contour:
   type: string
 {% endconfiguration %}
 
+Languages supported by this integration.
```
fabaff: Instead of copying over the content from the source we should simply add a link to https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/microsoft/tts.py#L20 in the `description:`. This way we don't need to maintain the list here too.

fabaff (changes requested):

source/_components/microsoft.markdown
aidbish: Updated as requested

fabaff (approved):

PR #9723, opened 2019-06-29
Added information on workaround

olskar:
**Description:**
Added information on workaround  if affected by bug  https://github.com/home-assistant/home-assistant/issues/23942

## Checklist:

- [X ] Branch: `next` is for changes and new documentation that will go public with the next [home-assistant](https://github.com/home-assistant/home-assistant) release. Fixes, changes and adjustments for the current release should be created against `current`.
- [X ] The documentation follows the [standards][standards].

### DIFF
--- a/source/_components/rflink.markdown
+++ b/source/_components/rflink.markdown
@@ -31,6 +31,10 @@ This integration is tested with the following hardware/software:
 
 ## {% linkable_title Configuration %}
 
+<p class='note warning'>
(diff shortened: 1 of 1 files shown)


frenck:
I don't think we should update documentation because of an open issue.
This becomes unmaintainable, as the docs need to be updated each time the state of the issue changes.

frenck:
Closing this PR, because of the above reasoning.

Nevertheless, thank you for taking the effort to open up a PR.

olskar:
I thought it would be good to tell users about a workaround since no one seems to be able to understand how this bug can be fixed. But I understand your reasoning, thank you for answering @frenck

PR #9737, opened 2019-06-30
No longer valid

aidbish:
https://github.com/home-assistant/home-assistant.io/issues/9712

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9737"></a>

PR #9740, opened 2019-06-30
Removal of navigation links PR9737

### LINKED ISSUE #9712: Remove hassio/external_storage
The following page is not valid and should be removed:

<https://www.home-assistant.io/hassio/external_storage/>

It also needs to be removed from the navigation.

aidbish:
Closes https://github.com/home-assistant/home-assistant.io/issues/9712


frenck:
@aidbish Please add that to #9737, thanks! 👍

PR #9803, opened 2019-07-07
Fix formatting of installation commands

Chris-Johnston:
**Description:**

This change fixes the formatting of the generic linux host installation commands so that they all begin with `$`. This ensures that they are rendered correctly in the documentation.

Previously, the commands under `$ sudo -i` all began with `#` (to indicate that the shell was running as root). While this is correct, the markdown formatting thinks that they are comments, so they are rendered in a light gray color. I think that it's better that they are shown as commands and not comments.

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9803"></a>

### DIFF
--- a/source/hassio/installation.markdown
+++ b/source/hassio/installation.markdown
@@ -83,7 +83,7 @@ Best practice for updating a Hass.io installation:
 SSH to your Hass.io system, or connect to the console, and run:
 
 ```bash
-$ hassio ha update --version=0.XX.X
+hassio ha update --version=0.XX.X
 ```
 
 ## Run the beta version on Hass.io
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
Actually, none of the commands produce output, so all preceding has to be removed.

See Markdown style guides rule 14:

https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md014---dollar-signs-used-before-commands-without-showing-output

frenck (approved):
Thanks for adjusting this so quickly @Chris-Johnston! 👍

PR #9824, opened 2019-07-09
Update emulated_hue.markdown

pattyland:
**Description:**

The current documentation states that port 80 is required for "Echo Dot 3". First of all this is unspecific because it counts for all new devices like e.g. Echo (2nd Gen.) and is not needed as long as an older device like a 2nd generation Dot remains active in the network.
Maybe an Info Block, which describes both ways briefly, makes sense?

## Checklist:

- [ x] Branch: `next` is for changes and new documentation that will go public with the next [home-assistant](https://github.com/home-assistant/home-assistant) release. Fixes, changes and adjustments for the current release should be created against `current`.

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9824"></a>

### DIFF
--- a/source/_components/emulated_hue.markdown
+++ b/source/_components/emulated_hue.markdown
@@ -49,7 +49,7 @@ emulated_hue:
 ```yaml
 # Amazon Echo example configuration.yaml entry
 emulated_hue:
-# Required for Echo Dot 3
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @pattyland! 👍

PR #9850, opened 2019-07-14
state trigger - entity_id details

akasma74:
I think it would be beneficial to underline in text that it is possible to use multiple entities in entity_id.
Not sure if it is applicable to any other triggers that accept entity_id (like numeric_state)

**Description:**

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9850"></a>

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -102,7 +102,7 @@ automation:
 
 ### State trigger
 
-Triggers when the state of a given entity changes. If only `entity_id` is given trigger will activate for all state changes, even if only state attributes change.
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @akasma74! 

I've removed the link to the YAML docs. While useful, it is misleading, since it does not show directly how to use this with multiple entities and can, therefore, be more confusing.

The part of any of the given entities (as in plural) is absolutely useful. Thank you for that.

👍

akasma74:
@frenck I tried to find an example where the mapping (dash) notation is used (as in Trigger section entities are separated by commas and I never actually noticed that) and the YAML page looked ok to me.
It describes and shows how to use mapping notation.

PR #9871, opened 2019-07-16
Allow more universal media_player attributes to be overridden by the user

babichjacob:
**Description:**
Currently, the universal media_player only allows for the is_volume_muted, state, source, source_list and volume_level attributes to be overridden in the attributes: section of the config.

This PR allows media_content_id, media_content_type, media_duration, media_image_url (and thereby entity_picture), media_title, media_artist, media_album_name, media_album_artist, media_track, media_series_title, media_season, media_episode, media_channel, media_playlist, app_id, app_name, media_position, and media_position_updated_at to be overridden in the same way.

Rationale: I needed to override app_name so that the custom Lovelace card mini-media-player would show my chosen source instead of passing through to the Assistant speaker's app_name that reports "Default Media Receiver". Because the other attributes listed above are just as easy to allow to be overridden as app_name, I included them in this PR instead of leaving them alone with the arbitrary restriction that they must take their value from the active child. 

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#24903

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9871"></a>

### DIFF
--- a/source/_components/universal.markdown
+++ b/source/_components/universal.markdown
@@ -63,7 +63,7 @@ commands:
   required: false
   type: string
 attributes:
-  description: "Attributes that can be overwritten. Possible entries are `is_volume_muted`, `state`, `source`, `source_list` and `volume_level`. The values should be an entity ID and state attribute separated by a pipe character (|). If the entity ID's state should be used, then only the entity id should be provided."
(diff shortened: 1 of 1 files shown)


babichjacob:
Cancelled parent PR.

PR #9878, opened 2019-07-17
Update solaredge_local.markdown

Petro31:
Remove sensors: section head.  This is an incorrect section header.  Adding the platform as a list item so new users can copy and paste into existing sensor sections.

**Description:**

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9878"></a>

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9878"></a>

### DIFF
--- a/source/_components/solaredge_local.markdown
+++ b/source/_components/solaredge_local.markdown
@@ -58,11 +58,10 @@ In case you would like to convert the values for example to kWh instead of the d
 
 {% raw %}
 ```yaml
-# Example configuration.yaml entry for template platform
-sensors:
-  platform: template
-  sensors:
-    solaredge_energy_this_year_template:
-      value_template: '{{(states('sensor.solaredge_energy_this_year') | float / 1000) | round(2)}}'
+# Example configuration.yaml entry for sensor template platform
+  - platform: template
+    sensors:
+      solaredge_energy_this_year_template:
+        value_template: '{{(states('sensor.solaredge_energy_this_year') | float / 1000) | round(2)}}'
 ```
 {% endraw %}
--- a/source/_components/solaredge_local.markdown
+++ b/source/_components/solaredge_local.markdown
@@ -59,9 +59,10 @@ In case you would like to convert the values for example to kWh instead of the d
 {% raw %}
 ```yaml
 # Example configuration.yaml entry for sensor template platform
+sensor:
   - platform: template
     sensors:
       solaredge_energy_this_year_template:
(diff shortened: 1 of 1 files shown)


frenck:
I'm thinking of not merging this since it will create an inconsistency in our documentation...

frenck:
as a counter-argument... depending on your configuration method, even your changed version is not copied & pasteable.

frenck:
Thought about this a bit more, I appreciate the way you think about the usability, however, consistency is an important factor as well.

Switching style on a single location, while keeping all the rest as is, is probably worse at this point. Hence I'm closing this PR.

VDRainer:
Is it better that users copy an absolut wrong example that doesn't work?
Sure there are always people in the community that help out!

source/_components/solaredge_local.markdown
```diff
@@ -58,11 +58,10 @@ In case you would like to convert the values for example to kWh instead of the d
...
-sensors:
-  platform: template
-  sensors:
-    solaredge_energy_this_year_template:
-      value_template: '{{(states('sensor.solaredge_energy_this_year') | float / 1000) | round(2)}}'
+# Example configuration.yaml entry for sensor template platform
+  - platform: template
+    sensors:
+      solaredge_energy_this_year_template:
+        value_template: '{{(states('sensor.solaredge_energy_this_year') | float / 1000) | round(2)}}'
```
otherwish: The outer or inner quotations should be different since it won't work as well. Maybe adding sensor: would bring it in line??

frenck:
> Is it better that users copy an absolut wrong example that doesn't work?

That depends on the situation, the change provided has the same effect for others as well.

Petro31:
Well at least we should remove the typo

frenck:
Aah, I see, the issue now!

However, this PR also removed the `sensor:` bit, that threw me off!

Petro31:
Yeah, I only a removed the 'sensor:' bit because some people don't understand the whole 'can't have more than 1 sensor section' restriction.

source/_components/solaredge_local.markdown
frenck: 👍

frenck (approved):
👍  Tweaked it a bit and this should be good to go!

Sorry for the confusion there! 🙏

frenck:
@Petro31 I get that, nevertheless, that is not something we should do on a single location. If we consider doing that, it should be decided more widely and changed widely.

Petro31:
@frenck yeah not a problem.  I'm good with whatever changes we make.

PR #9904, opened 2019-07-19
Update Arch Linux instructions to install from 'community' repo

etskinner:
**Description:** Arch Linux's Home Assistant package has recently been added to the official 'community' repository. This commit updates the instructions for Arch to reflect this change.

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9904"></a>

### DIFF
--- a/source/_docs/installation/archlinux.markdown
+++ b/source/_docs/installation/archlinux.markdown
@@ -5,7 +5,27 @@ description: "Installation of Home Assistant on your Arch Linux computer."
 
 [Arch Linux](https://www.archlinux.org/) is a lightweight and flexible Linux distribution for x86_64.
 
-Install the needed Python packages.
+### Official Repositories
+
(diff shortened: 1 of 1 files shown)


frenck:
We are trying to reduce the number of community created installation methods in our documentation, because they have concerns about people considering it an official supported method by us, which is not the case.

Secondly, there is no way we can control it, keep it up 2 dates or in sync with our current releases (at the time of writing this comment, the package is already 2 minor releases behind).

The currently place to display this community effort would be our community forum.

Thank you so much for your contribution ❤️

etskinner:
@frenck : Leaving this as-is would result in outdated instructions and a 'broken link' situation. Since the package has transitioned to official Arch repositiories, it is no longer available in the AUR like the documentation says. I respectfully request that you reconsider your decision.

frenck:
I’m not merging this in, instead part of the current docs will be removed, fixing your concern.

PR #9943, opened 2019-07-21
Rename huawei_lte traffic_statistics to monitoring_traffic_statistics

scop:
**Description:**

See home assistant commit for more info.

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant) (if applicable):** home-assistant/home-assistant#25377

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9943"></a>

### DIFF
--- a/source/_components/huawei_lte.markdown
+++ b/source/_components/huawei_lte.markdown
@@ -122,8 +122,8 @@ sensor:
     monitored_conditions:
       - device_information.SoftwareVersion
       - device_signal.rssi
-      - traffic_statistics.CurrentDownloadRate
-      - traffic_statistics.TotalConnectTime
+      - monitoring_traffic_statistics.CurrentDownloadRate
+      - monitoring_traffic_statistics.TotalConnectTime
 ```
 
 {% configuration %}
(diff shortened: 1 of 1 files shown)


source/_components/huawei_lte.markdown
```diff
@@ -122,8 +122,8 @@ sensor:
     monitored_conditions:
```
frenck, line 122: Monitored conditions should be removed overall, for more information see ADR-0003:

https://github.com/home-assistant/architecture/blob/master/adr/0003-monitor-condition-and-data-selectors.md

scop: Sure, I know, but these are just some changes I need to get out of the way before making larger changes. There will be quite a few, unique ids will change, all config will consolidate below top level huawei_lte entry, etc.

I will look into monitored_conditions as part of that, but I must also note that the ADR says "if that data is fetched in a single API request". In huawei_lte's case, it is not -- each category of information requires a separate http request to the router, and even worse, the set of things available from the same requests varies to some extent between routers. But as said, that's something for another near future changeset.

frenck (approved):
Thanks, @scop! 👍

PR #9952, opened 2019-07-23
Remove link from Logbook introduction

SeanPM5:
The Home Assistant demo website doesn't have Logbook, so there's no "live example" that can be seen anymore. Better to remove this link rather than causing readers confusion.

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9952"></a>


frenck (approved):
Thanks, @SeanPM5! 👍

PR #9970, opened 2019-07-25
shorten network instruction and add link

capstan1:
There already exists a very detailed instructions on how to set up network wifi in https://github.com/home-assistant/hassos/blob/dev/Documentation/network.md and https://github.com/home-assistant/hassos/blob/dev/Documentation/configuration.md

The shorter version without any details here does not help the user to solve his problem, he has to got to the other pages either way. When information are scattered throughout various pages this might confuse the user.
Also, the link to the to HassOS HowTo (with the configuration files) was only present in the second option, so the person using usbstick might not click it.

**Description:**

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9970"></a>

### DIFF
--- a/source/hassio/installation.markdown
+++ b/source/hassio/installation.markdown
@@ -33,10 +33,7 @@ The following will take you through the steps required to install Hass.io.
    - Flash the downloaded image to an SD card using [balenaEtcher][balenaEtcher]. If using a Pi we recommend at least a 32 GB SD card to avoid running out of space. On Virtual machine platforms, provide at least 32 GB of disk space for the VM.
    - Load the appliance image into your virtual machine software. Choose 64-bit Linux and UEFI boot.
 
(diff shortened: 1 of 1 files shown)


frenck:
I do not agree, this gives a bit of introduction on what is coming and helps the user to pre-orientate before reaching the more technical documentation part.

Furthermore, you talking about adding a link, but that was already there? 🤷‍♂

capstan1:
If that is the way it is wanted, okay.

And I should have written "moved link". If you stay with the current passage, you should 
1.) add the link to the first option too
or
2) move the link to the top or the bottom so that it's not placed at the second option only.

frenck:
I'm closing this PR since I believe it is not an improvement to what we currently have.

PR #9971, opened 2019-07-25
add details on usb-stick

capstan1:
A question I asked myself and that got asked frequently in the forums is weather the usb stick used for configuration needs to stay within the device. So I just added one line to make this more precise.

**Description:**

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9971"></a>

### DIFF
--- a/source/getting-started/index.markdown
+++ b/source/getting-started/index.markdown
@@ -57,12 +57,13 @@ We will need a few things to get started with installing Home Assistant. The lat
    - Create a folder named `network` in the root of the newly formatted USB-stick.
    - Within that folder create a file named `my-network` without extension.
    - Copy one of [the examples] to the `my-network` file and adjust accordingly.
-   - Plug the USB-stick into the Raspberry Pi 3. (The USB-stick needs to be inserted to the device during the following setup process only and can be disconnected afterwards.)
+   - Plug the USB-stick into the Raspberry Pi 3.
 
 1. Insert the SD card into your Raspberry Pi 3. If you are going to use an Ethernet cable, connect that too.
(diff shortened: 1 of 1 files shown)


source/getting-started/index.markdown
```diff
@@ -57,7 +57,7 @@ We will need a few things to get started with installing Home Assistant. The lat
    - Create a folder named `network` in the root of the newly formatted USB-stick.
    - Within that folder create a file named `my-network` without extension.
    - Copy one of [the examples] to the `my-network` file and adjust accordingly.
-   - Plug the USB-stick into the Raspberry Pi 3.
+   - Plug the USB-stick into the Raspberry Pi 3. (The USB-stick needs to be inserted to the device during the following setup process only and can be disconnected afterwards.)
```
frenck: I kinda think this is obvious, way to much text at this part and in a way: not true.

The network configuration is read early in the process (not that it really matters, but that is just details).

capstan1: Maybe I'm not the average user here but when I set up Hass.IO I was desperately looking for some USB drive I would not need later on. I only read later that it's not persistent.
Also, here:
https://community.home-assistant.io/t/hassos-wlan/59988?
https://community.home-assistant.io/t/hass-io-wifi/59778/2?

frenck: Well maybe add an additional step at the end:

```Markdown
1. If you have used a USB-stick for configuring the network, it can now be removed.
```

Which is much cleaner and follow a step-by-step procedure, instead of trying to squeeze in a note in a step in the middle.

frenck (approved):

PR #9984, opened 2019-07-27
Include note about OpenWRT unreliability

kylegordon:
OpenWRT SNMP only returns data from the ARP table - not all associated stations.

**Description:**

OpenWRT presence detection can be unreliable if using dumb AP mode. SNMP only returns data from the ARP table - not the associated stations list.

## Checklist:

- [current] Branch: `next` is for changes and new documentation that will go public with the next [home-assistant](https://github.com/home-assistant/home-assistant) release. Fixes, changes and adjustments for the current release should be created against `current`.

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/9984"></a>

### DIFF
--- a/source/_components/snmp.markdown
+++ b/source/_components/snmp.markdown
@@ -24,13 +24,7 @@ There is currently support for the following device types within Home Assistant:
 - [Switch](#switch)
 
 <div class='note warning'>
-  This device tracker needs SNMP to be enabled on the router. It could be that you need to install the SNMP support manually.
-  This can somewhat be mitigated with by installing fping on the device and performing a broadcast ping periodically.
-  
-  ```
-  opkg update; opkg install fping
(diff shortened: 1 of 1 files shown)


frenck:
I'm kinda in split on if we should add this or not. The page is not vendor-specific, but we are now adding some pretty vendor-specific notes to it.

My concern is, that if we start doing this, we could add a dozen other notes as well...

kylegordon:
I agree with your concerns. I do feel the information needs to be offered up somehow though - I spent weeks on and off wondering why OpenWRT presence detection was pitiful!

frenck:
Well, SNMP is full of caveats and vendor-specific things.

frenck:
While I do appreciate you took the time to create this PR, I'm going to close it, because of this reasoning. 👍 Thanks again.

PR #10031, opened 2019-08-01
Specify text language on some code blocks

agsimmons:
**Description:** There were some text blocks, generally containing things like config file data, which did not specify a "language" for the text block. This made the text appear like a single-line quoted string, but across multiple lines. Jekyll formats these differently than Github, so it appears to have been an oversight.

I may have missed some instances of this, but I tried to find them all.

I did not make changes to some areas where I was not sure if it would be the intended style. An example of this is in the ``source/_docs/z-wave/installation.markdown`` file in the ``Random Failures`` section. This section gives an example of an error message with this same formatting. Could I get confirmation on if instances like these should be modified?

There were a few cases I was not sure about the preferred style. I will make comments on these changes for review.

If this formatting is desired, it may be worth specifying it in the documentation standards.

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10031"></a>

### DIFF
--- a/source/_docs/autostart/init.d.markdown
+++ b/source/_docs/autostart/init.d.markdown
@@ -36,6 +36,9 @@ This logrotate script at `/etc/logrotate.d/homeassistant` will create an outage 
 
 ```
 /var/log/homeassistant/home-assistant.log
+```
+
+```text
 {
         rotate 7
         daily
--- a/source/_docs/autostart/init.d.markdown
+++ b/source/_docs/autostart/init.d.markdown
@@ -34,9 +34,7 @@ $ sudo service hass-daemon install
 
 This logrotate script at `/etc/logrotate.d/homeassistant` will create an outage of a few seconds every week at night. If you do not want this add `--log-rotate-days 7` to the `FLAGS` variable in the init script.
 
-```
-/var/log/homeassistant/home-assistant.log
-```
+File `/var/log/homeassistant/home-assistant.log`:
 
 ```text
 {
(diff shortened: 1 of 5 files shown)


source/_docs/ecosystem/scenegen.markdown
```diff
@@ -60,7 +60,7 @@ optional arguments:
 
 For basic operation just supply the url and optionally the api key (using the --key option) on the command line and scenegen will output a list of all lights and switches with their attributes. Optionally use the `--scenename` flag to explicitly set the scenename.
 
-```
+```text
```
agsimmons: Should this be ``bash``, as it contains a bash command?

source/_docs/autostart/init.d.markdown
```diff
@@ -36,6 +36,9 @@ This logrotate script at `/etc/logrotate.d/homeassistant` will create an outage
 
 ```
 /var/log/homeassistant/home-assistant.log
```
agsimmons: Is this formatting valid for specifying the location of the file which should contain the next block of text?

frenck: Adjusted, since it is indeed not part of that file.

source/_docs/ecosystem/scenegen.markdown
frenck: Adjusted (fixed some more occasions).

frenck (approved):
Thanks, @agsimmons! 👍

PR #10050, opened 2019-08-04
--init does not work on some platforms

WallyCZ:
... because this docker image does not contain init process. When this flag is pressent, following error occurs:
exec – failed: No such file or directory

Maybe this is releated to other platforms, but I cannot test it.

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10050"></a>


WallyCZ:
This is related only to libreelec, on other platforms it works probably fine...

PR #10102, opened 2019-08-10
Added bug warning about crashing

DubhAd:
Added bug warning about crashing when using stream and shell_command

**Description:**

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10102"></a>

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10102"></a>

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10102"></a>

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10102"></a>

### DIFF
--- a/source/_components/stream.markdown
+++ b/source/_components/stream.markdown
@@ -11,6 +11,12 @@ ha_qa_scale: internal
 
 The `stream` integration provides a way to proxy live streams through Home Assistant. The integration currently only supports proxying H.264 source streams to the HLS format and requires at least FFmpeg >= 3.2.
 
+<div class='note warning'>
+
+If you are also using the [shell_command component](/components/shell_command) there is currently a [bug in the library](https://github.com/home-assistant/home-assistant/issues/22999) used for `stream` that causes shell commands to crash Home Assistant.
+
+</div>
+
 ## Configuration
 
 To enable this component, add the following lines to your `configuration.yaml` file:


frenck:
We should not add notes about open bugs to our documentation. There is no way to keep that maintained properly. Also, after this one, we would still have 800 issues notes to add to the docs to go.

balloob:
I wonder if we can do something where there is another page/DB with known issues, and we can link them from the system health page. Entries could be as simple as:

 - integration
 - GH issue
 - version start
 - version end
 - closed/open (based on GH issue?)

Or maybe we just link to our device tracker ? Although that may be harder to find.

frenck:
Teaching/guiding people on where to look for open issues is definitely a big plus.
Since we use proper integration tags on issues nowadays, we could add a link that filters the issue list based on the integration that throws the error?

balloob:
well we should have another label with "Legit issue" or something. Because there is a lot of misinformation in our tracker , things caused by other parts of the system misbehaving etc.

frenck:
@balloob In that case, I suggest introducing the "Needs Triage", "Confirmed", "Invalid" labels or similar.

SeanPM5:
Another idea along these same lines is to add links to the integration tags in the sidebar of each integration page on the website. Here is a little mockup.

webmockup

The issue tracker link in this case would go to [/labels/integration/broadlink](https://github.com/home-assistant/home-assistant/labels/integration%3A%20broadlink). That part seems easy enough since it's just a static link, but displaying the issue counts might be more difficult.

I think increasing the visibility of this link would cut down on duplicate bug reports, and it also gives users an idea of how well-maintained an integration is.

DubhAd:
We need ***SOMETHING*** so that the poor sods using HA don't keep tripping over ***KNOWN PROBLEMS***

A _known issues_ page would be good, but assuming folks can find the right search keys for the issues list... is daft

frenck (approved):

frenck (approved):

PR #10119, opened 2019-08-12
Remove initial_state attributes from the examples

derandiunddasbo:
Recently I had an extensive discussion about the pros and cons of using the `initial_state` attribute in automations and in the end I had to admit, that there are good arguments against using it unless absolutely necessary and only if one really knows what he or she is doing.

In the examples above I can't see a good reason for using `initial_state` and thus I'd propose to delete it from both examples.

...and they were inkonsistent in both examples anyways.  ;-) One of them used 'on' and the other used true.

**Description:**

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10119"></a>

### DIFF
--- a/source/_components/frontend.markdown
+++ b/source/_components/frontend.markdown
@@ -86,7 +86,6 @@ Set a theme at the startup of Home Assistant:
 ```yaml
 automation:
   - alias: 'Set theme at startup'
-    initial_state: 'on'
     trigger:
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @derandiunddasbo! 👍

PR #10120, opened 2019-08-12
Add info about customizing entity_picture for a person

JeffLIrion:
**Description:**

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10120"></a>

### DIFF
--- a/source/_components/person.markdown
+++ b/source/_components/person.markdown
@@ -93,8 +93,8 @@ You can customize the picture used for a person entity in the `customize:` secti
 
 ```yaml
 customize:
-  person.ada6789
-    entity_picture: "/local/ada.JPG"
+  person.ada6789:
+    entity_picture: "/local/ada.jpg"
 ```
 
 See the documentation about [hosting files](/components/http/#hosting-files) for more information about the `www` folder. 
--- a/source/_components/person.markdown
+++ b/source/_components/person.markdown
@@ -89,7 +89,7 @@ person:
 
 ### Customizing the picture for a person
 
-You can customize the picture used for a person entity in the `customize:` section of your configuration. For example:
+By following the instructions on the [customizing entities](/docs/configuration/customizing-devices#entity_picture) page, you can customize the picture used for a person entity in the `customize:` section of your configuration. For example:
 
 ```yaml
 customize:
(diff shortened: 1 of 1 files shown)


source/_components/person.markdown
```diff
@@ -86,3 +86,15 @@ person:
...
       - device_tracker.beacon
 ```
+
+### Customizing the picture for a person
+
+You can customize the picture used for a person entity in the `customize:` section of your configuration. For example:
+
+```yaml
+customize:
+  person.ada6789
```
frenck: invalid yaml

source/_components/person.markdown
```diff
@@ -86,3 +86,15 @@ person:
...
+    entity_picture: "/local/ada.JPG"
```
frenck: `.JPG` is a weird file extension, please use lowercase.

frenck (changes requested):
I find this a bit weird to add here.
Every entity can be customized this way. Why add it here? And not on every single page we have?

Furthermore, it is documented here: https://www.home-assistant.io/docs/configuration/customizing-devices/#entity_picture

source/_components/person.markdown
JeffLIrion: Fixed

source/_components/person.markdown
JeffLIrion: Fixed

JeffLIrion:
> I find this a bit weird to add here.
> Every entity can be customized this way. Why add it here? And not on every single page we have?
> 
> Furthermore, it is documented here: https://www.home-assistant.io/docs/configuration/customizing-devices/#entity_picture

My reason for adding it here is because I would guess that this is the most common case where users want to customize the picture. I think `picture` should be a configuration parameter for the person component, but until then users will need to customize the entity in this way. 

Would you rather add the link to the section that you provided instead of adding a section describing how to customize the `entity_picture`? If so, I'll make that change.

frenck:
@JeffLIrion Agreed, however, I do think we should at least link to the customization documentation.

JeffLIrion:
Link added!

frenck (approved):
Thanks, @JeffLIrion! 👍

PR #10193, opened 2019-08-20
Update rest:switch for body_state parameter

quielb:
Update for new parameter body_state.

**Description:**
Update doc for new parameter that sends data in body of get on status update of rest switch

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#26096

## Checklist:

<a href="https://gitpod.io/#https://github.com/home-assistant/home-assistant.io/pull/10193"></a>

### DIFF
--- a/source/_components/switch.rest.markdown
+++ b/source/_components/switch.rest.markdown
@@ -54,9 +54,7 @@ body_off:
 body_state:
   description: "Allow for data in the body of the status get.  Some IoT devices don't return data when the end point is hit.  They require some query in the body of the get.  This value can be a [template](/topics/templating/)."
   required: false
-  type: string
-  default: None
-  
+  type: string  
 is_on_template:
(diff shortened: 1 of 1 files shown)


source/_components/switch.rest.markdown
```diff
@@ -51,6 +51,12 @@ body_off:
   required: false
   type: string
   default: OFF
+body_state:
+  description: "Allow for data in the body of the status get.  Some IoT devices don't return data when the end point is hit.  They require some query in the body of the get.  This value can be a [template](/topics/templating/)."
+  required: false
+  type: string
+  default: None
```
klaasnicolaas: Is there a default in the code?

quielb:
There is no default in the code.  If the option isn't defined then the parameter in the code should have a value of None.  The websession.get allows for the data parameter to be None.

klaasnicolaas:
in that case, I think we can remove the `default` variable

quielb:
removed

quielb:
No need for this.  I found time and have created a platform for the IoT device that I needed this feature for.  I am currently running it as a custom_component.  When I complete the Integration I will submit a PR for inclusion.

PR #10219, opened 2019-08-24
Remove hashes in front of commands

Derkades:
On the website they look like commented lines:
image

### DIFF
--- a/source/_docs/installation/freenas.markdown
+++ b/source/_docs/installation/freenas.markdown
@@ -10,24 +10,24 @@ This has been tested on FreeNAS 11.2 and should also work on FreeBSD 11.x as wel
 Create the user and group that Home Assistant will run as. The user/group ID of `8123` can be replaced if this is already in use in your environment.
 
 ```bash
-# pw groupadd -n homeassistant -g 8123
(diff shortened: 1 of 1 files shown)


fabaff:
> On the website they look like commented lines:

Then the CSS needs to be adjusted. `#` and `$` are used to distinguish between commands which should be run as root and those for users.

Derkades:
It makes sense to display them as comments for bash scripts. They are actually comments

frenck (approved):
Thanks, @Derkades! 👍 

@fabaff These are being removed for many reasons.
1) Copy/paste made easy.
2) Conflicts with Markdown rules MD014.
3) Is requested not to in our own developer docs.
4) Highlighting issues, as `#` in shell scripts are used as comments.
5) If `$` & `#` root/user? TBH if that is the indicator, the documentation around it is incomplete and unclear. Most of our user base will not know the difference.

DubhAd:
@fabaff Looks like our own documentation standards on this were updated by Frenck last month - it's caught quite a few of us out having the standard reversed ;)

frenck:
@DubhAd As explained to you in the other PR as well, this has been applied by me & @klaasnicolaas for quite some time now, and have been actively removing it from many places way longer already.

PR #10221, opened 2019-08-25
Updated Glances instructions

genestealer:
Updated to correct to new Glances API instructions and how to install for Linux and Windows.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/glances.markdown
+++ b/source/_components/glances.markdown
@@ -32,7 +32,7 @@ Glances web server started on http://0.0.0.0:61208/
 Tested on Windows Server 2019, Python 3.7.4 32bit, Glances 3.1.
 
 <div class='note'>
-Note: Glances 3.1.1 exits on running, so downgrade to Glances 3.1
+Note: Glances 3.1.1 exits on running, Glances 3.1 is tested to work. 
 </div>
 
 #### Install Python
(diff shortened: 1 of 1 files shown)


genestealer:
Issues with Windows instructions

klaasnicolaas:
Perhaps also mention that there is an add-on from Glances for Hass.io users

frenck:
I truly think we should not merge this.

While I appreciate the effort, we are not a website that provides documentation on how to install Glances. We should prefer linking to the Glances website in those cases.

The reasoning is that this is hard to maintain, and there are way too many variants. Besides that, this documentation also pins a version, keeping that updated, is most likely not going to happen.

frenck:
This has been discussed on Discord with some of our maintainers, and we agree that this type of documentation does not belong in the Home Assistant docs. Therefore, closing up this PR.

Nevertheless, thank you for contributing, that is greatly appreciated 👍

genestealer:
No problem I understand I also appreciate you taking the time to have a discussion before making a decision. Unfortunately the information I put into the update isn't available on the glances website, they cater very little to the windows environment. I will see whether or not I can update their documentation, in addition to perhaps adding the configuration example in the HA forum.

PR #10243, opened 2019-08-27
Document 'name' config parameter in rejseplanen

KennethLavrsen:
Document the undocumented but fully working name config. This is an optional config but essential to know to define more than one instance of the rejseplanen sensor and getting a predictable result. Otherwise the sensor name defaults to sensor.next_departure_# and the numbers can swap around after an HA restart

**Description:**

## Checklist:

### DIFF
--- a/source/_components/rejseplanen.markdown
+++ b/source/_components/rejseplanen.markdown
@@ -54,7 +54,7 @@ stop_id:
   required: true
   type: string
 name:
-  description: The name of the sensor. Entity ID for the sensor will be created based on this name. Example - Glostrup St becomes sensor.glostrup_st. It is optional but recommended if you define more than one sensor
+  description: The name of the sensor. Entity ID for the sensor will be created based on this name. Example - Glostrup St becomes sensor.glostrup_st. It's optional but recommended if you define more than one sensor.
   required: false
   type: string
   default: next_departure
--- a/source/_components/rejseplanen.markdown
+++ b/source/_components/rejseplanen.markdown
@@ -54,10 +54,10 @@ stop_id:
   required: true
   type: string
 name:
-  description: The name of the sensor. Entity ID for the sensor will be created based on this name. Example - Glostrup St becomes sensor.glostrup_st. It's optional but recommended if you define more than one sensor.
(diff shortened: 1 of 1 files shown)


source/_components/rejseplanen.markdown
```diff
@@ -45,6 +45,7 @@ Add a sensor to your `configuration.yaml` file as shown in the example:
 # Example configuration.yaml entry
 sensor:
   - platform: rejseplanen
+    name: 'Bus Stop Name'
```
klaasnicolaas: Please remove here the `name` variable

klaasnicolaas (changes requested):

source/_components/rejseplanen.markdown
KennethLavrsen: Why?
I added it intentionally to show an good example of how to use most common variables and even if name is implemented as optional in the code it is in my view recommended as the default is pretty useless. Ones you define two sensors they keep on getting auto named next_departure and next_departure_2 and they swap around at each restart of HA which again means that you cannot use these sensors for anything that replies on which is which - 99% of the cases. This name config should be in the example in my opinion. It adds value.

klaasnicolaas: That's our policy. We do not want to saddle users with optional options in the first place, so that users can use the first example as a copy paste. In the advanced example, the name value can remain as usual, that's fine. But please remove it here.

klaasnicolaas: If it is so important to set a name config as default, then this will have to be set to code layer and set the name value as required.

KennethLavrsen: This is the first PR I ever do on github for HA. Actually the first PR I ever do on github in any project.. I edited the patch. I am not sure I did it correctly. I did as you suggested. Let me know if I need to do more to get the edit through

At least the 'name' config gets documented now. I had lots of trouble with the rejseplanen integration and even raised a bug report on it (inconsistent naming of multiple sensors). But then I read the code and saw that the name config was actually there - just not documented. And adding the name to my sensors meant that the rejseplanen feature went from unusable to brilliant. It is important. Changing it to mandatory would be a breaking change. I am not sure it is worth making a breaking change just to comply with a policy. I would not do that if it was my decision.

klaasnicolaas (dismissed):

klaasnicolaas:
Congratz with your first PR on Github 🎉 We will merge it after finish the new build process 👍

fabaff (approved):

PR #10251, opened 2019-08-29
Update arp_ping default value

slackr31337:
**Description:**

Fix arp_ping default value to true. Add HA version 0.98 info.

## Checklist:

### DIFF
--- a/source/_components/mikrotik.markdown
+++ b/source/_components/mikrotik.markdown
@@ -18,7 +18,7 @@ There is currently support for the following device types within Home Assistant:
 
 ## Configuring `mikrotik` hub
 
-You have to enable accessing the RouterOS API on your router to use this platform.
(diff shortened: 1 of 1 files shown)


source/_components/mikrotik.markdown
```diff
@@ -80,7 +80,7 @@ method:
 arp_ping:
   description: Use ARP ping with DHCP method for device scanning.
   required: false
-  default: false
+  default: true
```
klaasnicolaas, line 83: I see [here](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/mikrotik/__init__.py#L52) that it's `False`

frenck:
This PR is incorrect, since indeed, as @klaasnicolaas pointed out, it is `false` by default.

PR #10262, opened 2019-08-31
MQTT Sensor: Warn about synchronization/double trigger issues.

ghost:
**Description:**
Add a warning in the documentation that the state and attributes of a sensor will not change synchronously if they share the same MQTT topic and automations may be triggered twice.  It is not considered a bug, so the issue should be documented.  Reference closed issues  home-assistant/home-assistant#21699, and home-assistant/home-assistant#21082.   (Also happens with force_update: false)

## Checklist:

- [ x] The documentation follows the [standards][standards].

### DIFF
--- a/source/_components/sensor.mqtt.markdown
+++ b/source/_components/sensor.mqtt.markdown
@@ -171,7 +171,7 @@ sensor:
 ```
 {% endraw %}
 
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @eyager1! 👍

PR #10264, opened 2019-08-31
Incorrect entity_id under data:

dtorner:
**Description:**

## Checklist:


frenck:
You are actually changing something, that is incorrect to something that is correct... 🤷‍♂

frenck:
I took a quick peek at how we do this in other `input_*` type integrations, as we use the format as it was already (which is also correct). I think we should stick to the more consistent variant in this case.

Nevertheless, thank you for the PR 👍

PR #10275, opened 2019-09-02
Switch to full commonmark + general upgrades

frenck:
**Description:**

For us, it is not needed to stick on the (outdated) GitHub commonmark variant.
This PR removes `jekyll-commonmark-ghpages` and replaces it with `jekyll-commonmark`, causing a major upgrade of the `commonmarker` gem (depedency).

The newer release of `commonmarker` has now disabled the `UNSAFE` option by default, however, we use quite a bit of HTML in our Markdown, and we need that option. So that is added in.

Futhermore, `highlighter: none` had been added as a site option, to prevent `commonmarker` to interfere with our current site (and behave as in `jekyll-commonmark-ghpages`).

Lastly, `sinatra` had a patch update: `2.0.6` -> `2.0.7`

**Pull request in home-assistant (if applicable):** n/a

## Checklist:


fabaff (approved):
LGTM

PR #10289, opened 2019-09-04
Document mandatory emulated_hue port setting

clach04:
Since 2019 August/September firmware updates for Echo and Echo Dot v1.

## Checklist:

### DIFF
--- a/source/_integrations/emulated_hue.markdown
+++ b/source/_integrations/emulated_hue.markdown
@@ -55,6 +55,9 @@ emulated_hue:
 ```yaml
 # Amazon Echo example configuration.yaml entry
 emulated_hue:
+# listen_port set to 80 is required since 2019-08 when new Firmware for Echo and Echo Dot v1 was pushed out by Amazon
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
We should not document this in the example, but in the actual documentation itself.

clach04:
@frenck are these changes what you had in mind?

There is a discussion thread on this at https://community.home-assistant.io/t/emulated-hue-suddenly-stopped-working-with-echo-dot-v2/133956/59 - an alternative discussed there is to use ip tables to remap ports (but the discovery service will still report the default 8300):

    sudo iptables -t nat -A PREROUTING -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 8300

I choose not to document that as the configuration would be inconsistent and then not work with well behaved clients!

frenck:
I was just suggesting to remove the instructions from the YAML and add it to the documentation (instead of YAML comments).

frenck:
Rebased onto `current` branch, resolved merge conflicts.

frenck (approved):

PR #10292, opened 2019-09-04
Set example MQTT config keys to lowercase

derkmdt:
**Description:**

## Checklist:


frenck:
@derkmdt Could you please explain why?

This is not a YAML configuration, but an explanation for the adding an integration dialog in the frontend...

derkmdt:
Oeps, was using this for the config and it didn’t work. So thought I push a commit 🤭 But it’s wrong as you also say.

PR #10308, opened 2019-09-08
Added 'availability_template' to Template Sensor platform

grillp:
Added documentation `availability_template` configuration option for Template Sensor platform components.

https://github.com/home-assistant/home-assistant/pull/26516

## Checklist:

### DIFF
--- a/source/_components/template.markdown
+++ b/source/_components/template.markdown
@@ -28,10 +28,6 @@ sensor:
         friendly_name: "Sun angle"
         unit_of_measurement: 'degrees'
         value_template: "{{ state_attr('sun.sun', 'elevation') }}"
-        availability_template: >-
-          {%- if not is_state('dependant_device.state', 'unavailable') %}
-            true
-          {% endif %}
 
       sunrise:
         value_template: "{{ state_attr('sun.sun', 'next_rising') }}"
--- a/source/_components/template.markdown
+++ b/source/_components/template.markdown
@@ -18,6 +18,7 @@ The `template` platform supports sensors which get their values from other entit
 The configuration of Template Sensors depends on what you want them to be. Adding the following to your `configuration.yaml` file will create two sensors, one for the current sun angle and one for the time of the next sunrise:
 
 {% raw %}
+
 ```yaml
 # Example configuration.yaml entry
 sensor:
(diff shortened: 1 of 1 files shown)


source/_components/template.markdown
```diff
@@ -27,10 +28,15 @@ sensor:
         friendly_name: "Sun angle"
         unit_of_measurement: 'degrees'
         value_template: "{{ state_attr('sun.sun', 'elevation') }}"
+        availability_template: >-
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_components/template.markdown
```diff
@@ -68,15 +74,11 @@ sensor:
         description: Defines a template for the entity picture of the sensor.
         required: false
         type: template
-      attribute_templates:
-        description: Defines templates for attributes of the sensor.
+      availability_template:
+        description: Defines a template to get the `available` state of the component. If the template returns `true` the device is `available`. If the template returns any other value, the device will be `unavailable`. If `availability_template` is not configured, the component will always be `available`
```
frenck:
```suggestion
        description: Defines a template to get the `available` state of the component. If the template returns `true`, the device is `available`. If the template returns any other value, the device will be `unavailable`. If `availability_template` is not configured, the component will always be `available`.
```

frenck (changes requested):
Hi there @grillp, I've reviewed your PR and left some comments, could you please take a look? Thanks! 👍

source/_components/template.markdown
```diff
@@ -253,38 +264,7 @@ sensor:
         value_template: "{{ states('sensor.power_consumption') }}"
         unit_of_measurement: 'kW'
 ```
-{% endraw %}
-
-### Add Custom Attributes
```
frenck, line 258: I guess this was removed by accident as well...

frenck (changes requested):

source/_components/template.markdown
grillp: Will fix

grillp: Completed..

frenck (approved):

PR #10310, opened 2019-09-08
Added 'availability_template' to all Template Switch platform

grillp:
Added documentation `availability_template` configuration option for all Template Switch platform components.

https://github.com/home-assistant/home-assistant/pull/26513

## Checklist:

### DIFF
--- a/source/_integrations/switch.template.markdown
+++ b/source/_integrations/switch.template.markdown
@@ -59,7 +59,7 @@ switch:
         required: true
         type: template
       availability_template:
-        description: Defines a template to get the `available` state of the component. If the template returns `true` the device is `available`. If the template returns any other value, the device will be `unavailable`. If `availability_template` is not configured, the component will always be `available`
+        description: Defines a template to get the `available` state of the component. If the template returns `true`, the device is `available`. If the template returns any other value, the device will be `unavailable`. If `availability_template` is not configured, the component will always be `available`.
         required: false
         type: template
         default: true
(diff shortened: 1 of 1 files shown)


source/_components/switch.template.markdown
```diff
@@ -35,7 +36,12 @@ switch:
           service: switch.turn_off
           data:
             entity_id: switch.skylight_close
+        availability_template: >-
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_components/switch.template.markdown
```diff
@@ -56,6 +62,11 @@ switch:
         description: Defines a template to set the state of the switch.
         required: true
         type: template
+      availability_template:
+        description: Defines a template to get the `available` state of the component. If the template returns `true` the device is `available`. If the template returns any other value, the device will be `unavailable`. If `availability_template` is not configured, the component will always be `available`
```
frenck:
```suggestion
        description: Defines a template to get the `available` state of the component. If the template returns `true`, the device is `available`. If the template returns any other value, the device will be `unavailable`. If `availability_template` is not configured, the component will always be `available`.
```

frenck (changes requested):
Same! I've reviewed it and left some comments @grillp. If you could adjust a bit, that would be great! 🎖

frenck (dismissed):

PR #10314, opened 2019-09-08
Added 'availability_template' to Template Cover platform

grillp:
Added documentation `availability_template` configuration option for Template Cover platform components.

https://github.com/home-assistant/home-assistant/pull/26509

## Checklist:

### DIFF
--- a/source/_components/cover.template.markdown
+++ b/source/_components/cover.template.markdown
@@ -19,6 +19,7 @@ To enable Template Covers in your installation,
 add the following to your `configuration.yaml` file:
 
 {% raw %}
+
 ```yaml
 # Example configuration.yaml entry
 cover:
@@ -33,7 +34,12 @@ cover:
           service: script.close_garage_door
         stop_cover:
           service: script.stop_garage_door
+        availability_template: >-
+          {%- if not is_state('dependant_device.state', 'unavailable') %}
+            true
+          {% endif %}
 ```
+
 {% endraw %}
 
 {% configuration %}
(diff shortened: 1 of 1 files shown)


source/_components/cover.template.markdown
```diff
@@ -33,7 +34,12 @@ cover:
           service: script.close_garage_door
         stop_cover:
           service: script.stop_garage_door
+        availability_template: >-
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_components/cover.template.markdown
```diff
@@ -62,6 +68,11 @@ cover:
         description: Defines a template to specify which icon to use.
         required: false
         type: template
+      availability_template:
+        description: Defines a template to get the `available` state of the component. If the template returns `true` the device is `available`. If the template returns any other value, the device will be `unavailable`. If `availability_template` is not configured, the component will always be `available`
```
frenck:
```suggestion
        description: Defines a template to get the `available` state of the component. If the template returns `true`, the device is `available`. If the template returns any other value, the device will be `unavailable`. If `availability_template` is not configured, the component will always be `available`.
```

frenck (changes requested):
Last one, @grillp! 👍

frenck (approved):

PR #10316, opened 2019-09-08
Corrected example and scan interval info

Destix:
slave: 1 option in platform configuration causes an error on 0.98.2 
added scan_interval: info
also it seems to me ha_iot_class: should be Local Pull because Modbus all about Pull data from slaves by master

**Description:**

## Checklist:

### DIFF
--- a/source/_components/switch.modbus.markdown
+++ b/source/_components/switch.modbus.markdown
@@ -18,7 +18,6 @@ To use your Modbus switches in your installation, add the following to your `con
 # Example configuration.yaml entry
 switch:
   platform: modbus
-  scan_interval: 10
   coils:
     - name: Switch1
       hub: hub1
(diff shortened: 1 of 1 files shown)


source/_components/switch.modbus.markdown
```diff
@@ -18,7 +18,7 @@ To use your Modbus switches in your installation, add the following to your `con
 # Example configuration.yaml entry
 switch:
   platform: modbus
-  slave: 1
+  scan_interval: 10
```
fabaff: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps users to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

fabaff (dismissed):

frenck (approved):
Thanks for the PR & the changes after the review! 👍 
Good to  merge 🎉

PR #10342, opened 2019-09-11
Plex config changes

jjlawren:
**Description:**
* Default value of `host` removed
* New `server` config option added
* Some descriptions updated or fixed to reflect current behavior

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#26583

## Checklist:

### DIFF
--- a/source/_components/plex.markdown
+++ b/source/_components/plex.markdown
@@ -36,8 +36,17 @@ If your server enforces SSL connections, write "`on`" or "`true`" in the _"Use S
 You can also enable the plex platform directly by adding the following lines to your `configuration.yaml`:
 
 ```yaml
-# Example configuration.yaml entry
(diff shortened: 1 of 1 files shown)


MartinHjelmare (dismissed):

jjlawren:
@klaasnicolaas, the YAML example was edited on purpose to show the new minimal config. Either `host` or `token` must be included and an empty config will now error on startup. Only providing `token` is what I'd expect to be the simplest way to configure the component for most users.

klaasnicolaas:
I saw in the code that both were optional in the list, which is why I removed it. But given your explanation, it is better to add this. You could also put 2 examples with both options 😉

frenck (approved):
LGTM! 👍

PR #10347, opened 2019-09-12
username-host field - clarify use of '%'

KaiboshOz:
**Description:**
Clarify that '%' is used in host field when wanting to connect *from* multiple hosts

## Checklist:

### DIFF
--- a/source/_addons/mariadb.markdown
+++ b/source/_addons/mariadb.markdown
@@ -41,7 +41,7 @@ logins:
       required: true
       type: string
     host:
-      description: Host for account. If you need to connect to this account from multiple hosts, use '%'.
+      description: Host for account. Use '%', to accept connections for this account from any host.
       required: true
       type: string
     password:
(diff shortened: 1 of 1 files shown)


frenck:
Actually, both the original & the changed version are incorrect.
`%` indicates MySQL to accept any host. Configuring multiple hosts in MySQL is also possible and works different.

frenck (changes requested):
Actually I was trying to tell that this should be changed in general. 
Using `%` is just bad practice in general, which we should not instruct people to use IMHO.

KaiboshOz:
I was only prompted to submit the doco update after needing to connect to the database from my desktop, and spending a fair whack of time searching and ending up following a bunch of red herrings before finally landing on a suggestion to use `%` which worked and figuring that it's something that should be in the addon doc as at least some other people are likely to need to do the same thing.

Happy to close this if others disagree, or to modify the update - perhaps a warning to use `%` with caution; or @frenck if you can point me in the right direction regarding the specific formatting for the addon config - as I'm ashamed to say when it comes to database, I'm used to GUIs and interfaces and the MariaDB doc wasn't much help with marrying up the actual DB config with the addon config for me to limit it to accept specific hosts while still allowing the HA/Docker connection.

frenck (approved):

PR #10351, opened 2019-09-12
Update bluetooth_le_tracker.markdown

bnounours:
set track_new_devices default to false, as in the version 0.97.2 it is false.
I had checked the code to understand why nothing was detected and the conclusion is the default value is false

**Description:**

## Checklist:


frenck (approved):
Thanks, @bnounours! Well spotted 👀 👍

PR #10367, opened 2019-09-14
Include port number in 'adb connect' command

dwinnnspeaks:
the command `adb connect <ipaddress>` doesn't work unless port is specified. The command `adb connect<ipaddress>:5555` works.

**Description:**

## Checklist:


frenck (commented):
Uh, yeah it does at my end. Also according to ADB's documentation.
The documentation examples of ADB do not even use a port.

Could you please point me to a resource confirming this?

dshokouhi:
@frenck I think this only impacts some install types, or maybe certain adb versions.  My Ubuntu install does not care but Windows does.  Still I don't think it hurts to add it to cover all cases.

image

frenck (approved):
Hmm, interesting, well, I guess there is no harm to put it in.
Thanks @dwinnnspeaks! 👍

PR #10380, opened 2019-09-16
Modernize Huawei LTE, add UI config, unauthenticated mode, switch

scop:
**Description:**

See main HA PR for details.

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#26675 and home-assistant/home-assistant#28188 and home-assistant/home-assistant#28214

## Checklist:

### DIFF
--- a/source/_integrations/huawei_lte.markdown
+++ b/source/_integrations/huawei_lte.markdown
@@ -74,8 +74,6 @@ To enable the component, add the following lines to your
 # Example configuration.yaml entry
 huawei_lte:
   - url: http://192.168.100.1/
-    username: YOUR_USERNAME
-    password: YOUR_PASSWORD
 ```
 For routers configured via the UI, each configuration item for the
 same router in YAML overrides and updates the values set in UI
--- a/source/_integrations/huawei_lte.markdown
+++ b/source/_integrations/huawei_lte.markdown
@@ -26,19 +26,17 @@ or using YAML.
 
 The integration can be run with or without
 authentication. Authenticated mode enables all available integration
-features, but may interfere with accessing the device web interface
-from another source such as a browser while the integration is active,
-or vice versa. The exact list of features requiring authentication to
-work varies by device and firmware version. The integration will try
-to use all configured ones and fail gracefully if it detects one
-requiring authentication in unauthenticated mode.
+features and entities, but may interfere with accessing the device web
+interface from another source such as a browser while the integration
+is active, or vice versa. The exact list of features requiring
+authentication to work varies by device and firmware version. The
+integration will try to use all configured ones and fail gracefully if
+it detects one requiring authentication in unauthenticated mode.
 
-Newly discovered entities from this integration are disabled by
-default. This is because the target provide for a large amount of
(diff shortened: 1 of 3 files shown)


fabaff (dismissed):

source/_integrations/huawei_lte.markdown
```diff
@@ -1,161 +1,122 @@
...
+`configuration.yaml` file:
 
 ```yaml
-notify:
-  - platform: huawei_lte
-    recipient: "+15105550123"
+huawei_lte:
+  - url: http://192.168.100.1/
+    username: YOUR_USERNAME
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_integrations/huawei_lte.markdown
```diff
@@ -1,161 +1,122 @@
...
+    password: YOUR_PASSWORD
```
frenck: Same as above

source/_integrations/huawei_lte.markdown
```diff
@@ -1,161 +1,122 @@
...
+    notify:
```
frenck: Same as above

frenck (changes requested):
Thanks for the PR @scop! Left a small comment, could you take a look? Thanks! 👍

source/_integrations/huawei_lte.markdown
scop: Removed.

source/_integrations/huawei_lte.markdown
scop: While username and password are really optional, in practice I believe the vast majority of people are going to be using authenticated mode, i.e. with username + password. Without them, in unauthenticated mode, many useful integration features are disabled.

So I left username and password there for now, but I don't feel that strongly about this. If you still think they should be removed, please feel free to go ahead and remove or let me know and I'll take care of it.

source/_integrations/huawei_lte.markdown
scop: Same as above too :)

source/_integrations/huawei_lte.markdown
frenck: I see your point of view, but in general, we follow these rules to get a more consistent experience in the documentation. IMHO, it makes no sense to accept it in one place and deny it in another.

scop: Ok, removed.

scop:
With home-assistant/home-assistant#28214 just merged as well, I've merged #10992 here as well, hope you don't mind.

frenck (approved):

scop:
Wondering if the netlify preview check is stuck or something...?

frenck:
@scop No, it is not 😉 

Old and new Netlify profiles on this PR still.

scop:
Ah, just inpatient then :)  Thanks all!

PR #10387, opened 2019-09-18
Fix incorrect warning

cogneato:
ACL is not required if using HA users and no anonymous access.

**Description:**

## Checklist:

### DIFF
--- a/source/_addons/mosquitto.markdown
+++ b/source/_addons/mosquitto.markdown
@@ -23,7 +23,7 @@ Set up [Mosquitto](https://mosquitto.org/) as MQTT broker.
 
 <div class='warning note'>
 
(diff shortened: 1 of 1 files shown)


fabaff (approved):

wsoyka:
@cogneato Sorry for the necro, but what are "legacy users"? This term is not specified further in the file and doesn't seem to be standard HA lingo.

From what I understand legacy users could mean non-HA users, i.E. users that are added to the mqtt addon's config via the logins array, is that correct?

frenck:
@wsoyka, please do not comment on older and closed PR's. Instead, raise a new issue in case you think there is a problem, or reach out on Discord in case you are in need for support.

PR #10399, opened 2019-09-20
Newer Alexa devices require port 80

scotthibbs:
Added a note because the new Alexa devices add each entity as a light by default resulting in "value is out of range for device..." and the fix for it found in the forums.

**Description:**

## Checklist:

### DIFF
--- a/source/_components/emulated_hue.markdown
+++ b/source/_components/emulated_hue.markdown
@@ -65,7 +65,7 @@ host_ip:
   required: false
   type: string
 listen_port:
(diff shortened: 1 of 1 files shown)


scotthibbs:
Overnight, Alexa changed all the switches I added manually BACK to lights to get the "That value is out of range for X" error again. 

At this time it would be best to take out the fix I described from the forums. However, it is still true Alexa now needs port 80.

frenck (approved):
Thanks, @scotthibbs! 👍

PR #10402, opened 2019-09-20
correct templating example

Mariusthvdb:
Current example shows the same preferred template twice, without the template that would be the one to substitute....

**Description:**

## Checklist:


fabaff (dismissed):
Using `is_state()` is preferred over `states.switch.source.state` -> https://developers.home-assistant.io/docs/en/documentation_standards.html#templates

Mariusthvdb:
Of course, that is what I am trying to make clear. As it stands 2 exact same  templates of the preferred format are shown.

Changing the first in the 'unpreferred' format should make that clear to the reader

frenck (approved):
Thanks, @Mariusthvdb! That is actually a correct fix.

PR #10456, opened 2019-09-25
Mention ongoing broken-ness of plex play_media service

pjv:
... which has not worked since back in the early 0.8x days

### DIFF
--- a/source/_components/plex.markdown
+++ b/source/_components/plex.markdown
@@ -170,6 +170,7 @@ Plays a song, playlist, TV episode, or video on a connected client.
 
   If this occurs, check the setting `Server`>`Network`>`Secure connections` on your Plex Media Server: if it is set to `Preferred` or `Required`, you may need to manually set the `ssl` and `verify` booleans to, respectively, `true` and `false`.
 * Movies must be located under 'Movies' section in the Plex library to properly get 'playing' state.
+* The `play_media` service is broken and [has been broken for a long time](https://github.com/home-assistant/home-assistant/issues/19623).
 
 ## Sensor
 


frenck:
We actually do not log issues in our documentation. We use two things for this:

- GitHub issues
- https://alerts.home-assistant.io

Nevertheless, thank you for your contribution! 👍

pjv:
@frenck That would make sense if it was a transient issue that was being actively worked on. In this case, the page has, for months, erroneously listed a service (`play_media`) which TOTALLY DOES NOT WORK AT ALL for this component. For anyone. Anywhere. 

It's not a transient issue. There is no active issue that any developers are paying attention to. It's not the case that it works for some people and not for others. It totally does not work and can not work.

I guess another way to deal with it would be to delete the section of the page that documents this non-functional service, but it can't be a proper solution to have a whole section of the documentation for this component telling people that it can do something that it absolutely, totally, CANNOT do.

There are people, myself among them, who are spending HOURS following this incorrect documentation and trying to get something to work which CANNOT work. That is bad. Making a note on the documentation to save people from wasting so much time and suffering so much frustration would be good. 

If you are deciding that you prefer to hold to some principle rather than do a good thing, then I'm afraid that you too are bad.

dshokouhi:
@pjv what is wrong with submitting an alert? It does precisely what you wish to achieve. Modify the docs page alerting the users about the issue.  Maybe not in your exact words but it does mention it on this page.

frenck:
> That would make sense if it was a transient issue that was being actively worked on.

I'm sorry to hear that, however, I don't see how that is relevant for this PR/

> In this case, the page has, for months, erroneously listed a service (play_media) which TOTALLY DOES NOT WORK AT ALL for this component. For anyone. Anywhere.

That is not erroneously if it is supposed to work: it is a bug.

> It's not a transient issue. There is no active issue that any developers are paying attention to.

This is an open-source project which largely relies on contributions that people make in their spare time. We cannot demand or supply active developers from a shelve. However, feel free to provide a solution if you can 

> It's not the case that it works for some people and not for others. It totally does not work and can not work.

Well, that is what alerts are for: https://alerts.home-assistant.io, which is much more effective, was designed for it, shows in the documentation (and will soon be visible inside Home Assistant itself as well).

> I guess another way to deal with it would be to delete the section of the page that documents this non-functional service, but it can't be a proper solution to have a whole section of the documentation for this component telling people that it can do something that it absolutely, totally, CANNOT do.

No. As pointed out already.

> There are people, myself among them, who are spending HOURS following this incorrect documentation and trying to get something to work which CANNOT work. That is bad. Making a note on the documentation to save people from wasting so much time and suffering so much frustration would be good.

Sorry to hear that, I agree that is not fun to deal with; and even more reason to issue an alert maybe?

> If you are deciding that you prefer to hold to some principle rather than do a good thing

I think the good thing is to guide you to the appropriate channels, which I did. Adding notes about a bug to the documentation is hard to maintain. We have almost 800 issues right now, adding 800 notes to the documentation manually is not manageable. Furthermore, they have to be removed again if a bug is fixed.

Hence, we have other channels, like the Home Assistant Alerts, to handle that so we can focus, be informative, and keep thing maintainable (which result in everybody being able to spend time on things that matter).

>  then I'm afraid that you too are bad.

That is just rude and totally uncalled for sir. 👎

PR #10465, opened 2019-09-27
Added link to Material Design Icons page

milothomas:
Added link to Material Design Icons page under the Additional Lovelace Resources

**Description:**

## Checklist:

### DIFF
--- a/source/lovelace/index.markdown
+++ b/source/lovelace/index.markdown
@@ -31,4 +31,4 @@ To try it yourself, check out [the demo](https://demo.home-assistant.io).
 
 * [Awesome HA](https://www.awesome-ha.com/#lovelace-ui)
 * [Community Custom Cards](https://github.com/custom-cards)
-* [Material Design Icons](https://cdn.materialdesignicons.com/4.4.95/)
+* [Material Design Icons](https://materialdesignicons.com/tag/community)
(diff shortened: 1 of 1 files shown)


source/lovelace/index.markdown
```diff
@@ -31,3 +31,4 @@ To try it yourself, check out [the demo](https://demo.home-assistant.io).
 
 * [Awesome HA](https://www.awesome-ha.com/#lovelace-ui)
 * [Community Custom Cards](https://github.com/custom-cards)
+* [Material Design Icons](https://cdn.materialdesignicons.com/4.4.95/)
```
frenck: Maybe we should avoid having the version in the URL? Since that means we need to keep it updated in the future.
How about linking it to just: <https://materialdesignicons.com/> ?

frenck (changes requested):

source/lovelace/index.markdown
milothomas: That makes sense, how about showing the community by default since those are the ones we can use in HA.
[https://materialdesignicons.com/tag/community](https://materialdesignicons.com/tag/community)

frenck (approved):
Thanks, @milothomas! 👍

PR #10478, opened 2019-09-29
Adding proto to xiaomi_aqara configuration.

sramanujan:
**Description:**
Addition an optional configuration key `proto` to let the module know how to communicate with the gateway for 2.0/3.0 protocol devices (like air conditioning companion)

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#27042

## Checklist:

- [ y] Branch: `next` is for changes and new documentation that will go public with the next Home Assistant release. Fixes, changes and adjustments for the current release should be created against `current`.
- [ y] The documentation follows the [standards][standards].

### DIFF
--- a/source/_components/xiaomi_aqara.markdown
+++ b/source/_components/xiaomi_aqara.markdown
@@ -112,6 +112,11 @@ gateways:
       required: false
       type: boolean
       default: false
+    proto:
+      description: Protocol to use for the HA <-> gateway communications. This is only required if the gateway doesn't use 1.0 protocol. 
+      required: false
+      type: string
+      default: 1.0
 discovery_retry:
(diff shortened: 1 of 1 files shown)


rytilahti (changes requested):
Please add a reference to some information about available protocol versions. Now it's unclear how users should know which version to use.

sramanujan:
> Please add a reference to some information about available protocol versions. Now it's unclear how users should know which version to use.

I'm not very clear on this aspect as well. I only know that my Xiaomi Gateway (acpartner.v3) does not work with the code in this module which otherwise works for `proto == '1.0'`. Can we get to know who wrote the code which checks for `proto` in xiaomi_gateway to help us out here?

PR #10513, opened 2019-10-01
Update transmission.markdown

joshmckinney:
Added default new sensors and removed the no longer valid "monitored_conditions".
[__init__py reference](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/transmission/__init__.py)
See TransmissionData class,

**Description:**

## Checklist:

### DIFF
--- a/source/_components/transmission.markdown
+++ b/source/_components/transmission.markdown
@@ -71,6 +71,7 @@ scan_interval:
 ## Default Sensors
 
 Default sensors are now added with the following schema "sensor.transmission(or name if used)_variable_here"
+
 - sensor.transmission_active_torrents
 - sensor.transmission_completed_torrents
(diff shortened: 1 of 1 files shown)


frenck:
This PR is actually incorrect, while the referenced code indeed has it removed, it has not been released yet.
So our current documentation still matches the current stable release. There is a change for the documentation pending for the upcoming release, which makes this PR obsolete.

For more information, see: home-assistant/home-assistant#26434

Nevertheless, thank you for your contribution 👍

joshmckinney:
My apologies looks as if the confusion came down to pulling the latest docker image and not the stable.

frenck:
@joshmckinney No need to apologize, things happen 👍

PR #10519, opened 2019-10-01
Accessibility improvements to home-assistant.io

voxic:
Did an Accessibility audit of home-assistant.io front page.
Did some improvements
<html> element does not have a [lang] attribute - FIX
If a page doesn't specify a lang attribute, a screen reader assumes that the page is in the default language that the user chose when setting up the screen reader. If the page isn't actually in the default language, then the screen reader might not announce the page's text correctly.

**Description:**

## Checklist:


klaasnicolaas:
You need to perform a rebase to the `current` branch

PR #10581, opened 2019-10-03
Update MQTT Device Tracker mapping payloads to zones/states

raman325:
**Description:**
I currently use the SmartThings MQTT bridge featured on the blog here to publish state changes from my SmartThings connected devices to HomeAssistant. I am currently unable to create an MQTT Device Tracker in HA that can connect to a SmartThings presence sensor because the bridge publishes the states of "present" and "not present", but HA expects "home" and "not_home".

As such, I have added a new optional map parameter to the component that will allow users to map topic payloads to states that HA would expect.

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#27143

## Checklist:

### DIFF
--- a/source/_integrations/climate.mqtt.markdown
+++ b/source/_integrations/climate.mqtt.markdown
@@ -123,7 +123,6 @@ modes:
   type: list
 action_topic:
   description: The MQTT topic on which to listen for the current action state of the HVAC. Expects `idle`, `cooling`, `heating`, `drying`, or `off`.
-_components/climate.mqtt.markdown
   required: false
   type: string
 action_template:
--- a/source/_integrations/climate.mqtt.markdown
+++ b/source/_integrations/climate.mqtt.markdown
@@ -123,6 +123,7 @@ modes:
   type: list
 action_topic:
   description: The MQTT topic on which to listen for the current action state of the HVAC. Expects `idle`, `cooling`, `heating`, `drying`, or `off`.
+_components/climate.mqtt.markdown
   required: false
   type: string
 action_template:
--- a/source/_integrations/climate.mqtt.markdown
+++ b/source/_integrations/climate.mqtt.markdown
@@ -121,14 +121,6 @@ modes:
   required: false
   default: ['auto', 'off', 'cool', 'heat', 'dry', 'fan_only']
   type: list
-action_topic:
-  description: The MQTT topic on which to listen for the current action state of the HVAC. Expects `idle`, `cooling`, `heating`, `drying`, or `off`.
-  required: false
-  type: string
-action_template:
-  description: A template to render the value received on the `action_topic` with.
-  requred: false
-  type: template
 temperature_command_topic:
   description: The MQTT topic to publish commands to change the target temperature.
   required: false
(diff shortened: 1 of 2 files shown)


source/_components/device_tracker.mqtt.markdown
```diff
@@ -22,17 +22,36 @@ device_tracker:
     devices:
       paulus_oneplus: 'location/paulus'
       annetherese_n4: 'location/annetherese'
+    payload_home: present
```
klaasnicolaas: In the first example we want only required config variables, but feel free to add an full configuration example at the bottom 😉

klaasnicolaas (changes requested):

source/_components/device_tracker.mqtt.markdown
raman325: done, thanks!

source/_components/device_tracker.mqtt.markdown
```diff
@@ -26,15 +26,45 @@ device_tracker:
 
 {% configuration %}
 devices:
-  description: List of devices with their topic.
+  description: A dictionary where the keys are devices and the values are their corresponding topic.
   required: true
-  type: list
+  type: map
```
klaasnicolaas: I think this must be a list

source/_components/device_tracker.mqtt.markdown
```diff
@@ -26,15 +26,45 @@ device_tracker:
...
+  default: `home`
+payload_not_home:
+  description: The payload that represents the not home state for the device.
+  required: false
+  type: string
+  default: `not_home`
+payload_custom:
+  description: A dictionary to map payload values to custom zones (e.g. not `home` and `not_home`). The keys are the payloads and the values are the corresponding locations that the payload represents.
+  required: false
+  type: string
```
klaasnicolaas: and maybe this one also

source/_components/device_tracker.mqtt.markdown
raman325: the example doesn't show it as a list, nor does the code:
`vol.Required(CONF_DEVICES): {cv.string: mqtt.valid_subscribe_topic}`

I can change both, but I was following the original convention

source/_components/device_tracker.mqtt.markdown
raman325: see my other comment

raman325: oops I committed a change because I thought I made a mistake, did not see that you had changed this. Let me know if you would like me to change it back, but both this and `devices` look like maps to me

raman325:
I am not clear on what character the build error is referring to

source/_components/device_tracker.mqtt.markdown
raman325: Actually I looked at some examples and now I understand. Will change accordingly

frenck:
This PR needs to be rebased, however, there are a couple of merge commits in here as well. That makes it a bit hard. I'm going to try to fix this, but possibly, I'll mess up. If that is the case, I'll open up a new PR 😉

frenck (approved):
Thanks for the PR @raman325! LGTM

Happy Hacktoberfest 🎉

PR #10593, opened 2019-10-03
Fix links to `device_class` for MQTT sensor

emontnemery:
**Description:**
Fix links to `device_class`

## Checklist:

### DIFF
--- a/source/_integrations/sensor.mqtt.markdown
+++ b/source/_integrations/sensor.mqtt.markdown
@@ -89,9 +89,9 @@ unique_id:
   required: false
   type: string
 device_class:
-  description: The type/class of the sensor to set the icon in the frontend.
+  description: The [type/class](/integrations/sensor/#device-class) of the sensor to set the icon in the frontend.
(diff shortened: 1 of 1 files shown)


emontnemery:
The link generated by `type: device_class` is a bit misleading, could some magic be done so we could specify which list of device classes to link to?

frenck:
Hmmm... ok we definitely should improve on this, however, I think we should avoid temporary changing it to the string type.

If we change it now, they will be harder to find if we have a real fix for it.

emontnemery:
@frenck OK, makes sense, I reverted to `device_class`

Btw, and unrelated to this PR, the MQTT docs is really hard to read due to so many options.
Do you think it would make sense to simply sort the options alphabetically?

frenck (approved):
Thanks, @emontnemery!

Happy Hacktoberfest 🎉

frenck:
About the additional comment, I like it!
Will hook up on Discord about it.

PR #10597, opened 2019-10-04
Update twilio_sms.markdown - Reflect lowercase requirements in documentation.

danwetherald:
**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/twilio_sms.markdown
+++ b/source/_integrations/twilio_sms.markdown
@@ -16,7 +16,7 @@ To use this notification platform in your installation, add the following to you
 ```yaml
 # Example configuration.yaml entry
 notify:
-  - name: NOTIFIER_NAME
+  - name: notifier_name
     platform: twilio_sms
     from_number: E164_PHONE_NUMBER or SENDER_ID
 ```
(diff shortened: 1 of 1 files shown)


frenck:
Hi there @dan003400!

The capitalized names in our documentation indicate placeholders that need to be replaced by your own.

This change breaks that standard, and therefore I cannot merge this in. Nevertheless, thank you for taking the time to create a contribution! 👍 

Happy Hacktoberfest 🎉

danwetherald:
That makes sense, but this was very frustrating for my self and others. I believe this is a bad standard that can cause confusion and lead users to make their own custom variables in all caps. 

https://community.home-assistant.io/t/0-86-notify-problem/94074/12

The error received when all caps are used is no where near helpful, as there is no error when loading the entity but later on causes an error when referencing that entity.

Is there somewhere in the getting started that explains why the examples use all caps when that is actually not how you should name you'r entities?

Thanks,

Dan

frenck:
The following things are not allowed in entity IDs:

- No capital letters
- No leading or ending underscores
- No double underscores

I do not know a direct place where this is described at this moment. However, we are happy to accept contributions to improve our documentation.

PR #10604, opened 2019-10-04
Clarify how to connect to Elk over serial port

colohan:
* Update docs to use the default Elk Baud rate of 115200 for serial connections.
* Use /dev/ttyUSB0, which is the serial device used by USB-to-serial converters by default.
* Don't use a username and password over a serial connection, as Elk doesn't require them over a hardwired serial connection.
* Give example of liberal use of exclusions, showing users how to avoid Entity registry spam caused by Elk.

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/elkm1.markdown
+++ b/source/_integrations/elkm1.markdown
@@ -43,7 +43,7 @@ elkm1:
 
 {% configuration %}
 host:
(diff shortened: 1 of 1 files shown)


frenck (approved):
Sorry for the long wait @colohan!

It looks great, thank you very much! 👍

PR #10609, opened 2019-10-04
Update index.markdown

inverse:
Since ui-schema is deprecated this would probably be a better fit.


frenck (approved):
Thanks, @inverse! That is helpful! 👍 

Happy Hacktoberfest 🎉

PR #10612, opened 2019-10-05
Update link to speedtestservers

LEJOUI:
The link to the speedtestservers seems to be broken. The new link does work. Also the server_id seems to be mandatory at this point. So this sets the server_id to be required and in the example configuration.yaml inserts an example server_id.

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/speedtestdotnet.markdown
+++ b/source/_integrations/speedtestdotnet.markdown
@@ -43,7 +43,7 @@ monitored_conditions:
       description: "The upload speed (Mbit/s)."
 server_id:
   description: Specify the speed test server to perform the test against.
-  required: true
+  required: false
   type: integer
 scan_interval:
   description: "Minimum time interval between updates. Supported formats: `scan_interval: 'HH:MM:SS'`, `scan_interval: 'HH:MM'` and Time period dictionary (see example below)."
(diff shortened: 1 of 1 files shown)


source/_integrations/speedtestdotnet.markdown
```diff
@@ -43,7 +43,7 @@ monitored_conditions:
       description: "The upload speed (Mbit/s)."
 server_id:
   description: Specify the speed test server to perform the test against.
-  required: false
+  required: true
```
frenck: This option is not required.

Code reference:
<https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/speedtestdotnet/__init__.py#L26>

frenck (changes requested):
Hi there @LEJOUI, thanks for opening a PR!

Welcome to the Home Assistant Community 🎉 

I've reviewed it and left a comment. Could you please take a look? Awesome! 👍

source/_integrations/speedtestdotnet.markdown
LEJOUI: The code indeed states it's not required. I'll create an issue about this instead. Speedtest.net is somehow not working over here without server_id

frenck: That is fine, but it still means this change has to be reverted right now.

klaasnicolaas (approved):

frenck:
Thanks, @LEJOUI! 👍 

Happy Hacktoberfest 🎉

frenck (approved):

PR #10615, opened 2019-10-05
Added alternative to using now() as a trigger

kloggy:
Based on this forum thread https://community.home-assistant.io/t/automation-condition-if-a-state-of-an-entity-hasnt-changed/140791/4

and the fact that I regularly see people say how you can’t use now() as a trigger but no one has ever offered an alternative.

Also, I had to remove some {% raw %} and {% endraw %} tags. I don't know if this is correct but my addition wouldn't appear correctly without doing this and I noticed not all code snippets were using it so assumed some had been left over from a previous edit. Apologies if this is the wrong thing to do.

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -58,8 +58,6 @@ automation:
 
 Triggers when numeric value of an entity's state crosses a given threshold. On state change of a specified entity, attempts to parse the state as a number and triggers once if value is changing from above to below or from below to above the given threshold.
 
-{% raw %}
-
 ```yaml
 automation:
   trigger:
@@ -78,16 +76,12 @@ automation:
       seconds: 5
 ```
 
-{% endraw %}
-
 <div class='note'>
 Listing above and below together means the numeric_state has to be between the two values.
 In the example above, a numeric_state that goes to 17.1-24.9 (from 17 or below, or 25 or above) would fire this trigger.
 </div>
 
 The `for:` can also be specified as `HH:MM:SS` like this:
-
-{% raw %}
 
 ```yaml
 automation:
@@ -104,11 +98,7 @@ automation:
     for: "01:10:05"
 ```
 
-{% endraw %}
-
 You can also use templates in the `for` option.
-
-{% raw %}
 
 ```yaml
 automation:
@@ -127,8 +117,6 @@ automation:
       message: >
         {{ trigger.to_state.name }} too high for {{ trigger.for }}!
 ```
-
-{% endraw %}
 
 The `for` template(s) will be evaluated when an entity changes as specified.
 
@@ -152,8 +140,6 @@ automation:
 
 You can also use templates in the `for` option.
 
-{% raw %}
-
 ```yaml
 automation:
   trigger:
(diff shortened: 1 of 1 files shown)


klaasnicolaas:
Please put the `raw` and `endraw` tags back

kloggy:
Sorry, I guess that means I was wrong to remove them. I tried several things in order to include my edit in exactly the format it is in now but it would not format the document correctly. I was about to give up when I noticed the inconsistent raw and endraw tags.

As I said, sorry to have removed them. However I don't know how to revert changes I have made. Maybe it would be best for you to reject the entire edit if that is possible and I will try and start again.

Once again, sorry but I was trying to help.

klaasnicolaas:
Just make a new commit where you placed the tags back

kloggy:
@klaasnicolaas 
I'm sorry but I cannot find how to edit my changes. As you can guess I am not experienced in the use of GitHub.

kloggy:
As an aside, what do the {% raw %} and {% endraw %} tags do and why are they not around every code snippet in the doc and finally, why do they appear as actual text in the doc?

image

klaasnicolaas:
Just edit [this file](https://github.com/kloggy/home-assistant.io/blob/patch-1/source/_docs/automation/trigger.markdown) on your fork and commit to this PR. 

We use these raw tags because JInja2 templating must be wrapped outside the markdown code.

kloggy:
@klaasnicolaas 
Thanks for the clarification. That explains why I couldn't get the edit to work in the first place. My addition includes Jinja 2 templating!!

Sorry it was such an ordeal. I hope the changes are acceptable.

source/_docs/automation/trigger.markdown
```diff
@@ -279,6 +279,22 @@ The `for` template(s) will be evaluated when the `value_template` becomes `true`
...
+
+As an alternative, providing you include the sensor [time](/integrations/time_date/) in your configuration, you can use the following template:
+
+{% raw %}
+
+```yaml
+automation:
+  trigger:
+    platform: template
+    value_template: "{{ (as_timestamp(states.sensor.time.last_changed) - as_timestamp(states.YOUR.ENTITY.last_changed)) > 300 }}"
```
frenck, line 291: We should not use `states.sensor.time.last_changed` as per https://www.home-assistant.io/docs/configuration/templating/#states

frenck (changes requested):
Thank for the PR @kloggy, I've left a review comment, could you take a look? Thanks! 👍

source/_docs/automation/trigger.markdown
kloggy: Interestingly I never use the form `states.sensor.time.last_changed` in my code but that is how it was suggested in the forum so I bowed to what I thought was their better knowledge.

kloggy: Err... Do I need to do something to 'accept' your suggested change?
(I'm still learning my way around Github...)

EDIT: Except... how do you get 'last_changed' in any other form?

frenck: Sorry, you are right, not sure where my mind was when I wrote that comment. 🤒

frenck (approved):
Thanks, @kloggy! 👍 

Happy Hacktoberfest 🎉

Issue #10491, opened 2019-09-29
🔗 Fixing broken links on our website


frenck:
Our website has been aging a bit over the last couple of years, unfortunately, this means a lot of links to external sites, have broken down. E.g., due to the website no longer exists, page has moved and bunch of other reasons. This is unfortunate for the users of our website.

This GitHub issue serves as a placeholder to get all those links fixed (or removed if fixing seems impossible).

**Want to help us out?**

Download [Xenu](http://home.snafu.de/tilman/xenulink.html) (for free) and scan our website for broken links and help us fixing them. 👍 🎖 🎉 🍾 
For those who prefer to skip scanning: [Home Assistant.zip](https://github.com/home-assistant/home-assistant.io/files/3720759/Home.Assistant.zip) (last update 12-10-2019)

(Any other tool is allowed as well of course).

**Notes**
- `credits.markdown` is a generated file. Unfortunately, we cannot fix broken links there at this moment.

**Please reference this issue in your PR**.

klaasnicolaas:
O and btw scanning takes some hours 🤣 What if we do the scanning and publish the Xenu file here?

srg74:
https://www.home-assistant.io/blog/ - error https://github.com/emontomery -> should be https://github.com/emontnemery

srg74:
on page https://www.home-assistant.io/components/qwikswitch/ link https://www.home-assistant.io/components/qwikswitch/components/binary_sensor is broken

frenck:
@srg74 Nice sir! 👍 

On all pages, there is a link in the top right "Edit on GitHub", which allows you to change it and open a PR! That way, you can contribute and help to fix those! 🏅 

Happy Hacktoberfest 🎉

srg74:
Not on a first one :(

frenck:
@srg74 Here you go: <https://github.com/home-assistant/home-assistant.io/blob/current/source/_posts/2019-09-18-release-99.markdown>

klaasnicolaas:
In the main message I have put a zip file with an export of my scan with Xenu, so you can get started right away. To show only the broken links, click on **View** -> **Show broken links only**. Of course it is wise to do your own scan in a few days to keep it up to date 😉

claughinghouse:
> O and btw scanning takes some hours 🤣 What if we do the scanning and publish the Xenu file here?

[...]

PR #10617, opened 2019-10-05
Fix broken links

scheric:
**Description:**
Fixing some links.
 #10491

## Checklist:


scheric:
@frenck  Should these PRs only have one component in it or should I create more PR.

fabaff:
Usually smaller pull requests are preferred as they require less time to review.

source/_docs/scripts/conditions.markdown
```diff
@@ -241,7 +241,7 @@ A better weekday condition could be by using the [Workday Binary Sensor](/integr
 
 ### Zone condition
 
-Zone conditions test if an entity is in a certain zone. For zone automation to work, you need to have setup a device tracker platform that supports reporting GPS coordinates. Currently this is limited to the [OwnTracks platform](/integrations/device_tracker.owntracks/) and the [iCloud platform](/integrations/device_tracker.icloud/).
+Zone conditions test if an entity is in a certain zone. For zone automation to work, you need to have setup a device tracker platform that supports reporting GPS coordinates. Currently this is limited to the [OwnTracks platform](/integrations/device_tracker.owntracks/) and the [iCloud platform](/integrations/icloud/).
```
fabaff: For stuff like that we have already [redirects](https://github.com/home-assistant/home-assistant.io/blob/next/source/_redirects#L339).

klaasnicolaas: Does not mean that this adjustment is not correct, once it has changed everywhere, the redirect is no longer necessary 😉

klaasnicolaas:
What @fabaff already indicates, we prefer to have small PRs so that it is easy to review. But in this case it may be more than 1 link modification, if there are 15 then I think it's good too 😄

source/_docs/scripts/conditions.markdown
scheric: But that means that the link should work.
If i click on [iCloud platform](https://www.home-assistant.io/integrations/device_tracker.icloud/) or [OwnTracks platform](https://www.home-assistant.io/integrations/device_tracker.owntracks/) in https://www.home-assistant.io/docs/scripts/conditions/#zone-condition I get redirected to Page Not Found.

scheric: Same with https://www.home-assistant.io/docs/mqtt/broker/#cloudmqtt
 [OwnTracks](https://www.home-assistant.io/integrations/device_tracker.owntracks)

frenck: While not broken, It is OK (and personally to me preferred) to change those. If we can reduce redirects, that is always a win. Both for client<->server round trips, but also SEO wise.

frenck (approved):
LGTM, Thanks @scheric!

Happy Hacktoberfest 🎉

PR #10628, opened 2019-10-06
Add Netlify header control to improve CouldFlare caching behavior

frenck:
### Description:

This PR aims to improve the caching logic we have with Cloudflare & Netlify.

### Current situation

Currently, Netlify's cache-control headers is set to: `max-age=0, must-revalidate, public`.
This destroys the use of caching in CloudFlare, however, in Cloudflare we enforce browser caching on the client-side.

Right now, if we make changes to the site, it may take up to 5 days for the user to see the change (due to the enforce browser cache).

Our current security headers score:
image

### Wanted situation:

This is actually not really the situation it should be.

Furthermore, this does not leverage the true power of Cloudflare. Cloudflare should be the one caching and taking all the hits. Not the browser of the client. Cloudflare can handle it, it was built for it.

From Netlify's end, we want to reduce bandwidth as much as possible, but want to show the newest site as well.

And while we are obviously going to modify headers, let's improve the security headers in the same go.

### Tasks:

- [x] Add Netlify header control file
- [x] Adjust Jekyll to publish the header file in the generated site
- [x] Add [`s-max-age`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Expiration) cache headers to improve edge caching on Cloudflare.
- [x] Adjust Jekyll to have some form of cache-busting on the Javascript & CSS assets.
- [x] Add some security headers
   - [x] `X-Frame-Options`
   - [x] `X-XSS-Protection`
   - [x] `Referrer-Policy`
   - [x] `Feature-Policy`
   - [x] `Content-Security-Policy`
   - [x] `X-Content-Type-Options`
- [ ] Remove Cloudflare's client browser & edge cache enforcement, giving control to the website to finegrain caching policies.

**Pull request in home-assistant (if applicable):** n/a

## Checklist:

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -134,5 +134,6 @@ defaults:
 
 # Support for files Jekyll will normally exclude
 include:
-  - ".well-known"
+  - "_headers"
   - "_redirects"
+  - ".well-known"
--- a/source/_headers
+++ b/source/_headers
@@ -1,5 +1,11 @@
 /*
   Cache-Control: public, max-age: 0, s-max-age=3600, must-revalidate
+  Content-Security-Policy: form-action https:
+  Feature-Policy: vibrate 'none'; geolocation 'none'; midi 'none'; microphone 'none'; camera 'none'; magnetometer 'none'; gyroscope 'none'; speaker 'none'; vibrate 'none'; payment 'none'
+  Referrer-Policy: strict-origin-when-cross-origin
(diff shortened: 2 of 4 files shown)


frenck:
That is more like it.

image

frenck:
GTMetrix is still happy

image

balloob (approved):

PR #10677, opened 2019-10-09
Small documentation fix for history_graph

kilrogg:
"title"-field was incorrectly documented as "name"


fabaff (approved):

frenck:
This change should not have been merged. This is not a Lovelace card, but an integration.

<https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/history_graph/__init__.py#L24>

PR #10698, opened 2019-10-10
Add note about not supporting two-factor

meichthys:
This integration doesn't support two-factor authenticated Reddit accounts. As of 2019-10-10, the package behind the integration (praw==6.3.1) supports short-term two-factor auth, but the totp token expires in 1 hour which would cause the integration to fail.

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/reddit.markdown
+++ b/source/_integrations/reddit.markdown
@@ -14,6 +14,10 @@ The Reddit sensor integrates data from [Reddit](https://reddit.com/) to monitor 
 
 To set up this sensor, you will need to generate a `client_id` and `client_secret` for the user account you will use to connect. Follow the first steps in [this Wiki page](https://github.com/reddit-archive/reddit/wiki/OAuth2-Quick-Start-Example).
 
+<div class='note'>
+This integration does not support Reddit two-factor authentication. If you use two-factor authentication for your Reddit account, create a separate Reddit account without two-factor authentication for use with homeassistant.
+</div>
+
 ## Configuration
 
 To enable this platform, add the following to your `configuration.yaml` file:


walebant:
Hi, can I work on this?

meichthys:
@wale-bant , I don't think there's anything to work on. 

I suppose it would be possible to create a totp add-on for Homeassistant that could provide two-factor tokens, but that would be out of the scope of this pull-request.

source/_integrations/reddit.markdown
```diff
@@ -14,6 +14,10 @@ The Reddit sensor integrates data from [Reddit](https://reddit.com/) to monitor
 
 To set up this sensor, you will need to generate a `client_id` and `client_secret` for the user account you will use to connect. Follow the first steps in [this Wiki page](https://github.com/reddit-archive/reddit/wiki/OAuth2-Quick-Start-Example).
 
+<div class='note'>
+This integration does not support Reddit two-factor authentication. If you use two-factor authentication for your Reddit account, create a separate Reddit account without two-factor authentication for use with homeassistant.
```
frenck, line 18: > This integration does not support Reddit's two-factor authentication. If you use two-factor authentication for your Reddit account, create a separate Reddit account without two-factor authentication for use with Home Assistant.

frenck (changes requested):
Thanks for the PR @meichthys!

Two typos in there, if you can adjust those, we can merge it in! 👍

source/_integrations/reddit.markdown
meichthys: Hi @frenck, I agree with your suggested changes, but i can't figure out how to make those changes since i've already deleted my fork of home-assistant.io...

frenck: In that case, you can't 🤷‍♂ 

The best thing to do right now, I guess, is to open up a new PR. With a small change like this, it would be a tiny copy & paste action. 

Thanks already! 👍

meichthys: Ok, Will do. Lessons Learned!

PR #10707, opened 2019-10-11
Update credits.markdown

RowdyDog12:
**Description:**
Fixed a developer link on the credits page.

## Checklist:


frenck:
This is a generated file and should not be changed.
Even if we change it, the next generation of the file would overwrite it again.

Nevertheless, thank you for your contribution 👍

RowdyDog12:
Ahh sorry that I did not catch that part!  Thanks for seeing that.

PR #10726, opened 2019-10-11
Update control section in group

Misiu:
`control: hidden` works fine for groups in Lovelace

**Description:**

**Pull request in home-assistant (if applicable):** https://github.com/home-assistant/home-assistant-polymer/pull/2478

## Checklist:

### DIFF
--- a/source/_integrations/group.markdown
+++ b/source/_integrations/group.markdown
@@ -55,7 +55,7 @@ view:
   type: boolean
   default: false
 control:
-  description: "If value set to `hidden` the group switch will be hidden."
+  description: "**This is only a feature in generated mode of Lovelace** If value set to `hidden` the group switch will be hidden."
   required: false
   type: string
 {% endconfiguration %}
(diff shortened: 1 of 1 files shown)


frenck:
Still, I think we should not add this. it is likely we are removing the old state UI before HA 1.0.
The backend should not control or influence the UI. 

So while maybe true, I'm kinda doubting we should remove this notice.

iantrich:
This is only a feature in generated mode of Lovelace. It does nothing once you take control of customizing your UI. At the very least that should be mentioned.

frenck (changes requested):
@Misiu Would you please update this PR with the feedback given? Thanks! 👍

Misiu:
@frenck I've added info about the generated mode. Is that ok?

> This is only a feature in generated mode of Lovelace. It does nothing once you take control of customizing your UI. At the very least that should be mentioned.

@iantrich I didn't know about that, so if I create the UI using YAML then I won't be able to hide the switch in a group? If so then this is confusing and I think it should be unified.

frenck (approved):
Thanks, @Misiu! 👍 

Happy Hacktoberfest 🎉

PR #10736, opened 2019-10-12
🚧 Add note about NS API changes

klaasnicolaas:
**Description:**
NS announced at the beginning of this year that they would adjust the available API. The current integration only worked with the old API.

The API has been fully deprecated since October 1, so it no longer works in Home Assistant until the python lib has been updated to work with new NS API.

## Checklist:


frenck:
As discussed on Discord, let's make this an alert, since all users of this integration are currently affected.

PR #10750, opened 2019-10-12
changed blockquote to italic in Buienradar

peternijssen:
**Description:**
Changing the blockquote to italic, because the blockquote acts weird by adding \201C and \201D. See https://www.home-assistant.io/integrations/buienradar/

A similar change was done here, which contained the same issue: https://github.com/home-assistant/home-assistant.io/pull/9974

If we actually fix the blockquotes, then feel free to close this PR. I have no clue though how to fix it otherwise ;)

## Checklist:

### DIFF
--- a/source/_integrations/buienradar.markdown
+++ b/source/_integrations/buienradar.markdown
@@ -113,5 +113,7 @@ version of the name will be used as the name of the entity.
 With the default of "Buienradar loop" the entity name becomes
 `camera.buienradar_loop`.
 
+<div class='note'>
 [Usage statement:](https://www.buienradar.nl/overbuienradar/gratis-weerdata)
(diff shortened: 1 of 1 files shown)


source/_integrations/buienradar.markdown
```diff
@@ -113,5 +113,7 @@ version of the name will be used as the name of the entity.
 With the default of "Buienradar loop" the entity name becomes
 `camera.buienradar_loop`.
 
+<div class='note'>
```
frenck: I don't think it needs to be quoted or needs an alert box.

The most classical and elegant way would be marking that whole sentence italic.

frenck (changes requested):

source/_integrations/buienradar.markdown
peternijssen: Done! not sure if you had in mind how I currently did it.

frenck (approved):
Yep! Stylish, sweet, elegant and still there, thanks @peternijssen! 👍 

Happy Hacktoberfest 🎉

PR #10768, opened 2019-10-13
MQTT Device Tracker: Updated documentation with source type

tenfire:
**Description:**

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#27595

## Checklist:

### DIFF
--- a/source/_integrations/device_tracker.mqtt.markdown
+++ b/source/_integrations/device_tracker.mqtt.markdown
@@ -22,7 +22,6 @@ device_tracker:
     devices:
       paulus_oneplus: 'location/paulus'
       annetherese_n4: 'location/annetherese'
-    source_type: router
 ```
 
 {% configuration %}
@@ -38,6 +37,7 @@ source_type:
   description: Source type of the device tracker.
   required: false
   type: string
+  default: gps
 {% endconfiguration %}
 
 ## Usage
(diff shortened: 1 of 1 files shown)


source/_integrations/device_tracker.mqtt.markdown
```diff
@@ -33,6 +34,10 @@ qos:
   description: The QoS level of the topic.
   required: false
   type: integer
+source_type:
+  description: Source type of the device tracker.
+  required: false
```
frenck, line 38: This has a default, please add that to this block.

source/_integrations/device_tracker.mqtt.markdown
```diff
@@ -22,6 +22,7 @@ device_tracker:
     devices:
       paulus_oneplus: 'location/paulus'
       annetherese_n4: 'location/annetherese'
+    source_type: router
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

frenck (changes requested):
Thanks for the PR @tenfire! 

I've reviewed it and left some comments. Could you please take a look? Thanks! 👍

source/_integrations/device_tracker.mqtt.markdown
tenfire: I have now pushed these changes

frenck (approved):

PR #10775, opened 2019-10-14
fix grouping and sorting for versions >= 0.100.x

dwradcliffe:
**Description:**

Fixes the grouping and sort order of the versions list on the integrations page for versions after 0.100.x.

When grouped by the first character of the minor version, it groups 0.100 with 0.1. We can instead chop off the last character and have an accurate group. I also fixed the sort order and was able to remove a custom sort key that wasn't needed anymore.

image

## Checklist:

### DIFF
--- a/plugins/filters.rb
+++ b/plugins/filters.rb
@@ -100,13 +100,10 @@ module Jekyll
         if minor.length == 1
           "#{major}.X"
         else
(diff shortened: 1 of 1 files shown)


frenck (approved):
Nice! Thanks, @dwradcliffe! 👍 

Happy Hacktoberfest 🎉

PR #10813, opened 2019-10-16
Add documentation for creating AWS Lambda proxy for Alexa Intents

lpomfrey:
**Description:** The current documentation for configuring Alexa Custom skils for intents doesn't work with the new auth system. This adds documentation for setting up a proxying AWS Lambda function to forward requests to /api/alexa in the same way as the documented Alexa Smart Home configuration

**Pull request in home-assistant (if applicable):** N/A
## Checklist:


bbrendon:
I've looked it over and I'm confused....

The page is about a Custom Skill but the new documentation discusses a Smart Home Skill. From what I understand these two things are different.

I see phrases like "Smart Home API " and "Smart Home Skill". I think the terms "Custom Skill" or "Smart Home Skill API" should be used.

Frenk, basically the way I understand it "Smart Home Skill API" != "Custom Skill".

Anyway, hope that's helpful. So if it were me, I would get another set of eyes on this but based on what I know, I think it needs some edits.

balloob:
@bbrendon is right. This is the page about custom skills, not the smart home docs. Manual set up for smart home docs is here: https://www.home-assistant.io/integrations/alexa.smart_home/

lpomfrey:
Those will be hangovers from copying most of this out of the Smart Home docs, I thought I'd caught them all. Should be fixed in the latest commit.

source/_integrations/alexa.intent.markdown
```diff
@@ -17,17 +17,11 @@ The built-in Alexa integration allows you to integrate Home Assistant into Alexa
...
+- Amazon Developer Account. You can sign on [here](https://developer.amazon.com).
+- An [AWS account](https://aws.amazon.com/free/) is need if you want to use Alexa Custom Skill API. Part of your Alexa Custom Skill will be hosted on [AWS Lambda](https://aws.amazon.com/lambda/pricing/). However you don't need worry the cost, AWS Lambda allow free to use up to 1 millions requests and 1GB outbound data transfer per month.
+- The Alexa Custom Skill API also needs your Home Assistant instance can be accessed from Internet. We strongly suggest you host HTTPS server and use validation certificate. Read more on [our blog](/blog/2015/12/13/setup-encryption-using-lets-encrypt/) about how to set up encryption for Home Assistant. When running Hass.io, using the [Let's Encrypt](/addons/lets_encrypt/) and [Duck DNS](/addons/duckdns/) add-ons is the easiest method.
```
balloob: Home Assistant Cloud is the easiest. Second is using the DuckDNS add-on because it has Let's Encrypt built-in.

source/_integrations/alexa.intent.markdown
```diff
@@ -36,12 +30,89 @@ To get started with Alexa skills:
...
+
+#### Create an IAM Role for Lambda 
+
+The first thing you need to do after you sign in to the [AWS console](https://console.aws.amazon.com/) is to create an IAM Role for Lambda execution. AWS has very strict access control, you have to explicitly define and assign the permissions.
+
+- Click `Service` in top navigation bar, expand the menu to display all AWS services, click `IAM` under `Security, Identity, & Compliance` section to navigate to IAM console. Or you may use this [link](https://console.aws.amazon.com/iam/home)
+- Click `Roles` in the left panel, then click `Create role`, select `AWS Service` -> `Lambda` in the first page of the wizard, then click `Next: Permissions`
+- Select `AWSLambdaBasicExecutionRole` policy, then click `Next: Tags`. (Tips: you can use the search box to filter the policy)
+
+  <img src='/images/integrations/alexa/create_iam_role_attach_permission.png' alt='Screenshot: Attach permission policy to IAM role'>
```
balloob, line 52: Want to wrap this in `<p class='img'></p>`

source/_integrations/alexa.intent.markdown
```diff
@@ -36,12 +30,89 @@ To get started with Alexa skills:
...
 
 You can use this [specially sized Home Assistant logo][large-icon] as the large icon and [this one][small-icon] as the small one.
 
+### Create Your Lambda Function
+
+The Alexa Custom skill will trigger a AWS Lambda function to process the request, we will write a small piece of code hosted as a Lambda function to basically redirect the request to your Home Assistant instance, then the Alexa integration in Home Assistant will process the request and send back the response. Your Lambda function will deliver the response back to Alexa.
+
+OK, let's go. You first need sign in your [AWS console](https://console.aws.amazon.com/), if you don't have an AWS account yet, you can create a new user [here](https://aws.amazon.com/free/) with 12-month free tier benefit. You don't need worry the cost if your account has already passed the first 12 months, AWS provides up to 1 million Lambda requests, 1GB of outbound data and unlimited inbound data for free every month for all users. See [Lambda pricing](https://aws.amazon.com/lambda/pricing/) for details.
+
+#### Create an IAM Role for Lambda
```
balloob, line 43: Instead of these instructions, someone should just create a GitHub repo with a serverless.com implementation. Then it becomes as easy as `npm install && aws login && npm run deploy`

source/_integrations/alexa.intent.markdown
```diff
@@ -36,12 +30,89 @@ To get started with Alexa skills:
...
+  * **US West (Oregon)** region for Japanese and English (AU) skills. 
+- Click `Functions` in the left navigation bar, display list of your Lambda functions.
+- Click `Create function`, select `Author from scratch`, then input a `Function name`.
+- Select *Python 3.6* or *Python 3.7* as `Runtime`.
+- Select *Use an existing role* as `Execution role`, then select the role you just created from the `Existing role` list.
+- Click `Create function`, then you can config detail of Lambda function.
+- Under `Configuration` tab, expand `Designer`, then click `Alexa Skills Kit` in the left part of the panel to add a Alexa Skills Kit trigger to your Lambda function.
+- Scroll down little bit, you need to input the `Skill ID` from the skill you created in previous step. (You may need to switch back to the Alexa Developer Console to copy the `Skill ID`.)
+- Click your Lambda Function icon in the middle of the diagram and scroll down, you will see a `Function code` window.
+- Clear the example code and copy the Python script from: <https://gist.github.com/lpomfrey/97381cf4316553b03622c665ae3a47da>
```
balloob: This is not how links work in Markdown.

bbrendon:
Is it possible to write this so it covers with and without lambda?

balloob:
No. Alexa requires a lambda :(

source/_integrations/alexa.intent.markdown
lpomfrey: This whole page (and the smart home one) are prefaced by a section introducing Home Assistant Cloud and indicating that these are the "manual" instructions if you choose not to do that, so re-mentioning Home Assistant Cloud may be confusing. I can add another mention of it if you'd prefer though?

source/_integrations/alexa.intent.markdown
lpomfrey: It is according to this: https://www.markdownguide.org/basic-syntax#urls-and-email-addresses
But I can change it to use the `[text](url)` syntax here and on the Smart Home section if that's preferred?

lpomfrey:
Strictly it doesn't require a Lambda. It just needs a way of pulling the auth token from the request body and adding it to and `Authorization: Bearer <token>` header so Home Assistant recognises it.

I used a Lambda function since the Smart Home integration already used it, so it seemed the obvious choice. But there's no reason you couldn't use a locally hosted proxy (it would only need to be something like a small Python WSGI application that does essentially the same as the lambda), or possibly some sort of nginx processing using it's Lua module, or any number of other ways. 

Alternatively it could be handled in Home Assistant itself, maybe if some kind of view middleware concept was introduced?

source/_integrations/alexa.intent.markdown
balloob: At least let's just mention the duckdns add-on.

source/_integrations/alexa.intent.markdown
balloob: Oh, didn't know. Yeah let's use the other syntax, it seems more common.

bbrendon:
@lpomfrey  Thanks for the excellent explanation on Lambda. I didn't realize the new auth system added it as a requirement. I'm probably using the api_password still for this as I set it up many years ago and haven't really looked it since.

balloob:
@bbrendon api password support in the url is being removed in the next release. Has been deprecated for 6 months.

You're right @lpomfrey that for smart homes we require a lambda, for custom intents it's not needed. I am surprised no one has ever updated [the code that deals with it](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/alexa/intent.py#L47)…

balloob (approved):

PR #10819, opened 2019-10-16
Add support for SAJ inverters connected via WiFi

fredericvl:
**Description:**

Currently the SAJ component only works with inverters connected via an ethernet webserver module.
Apparently the WiFi webserver module has another web interface and thus also other URL's to get the data from.
Also the WiFi webserver requires a username and password where the ethernet webserver does not.
This PR adds 3 extra optional configuration parameters:
- type : `ethernet` or `wifi` (default value: `ethernet` so not breaking for existing configurations)
- username (only used when type is `wifi` but can be skipped if the inverter still has the default credentials set: admin/admin)
- password (only used when type is `wifi` but can be skipped if the inverter still has the default credentials set: admin/admin)

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#27742

## Checklist:


source/_integrations/saj.markdown
```diff
@@ -12,36 +12,62 @@ The `saj` sensor will poll a [SAJ](https://www.saj-electric.com/) solar inverter
...
+  - platform: saj
+    host: IP_ADDRESS_OF_DEVICE
+```
+
+```yaml
+# Example configuration.yaml entry for WiFi module
 sensor:
   - platform: saj
     host: IP_ADDRESS_OF_DEVICE
+    type: wifi
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

frenck (changes requested):
Hi there, @fredericvl! Thanks for the PR!

I've reviewed it and left a comment, could you please take a look?

Furthermore, since this PR depends on a parent PR in our main codebase (and thus the next release)., this PR should be targeting the `next` branch. Please change the target branch to `next` and rebase onto `next` if needed. (Feel free to ask for help in case you are unfamiliar with the process).

source/_integrations/saj.markdown
fredericvl: Now if a user has an inverter connected via ethernet cable he can indeed just copy-&-paste the sample config (without type, username, password) and it will work.
But if a user has an inverter connected via WiFi it will not work with the sample config and he needs to specify at least the type "wifi".

Just asking for your opinion, wouldn't it be better that the "type" parameter is required then (breaking change) ?

frenck: That is a code decision and not related directly to the documentation.
The documentation is here to reflect the state of the code.

I think it would wise to discuss this on the code base PR.

frenck (approved):
LGTM! Thanks, @fredericvl! 👍 

Happy Hacktoberfest 🎉

PR #10820, opened 2019-10-16
Clarification of time formats

akasma74:
This PR superceeds #10796 (couldn't rebase that).
I couldn't find anything in the current docs on time formats and thought it would be useful to give more details (rather than time | integer description) about ones available.

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/generic_thermostat.markdown
+++ b/source/_integrations/generic_thermostat.markdown
@@ -85,7 +85,6 @@ precision:
   default: "`0.5` for Celsius and `1.0` for Fahrenheit."
 {% endconfiguration %}
 
-Notes  
 Time for `min_cycle_duration` and `keep_alive` must be set as "hh:mm:ss" or it must contain at least one of the following entries: `days:`, `hours:`, `minutes:`, `seconds:` or `milliseconds:`. Alternatively, it can be an integer that represents time in seconds.
 
 Currently the `generic_thermostat` climate platform supports 'heat', 'cool' and 'off' hvac modes. You can force your `generic_thermostat` to avoid starting by setting HVAC mode to 'off'.
--- a/source/_integrations/generic_thermostat.markdown
+++ b/source/_integrations/generic_thermostat.markdown
@@ -55,7 +55,7 @@ ac_mode:
 min_cycle_duration:
   description: Set a minimum amount of time that the switch specified in the *heater* option must be in its current state prior to being switched either off or on.
   required: false
-  type: time
+  type: [time, integer]
 cold_tolerance:
   description: Set a minimum amount of difference between the temperature read by the sensor specified in the *target_sensor* option and the target temperature that must change prior to being switched on. For example, if the target temperature is 25 and the tolerance is 0.5 the heater will start when the sensor equals or goes below 24.5.
   required: false
@@ -69,7 +69,7 @@ hot_tolerance:
 keep_alive:
   description: Set a keep-alive interval. If set, the switch specified in the *heater* option will be triggered every time the interval elapses. Use with heaters and A/C units that shut off if they don't receive a signal from their remote for a while. Use also with switches that might lose state. The keep-alive call is done with the current valid climate integration state (either on or off).
   required: false
-  type: time
+  type: [time, integer]
 initial_hvac_mode:
   description: Set the initial HVAC mode. Valid values are `off`, `heat` or `cool`. Value has to be double quoted. If this parameter is not set, it is preferable to set a *keep_alive* value. This is helpful to align any discrepancies between *generic_thermostat* and *heater* state.
   required: false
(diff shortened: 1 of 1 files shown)


source/_integrations/generic_thermostat.markdown
```diff
@@ -55,7 +55,7 @@ ac_mode:
 min_cycle_duration:
   description: Set a minimum amount of time that the switch specified in the *heater* option must be in its current state prior to being switched either off or on.
   required: false
-  type: [time, integer]
```
frenck: This change is not correct, it accepts both time or an integer (which will be interpreted as seconds).

source/_integrations/generic_thermostat.markdown
```diff
@@ -69,7 +69,7 @@ hot_tolerance:
 keep_alive:
   description: Set a keep-alive interval. If set, the switch specified in the *heater* option will be triggered every time the interval elapses. Use with heaters and A/C units that shut off if they don't receive a signal from their remote for a while. Use also with switches that might lose state. The keep-alive call is done with the current valid climate integration state (either on or off).
   required: false
-  type: [time, integer]
+  type: time
```
frenck: This change is not correct, it accepts both time or an integer (which will be interpreted as seconds).

source/_integrations/generic_thermostat.markdown
```diff
@@ -85,12 +85,14 @@ precision:
   default: "`0.5` for Celsius and `1.0` for Fahrenheit."
 {% endconfiguration %}
 
-A full configuration example looks like the one below. `min_cycle_duration` and `keep_alive` must contain at least one of the following entries: `days:`, `hours:`, `minutes:`, `seconds:` or `milliseconds:`.
+__Notes__
```
frenck: Let's remove this, there is no need to create an emphasis here.

source/_integrations/generic_thermostat.markdown
```diff
@@ -85,12 +85,14 @@ precision:
...
 
-A full configuration example looks like the one below. `min_cycle_duration` and `keep_alive` must contain at least one of the following entries: `days:`, `hours:`, `minutes:`, `seconds:` or `milliseconds:`.
+__Notes__  
+Time for `min_cycle_duration` and `keep_alive` must be set as wrapped in quotes representation of time in form of "hh:mm:ss", time in seconds (without quotes) or it must contain at least one of the following entries: `days:`, `hours:`, `minutes:`, `seconds:` or `milliseconds:`
 
 Currently the `generic_thermostat` climate platform supports 'heat', 'cool' and 'off' hvac modes. You can force your `generic_thermostat` to avoid starting by setting HVAC mode to 'off'.
 
 Please note that when changing the preset mode to away, you will force a target temperature change as well that will get restored once the preset mode is set to none again.
 
+__Example__
```
frenck:
```suggestion
## Full configuration example

```

frenck (changes requested):
Hi @akasma74, thanks for the PR!

I've reviewed it and left some comments, could you please take a look? Thanks! 👍

source/_integrations/generic_thermostat.markdown
akasma74: Well, I explained that in the notes below. Can revert it back, but it looks ugly as time is not defined and that's the maing reason of all this fuss ;)

source/_integrations/generic_thermostat.markdown
akasma74: same as above

source/_integrations/generic_thermostat.markdown
akasma74: no problem, just wanted to make it as a separate paragraph

akasma74:
@frenck I've made some changes, hope it's better now.

source/_integrations/generic_thermostat.markdown
frenck: That is not how paragraphs work ;)
But in all honesty, a simple note doesn't need a paragraph. Especially since this note concerns the configuration that is listed above.

frenck (approved):
Thanks, @akasma74! 👍 

Happy Hacktoberfest 🎉

akasma74:
Thanks for approving the PR.
BTW, I was always curious why there is no mention of services to work with generic_thermostat on this page (and tbh it's not that obvious where to look for them) - perhaps it's a good idea to add something like this before the config exapmple?

> For the services available to work with `generic_thermostat` entities please refer to [this page](https://www.home-assistant.io/integrations/climate/).

Shall I add it to this PR or it's better to create a separate one (I support the former, reflecting this in PR's title)?

frenck:
We are looking for more generic ways to resolve this. Since every entity has one of the core entities as the base (e.g., light, climate, switch), and thus inherits the services. Adding this to every single integration or platform seems a bit cumbersome 😉

akasma74:
Agree, the more generic way would be ideal.
Just wanted to check that you guys have it on your list as currently it's a bit difficult to use for some and definitely can be improved ;)

PR #10830, opened 2019-10-16
Add notice about imports

noxhirsch:
Sometimes people are confused, because their Python scripts with imports don't work. I added a notice about that topic.

**Description:**

## Checklist:


bbrendon:
I haven't used this feature in a long time but I seem to recall there are some imports that are allowed.

frenck:
Yeah, build-ins works, however, I think this might be a valid statement to make. As soon as you need imports, you probably better not using this anyway.

Thanks, @noxhirsch! 👍 

Happy Hacktoberfest 🎉

frenck (approved):

PR #10836, opened 2019-10-16
Update with added options for multiple event types and emulated key hold events.

bendavid:
Corresponding to https://github.com/home-assistant/home-assistant/pull/27761

### DIFF
--- a/source/_integrations/keyboard_remote.markdown
+++ b/source/_integrations/keyboard_remote.markdown
@@ -28,15 +28,18 @@ type:
 emulate_key_hold:
   description: Emulate key hold events when key is held down.  (Some input devices do not send these otherwise.)
   required: false
-  type: boolean  
+  type: boolean
+  default: false
 emulate_key_hold_delay:
   description:  Number of milliseconds to wait before sending first emulated key hold event
   required: false
   type: positive_int
+  default: 350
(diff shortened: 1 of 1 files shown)


source/_integrations/keyboard_remote.markdown
```diff
@@ -22,10 +22,25 @@ keyboard_remote:
...
-device_description:
+emulate_key_hold:
+  description: Emulate key hold events when key is held down.  (Some input devices do not send these otherwise.)
+  required: false
+  type: boolean
+  default: false
+emulate_key_hold_delay:
+  description:  Number of milliseconds to wait before sending first emulated key hold event
+  required: false
+  type: positive_int
```
frenck: `positive_int` is not a valid type for the documentation, use `integer` instead.

frenck (changes requested):
Please make sure the build passes.

source/_integrations/keyboard_remote.markdown
bendavid: This has been changed to float in the end.

bendavid:
I believe all of the previous comments have been addressed now.

frenck (dismissed):

frenck (approved):

PR #10901, opened 2019-10-20
Zone: error

aviel08:
"zone:" not needed as a header of the zone.yaml file

**Description:**

## Checklist:


source/_integrations/zone.markdown
```diff
@@ -14,7 +14,6 @@ The `zone` integration uses YAML sequences to configure multiple zones:
 
 ```yaml
-zone:
```
frenck, line 17: uh? This is not the zone file, as written on multiple places, even in the line above this one.

frenck (changes requested):

PR #10903, opened 2019-10-20
Fix Client ID in setup instructions

peteclarkez:
Hi 
I've just setup an alexa skill using these instructions.
All seemed to work well.

2 Issues
1. The Client ID for EU doesn't seem to be correct.
I've fixed it here.

2. Error linking lambda and skill

When attaching the lamdba reference to the Skill I came across the issue and solution below (found from the error msg)
https://forums.developer.amazon.com/questions/183361/failed-to-save-skill-information.html
I could add a note for that , but I'm assuming it's something local to what I did.

Hope it helps
Peter

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/alexa.smart_home.markdown
+++ b/source/_integrations/alexa.smart_home.markdown
@@ -168,7 +168,7 @@ Alexa can link your Amazon account to your Home Assistant account. Therefore Hom
   * `Access Token URI`: https://[YOUR HOME ASSISTANT URL:PORT]/auth/token
   * `Client ID`:
     - https://pitangui.amazon.com/ if you are in US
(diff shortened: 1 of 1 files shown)


peteclarkez:
Also there I commented on the forum about this here
https://community.home-assistant.io/t/alexa-smart-home-account-linking/139646/16

pedrolamas:
I can attest to @peteclarkez changes as I had the same issue and that was only fixed after replacing the ".co.uk" with ".com" as per his PR.

frenck (approved):
Thanks, @peteclarkez! 👍 

Happy Hacktoberfest 🎉

PR #10915, opened 2019-10-21
Lovelace: custom badges and state-badge actions

iantrich:
**Description:**

https://github.com/home-assistant/home-assistant-polymer/pull/3867
https://github.com/home-assistant/home-assistant-polymer/pull/4028

## Checklist:


iantrich:
So, I've added custom badges here and need them listed on the sidebar as cards are but I'm not sure how to accomplish this; will need some guidance

iantrich:
I need further help on this. I'm not sure how to structure the docs so that I can get the badges in the sidebar.

frenck:
@iantrich The main problem with this is, that Jekyll does not support nested collections (which you are trying to do as it seems).

I see 2 possible options:
1. Add a `_badges` folder next to `_lovelace` but that will disconnect it from the user experience maybe.
2. Add a `type` kinda variable to the frontmatter in every markdown in the `_lovelace` collection (e.g., `card`, `badge`) and filter the collection when creating those 2 separate menu's. In order to reduce effort on changing all of them, the default of the frontmatter config can be set to `card` in the Jekyll config, so it only needs to be overridden for badges.

I have to be honest here: Both aren't pretty IMHO. 😢

iantrich:
I might just throw this in the views markdown for now 🤷‍♂

frenck:
Well, that is a good start, right? :)

frenck:
Working first, pretty later :)

frenck (approved):
👍

PR #10922, opened 2019-10-21
Correct grammar and improve clarity

Tediore:
The current version of this documentation refers to the Raspberry Pi 3 B+ as "the latest version" which is no longer correct with the release of the Pi 4. I changed this to explicitly refer to the 3 B+ instead of saying "the latest version." Do we want to continue saying "the latest version" and change the other references to the Pi 4 (and change the Amazon affiliate link)? The Hass.io installer for the Pi 4 is still an RC which is why I ask.

**Description:**

## Checklist:

### DIFF
--- a/source/getting-started/index.markdown
+++ b/source/getting-started/index.markdown
@@ -7,7 +7,7 @@ description: "Getting started: How to install Home Assistant."
 
 Note for contributors:
 
-The getting started guide aims at getting new users get Home Assistant  up and
+The getting started guide aims at getting new users get Home Assistant up and
 running as fast as possible. Nothing else. All other things should not be
 written down, as it creates a spaghetti of links and the user will lose focus.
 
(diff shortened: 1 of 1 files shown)


source/getting-started/index.markdown
```diff
@@ -49,21 +49,21 @@ We will need a few things to get started with installing Home Assistant. The lat
...
-1. Put the SD card in your SD card reader.
+1. Put the SD card in your card reader.
 2. Open balenaEtcher, select the Hass.io image and flash it to the SD card.
-3. Unmount the SD card and remove it from your SD card reader.
-4. Only if you want to configure WiFi or a Static IP (requires USB stick):
-   - Format a USB-Stick to FAT32 with volume-name `CONFIG`.
-   - Create a folder named `network` in the root of the newly formatted USB-stick.
-   - Within that folder create a file named `my-network` without extension.
+3. Unmount the SD card and remove it from your card reader.
+4. Follow this step if you want to configure Wi-Fi or a static IP address (this step requires a flash drive). Otherwise, move to step 5.
```
frenck: I would not prefer to use a flash drive since that is less specific. An SD card is a flash drive as well.

frenck (changes requested):
Awesome, @Tediore! One small comment, let me know what you think 👍

source/getting-started/index.markdown
Tediore: Ah, good point, I'll change it. I thought USB stick was too general, but given the context in the document, it's obvious what it means. Or "USB flash drive" maybe?

Any thoughts on referring to the Raspberry Pi 3 B+ vs the Pi 4? Explanation in my original comment above.

frenck:
I would refrain from recommending the Raspberry Pi 4 at this point since HassOS for that device is still not declared stable.

Tediore:
My thoughts exactly, thank you.

Tediore:
Updated. Thoughts?

frenck (approved):
LGTM! Great work @Tediore! 👍 

Happy Hacktoberfest 🎉

PR #10942, opened 2019-10-22
Add support for more Orange Pi Devices

pascallj:
**Description:**
Use only one 'pin_mode' per device instead of groups. This way it will be future proof if the library changes something for just one device instead of a group.

This is backwards compatible as the 'Lite', 'One' and 'PC' are internally handled the same way. 'Prime' is now handled slightly different due to changes in the library, so users of the Orange Pi Prime are advised to change their configuration to this new value.

Also fixes a link anchor.

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#28109

## Checklist:

### DIFF
--- a/source/_integrations/orangepi_gpio.markdown
+++ b/source/_integrations/orangepi_gpio.markdown
@@ -30,7 +30,7 @@ binary_sensor:
 
 {% configuration %}
 pin_mode:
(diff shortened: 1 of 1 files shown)


frenck (approved):

PR #11019, opened 2019-10-27
Update fritzbox_netmonitor additional package information

gerard33:
**Description:**
With Hass.io installing additional packages is not necessary.

## Checklist:


fabaff (approved):

frenck:
This is now not true anymore, this also applies to Docker instances. Although I appreciate you've made it more specific, it also causes it to be less true.

PR #11061, opened 2019-10-31
Lovelace: add visible example for views

iantrich:
**Description:**

## Checklist:

### DIFF
--- a/source/lovelace/views.markdown
+++ b/source/lovelace/views.markdown
@@ -68,6 +68,121 @@ user:
   type: string
 {% endconfiguration %}
 
+#### Example
+
+View config:
+
+```yaml
+- title: Living room
+  badges:
+    - device_tracker.demo_paulus
+    - entity: light.ceiling_lights
+      name: Ceiling Lights
+      icon: mdi:bulb
+    - entity: switch.decorative_lights
+      image: /local/lights.png
+```
+
+## Paths
+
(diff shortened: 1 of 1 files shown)


source/lovelace/views.markdown
```diff
@@ -64,10 +64,125 @@ If you define `visible` as objects instead of a boolean to specify conditions fo
...
+
+You can specify the visibility of views as a whole or per-user.
+
+### Example
+
+```yaml
+views:
+  - title: Ian
+    visible:
+      - user: iantrich_id
```
bramkragten: Use a hex string, so people know they have to search for a long weird string.

source/lovelace/views.markdown
```diff
@@ -64,10 +64,125 @@ If you define `visible` as objects instead of a boolean to specify conditions fo
...
+### Example
+
+```yaml
+- title: Garden
+  icon: mdi:flower
+```
+
+## Visible
+
+You can specify the visibility of views as a whole or per-user.
```
bramkragten: Maybe explain that it just isn't visible in the tab bar, but you can still reach it by url?

source/lovelace/views.markdown
iantrich: done

source/lovelace/views.markdown
iantrich: done

bramkragten (approved):

PR #11063, opened 2019-10-31
Improve scene docs

Santobert:
**Description:**
Improve scene docs

## Checklist:


dshokouhi (changes requested):
We should also update the 2 examples, looks like they are missing `state` for the `media_player`

Santobert:
I did that here already  #11030

But yea, you are right. I'll move the changes to this PR. Stick to the scope :D

balloob (approved):

PR #11084, opened 2019-11-02
Update ios.markdown

filip-van-hoeckel:
**Description:**

## Checklist:

### DIFF
--- a/source/_docs/ecosystem/ios.markdown
+++ b/source/_docs/ecosystem/ios.markdown
@@ -31,7 +31,7 @@ The Home Assistant for iOS app supports the new authentication system introduced
 
 ### Automated Setup
 
-The `ios` integration will automatically be loaded under the following circumstances:
+The `ios` integration may or may not be automatically loaded under the following circumstances:
 
 1. The [`discovery`](/integrations/discovery) integration is enabled.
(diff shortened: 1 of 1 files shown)


fabaff (approved):

source/_docs/ecosystem/ios.markdown
```diff
@@ -31,7 +31,7 @@ The Home Assistant for iOS app supports the new authentication system introduced
 
 ### Automated Setup
 
-The `ios` integration will automatically be loaded under the following circumstances:
+The `ios` integration may or may not be automatically loaded under the following circumstances:
```
frenck, line 34: This is just not needed, and not true.
It should automatically load in those circumstances if not, it should be reported as a bug.

This PR almost feels like a troll (including an another PR from @filip-van-hoeckel as well, adding the term "n00bs".

I'm reverting this one.

filip-van-hoeckel: It should != it will. Having users rely on incorrect or missing documentation is not helping anyone move forward.

frenck: There is no need to confuse even more. If it doesn't, it is a bug and must be solved.

PR #11091, opened 2019-11-03
Improved doorbird setup documentation

Tommatheussen:
**Description:**
As I'm always following practices of giving least permissions needed for my devices, I had created a user with only the API-operator, as it was the only permission listed in the documentation.
This caused no video streams to show up in HA.

I've updated the documentation to reflect which permissions are needed for which functionality, as well as moved the user creation documentation to the setup (made more sense imo)

**Pull request in home-assistant (if applicable):** None

## Checklist:

### DIFF
--- a/source/_integrations/doorbird.markdown
+++ b/source/_integrations/doorbird.markdown
@@ -53,7 +53,7 @@ devices:
       required: true
       type: string
     username:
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @Tommatheussen! 👍

PR #11101, opened 2019-11-05
Adding a small tip about DNS propagation.

brandoncorry:
**Description:**
I recently had to redo the setup for Google Assistant. Everything was correct (URLs in Google, ports, port forwarding, certs, etc.) but the process was failing when trying to connect with Google Assistant via the app to add the devices. While doing many troubleshooting steps, after some time, it would randomly work once, then not again for some time. More troubleshooting with failures, like analyzing tcpdumps deep, then randomly work for no reason. I had updated my DNS about 12-15 hours prior, which I thought was enough time, but it looks like it needed longer. Simply waiting was sufficient. I am hoping that this tip might save others some time pulling their hair out. :)

## Checklist:

- [X ] Branch: `next` is for changes and new documentation that will go public with the next Home Assistant release. Fixes, changes and adjustments for the current release should be created against `current`.
- [X ] The documentation follows the [standards][standards].

### DIFF
--- a/source/_integrations/google_assistant.markdown
+++ b/source/_integrations/google_assistant.markdown
@@ -22,7 +22,7 @@ The Google Assistant integration requires a bit more setup than most due to the 
 
 <div class='note warning'>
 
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @brandoncorry! 👍

PR #11102, opened 2019-11-05
fix http config entry for DuckDNS addon

basnijholt:
**Description:**
I spend hours yesterday struggling with this and reading through many many discussions.

The example now will actually work.

Apparently one shouldn't have the port number inside `base_url` and `https://` isn't needed.

For clarity I've added the `server_port` entry with a comment that `8123` is the default.

## Checklist:

### DIFF
--- a/source/_addons/duckdns.markdown
+++ b/source/_addons/duckdns.markdown
@@ -50,7 +50,8 @@ Use the following configuration in Home Assistant to use the generated certifica
 
 ```yaml
 http:
-  base_url: https://my-domain.duckdns.org:8123
+  base_url: my-domain.duckdns.org
+  server_port: 8123  # 8123 is also the default value
   ssl_certificate: /ssl/fullchain.pem
   ssl_key: /ssl/privkey.pem
 ```


source/_addons/duckdns.markdown
```diff
@@ -50,7 +50,8 @@ Use the following configuration in Home Assistant to use the generated certifica
 
 ```yaml
 http:
-  base_url: https://my-domain.duckdns.org:8123
+  base_url: my-domain.duckdns.org
+  server_port: 8123  # 8123 is also the default value
```
frenck, line 54: We don't add optional/default values to examples.

source/_addons/duckdns.markdown
```diff
@@ -50,7 +50,8 @@ Use the following configuration in Home Assistant to use the generated certifica
 
 ```yaml
 http:
-  base_url: https://my-domain.duckdns.org:8123
+  base_url: my-domain.duckdns.org
```
frenck, line 53: This change is incorrect.
It should be a full URI, also see: <https://www.home-assistant.io/integrations/http#base_url>

frenck (changes requested):

frenck:
I think we are dealing with an issue here, instead of a documentation error.

source/_addons/duckdns.markdown
basnijholt: You are right. I have no clue why it didn’t work for me. Now it does...

PR #11115, opened 2019-11-06
Automation example for Lutron integration

JonGilmore:
**Description:**
added an example of how to trigger an automation from a keypad button press. This was loosely explained previously, but takes prior knowledge to understand. An example is much easier to understand and implement.
@klaasnicolaas - sorry about the messed up PR, here's a new one against `current`. Thanks!

## Checklist:

### DIFF
--- a/source/_integrations/lutron.markdown
+++ b/source/_integrations/lutron.markdown
@@ -73,3 +73,18 @@ After setup, scenes will appear in Home Assistant using the area, keypad and but
 ## Occupancy Sensors
 
 Any configured Powr Savr occuancy sensors will be added as occupancy binary sensors. Lutron reports occupancy for an area, rather than reporting individual sensors. Sensitivity and timeouts are controlled on the sensors themselves, not in software.
+
+## Example Automations
+``` yaml
+- alias: "keypad button pressed notification"
+  trigger:
+    - platform: event
+      event_type: lutron_event
+      event_data:
+        id: office_pico_on
+        action: single
+  action:
+    - service: notify.telegram
+      data_template:
+        message: "pico just turned on!"
+```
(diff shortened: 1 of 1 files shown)


source/_integrations/lutron.markdown
```diff
@@ -73,3 +73,18 @@ After setup, scenes will appear in Home Assistant using the area, keypad and but
...
+  trigger:
+    - platform: event
+      event_type: lutron_event
+      event_data:
+        id: office_pico_on
+        action: single
+  action:
+    - service: notify.telegram
+      data_template:
+        message: "pico just turned on!"
```
frenck, line 90: Why is this a template? There are no variables in it... Couldn't we just use the `message` directly?

source/_integrations/lutron.markdown
```diff
@@ -73,3 +73,18 @@ After setup, scenes will appear in Home Assistant using the area, keypad and but
 ## Occupancy Sensors
 
 Any configured Powr Savr occuancy sensors will be added as occupancy binary sensors. Lutron reports occupancy for an area, rather than reporting individual sensors. Sensitivity and timeouts are controlled on the sensors themselves, not in software.
+
+## Example Automations
```
frenck, line 77: Add an empty line after a header.

frenck (changes requested):

source/_integrations/lutron.markdown
JonGilmore: :D its because I missed it in the copy/paste from my automation example. I've removed the template and switched it to just `data:`

source/_integrations/lutron.markdown
JonGilmore: check the latest commit, missed this in the conventions. thanks for the review.

frenck (approved):
Thanks, @JonGilmore! 👍

PR #11133, opened 2019-11-09
Remove host/netsh commands from docker directives

datdamnzotz:
Added -p 8123:8123 to the Windows docker command line.  Seems more reliable and easier for users.
Add ports to docker-compose.yaml files for easier launches..

**Description:**

## Checklist:

### DIFF
--- a/source/_docs/installation/docker.markdown
+++ b/source/_docs/installation/docker.markdown
@@ -39,17 +39,19 @@ $ docker run --init -d --name="home-assistant" -e "TZ=America/Los_Angeles" -v /P
 Alternatively, `docker-compose` works with any recent release of `docker-ce` on macOS. Note that (further down this page) we provide an example `docker-compose.yml` however it differs from the `docker run` example above. To make the .yml directives match, you would need to make _two_ changes: first add the equivalent `ports:` directive, then _remove_ the `network_mode: host` section. This is because `Port mapping is incompatible with network_mode: host:`. More details can be found at [Docker networking docs](https://docs.docker.com/network/). Note also the `/dev/tty*` device name used by your Arduino etc. devices will differ from the Linux example, so the compose `mount:` may require updates.
 
 ### Windows
+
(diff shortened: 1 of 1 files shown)


source/_docs/installation/docker.markdown
```diff
@@ -39,16 +39,20 @@ $ docker run --init -d --name="home-assistant" -e "TZ=America/Los_Angeles" -v /P
...
 ### Windows
+Docker containers are completely isolated from its Windows host system. So when you delete a container, all the changes you made to that container are also removed. If you want to have configuration files or other assets remain persistent, try mounting Windows folders on containers.
+
+Before proceeding make sure you have shared out a drive for docker to mount to.  This will allow saving of config files to persist on the local machine rather than in the docker container. (which may be destroyed when upgraded)
+https://docs.docker.com/docker-for-windows/#shared-drives
+https://docs.docker.com/docker-for-windows/troubleshoot/#verify-domain-user-has-permissions-for-shared-drives-volumes
 
 ```powershell
-$ docker run --init -d --name="home-assistant" -e "TZ=America/Los_Angeles" -v /PATH_TO_YOUR_CONFIG:/config --net=host homeassistant/home-assistant:stable
+$ docker run --init -d --name="home-assistant" -e "TZ=America/Los_Angeles" -v /PATH_TO_YOUR_CONFIG:/config -p 8123:8123 homeassistant/home-assistant:stable
```
frenck: This is an incorrect change that will break things. Home Assistant requires to be running on the host network in order to have all features available.

While there are certainly are situations where this would be enough, we don't want users to run into surprises here.

frenck (changes requested):
Hi there, @datdamnzotz!

Thanks for the PR, however, we need to advise to run Home Assistant on the host network for various reasons.

The shares part is a really useful addition though, so maybe you could revert the host network/port mapping part? 👍

source/_docs/installation/docker.markdown
datdamnzotz: rgr that.  makes sense if you are using it in other ways.  reverted.

frenck (approved):
Thanks for the quick adjustment @datdamnzotz! 👍

PR #11181, opened 2019-11-14
Update to clarify path declaration in lovelace config

poldim:
This was not very clear as many people were having this issue:

https://community.home-assistant.io/t/google-cast-error-unable-to-find-a-view-with-path-xxxx/138180/48

**Description:**

## Checklist:

- [ x] Branch: `next` is for changes and new documentation that will go public with the next Home Assistant release. Fixes, changes and adjustments for the current release should be created against `current`.

### DIFF
--- a/source/_integrations/cast.markdown
+++ b/source/_integrations/cast.markdown
@@ -13,7 +13,7 @@ You can enable the Cast integration by going to the Integrations page inside the
 
 ## Home Assistant Cast
 
-Home Assistant has its own Cast application to show the Home Assistant UI on any Chromecast device.  You can use it by adding the [Cast entity row](/lovelace/entities/#cast) to your Lovelace UI, or by calling the `cast.show_lovelace_view` service.  The service takes the path of a Lovelace view and an entity ID of a Cast device to show the view on.
+Home Assistant has its own Cast application to show the Home Assistant UI on any Chromecast device.  You can use it by adding the [Cast entity row](/lovelace/entities/#cast) to your Lovelace UI, or by calling the `cast.show_lovelace_view` service.  The service takes the path of a Lovelace view and an entity ID of a Cast device to show the view on.  The `path` has to be defined in your lovelace yaml for each view as outlined in the [views documentation](/lovelace/views/#path).
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


Villhellm:
Does it *have* to be defined? Paths are also autogenerated if you don't define it in the view. Just open the view and look at the browser url

frenck:
@Villhellm They don't have to be defined. It is definitely helpful to do so.

source/_integrations/cast.markdown
```diff
@@ -13,7 +13,7 @@ You can enable the Cast integration by going to the Integrations page inside the
 
 ## Home Assistant Cast
 
-Home Assistant has its own Cast application to show the Home Assistant UI on any Chromecast device.  You can use it by adding the [Cast entity row](/lovelace/entities/#cast) to your Lovelace UI, or by calling the `cast.show_lovelace_view` service.  The service takes the path of a Lovelace view and an entity ID of a Cast device to show the view on.
+Home Assistant has its own Cast application to show the Home Assistant UI on any Chromecast device.  You can use it by adding the [Cast entity row](/lovelace/entities/#cast) to your Lovelace UI, or by calling the `cast.show_lovelace_view` service.  The service takes the path of a Lovelace view and an entity ID of a Cast device to show the view on.  The `path` has to be defined in your lovelace yaml for each view as outlined in the [views documentation](/lovelace/views/#path).
```
frenck:
```suggestion
Home Assistant has its own Cast application to show the Home Assistant UI on any Chromecast device.  You can use it by adding the [Cast entity row](/lovelace/entities/#cast) to your Lovelace UI, or by calling the `cast.show_lovelace_view` service. The service takes the path of a Lovelace view and an entity ID of a Cast device to show the view on. A `path` has to be defined in your Lovelace YAML for each view, as outlined in the [views documentation](/lovelace/views/#path).
```

frenck (changes requested):

poldim:
> Does it _have_ to be defined? Paths are also autogenerated if you don't define it in the view. Just open the view and look at the browser url

@Villhellm So if a path is not defined, can you successfully cast via the service call? 

When I didn't initially have them defined they were autogenerated as just numbers.  I tried casting to the number as the path and that generated some error.

frenck (approved):
Thanks, @poldim! Let's just add this in, as it is a general good practice as well.

PR #11189, opened 2019-11-15
clarify options' use

akasma74:
The current description looks outdated to me because:
1. In Lovelace the time/date is not displayed when we open default_view
2. If only - 'time' specified, no `sensor.date` or anything else exists (and it's not clear as the configuration variable name is `display_options` (maybe now it's better to make it deprecated and change to `options` or something)?

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/time_date.markdown
+++ b/source/_integrations/time_date.markdown
@@ -10,9 +10,6 @@ ha_qa_scale: internal
 ---
 
 The time and date (`time_date`) sensor platform adds one or more sensors to your Home Assistant state machine.
-It also displays the time in various formats, the date, or both in non-Lovelace mode.  
-Note that you have to add all options you need access to even if you don't want them to be displayed.
-
 To have these sensors available in your installation, add the following to your `configuration.yaml` file (each option creates a separate sensor that contains appropriate data):
 
 ```yaml
--- a/source/_integrations/time_date.markdown
+++ b/source/_integrations/time_date.markdown
@@ -9,9 +9,11 @@ ha_release: pre 0.7
 ha_qa_scale: internal
 ---
 
-The time and date (`time_date`) sensor platform simple displays the time in various formats, the date, or both.
+The time and date (`time_date`) sensor platform adds one or more sensors to your Home Assistant state machine.
+It also displays the time in various formats, the date, or both in non-Lovelace mode.  
+Note that you have to add all options you need access to even if you don't want them to be displayed.
 
-To enable this sensor in your installation, add the following to your `configuration.yaml` file:
+To have these sensors available in your installation, add the following to your `configuration.yaml` file (each option creates a separate sensor that contais appropriate data):
 
 ```yaml
 # Example configuration.yaml entry
--- a/source/_integrations/time_date.markdown
+++ b/source/_integrations/time_date.markdown
@@ -13,7 +13,7 @@ The time and date (`time_date`) sensor platform adds one or more sensors to your
 It also displays the time in various formats, the date, or both in non-Lovelace mode.  
 Note that you have to add all options you need access to even if you don't want them to be displayed.
 
-To have these sensors available in your installation, add the following to your `configuration.yaml` file (each option creates a separate sensor that contais appropriate data):
+To have these sensors available in your installation, add the following to your `configuration.yaml` file (each option creates a separate sensor that contains appropriate data):
 
 ```yaml
 # Example configuration.yaml entry


source/_integrations/time_date.markdown
```diff
@@ -9,9 +9,11 @@ ha_release: pre 0.7
 ha_qa_scale: internal
 ---
 
-The time and date (`time_date`) sensor platform simple displays the time in various formats, the date, or both.
+The time and date (`time_date`) sensor platform adds one or more sensors to your Home Assistant state machine.
+It also displays the time in various formats, the date, or both in non-Lovelace mode.
```
frenck: Let's remove the reference of the non-Lovelace mode.

source/_integrations/time_date.markdown
```diff
@@ -9,9 +9,11 @@ ha_release: pre 0.7
 ha_qa_scale: internal
 ---
 
-The time and date (`time_date`) sensor platform simple displays the time in various formats, the date, or both.
+The time and date (`time_date`) sensor platform adds one or more sensors to your Home Assistant state machine.
+It also displays the time in various formats, the date, or both in non-Lovelace mode.  
+Note that you have to add all options you need access to even if you don't want them to be displayed.
```
frenck: I'm not sure what this means?

source/_integrations/time_date.markdown
```diff
@@ -9,9 +9,11 @@ ha_release: pre 0.7
...
 
-To enable this sensor in your installation, add the following to your `configuration.yaml` file:
+To have these sensors available in your installation, add the following to your `configuration.yaml` file (each option creates a separate sensor that contais appropriate data):
```
frenck:
```suggestion
To have these sensors available in your installation, add the following to your `configuration.yaml` file (each option creates a separate sensor that contains appropriate data):
```

frenck (changes requested):
Thanks, for the PR @akasma74!

Left a couple of comments, could you take a look? Thanks! 👍

source/_integrations/time_date.markdown
akasma74: That's the main point here. This integration was used to display time/tima and date in pre-Lovelace times.
And to be able to do so it creates some sensors.
Now you can see time displayed only in non-Lovelace, there is no time displayed in Lovelace afaik.
But the sensors are still in state machine.

source/_integrations/time_date.markdown
akasma74: See my previous comment.
If you want to access sensor.date, you need to add:
```
sensor:
    platform: time_date
    display_options:
       - date
```
even if you have no intention to see this date displayed or if you use HA in Lovelace mode.
That `display_options` is misleading and now means just `options` or `sensors`

source/_integrations/time_date.markdown
frenck: There is no difference. You can still add the to Lovelace.
States pages were generated, Lovelace is a manual job (unless running in automated mode). You could add this note you made there, to every single integration and entity throughout the documentation, hence me suggesting to remove it.

source/_integrations/time_date.markdown
frenck: The definition of sensors/entities/integrations has nothing to do with displaying them. Those are 2 separate concerns. Hence my question, what this means.
Please remove it.

source/_integrations/time_date.markdown
akasma74: Ah, I understand.
If it's pointless to have such a note in every single integration, where can we have it?
I believe it's worth to  convey this concept to the users.

source/_integrations/time_date.markdown
akasma74: > The definition of sensors/entities/integrations has nothing to do with displaying them. 

agree.
but what happens when you define
```
sensor:
   platform: time_date
```
?

> Those are 2 separate concerns. Hence my question, what this means.

from my point of view it means that currently to have a sensor that contains date I have to write
```
sensor:
    platform: time_date
    display_options:
       - date
```
instead of more intuitive
```
sensor:
    platform: time_date
    options:
       - date
```
exactly because "The definition of sensors/entities/integrations has nothing to do with displaying them", but in this integration definition of a sensor is tied to ```display_options``` variable, which is pretty misleading.

source/_integrations/time_date.markdown
frenck: It is the basic core concept of Lovelace?

source/_integrations/time_date.markdown
frenck: That name could use an update, however, my comment remains the same.

source/_integrations/time_date.markdown
akasma74: You mean that in Satest mode everything is added automatically and one need to hide unnecessary stuff when in Lovelace only the manually added stuff is displayed?
Yes, that's correct.
Perhaps the main reason of this PR was the way to define sensors - I assumed that everything is in the state machine, but only sensors listed in `display_options` will be displayed and it's only applicable to the States mode.

source/_integrations/time_date.markdown
akasma74: Does it mean that it's worth to have the wording as you requested (I think it's already is) and to create a new Feature Request to rename `display_options` (but there is no such thing for HA on Github afaik) referencing this PR as the cause?

frenck: I think that that discussion is out of scope for this PR.

akasma74: I think it is directly related to this PR and I'm asking here because I know no other means to get information from the team behind HA (not because I just want to talk).

frenck: The documentation should reflect the code base, how things are named in code is not a discussion for the documentation.

akasma74: Sure. Could you tell me _where_ I can discuss it with HA developers?

frenck: Discord 👍

frenck (approved):
Thanks, @akasma74! 👍

PR #11199, opened 2019-11-16
Add note for disco lights when HA is running

LordMike:
See https://github.com/home-assistant/home-assistant/issues/28642#issuecomment-552678145

Adds a note that you should stop HA when applying a Disco-lights "fix" to the AEOTEC Z-Stick.


frenck (approved):
Thanks, @LordMike! 👍

PR #11225, opened 2019-11-18
Remove warning about api_password

Villhellm:
**Description:**
Since api_password doesn't exist anymore this warning is no longer necessary.


frenck (approved):
Thanks, @Villhellm! 👍

PR #11228, opened 2019-11-19
Update 'other machine types' section in docs

dnguyen800:
**Description:**

I noticed the list of machine types supported by `hassio_install.sh` wasn't up-to-date with the list on [Github](https://github.com/home-assistant/hassio-installer). I updated the list and added a reference to the Github page. Maybe the duplicate list should be removed from here and instead add a reference to the Github?

## Checklist:


frenck:
It would be nice to link to the GitHub page @dnguyen800, however, this is a very essential part of an installation procedure. We don't want to send "new" people to all kinds of places, so in this case, I think this is actually fine.

frenck (approved):
Thanks! 👍

PR #11230, opened 2019-11-19
Update http documentation for trusted_proxies CIDR

hairychris:
**Description:**

It took me a little while to setup ip_bans properly with Docker, I also didn't realise you could use CIDR in http config. This should make life a little easier for people.

## Checklist:

- [*] Branch: `next` is for changes and new documentation that will go public with the next Home Assistant release. Fixes, changes and adjustments for the current release should be created against `current`.
- [*] The documentation follows the [standards][standards].

### DIFF
--- a/source/_integrations/http.markdown
+++ b/source/_integrations/http.markdown
@@ -114,7 +114,10 @@ http:
     - https://www.home-assistant.io
   use_x_forwarded_for: true
   trusted_proxies:
+    # specify a single IP
     - 10.0.0.200
+    # you can also use CIDR to specify a whole range
+    - 172.17.0.0/16
   ip_ban_enabled: true
   login_attempts_threshold: 5
 ```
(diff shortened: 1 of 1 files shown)


source/_integrations/http.markdown
```diff
@@ -154,6 +157,9 @@ Please note, that sources from `trusted_networks` won't be banned automatically.
 
 </div>
 
+If you are using a load balancer (e.g. Nginx, Traefik etc) with Docker then all requests will appear to come from the IP of your load balancer and therefore IP block all clients. To get the real IP you must set `use_x_forwarded_for` to true and add Dockers IP range to trusted_proxies (the default Docker IP range is 172.17.0.0/16 but it will use other ranges if this is already taken on your network).
```
frenck, line 160: This is partially true. I would not recommend anybody to use a CIDR for this, because it simply opens up the attack surface. In the setting of Home Assistant assuming that users using a proxy, are probably just using a single proxy and thus IP (I highly doubt a cluster of proxies).

hairychris: So the problem I was trying to solve is that the reverse proxy / load balancer might have a different IP address each time Docker starts. This is to my knowledge the default situation with HASS.io

Is there a better solution to this issue?

frenck:
On hass.io the IP for add-on stick (unless re-installed or restored)

hairychris:
Ok great will fix that up to suggest a specific IP and add a note warning that IP ranges can increase attack surface. 172 is a local range, I believe it would only effect those on an untrusted LAN and even then only if it uses the 172 range.

Edit: on further thought Docker will not choose an IP range that is in use.

Can you confirm if the IP will still be sticky on a NUC with hass.io installed?

frenck:
Hass.io add-on IPs stick.

frenck:
Closing this PR, it is running stale a bit. And we should really avoid driving people towards opening up whole subnets to this.

PR #11232, opened 2019-11-19
Some troubleshooting tips

rschuiling:
It was not described what to expect after configuring Life360 (i.e. a device tracker showing up). Also added some troubleshooting tips and a reference to the Device Tracker docs.

**Description:**

## Checklist:


pnbruckner:
Thanks. Nice addition!

If you don't mind, while you're at it, it would probably also be useful to add a note that it is normal for it to say on the Life360 Integrations page, after adding an account, "This integration has no devices."

frenck (approved):
Thanks, @rschuiling! 👍

PR #11251, opened 2019-11-21
Add remote.toggle

QbaF:
Missed `remote.toggle` service description.
Here mentioned: 
https://www.home-assistant.io/integrations/remote/
and working (but not 100%ok)
https://community.home-assistant.io/t/harmony-activities-volume-slider-and-lovelace/72419/109?u=qbaf

**Description:**
Service remote.toggle do not always toggle activity indicated in service_data. It does toggle active activity. If no one is activated, it toggle correctly.

### DIFF
--- a/source/_integrations/harmony.markdown
+++ b/source/_integrations/harmony.markdown
@@ -95,6 +95,10 @@ Start an activity. Will start the default `activity` from configuration.yaml if 
 | `entity_id`            |       no | Entity ID to target.
 | `activity`             |      yes | Activity ID or Activity Name to start.
 
+### Service `remote.toggle`
+
+..........
+
 ##### Example
 
 In the file 'harmony_REMOTENAME.conf' you can find the available activities, for example:
--- a/source/_integrations/harmony.markdown
+++ b/source/_integrations/harmony.markdown
@@ -99,6 +99,10 @@ Start an activity. Will start the default `activity` from configuration.yaml if 
 
 ..........
 
+| Service data attribute | Optional | Description |
+| ---------------------- | -------- | ----------- |
+| `entity_id`            |       no | Entity ID to target.
+
 ##### Example
 
 In the file 'harmony_REMOTENAME.conf' you can find the available activities, for example:


klaasnicolaas:
Only adding some information about the service and then we can merge it 😉

source/_integrations/harmony.markdown
```diff
@@ -95,6 +95,14 @@ Start an activity. Will start the default `activity` from configuration.yaml if
 | `entity_id`            |       no | Entity ID to target.
 | `activity`             |      yes | Activity ID or Activity Name to start.
 
+### Service `remote.toggle`
+
+..........
```
frenck, line 100: The description is missing, as pointed out by @klaasnicolaas already.

Please add one, thanks! 👍

frenck (changes requested):

source/_integrations/harmony.markdown
QbaF: Yes, but I suppose that the service firstly should start working correctly.
Maybe there should be an Issue reported? Sorry but I am very new here. And even with this supplementing content suggestion, I am not sure that I am in the right place. My intention was to draw someone's attention that I found service working not accordingly to my logic expectation and with missing documentation. I was asked to do that, it wasn't my idea;)

frenck: Oh, I'm sorry, I've missed a thing as well.

You are editing the Harmoney page, which is remote. All entities in Home Assistant inherit the services from the parent entity. So a Harmony gets all services from Remote.

However, we only document services to the places they belong to. So the service is `remote.toggle`, which is part of the remote integration, not the Harmony. It is usable due to Harmony being an remote, but it should not be documented on the Harmony integration.

The reasoning for this is: If we start doing that, we are duplicating a tremendous amount of documentation everywhere, which would be hard to keep in sync.

So, Thanks for opening up a PR, but I'm going to close this one.


If you think there is a bug with the service, please report it @ <https://github.com/home-assistant/home-assistant/issues>

Thanks! 👍

QbaF: I can understand. But in this case on this page shouldn't be also: `remote.turn_on`, `remote.turn_off`, `remote.send_command`. 
But in this case I suggest making better descriptions of this services on page [/integrations/remote/](https://www.home-assistant.io/integrations/remote/)

Indeed I think there is a bug in `remote.toggle` but until I can not find anywhere description how designer intended it to work, it is only my guess/logic conception.

frenck:
I agree that our documentation needs more love, and feel free to join in on that.

For general support questions, please use our community forum or Discord chat.
In case of a bug, use the reporting URL provided earlier.

PR #11253, opened 2019-11-22
remove 'from' clause for updater notification example

jamesgoodhouse:
**Description:**
Depending on the state of your system the state for the updater may be `off` or it may be `unavailable` (if the system has restarted recently, it'll be `unavailable`). Therefore, we can't assume the automation should on fire when going from `off` to `on`.

## Checklist:


jamesgoodhouse:
Moved to `current` branch in different pr (https://github.com/home-assistant/home-assistant.io/pull/11254)

PR #11254, opened 2019-11-22
remove 'from' clause for updater notification example

jamesgoodhouse:
**Description:**
Depending on the state of your system the state for the updater may be `off` or
it may be `unavailable` (if the system has restarted recently,
it'll be `unavailable`). Therefore, we can't assume the automation
should on fire when going from `off` to `on`.

## Checklist:

### DIFF
--- a/source/_integrations/updater.markdown
+++ b/source/_integrations/updater.markdown
@@ -70,7 +70,6 @@ automation:
   trigger:
     - platform: state
       entity_id: binary_sensor.updater
-      from: 'off'
       to: 'on'
   action:
     - service: notify.notify


frenck:
I do not totally agree. It depends a bit on where the notification is heading.

For example, a more extreme one, I absolutely don't want to have my SMS credits wasted each time I restart Home Assistant, while I already had that message.

jamesgoodhouse:
That’s a good callout. My problem is I’m missing the notification from time to time when tinkering and restarting.

Maybe a good option would to be adding a note or callout for the docs to state the edge case of not getting notified and that removing the from clause will result in always being notified, but that it will be noisier?

frenck:
I think this specific location of the docs is not the right place for that.

The examples provided on integration pages should have sane/common defaults. I believe it currently has.

Learning about automations and how they work, is something totally different. We cannot put that into each example we have.

jamesgoodhouse:
I can get on board with that. Thanks for your feedback @frenck!

PR #11261, opened 2019-11-23
Add integration here_weather

eifinger:
**Description:**

Adds documentation for the new integration `here_weather`

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#28910

## Checklist:


eifinger:
The different modes will result in different sensors being created when using the sensor platform.

Should I add an extensive list of which mode will create which sensors (information taken from https://developer.here.com/documentation/weather/dev_guide/topics/resource-type-weather-items.html) or should I just leave in the link in the `mode` description?

PR #11263, opened 2019-11-23
File name refactor and permissions revision

tiagofreire-pt:
**Description:**

## Checklist:

### DIFF
--- a/source/_docs/ecosystem/certificates/tls_self_signed_certificate.markdown
+++ b/source/_docs/ecosystem/certificates/tls_self_signed_certificate.markdown
@@ -15,28 +15,47 @@ Change to your Home Assistant [configuration directory](/getting-started/configu
 The certificate **must** be `.pem` extension.
 
 ```bash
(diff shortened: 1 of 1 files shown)


frenck:
So basically, this PR renames the file from `certificate.pem` to `fullchain.pem`.

Could you please elaborate on why? Throughout the system, we refer to it as the certificate file. Which, naming it `fullchain` might be extra confusing for people (SSL is hard for most already).

So am I missing something?

tiagofreire-pt:
@frenck, just harmonizing the names for the standard configuration and also the same for duckdns plug-in. I see no need for different names for this kind of certificates, creating some confusion around the topic for new users. 

With this approach both scenarios could be interchangeable, no matter the yaml used on the configuration.yaml file.

Also, the permissions should be 600 instead 755, allowing only the owner to read and write those files.

frenck (approved):
Thanks for clearing that up!

👍

PR #11301, opened 2019-11-26
New integration: aten_pe

mtdcr:
**Description:**
This documents the ATEN PE component for ATEN Rack PDUs

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#27960

## Checklist:

### DIFF
--- a/source/_integrations/aten_pe.markdown
+++ b/source/_integrations/aten_pe.markdown
@@ -8,9 +8,7 @@ ha_release: 0.103
 
 The `aten_pe` integration lets you control [ATEN Rack PDUs](https://www.aten.com/eu/en/products/energy-intelligence-pduupsracks/rack-pdu/) from Home Assistant.
 
-In order to use it, SNMP must be enabled on your PDU. For an outlet to appear in Home Assistant, its name must have been configured.
-
-It is recommended to use SNMPv3 to protect your credentials from eavesdropping.
+In order to use it, SNMP must be enabled on your PDU. It is recommended to use SNMPv3 to protect your credentials from eavesdropping.
 
 Tested devices:
   * [PE8324G](https://www.aten.com/eu/en/products/energy-intelligence-pduupsracks/rack-pdu/pe8324/)
--- a/source/_integrations/aten_pe.markdown
+++ b/source/_integrations/aten_pe.markdown
@@ -1,0 +1,54 @@
+---
+title: "ATEN Rack PDU"
+description: "Instructions on how to integrate ATEN Rack PDUs into Home Assistant."
+ha_category:
+  - Switch
+ha_release: 0.103
+---
+
+The `aten_pe` integration lets you control [ATEN Rack PDUs](https://www.aten.com/eu/en/products/energy-intelligence-pduupsracks/rack-pdu/) from Home Assistant.
+
+In order to use it, SNMP must be enabled on your PDU. For an outlet to appear in Home Assistant, its name must have been configured.
+
+It is recommended to use SNMPv3 to protect your credentials from eavesdropping.
+
+Tested devices:
+  * [PE8324G](https://www.aten.com/eu/en/products/energy-intelligence-pduupsracks/rack-pdu/pe8324/)
+
+To set it up, add the following information to your `configuration.yaml` file:
+
+```yaml
+switch:
+  - platform: aten_pe
+    host: 192.168.0.60
+```
+
+{% configuration %}
+host:
+  description: The IP/host which to control.
+  required: true
+  type: string
+port:
+  description: The port on which to communicate.
+  required: false
+  type: string
+  default: 161
+community:
+  description: community string to use for authentication (SNMP v1 and v2c).
+  required: false
+  type: string
+  default: private
+username:
+  description: Username to use for authentication.
+  required: false
+  type: string
+  default: administrator
+auth_key:
+  description: Authentication key to use for SNMP v3.
+  required: false
+  type: string
+priv_key:
+  description: Privacy key to use for SNMP v3.
+  required: false
+  type: string
+{% endconfiguration %}


source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
+---
+title: "ATEN Rack PDU"
+description: "Instructions on how to integrate ATEN Rack PDUs into Home Assistant."
```
frenck, line 3: Missing `logo` frontmatter option.
It makes a huge difference for a user on the documentation if there is a logo. Please consider adding one.

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+
+Tested devices:
+  * [PE8324G](https://www.aten.com/eu/en/products/energy-intelligence-pduupsracks/rack-pdu/pe8324/)
+
+To set it up, add the following information to your `configuration.yaml` file:
+
+```yaml
+switch:
+  - platform: aten_pe
+    host: mypdu.lan.
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

Secondly, I notice the period (`.`) at the end. Is that needed? It might need some additional notes on that in that case.

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+    authkey: !secret mypdu_authkey
```
frenck:
```suggestion
    authkey: MYPDU_AUTHORIZATION_KEY
```

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+    privkey: !secret mypdu_privkey
```
frenck:
```suggestion
    privkey: MYPDU_PRIVATE_KEY
```

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+  description: community string to use for authentication (SNMP v1 and v2c).
+  required: false
+  type: string
+  default: 'private'
+username:
+  description: Username to use for authentication.
+  required: false
+  type: string
+  default: 'administrator'
+auth_key:
```
frenck, line 44: Key name seems incorrect.

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+  description: Authentication key to use for SNMP v3.
+  required: false
+  type: string
+  default: no key
+priv_key:
```
frenck, line 48: Key name seems incorrect.

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+host:
+  description: The IP/host which to control.
+  required: false
+  type: string
+  default: 'localhost'
+port:
+  description: The port on which to communicate.
+  required: false
+  type: string
+  default: '161'
```
frenck:
```suggestion
  default: 161
```

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+    authkey: !secret mypdu_authkey
+    privkey: !secret mypdu_privkey
+```
+
+{% configuration %}
+host:
+  description: The IP/host which to control.
+  required: false
+  type: string
+  default: 'localhost'
```
frenck:
```suggestion
  default: localhost
```

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+port:
+  description: The port on which to communicate.
+  required: false
+  type: string
+  default: '161'
+community:
+  description: community string to use for authentication (SNMP v1 and v2c).
+  required: false
+  type: string
+  default: 'private'
```
frenck:
```suggestion
  default: private
```

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+username:
+  description: Username to use for authentication.
+  required: false
+  type: string
+  default: 'administrator'
```
frenck:
```suggestion
  default: administrator
```

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+auth_key:
+  description: Authentication key to use for SNMP v3.
+  required: false
+  type: string
+  default: no key
```
frenck:
```suggestion
```

source/_integrations/aten_pe.markdown
```diff
@@ -0,0 +1,59 @@
...
+priv_key:
+  description: Privacy key to use for SNMP v3.
+  required: false
+  type: string
+  default: no key
```
frenck:
```suggestion
```

frenck (changes requested):
Reviewed your PR @mtdcr, and left some comments and suggestions. Could you please take a look? Cool, thanks! 👍

source/_integrations/aten_pe.markdown
mtdcr: There may be copyright or trademark issues involved, which I'd like to avoid.

source/_integrations/aten_pe.markdown
mtdcr: While `host` is optional, the only sane default can be `localhost`, which will in turn be used only rarely with this integration, i.e. only when port forwarding is involved.

A period at the end of a host marks a FQDN. This is standard syntax in DNS. It allows the resolver to skip iterating over search domains. It's not required for or special to this integration. I can remove it though, if you think it causes confusion.

source/_integrations/aten_pe.markdown
mtdcr: Right. I copied and pasted the options from the SNMP switch documentation, but the code deviates from it. I think I should adjust the code and the example above instead.

mtdcr:
@frenck: I updated the documentation according to your remarks (and aligned the code, too). Please take a look at the comments I left at the two unresolved hunks.

source/_integrations/aten_pe.markdown
frenck: If the host is optional in the integration, it need to be removed from the initial example. Optional configuration options are not allowed in the initial and minimal example.

I understand the `.`, however, consider a non-technical user, as this is documentation.

Considering your comment, I'm kinda thinking `host` should not be optional.

source/_integrations/aten_pe.markdown
frenck: The logo is not used for commercial usage and only services for identification purposes.

frenck: The logo is not used for commercial usage and only serves for identification purposes.

mtdcr: I can follow your argument, but I am not a lawyer, and I don't even know which jurisdictions would apply. A person who's comfortable with adding a logo can do it later. I already learned my lessons the hard way.

source/_integrations/aten_pe.markdown
mtdcr: I'm going to make host non-optional and use the default ip-address of the device in the example.

mtdcr: Btw., auth_key and priv_key are also optional (even though good practice). I'll remove them.

frenck:
In general, please don't squash and force push.
It is hard for reviewers to see what changed this way.
Thanks!

frenck (approved):

PR #11305, opened 2019-11-27
Add initial list of compatible apps to Mobile App page

SeanPM5:
**Description:**
When you try to set up the Mobile App integration from within HA, the config flow says "Open the mobile app to set up the integration with Home Assistant. **See the docs for a list of compatible apps.**"

It links here and there is no list of compatible apps, so I am rectifying that now. Just adding the two official apps for now since I'm not 100% sure which third party apps use `mobile_app:` (I think Ariela, HA Client, HassApp?). Can be expanded upon later.

## Checklist:


frenck (approved):
Thanks, @SeanPM5! 👍

PR #11356, opened 2019-12-03
Remove hardware that is not recommended

cogneato:
I think these should be removed for a better user experience. If it isn't recommended then they shouldn't be listed. Maybe keep them in the assets for people who like to dig around?

**Description:**
Remove hardware from the list which encourages a bad experience.

## Checklist:


frenck (approved):
Agreed. 👍  Thanks!

PR #11396, opened 2019-12-08
Add stackdriver documentation

frog32:
**Description:**
Add documentation for stackdriver

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#29590

## Checklist:

- [x] Add image


source/_integrations/google_stackdriver.md
```diff
@@ -0,0 +1,52 @@
...
+description: "Setup for Google Stackdriver integration"
+logo: google_stackdriver.png
+ha_category:
+  - History
+ha_release: 0.104
+---
+
+The `google_stackdriver` integration allows you to stream logs to [Google Stackdriver Logging](https://cloud.google.com/logging/). The current [free tier](https://cloud.google.com/free/) of GCP should allow you to log up to 50 GiB.
+
+## First time setup
```
frenck:
```suggestion
## First-time setup
```

source/_integrations/google_stackdriver.md
```diff
@@ -0,0 +1,52 @@
+---
+title: "Google Stackdriver"
+description: "Setup for Google Stackdriver integration"
+logo: google_stackdriver.png
+ha_category:
+  - History
+ha_release: 0.104
```
frenck: `ha_iot_class` is missing.

source/_integrations/google_stackdriver.md
```diff
@@ -0,0 +1,52 @@
...
+key_file:
+  description: The filename of the Google Service Account JSON file.
+  required: true
+  type: string
+level:
+  description: Log level that should be streamed to Stackdriver.
+  required: false
+  type: string
+labels:
+  description: Labels to add to every log entry. If streaming logs from multiple sources to Stackdriver this allows them to be separated by using filters.
```
frenck:
```suggestion
  description: Labels to add to every log entry. If streaming logs from multiple sources to Stackdriver, this allows them to be separated by using filters.
```

source/_integrations/google_stackdriver.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Configuration
+
+Add the following lines to your `configuration.yaml` file:
+
+```yaml
+# Example configuration.yaml entry
+google_stackdriver:
+  key_file: frog32ch_key.json
+  labels:
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_integrations/google_stackdriver.md
```diff
@@ -0,0 +1,52 @@
...
+  required: false
+  type: string
+labels:
+  description: Labels to add to every log entry. If streaming logs from multiple sources to Stackdriver this allows them to be separated by using filters.
+  required: false
+  type: map
+{% endconfiguration %}
+
+<div class='note warning'>
+Depending on how much logging is done by your setup and the log level configured for Stackdriver this could exhaust the free tier. To avoid getting surprises it is recommended to set up an <a href="https://cloud.google.com/stackdriver/pricing#alert-usage">alert</a> to get notified before the free tier is exhausted.
```
frenck: Please use Markdown as much as possible.

```suggestion

Depending on how much logging is done by your setup and the log level configured for Stackdriver, this could exhaust the free tier. To avoid getting surprises it is recommended to set up an [alert](https://cloud.google.com/stackdriver/pricing#alert-usage) to get notified before the free tier is exhausted.  

```

source/_integrations/google_stackdriver.md
```diff
@@ -0,0 +1,52 @@
...
+## First time setup
+
+This assumes you already have a Google Cloud project. If you don't, please create one in the [Google Cloud Console](https://console.cloud.google.com/projectcreate)
+
+You need to create a Service Account key in the [Google Cloud API Console](https://console.cloud.google.com/apis/credentials/serviceaccountkey)
+- Choose a new "New Service Account", give it a name and leave the key type as JSON
+- Select the role: Logs Writer 
+
+This will download the Service Account JSON key to your machine. Do NOT share this with anyone. Place this file in your Home Assistant config folder.
+
```
frenck, line 22: Unneeded empty line.
```suggestion
```

frenck (changes requested):
Thanks for the PR @frog32!

I've reviewed it and left some comments. Could you please take a look? Thanks 👍

source/_integrations/google_stackdriver.md
frog32: I tried that at first but it does not work within the warning div. If you know how to create a warning with pure markdown I am happy to change that.

source/_integrations/google_stackdriver.md
frog32: I am not sure what's the right class. Certainly cloud but neither push nor polling is true since data (logs not even entity related data) only flows from Home Assistant to Stackdriver so no need to push or poll. The logs are transfered in batches so the right thing would be "batched push" but will change it to "Cloud Push".

source/_integrations/google_stackdriver.md
frog32: OK went with a minimal example. People using Stackdriver already would probably understand how to come up with valid yaml so no extended example needed.

frog32:
ha_iot_class and the markdown warning need further feedback the rest should be done.

source/_integrations/google_stackdriver.md
frenck: The above-suggested change can be applied and will work.

Markdown is only able to switch context between HTML & Markdown if there is an empty line. Hence, a new line after the opening div and before the closing div.

frog32:
Thanks @frenck for the review and clarifying the html/markdown issue.

frenck (dismissed):

PR #11409, opened 2019-12-09
Add note to alert about new NS Api

michaeldeborst:
You currently can not set up the NS integration due to api incompatibility for new users

**Description:**

## Checklist:


frenck:
We don't put these types of notices in the documentation. It is hard to maintain and keep up with.

We have our alert service for it, which is also shown on the documentation website dynamically.

<https://alerts.home-assistant.io>

PR #11413, opened 2019-12-09
Add example of condition inside action

bbbburns:
Added an example of a condition inside the action and a note that manual trigger bypasses the condition. Based on conversation in: https://community.home-assistant.io/t/automation-with-condition-but-always-triggering/120792/14

**Description:**

## Checklist:


frenck:
While this is a nice workaround, once should not put conditions inside the actions.

So for general documentation, we should stick with the best practices and therefore not add this.

Nevertheless, thanks for your contribution, that is always appreciated! 👍 
I've learned something 😉

PR #11444, opened 2019-12-12
Improve iCloud setup instructions

AlecRust:
**Description:**

Some confusions in the iCloud docs I encountered when setting it up:
- That this integration creates a device tracker for every device in Find My
- Still referenced Find My iPhone service instead of Find My
- Wasn't clear how the 2 Step Authentication code would be provided by the user

Have tidied these things up.

## Checklist:


frenck (approved):
Thanks, @AlecRust! 👍

PR #11452, opened 2019-12-14
rflink deprecation and other doc fixes

scop:
**Description:**

See individual commit messages for details.

home-assistant/home-assistant#29972 is related, but this can go in already now without waiting that to be applied.

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#8348 (old)

## Checklist:

### DIFF
--- a/source/_integrations/light.rflink.markdown
+++ b/source/_integrations/light.rflink.markdown
@@ -99,18 +99,6 @@ devices:
           required: false
           default: true
           type: boolean
-        aliasses:
-          description: (**deprecated**) Alternative RFLink ID's this device is known by.
-          required: false
-          type: [list, string]
-        group_aliasses:
-          description: "(**deprecated**) `aliases` which only respond to group commands."
-          required: false
-          type: [list, string]
-        no_group_aliasses:
-          description: "(**deprecated**) `aliases` which do not respond to group commands."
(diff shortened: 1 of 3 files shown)


frenck (changes requested):
Since this change relies on a change that is pending the `next` Home Assistant release, the documentation PR should be targeting the `next` branch.

Please change the target branch for this PR to `next` and rebase if needed.

scop:
As I tried to note in the initial comment, in my opinion this does _not_ depend on the HA change. It is related, but we can merge this already now to current.

frenck:
The `current` documentation should reflect the current state. By merging this, it doesn't do that. So please adjust to `next`.

scop:
Sure it does reflect the current state, there's no reason to promote/document the deprecated values any more. The docs as modified by this PR work right away with the unchanged, current version of HA. And note that before this change, the documentation is misleading (incorrectly marked deprecations) for both current and next.

I can rebase to next if that's what it takes to get this in (please confirm), but frankly to me that is doing some (a tiny bit of) work which will only serve to delay user access to the corrected docs, i.e. that's time spent to make things worse, for no gain.

scop:
Meh, rebased.

frenck (approved):

PR #11456, opened 2019-12-14
add raw to the delay template

Mariusthvdb:
add raw to the delay template. (credits for this to: @lddubeau )
might need something extra: right now, when visiting this page, no value is displayed for the delay, which wouldn't be correct, since it could suggest it to be valid syntax to do so.

Schermafbeelding 2019-12-14 om 16 50 30

Not sure how to change that in the html of this page, so ask for suggestion on behalf of the team.

please see https://community.home-assistant.io/t/confused-about-script-field-template-in-docs/155023/7?u=mariusthvdb for more info here

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/script.markdown
+++ b/source/_integrations/script.markdown
@@ -89,7 +89,7 @@ script: 
           brightness: 100
       - delay:
           # supports seconds, milliseconds, minutes, hours
-          minutes: {{ minutes }}
+          minutes: {% raw %}{{ minutes }}{% endraw %}
       - alias: Living room lights on
         service: light.turn_on
         data:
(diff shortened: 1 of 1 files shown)


source/_integrations/script.markdown
```diff
@@ -89,7 +89,7 @@ script: 
           brightness: 100
       - delay:
           # supports seconds, milliseconds, minutes, hours
-          minutes: {{ minutes }}
+          minutes: {% raw %}{{ minutes }}{% endraw %}
```
frenck: Please put the raw tags around the whole code fenced block, not just inline.

frenck (changes requested):
Thanks, @Mariusthvdb!

Just a small note! Could you take a look? 👍

frenck (approved):

PR #11461, opened 2019-12-15
Add Apprise file attachment support

caronc:
**Description:**

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#29952

## Checklist:

### DIFF
--- a/source/_integrations/apprise.markdown
+++ b/source/_integrations/apprise.markdown
@@ -14,28 +14,48 @@ To use Apprise supported notifications, add the following to your `configuration
 ```yaml
 # Example configuration.yaml entry using URLs
 notify:
-  - platform: apprise
+  - name: NOTIFIER_NAME
+    platform: apprise
     url: YOUR_APPRISE_URLS
 ```
 
-You can also pre-define your own configuration files while storing them either remotely or locally. Simply just use the `config` option.
+Apprise URLs may seem strange at first, but they aren't complicated to create at all. They provide all of the information needed to activate a notification service in one single web-like URL. You specify more than one (causing them all to be notified) by just separating each URL with a comma and/or space.  To see how you can construct your own URL(s) [click here](https://github.com/caronc/apprise/wiki). For example, an [Apprise E-mail URL](https://github.com/caronc/apprise/wiki/Notify_email) configuration entry might look like this:
+
+```yaml
+# An Apprise E-Mail Example (using Google Mail)
+notify:
+  - name: by_email
+    platform: apprise
+    url: mailto://chuck:pass123@gmail.com
+```
+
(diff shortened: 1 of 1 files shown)


source/_integrations/apprise.markdown
```diff
@@ -14,28 +14,48 @@ To use Apprise supported notifications, add the following to your `configuration
 ```yaml
 # Example configuration.yaml entry using URLs
 notify:
-  - platform: apprise
+  - name: NOTIFIER_NAME
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_integrations/apprise.markdown
```diff
@@ -14,28 +14,48 @@ To use Apprise supported notifications, add the following to your `configuration
...
+    platform: apprise
     url: YOUR_APPRISE_URLS
 ```
 
-You can also pre-define your own configuration files while storing them either remotely or locally. Simply just use the `config` option.
+Apprise URLs may seem strange at first, but they aren't complicated to create at all. They provide all of the information needed to activate a notification service in one single web-like URL. You specify more than one (causing them all to be notified) by just separating each URL with a comma and/or space.  To see how you can construct your own URL(s) [click here](https://github.com/caronc/apprise/wiki). For example, an [Apprise E-mail URL](https://github.com/caronc/apprise/wiki/Notify_email) configuration entry might look like this:
```
frenck: We have now so much text above the `configuration` options table/list. That is not how it should be.

Please move the `configuration` table part right below the first example and put all other stuff below that.

frenck (changes requested):
Hi there, @caronc!

Thanks for the PR, I've looked at it, and left some comments. Could you please take a look? Thanks! 👍

source/_integrations/apprise.markdown
caronc: Thank you for your fast review!  I added the name: because i was having a look at how the [Slack Documentation](https://www.home-assistant.io/integrations/slack/) looked.  Regardless, I dropped the name as requested! :slightly_smiling_face:

source/_integrations/apprise.markdown
caronc: Done!

source/_integrations/apprise.markdown
frenck: Thanks! Yeah, we haven't been to strict in the past, so a lot to be done! 😉

frenck (approved):

PR #11473, opened 2019-12-16
Update on onyo.markdown

inputd:
**Description:**
The default source list for the Onkyo integration does not include all sources, and there is an easy way to determine exact source names. In my case, even though my receiver displays "Blu-ray/DVD" on screen, the "bd-dvd" source value was incorrect, and the correct input was "bd", which isn't in the default list, and is also pretty counter-intuitive.

## Checklist:


frenck (approved):
Thanks, @inputd! 👍

PR #11492, opened 2019-12-18
Update light.markdown

Jerommeke4:
The light.turn_on service and light.turn_off service nowadays seem to require the entity_id(s) to be set.

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/light.markdown
+++ b/source/_integrations/light.markdown
@@ -24,7 +24,7 @@ Most lights do not support all attributes. You can check the integration documen
 
 | Service data attribute | Optional | Description |
 | ---------------------- | -------- | ----------- |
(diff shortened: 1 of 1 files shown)


frenck (approved):
Technically, it is not. In our codebase it is optional ([reference](https://github.com/home-assistant/home-assistant/blob/9e5de1a106b934100c98aac1764f8a7a4b713fd5/homeassistant/helpers/config_validation.py#L727)). But I get what you are saying.

Thanks, @Jerommeke4. Let's merge this in.

frenck:
Oh, wait, second thought, right before I wanted to hit merge.

It is optional! It is not just an entity_id, it could be an area as well! In that case, the entity is not needed.

So, uh, I guess the current docs are correct after all!

PR #11502, opened 2019-12-20
Updated mention of services in top text to reflect new service names

atxbyea:
**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/vacuum.xiaomi_miio.markdown
+++ b/source/_integrations/vacuum.xiaomi_miio.markdown
@@ -20,7 +20,11 @@ Currently supported services are:
 - `clean_spot`
 - `set_fan_speed`
 - remote control of your robot.
-- `xiaomi_clean_zone`
+- `xiaomi_miio.vacuum_clean_zone`
+- `xiaomi_miio.vacuum_remote_control_start`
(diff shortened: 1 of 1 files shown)


source/_integrations/vacuum.xiaomi_miio.markdown
```diff
@@ -20,7 +20,11 @@ Currently supported services are:
 - `clean_spot`
 - `set_fan_speed`
 - remote control of your robot.
-- `xiaomi_clean_zone`
+- `xiaomi_miio.vacuum_clean_zone`
```
frenck, line 23: Why are these prefixed by the domain and the ones already in the docs not?

frenck (changes requested):
Left a small question, @atxbyea, could you help me answer that?

frenck:
@atxbyea Ping! 😄

atxbyea:
good question @frenck I just saw that list in the top of the docs, and saw that it referenced the "old" service name, and changed it to the "new" service name, probably it could just say `clean_zone` to be in line with the rest of them ?

frenck:
@atxbyea There is no policy on that, however, I think we should be consistent on the same page at least 😉

frenck:
Closing PR since it is running stale.

PR #11505, opened 2019-12-20
Update MS SQL Server documentation with info on ODBC Drivers

datafx:
Since switching from pymssql to pyodbc, we now need to use ODBC drivers to connect to SQL server.


frenck (approved):
Thanks, @datafx! 👍

PR #11507, opened 2019-12-20
Added missing quotes for value_template value

cinatic:
**Description:**

added missing quotes for value_template in numeric_state example (https://www.home-assistant.io/docs/scripts/conditions/#numeric-state-condition)

### DIFF
--- a/source/_docs/scripts/conditions.markdown
+++ b/source/_docs/scripts/conditions.markdown
@@ -206,7 +206,7 @@ The template condition tests if the [given template][template] renders a value e
 ```yaml
 condition:
   condition: template
-  value_template: "{% raw %}'{{ (state_attr('device_tracker.iphone', 'battery_level')|int) > 50 }}'{% endraw %}"
+  value_template: "{% raw %}{{ (state_attr('device_tracker.iphone', 'battery_level')|int) > 50 }}{% endraw %}"
 ```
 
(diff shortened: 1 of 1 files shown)


source/_docs/scripts/conditions.markdown
```diff
@@ -206,7 +206,7 @@ The template condition tests if the [given template][template] renders a value e
 ```yaml
 condition:
   condition: template
-  value_template: "{% raw %}{{ (state_attr('device_tracker.iphone', 'battery_level')|int) > 50 }}{% endraw %}"
+  value_template: "{% raw %}'{{ (state_attr('device_tracker.iphone', 'battery_level')|int) > 50 }}'{% endraw %}"
```
frenck: If you take away the `{% raw %]` tags, then this is what is left:

```
  value_template: "'{{ (state_attr('device_tracker.iphone', 'battery_level')|int) > 50 }}'"
```

Why does it need a quotation twice?

frenck (changes requested):

source/_docs/scripts/conditions.markdown
frenck: PS: it helps if you used the PR template that was provided when you opened up the PR. Since it will ask for a description, which is currently now missing from your PR. Could you elaborate?

frenck (approved):
Thanks,  @cinatic 👍

PR #11513, opened 2019-12-21
Update featured integrations

springstan:
**Description:**

- add more integrations to the featured section to reflect current state of Home Assistant and to show off more/popular smart home appliances/devices
- discussion about which integrations should be featured is welcome 😊

**Result:**
screenshot-components

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#30134<home-assistant PR number goes here>

## Checklist:

### DIFF
--- a/source/_integrations/almond.markdown
+++ b/source/_integrations/almond.markdown
@@ -6,7 +6,6 @@ ha_category:
   - Voice
 ha_iot_class: Local Polling
 ha_release: "0.102"
-featured: true
 ---
 
 [Almond](https://almond.stanford.edu/) is an open, privacy-preserving virtual assistant by [Stanford Open Virtual Assistant Lab](https://oval.cs.stanford.edu/). It allows you, among other things, to control Home Assistant using natural language. Once installed, it will be available on Lovelace via the microphone icon in the top right.
--- a/source/_integrations/almond.markdown
+++ b/source/_integrations/almond.markdown
@@ -6,6 +6,7 @@ ha_category:
   - Voice
 ha_iot_class: Local Polling
 ha_release: "0.102"
+featured: true
 ---
 
 [Almond](https://almond.stanford.edu/) is an open, privacy-preserving virtual assistant by [Stanford Open Virtual Assistant Lab](https://oval.cs.stanford.edu/). It allows you, among other things, to control Home Assistant using natural language. Once installed, it will be available on Lovelace via the microphone icon in the top right.
--- a/source/_integrations/homekit.markdown
+++ b/source/_integrations/homekit.markdown
@@ -5,6 +5,7 @@ ha_category:
   - Voice
 ha_release: 0.64
 logo: apple-homekit.png
+featured: true
 ---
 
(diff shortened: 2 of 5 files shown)


frenck:
TBH if we add featured integration, they should be at least platinum level on the QA scale.

"show off what more/popular smart home appliances/devices" -> So, what did you base that on? Since we don't have that data, which makes me interested.

MartinHjelmare:
I don't think we can use the quality scale to select what featured integrations to show, if the aim is to show popular integrations. Popularity among integrations isn't coupled to our quality scale unfortunately. If we would only go by quality scale it would be a meager selection since we have so few platinum integrations.

springstan:
> TBH if we add featured integration, they should be at least platinum level on the QA scale.

Good point, however most integrations do not list a quality scale, so it is rather difficult to select the best ones.

> "show off what more/popular smart home appliances/devices" -> So, what did you base that on? Since we don't have that data, which makes me interested.

TBH mostly based on my opinion. I wanted to add integrations that complement e.g. Homekit to Google Assistant or complete the feature integrations, e.g. Roomba/Unifi/Almond.
Roomba/Unifi/Almond are integrations that demonstrate the variety of Home Assistant's integrations.

frenck:
I do agree with both, and I'm intentionally triggering it a bit. Since well, we don't know what is popular, at the same time, we don't have many platinum integrations.

That said, marking and listing an integration as featured, is easily considered/interpreted as endorsed by a user. Therefore, the quality of the integration does matter somewhat as it the popularity.

For that reason, I would say Almond won't qualify as either of those at this moment? 

I hope I make sense with this brain rambling as it is Sunday morning for my head still... 😉

springstan:
> I do agree with both, and I'm intentionally triggering it a bit.

Questioning the reasoning of someone is always a good idea 👍

I completely agree, that these featured integrations could be easily interpreted in the two ways you mentioned. 
Sure I can remove Almond, I know it is still a work in progress but it seems very promising.
Do you have any particular integration in mind to replace Almond?

> I hope I make sense with this brain rambling as it is Sunday morning for my head still... 😉

No worries, you make sense to me 😄

frenck:
Looking at the list, and I really don't know.

[Somfy](https://www.home-assistant.io/integrations/somfy) maybe? It ain't platinum, but does work with the new OAuth2 authentication, which is easy to use. It is also a worldwide known and used product.


Some other rules/conditions to take into account: Is the integration instantly/usable by a new user? Does it have a config flow? Can it be used without requesting developer tokens from a third party?

springstan:
Yeah it is really difficult to pick one out of so many.

Good idea, I have been thinking about Somfy as well. I will replace Almond with Somfy in a new commit.

> Some other rules/conditions to take into account: Is the integration instantly/usable by a new user? Does it have a config flow? Can it be used without requesting developer tokens from a third party?

Do these rules only apply to additional integrations or to all featured integrations? Since some of them are harder to set up than others.

frenck:
There are no hard rules on this... But I think it is definitely something to take into consideration.

Let's add Somfy for now, and be done and look for a future iteration.

fabaff:
IMHO, we should keep it 4x4. This work on 1920x1080 at least without scrolling and scrolling through 16 integration on a mobile device is already way too long.

frenck:
Good thinking @fabaff! 👍

springstan:
Fair point @fabaff!
Should the current selection of featured integrations be discussed?

springstan:
Closing this PR in agreement with https://github.com/home-assistant/home-assistant.io/pull/11513#issuecomment-568259086

PR #11514, opened 2019-12-22
broadlink.send service in broadlink integration doc should include switch note

sprilukin:
Description for broadlink.send service should include a note that broadlink switch entity still should be created even if user plan to use only broadlink.send. Otherwise it's not obvious from the documentation.

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/broadlink.markdown
+++ b/source/_integrations/broadlink.markdown
@@ -464,6 +464,8 @@ script:
             - "JgBGAJSTFDUUNhM2ExITEhMSExITEhM2EzYTNhQRFBEUERQRFBEUNRQ2ExITNhMSExITNhMSExITEhM2ExITNhQ1FBEUNhMADQUAAA=="
 ```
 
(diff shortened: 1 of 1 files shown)


frenck:
I'm kinda thinking about not adding this.
The reasoning is that this goes for almost any service... And to be honest, I don't feel like we should add that note to each service call we document.

Or am I missing something here?

sprilukin:
I'm actually a new homeasisstant user - i'm using it just for a week.
I read different related docs but did found something similar to what i expect.
I had to debug the application to find the reason.

Is there any reference in doc which describes this behavior?
If so - this issue could be closed (link to the doc would be appreciated)

frenck:
I'm not sure of a reference to the docs, but services add functionality to entities, devices, and integrations.
When you do not add an integration, you obviously do not get the functionality for that integration available.

Oh! Welcome to the Home Assistant community!
Be sure to join our Discord chat server, as there are many people in there, that can quickly help you answer questions 👍

PR #11518, opened 2019-12-22
Aadd extract instruction to getting started guide

muhasaho:
It is not obvious that the downloaded file needs to be extracted before flashing it. I assumed balenaEtcher did it automatically and spent an entire afternoon trying to figure out why the raspberry pi was not booting up. Hopefully, this change will prevent someone from making the same mistake in the future.

**Description:**

## Checklist:

### DIFF
--- a/source/getting-started/index.markdown
+++ b/source/getting-started/index.markdown
@@ -42,7 +42,7 @@ We will need a few things to get started with installing Home Assistant. The Ras
 
 ### Software requirements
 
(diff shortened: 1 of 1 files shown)


frenck:
One should not have to extract it, but there has been issues with it in the past, so I agree. It surely cannot harm it.

Thanks, @muhasaho! 👍

frenck (approved):

PR #11564, opened 2019-12-29
Quotation marks are missing

PalmMx:
The hvac_mode state should be entered with quotation marks, like "heat".

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/climate.markdown
+++ b/source/_integrations/climate.markdown
@@ -95,7 +95,7 @@ automation:
       data:
         entity_id: climate.kitchen
         temperature: 24
-        hvac_mode: heat
+        hvac_mode: "heat"
 ```
 
 ```yaml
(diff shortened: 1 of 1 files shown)


frenck:
That seems unnecessary?

Those are perfectly normal strings, that are not ambiguous with any other value type, so are just interpreted by the parser as strings, quotes will not make a difference.

Could you elaborate more on your change and why?

PalmMx:
I did some automations with the following action to set the hvac_mode to "heat" or "off" respectively.

This one didn't work:
```
action:
  - service: climate.set_hvac_mode
    data:
      entity_id: climate.xxx
      hvac_mode: off
```

But this one did:
```
action:
  - service: climate.set_hvac_mode
    data:
      entity_id: climate.xxx
      hvac_mode: "off"
```

Without quotation marks I got an error in the HA log saying the action couldn't be executed.

FYI - the climate entities come from the generic_thermostat configuration where the initial hvac_mode has to be set with quotation marks as well. Maybe this is related to my issue?

frenck:
`off` has special meaning in YAML and works like a boolean (evaluates to `false`).

However, your changes are not related to `off`.

PalmMx:
Ok, I see.
Then maybe one of the examples should be with “off” instead of just heat.
But nevermind.

PR #11574, opened 2019-12-30
Correct SimpliSafe website link

bachya:
**Description:**

https://github.com/home-assistant/home-assistant.io/pull/11563 makes a somewhat nonsensical change to the SimpliSafe docs (turns "SimpliSafe home security" into a link to the SimpliSafe lock). Not sure what the intention is, but this isn't correct. This PR uses the base SimpliSafe URL.

**Pull request in home-assistant (if applicable):** N/A

## Checklist:

### DIFF
--- a/source/_integrations/simplisafe.markdown
+++ b/source/_integrations/simplisafe.markdown
@@ -8,7 +8,7 @@ ha_category:
   - Lock
 ---
 
-The `simplisafe` integration integrates [SimpliSafe home security](https://simplisafe.com/smart-lock)  (V2 and V3) systems into Home Assistant. Multiple SimpliSafe accounts can be accommodated.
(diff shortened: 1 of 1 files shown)


frenck:
Instead of doing a fully revert, you could have also just adjusted the URL to the homepage?

bachya:
We're reverting the previous commit. Now, when this is squashed into `current`, it will accurate represent that we reverted a previous change, rather than adding _another_ commit to return to a previous state. It will get you the same end result.

frenck:
@bachya? Same end result? I'm proposing a different end result.
I suggest we correct the URL to the main website of the manufacturer, instead of removing the link completely.

I fail to see how that relates to reverting this PR. Removing it by reverting it, and next add it in again with another PR, is not going to make things any clearer and causes more commits than actually needed (and more reviews as well).

Do you have any objections to just linking it to <https://simplisafe.com> ?

bachya:
@frenck I see. Thanks for clarifying. I can do that.

frenck (approved):
Thanks, @bachya! 👍

PR #11585, opened 2019-12-31
Update Freenas docs

tprelog:
**Description:**

I've an updated version of the rc.d script which has not made it to my FreeNAS resource that the current script is based upon. It does need bash installed though. 

And why not just use bash anyways? At least for the HA user. It helps keep things in line with other virtualenv instruction. I'm not trying to offend anyone but you don't see `activate.csh`, in docs for other things like AppDaemon or HASS-Configurator for example. I think in the end it's just easier to make a small change here to avoid frustration when people are trying follow a Linux guide down the road.

I suggested a few other pkgs to install from the start as well. I'm all about minimum required but having these few extras, I think will go a long way to ease some frustration when starting to add your components into HA.

## Checklist:

### DIFF
--- a/source/_docs/installation/freenas.markdown
+++ b/source/_docs/installation/freenas.markdown
@@ -13,13 +13,6 @@ Enter the Home Assistant jail. If you don't know which name you have given the j
 iocage exec HomeAssistant
 ```
 
-Create the user and group that Home Assistant will run as. The user/group ID of `8123` can be replaced if this is already in use in your environment.
-
-```bash
-pw groupadd -n homeassistant -g 8123
-echo 'homeassistant:8123:8123::::::/bin/csh:' | adduser -f -
-```
-
 Install the sugessted packages and virtualenv:
 
 ```bash
@@ -29,6 +22,13 @@ pkg install -y autoconf bash ca_root_nss gmake pkgconf python37 python37
 python3.7 -m ensurepip
 pip3 install --upgrade pip
 pip3 install --upgrade virtualenv
+```
+
+Create the user and group that Home Assistant will run as. The user/group ID of `8123` can be replaced if this is already in use in your environment.
+
+```bash
+pw groupadd -n homeassistant -g 8123
+echo 'homeassistant:8123:8123::::::/usr/local/bin/bash:' | adduser -f -
 ```
 
 Create the installation directory:
@@ -44,7 +44,7 @@ Install Home Assistant itself:
 su homeassistant
 cd /usr/local/share/homeassistant
 virtualenv -p python3.7 .
-source ./bin/activate.csh
+source ./bin/activate
 pip3 install homeassistant
 ```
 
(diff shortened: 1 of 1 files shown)


klaasnicolaas:
Can you please rebase this PR to the `current` branch?

tprelog:
I'll be honest here. I only understand some basics of the git life but I'm gonna give this a shot. 

I hope I don't break the internet :rofl:

klaasnicolaas:
If you dont't know how to do, we can also do it for you 😉

tprelog:
Yes please! At least this time. :smile: 

I would like to continue contributing in the future so I'll learn how to do this myself. In the future can I avoid this all together by always making pull request of this type against the current branch? I understand (mostly) how to fetch and merge upstream changes. It's the rebasing part I don't know.

_thank you_ for taking care of it this time!

frenck:
Rebased onto the `current` branch.

source/_docs/installation/freenas.markdown
```diff
@@ -13,24 +13,24 @@ Enter the Home Assistant jail. If you don't know which name you have given the j
...
+Install the suggested packages and virtualenv:
 
 ```bash
 pkg update
 pkg upgrade
-pkg install -y python37 py37-sqlite3 ca_root_nss
+pkg install -y autoconf bash ca_root_nss gmake pkgconf python37 py37-sqlite3
 python3.7 -m ensurepip
 pip3 install --upgrade pip
 pip3 install --upgrade virtualenv
```
frenck: So why are we still advertising `virtualenv` instead of the build-in `venv` ?

tprelog: That's a good question. I didn't even think about `venv`.

I think the first time I used a virtualenv, the guide I was following must have used `virtualenv`. Typically, I do prefer "minimum you can install" but never considered `virtualenv` to add any considerable overhead so I never looked at the alternatives. That has to be at least a few years in the past by now. So I guess it's just old habits. "If it ain't broke..." I never had a reason to switch.

If you think this should be changed then maybe we should change it. I really don't know any difference so I'd like to ask is there an advantage to `venv` beside it's just built in? I assume everything would still work the same, correct? 

On the opposite side is there a disadvantage to `virtualenv`

Thanks

frenck: There is no real disadvantage, especially if you consider our project is Python3 only.

An advantage is that you do not need an additional package (unless it is explicitly packaged by the package manager in a separate package, so that needs checking). Furthermore, `venv` does not copy in the Python binaries into the `venv`.

tprelog: I changed it over to venv.

Thanks for explaining, I changed my install script as well. Shaved some seconds off. :smile:

frenck (approved):
Nice!

This also makes it a lot closer to our `venv` install manuals, which is awesome!

Thanks, @tprelog! 👍 


Happy New Year 🍾

PR #11616, opened 2020-01-03
Update config to new Samsung TV integration

escoand:
**Description:**
Update description to new configuration options.

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#28306

## Checklist:


source/_integrations/samsungtv.markdown
```diff
@@ -20,39 +20,33 @@ To add a TV to your installation without relying on the [discovery component](/i
 
 ```yaml
-media_player:
-  - platform: samsungtv
-    host: IP_ADDRESS
+samsungtv:
+  - host: IP_ADDRESS
+    name: My Samsung TV
```
klaasnicolaas: In the first example we want only required config variables, please remove everything that is optional here.
You can make a more advanced example above the `supported models` part 😉

klaasnicolaas (changes requested):

frenck:
@escoand My main concern with this PR right now, is that it has no mention of configuring/setting up the integration via the integrations panel in the frontend.

I think that should be added. Also, add the `ha_config_flow: true` to the front matter of the page, to indicate this integration has a config flow.

frenck:
Maybe I'm missing something, but you can go to the integrations panel and just add the integration manually, right?

escoand:
Correct, as I wrote in line 18:

```markdown
You can add your TV by using the web interface via [ssdp](/integrations/ssdp) discovery or manual config.
For more control over the integration you can also use an entry in the configuration files.
```

Maybe this is a bit misleading as I used "manual config" also for it.

frenck:
That does not really want I meant with my original comment.

There is a manual setup (click path) for the user via integrations to set it up manually. (Click integrations -> add using the + -> etc.

The manual should be written from a perspective a user of any experience level can use it. Right now, I think we should add instructions on how to set it up via just the frontend (since that is now possible).

arsaboo:
Since the component now allows both config flow and yaml, maybe the [Ring docs](https://github.com/home-assistant/home-assistant.io/pull/11679) are a good reference.

escoand:
OK, what do you think about he new wording?

frenck (approved):
Thanks!

PR #11620, opened 2020-01-03
Update pi_hole.markdown

whattheschnell:
See https://github.com/hassio-addons/addon-pi-hole/issues/99.  API Key needs to be set to null due to another issue in the pihole addon.  

**Description:**

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#11620

## Checklist:


frenck:
I think we discussed this on Discord already, but this is documented explicitly in the add-on documentation.

In other setups, this can even cause issues. Therefore, I think this does not belong in the Home Assistant documentation.

PR #11627, opened 2020-01-03
add camera password line to sample config

cshields:
**Description:**

The need for the camera password line is explained in the uvc doc, but it is not included in the sample config for copypasting. Without it, the integration works but imagery does not show.

## Checklist:


klaasnicolaas:
I have doubts about this change. We apply the rule that in the first example only required config variables may be included. `Password` is optional (also in the python code) and would therefore not belong in the first example.

@frenck what do you think about it?

frenck:
Docs should match the codebase. If it is a bug, it should be reported so the codebase can be adjusted.

PR #11631, opened 2020-01-04
Point to Telegram chatbot page from Telegram config example

scop:
**Description:**

Was a bit confused because the example config for the bot is not explained where it is presented in the Telegram (not bot) page, specifically the `platform: polling` entry. Also, fix one occurrence where a Telegram platform is referred to as "method" to use the word "platform".

## Checklist:


frenck (approved):
Thanks, @scop! 👍

PR #11645, opened 2020-01-06
Use explicit links inside configuration blocks

### LINKED ISSUE #11643: Links not working inside {% configuration %} blocks
When links are used inside `{% configuration %}` blocks, they are not rendered correctly but instead the source is retained verbatim, see for example [addons/google_assistant](https://www.home-assistant.io/addons/google_assistant) (`[Google Actions Console][GActionsConsole]`) or [integrations/alert](https://www.home-assistant.io/integrations/alert/) (`[template][template]`):

Screenshot from 2020-01-06 09-27-06

tribut:
**Description:**

The converter used in the configuration plugin is not aware of references declared outside of the block and therefore fails to render reference links. This replaces them with explicit links.

Resolves #11643

## Checklist:

### DIFF
--- a/source/_addons/google_assistant.markdown
+++ b/source/_addons/google_assistant.markdown
@@ -42,15 +42,15 @@ That's it. You should no be able to use the Google Voice assistant.
 
 {% configuration %}
 client_secrets:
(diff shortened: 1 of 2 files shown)


frenck (approved):
Thanks, @tribut! 👍

PR #11652, opened 2020-01-07
Update archlinux installation process

myxor:
**Description:**

Extend installation process by installing the existing Arch Linux package

## Checklist:

### DIFF
--- a/source/_docs/installation/archlinux.markdown
+++ b/source/_docs/installation/archlinux.markdown
@@ -5,6 +5,26 @@ description: "Installation of Home Assistant on your Arch Linux computer."
 
 [Arch Linux](https://www.archlinux.org/) is a lightweight and flexible Linux distribution for x86_64.
 
+There are two ways to install Home Assistant on Arch Linux:
+
+## 1. Use existing package
+
+Install the package available for Arch Linux based distributions: [home-assistant-package](https://www.archlinux.org/packages/community/any/home-assistant/).
+
+Run 
+
+```sudo pacman -S home-assistant```
+
+to install home-assistant including all required dependencies.
+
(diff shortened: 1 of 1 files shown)


source/_docs/installation/archlinux.markdown
```diff
@@ -5,6 +5,26 @@ description: "Installation of Home Assistant on your Arch Linux computer."
 
 [Arch Linux](https://www.archlinux.org/) is a lightweight and flexible Linux distribution for x86_64.
 
+There are two ways to install Home Assistant on Arch Linux:
+
+## 1. Use existing package
```
fabaff, line 10: I'm not so sure that we should add this method. Definitively not as the first option and then only with a fat warning.

There are multiple issues with the Arch Linux package like with every other package that is provided by a distribution. First (which is the primary blocker to include it in Fedora) is the mixing of Python modules. The installer prepares them for the system-wide usage while during run-time Home Assistant is pull the additional packages. Home Assistant has no access to the system's installer, is not running as root and can't use the Python default installation path thus it's very likely that this mixing is causing problems later.

The Arch package is removing the pinning which indicates that Home Assistant and Arch Linux already have different versions. This will make it very hard to debug.

myxor: I wasn't aware of the issues you mentioned. As i am using the package on my installation and not having any issues so far i thought i would recommend this installation method.
If you think this pull request is wrong you can close it.

frenck:
Hi @myxor!

The main issue, is that the package is not maintained by Home Assitant, and therefore considered a community effort. Those should not go into our documentation, but can be shared with the community on the Home Assistant Community Forum.

Nevertheless, thank you for your contribution! 👍

PR #11658, opened 2020-01-07
Switch order of Python and ADB server instructions

JeffLIrion:
**Description:**

Put the Python ADB section before the ADB server section. This approach is easier to setup and easier to debug, and users should only option for the ADB server approach if the Python ADB approach is unsuccessful / unreliable. 

## Checklist:


frenck (approved):
Thanks, @JeffLIrion! 👍

PR #11666, opened 2020-01-07
Removing Windows

DubhAd:
We get a regular trickle of folks installing on Windows, who're then stuck on things like auto-starting, and almost everything else.

As much as I don't like removing community provided guides, I think this one needs to disappear.

**Description:**

## Checklist:


frenck:
This just removes the menu item. Should we remove the pages as well?

DubhAd:
Yes, but I wasn't sure whether this would be rejected or not - guide removal in [here](https://github.com/home-assistant/home-assistant.io/pull/11709)

frenck (approved):

PR #11677, opened 2020-01-08
Added migration documentation to scripts

hAmpzter:
**Description:**

Added "Migrating your scripts to `scripts.yaml`"-section to the documentation.

## Checklist:

### DIFF
--- a/source/_docs/scripts/editor.markdown
+++ b/source/_docs/scripts/editor.markdown
@@ -20,3 +20,23 @@ script: !include scripts.yaml
 ```
 
 The content that was under `script:` should now be moved to `scripts.yaml` to be editable.
+
+## Migrating your scripts to `scripts.yaml`
+
+If you want to migrate your old scripts to use the editor, you'll have to copy them to `scripts.yaml`. Make sure that `scripts.yaml` is named entries! For each automation that you copy over, you'll have to add an key per script. This can be any string as long as it's unique.
+
+For example, the below script will turn on a scene.
+
+```yaml
+# Example scripts.yaml entry
+'my_unique_id':  # <-- Required for editor to work, for automations created with the editor the id will be automatically generated.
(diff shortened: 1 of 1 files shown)


source/_docs/scripts/editor.markdown
```diff
@@ -20,3 +20,23 @@ script: !include scripts.yaml
 ```
 
 The content that was under `script:` should now be moved to `scripts.yaml` to be editable.
+
+## Migrating your scripts to `scripts.yaml`
+
+If you want to migrate your old scripts to use the editor, you'll have to copy them to `scripts.yaml`. Make sure that `scripts.yaml` is named entries! For each automation that you copy over, you'll have to add an key per script. This can be any string as long as it's unique.
```
frenck, line 26: I'm a bit in the middle on this PR.

Mainly because of this section.

> Make sure that `scripts.yaml` is named entries!

> For each automation that you copy over, you'll have to add an key per script.

This is not typical of the scripts.yaml file, but a consequence of the scripts integration.

> This can be any string as long as it's unique.

This is a consequence because we use YAML and not specifically odd for this case.

So this PR tries to describe a guide about YAML packed as a manual for scripts?

hAmpzter: Hey @frenck .

What is it you want to say with your PR comment that I should change with it?

It's mainly a description (equally to the one on https://www.home-assistant.io/docs/automation/editor/#migrating-your-automations-to-automationsyaml ) to migrate the settings for your scripts (if you have them as separate files) to scripts.yaml?!

frenck: I get that, however, it is 1-on-1 transferred and adjusted.
For example, the part of the unique ID in automation, is not the same as the need for a unique key in a mapping in YAML (which scripts use).

PR #11691, opened 2020-01-09
Modified example for view.visible property to include visible list example

sprilukin:
It looks like an original example for `visible` property is not complete: it does not actually include `visible` property by itself neither as a boolean nor as a list.
Now example really includes an example of a view `visible` property as a list.

**Description:**

## Checklist:


sprilukin:
visible property described later in this doc. closing PR.

PR #11696, opened 2020-01-10
Update solax.markdown

danpowell88:
**Description:**

Add a note for newer inverters that have updated firmware and require a reverse proxy for accessing their api

## Checklist:

### DIFF
--- a/source/_integrations/solax.markdown
+++ b/source/_integrations/solax.markdown
@@ -58,11 +58,13 @@ sensors:
 {% endraw %}
 
 ### Note
-Inverter models with newer firmware (and also those using devices like PocketWifi) no longer expose an API when connected to your wireless network, they do however continue to expose it on their own broadcasted SSID. To use this sensor in this case it is neccessary to setup a reverse proxy with something like Nginx and use a raspberry pi (or similar) with two network connections (one being wifi that connects to the inverters SSID).
+
+Inverter models with newer firmware (and also those using devices like PocketWifi) no longer expose an API when connected to your wireless network, they do however continue to expose it on their own broadcasted SSID. To use this sensor in this case it is necessary to set up a reverse proxy with something like Nginx and use a raspberry pi (or similar) with two network connections (one being wifi that connects to the inverters SSID).
+
 
 Example Nginx configuration
 
-``` text
+```text
 location / {
   proxy_pass http://5.8.8.8;
 }
--- a/source/_integrations/solax.markdown
+++ b/source/_integrations/solax.markdown
@@ -58,11 +58,11 @@ sensors:
 {% endraw %}
 
 ### Note
-Inverter models with newer firmware (and also those using devices like PocketWifi no longer expose an API when connected to SSID, they do however continue to expose it on their own broadcasted SSID, in order to use this sensor in this case it is neccessary to setup a reverse proxy with something like nginx and use a raspberry pi (or similar) with two network connections (one wifi that connects to the inverters SSID).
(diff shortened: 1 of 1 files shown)


danpowell88:
Not sure why the others commits are showing up as well???

source/_integrations/solax.markdown
```diff
@@ -56,3 +56,14 @@ sensors:
       value_template: "{{ (states('sensor.power_now') | float) - (states('sensor.exported_power') | float) }}"
 ```
 {% endraw %}
+
+### Note
+Inverter models with newer firmware (and also those using devices like PocketWifi no longer expose an API when connected to SSID, they do however continue to expose it on their own broadcasted SSID, in order to use this sensor in this case it is neccessary to setup a reverse proxy with something like nginx and use a raspberry pi (or similar) with two network connections (one wifi that connects to the inverters SSID).
```
frenck:
```suggestion

Inverter models with newer firmware and also those using devices like PocketWifi, no longer expose an API when connected to SSID. They do, however, continue to expose it on their own broadcasted SSID, to use this sensor, in this case, it is necessary to set up a reverse proxy with something like Nginx and use a raspberry pi (or similar) with two network connections (one wifi that connects to the inverters SSID).
```

source/_integrations/solax.markdown
```diff
@@ -56,3 +56,14 @@ sensors:
...
+
+Example nginx configuration
```
frenck:
```suggestion
Example Nginx configuration
```

source/_integrations/solax.markdown
```diff
@@ -56,3 +56,14 @@ sensors:
...
+
+```
```
frenck: Code block language type missing, maybe `text`?

frenck (changes requested):
Thanks for the PR @danpowell88!

I've reviewed it and left some comments and suggestions. Could you please take a look? Thanks! 👍

danpowell88:
@frenck updated, how does that look to you?

source/_integrations/solax.markdown
```diff
@@ -56,3 +56,14 @@ sensors:
       value_template: "{{ (states('sensor.power_now') | float) - (states('sensor.exported_power') | float) }}"
 ```
 {% endraw %}
+
+### Note
+Inverter models with newer firmware (and also those using devices like PocketWifi) no longer expose an API when connected to your wireless network, they do however continue to expose it on their own broadcasted SSID. To use this sensor in this case it is neccessary to setup a reverse proxy with something like Nginx and use a raspberry pi (or similar) with two network connections (one being wifi that connects to the inverters SSID).
+
+Example Nginx configuration
+
+``` text
```
frenck:
```suggestion
```text
```

source/_integrations/solax.markdown
```diff
@@ -56,3 +56,14 @@ sensors:
       value_template: "{{ (states('sensor.power_now') | float) - (states('sensor.exported_power') | float) }}"
 ```
 {% endraw %}
+
+### Note
+Inverter models with newer firmware (and also those using devices like PocketWifi) no longer expose an API when connected to your wireless network, they do however continue to expose it on their own broadcasted SSID. To use this sensor in this case it is neccessary to setup a reverse proxy with something like Nginx and use a raspberry pi (or similar) with two network connections (one being wifi that connects to the inverters SSID).
```
frenck:
```suggestion

Inverter models with newer firmware (and also those using devices like PocketWifi) no longer expose an API when connected to your wireless network, they do however continue to expose it on their own broadcasted SSID. To use this sensor in this case it is necessary to set up a reverse proxy with something like Nginx and use a raspberry pi (or similar) with two network connections (one being wifi that connects to the inverters SSID).

```

frenck (changes requested):

frenck (approved):
Thanks, @danpowell88! 👍

PR #11698, opened 2020-01-10
Add assume_state option at xiaomi_tv platform

ppanagiotis:
**Description:**

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#30645

## Checklist:

### DIFF
--- a/source/_integrations/xiaomi_tv.markdown
+++ b/source/_integrations/xiaomi_tv.markdown
@@ -36,6 +36,11 @@ name:
   required: false
   default: Xiaomi TV
   type: string
+assume_state:
+  description: Assume the state of tv or not.
+  required: false
+  default: true
+  type: boolean
 {% endconfiguration %}
 
 If you do not set a host in the configuration file, local TVs will automatically be discovered.
@@ -51,5 +56,6 @@ media_player:
 ```
 
 <div class='note info'>
-The platform will never turn your TV off. Instead, it will be put to sleep and woken up. This can be useful, because the selected source of the TV will remain the same. It will essentially turn your TV into a dumb TV.
+If the `assume_state` (default) is `true` the platform will never turn your TV off. Instead, it will be put to sleep and woken up. This can be useful because the selected source of the TV will remain the same. It will essentially turn your TV into a dumb TV.
+If the `assume_state` is `false` your TV will turn off and not able to turn on again using this module. You have to power the TV on using IR, Bluetooth remote control or HDMI-CEC.
 </div>


source/_integrations/xiaomi_tv.markdown
```diff
@@ -36,6 +36,11 @@ name:
   required: false
   default: Xiaomi TV
   type: string
+assume_state:
+  description: Assume state of tv or not
+  required: false
+  default: True
```
frenck:
```suggestion
  default: true
```

source/_integrations/xiaomi_tv.markdown
```diff
@@ -36,6 +36,11 @@ name:
   required: false
   default: Xiaomi TV
   type: string
+assume_state:
+  description: Assume state of tv or not
```
frenck:
```suggestion
  description: Assume the state of tv or not.
```

source/_integrations/xiaomi_tv.markdown
```diff
@@ -51,5 +56,6 @@ media_player:
 ```
 
 <div class='note info'>
-The platform will never turn your TV off. Instead, it will be put to sleep and woken up. This can be useful, because the selected source of the TV will remain the same. It will essentially turn your TV into a dumb TV.
+If the `assume_state` (default) is True the platform will never turn your TV off. Instead, it will be put to sleep and woken up. This can be useful, because the selected source of the TV will remain the same. It will essentially turn your TV into a dumb TV.
```
frenck:
```suggestion
If the `assume_state` (default) is `true` the platform will never turn your TV off. Instead, it will be put to sleep and woken up. This can be useful because the selected source of the TV will remain the same. It will essentially turn your TV into a dumb TV.
```

source/_integrations/xiaomi_tv.markdown
```diff
@@ -51,5 +56,6 @@ media_player:
...
+If the `assume_state` is False your TV will turn off and not able to turn on again using this module. You have to power the TV on using IR, bluetooth remote control or HDMI-CEC.
```
frenck:
```suggestion
If the `assume_state` is `false` your TV will turn off and not able to turn on again using this module. You have to power the TV on using IR, Bluetooth remote control or HDMI-CEC.
```

frenck (changes requested):
Thanks for the PR @ppanagiotis!

I've reviewed it and left some comments/suggestions. Could you please take a look? Thanks! 👍

PR #11706, opened 2020-01-11
✨ Add conditional-row to Lovelace entities card

iantrich:
**Description:**
https://github.com/home-assistant/home-assistant-polymer/pull/4569

## Checklist:

### DIFF
--- a/source/_lovelace/entities.markdown
+++ b/source/_lovelace/entities.markdown
@@ -241,12 +241,12 @@ conditions:
       description: HA entity ID.
       type: string
     state:
-      required: false
-      description: Entity state is equal to this value.*
+      required: false*
+      description: Entity state is equal to this value.
       type: string
     state_not:
-      required: false
-      description: Entity state is unequal to this value.*
+      required: false*
+      description: Entity state is unequal to this value.
       type: string
 row:
   required: true
--- a/source/_lovelace/entities.markdown
+++ b/source/_lovelace/entities.markdown
@@ -222,6 +222,42 @@ hide_if_unavailable:
   default: false
 {% endconfiguration %}
 
+### Conditional
+
+Special row that displays based on entity states.
+
+{% configuration %}
+type:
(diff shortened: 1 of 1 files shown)


source/_lovelace/entities.markdown
```diff
@@ -222,6 +222,42 @@ hide_if_unavailable:
...
+  required: true
+  description: List of entity IDs and matching states.
+  type: list
+  keys:
+    entity:
+      required: true
+      description: HA entity ID.
+      type: string
+    state:
+      required: false
```
bramkragten, line 244: I would place the asterisk here

source/_lovelace/entities.markdown
```diff
@@ -222,6 +222,42 @@ hide_if_unavailable:
...
+  type: string
+conditions:
+  required: true
+  description: List of entity IDs and matching states.
+  type: list
+  keys:
+    entity:
+      required: true
+      description: HA entity ID.
+      type: string
```
bramkragten: Aren't entities and state lists like with an `entity-filter`?

source/_lovelace/entities.markdown
```diff
@@ -222,6 +222,42 @@ hide_if_unavailable:
...
+    state:
+      required: false
+      description: Entity state is equal to this value.*
```
klaasnicolaas, line 245: You can't use a ⭐️ to indicate that something is required, Markdown doesn't like that.

klaasnicolaas (changes requested):

source/_lovelace/entities.markdown
iantrich: The `*` just highlights the additional information

iantrich: Ah, were you referring to @bramkragten suggestion to move the *, as that did indeed break the parser?

frenck (approved):

PR #11717, opened 2020-01-12
Fixed ws-url and needed steps.

linkztream:
Modified the ws-url to reflect the documentation from SPC-GATEWAY manual and added the two needed changes to customize.yaml + configuration.yaml to make the sensors produce usable output.

**Description:**

## Checklist:

### DIFF
--- a/source/_integrations/spc.markdown
+++ b/source/_integrations/spc.markdown
@@ -34,12 +34,25 @@ api_url:
   required: true
   type: string
 ws_url:
-  description: URL of the SPC Web Gateway websocket, e.g., `ws://<ip>:8088`.
+  description: URL of the SPC Web Gateway websocket, e.g., `ws://<ip>:8088/ws/spc`.
   required: true
   type: string
 {% endconfiguration %}
 
-Supported sensors will be automatically discovered and added, however they will be hidden by default.
+Supported sensors will be automatically discovered and added, however they will be hidden by default, and unless unhidden they will not produce any history or work in automations.
+
+To unhide them add/edit /config/customize.yaml and add an entry for all sensors you want to include like this, replacing the sensor name with the name you find in developer-tools -> States `<ha-url>:8123/developer-tools/states`:
+```
+binary_sensor.your_sensor: 
+  hidden: false
+```
+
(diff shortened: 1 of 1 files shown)


source/_integrations/spc.markdown
```diff
@@ -34,12 +34,25 @@ api_url:
...
   type: string
 ws_url:
-  description: URL of the SPC Web Gateway websocket, e.g., `ws://<ip>:8088`.
+  description: URL of the SPC Web Gateway websocket, e.g., `ws://<ip>:8088/ws/spc`.
   required: true
   type: string
 {% endconfiguration %}
 
-Supported sensors will be automatically discovered and added, however they will be hidden by default.
+Supported sensors will be automatically discovered and added, however they will be hidden by default, and unless unhidden they will not produce any history or work in automations.
```
frenck: Let's remove this part altogether.
With Lovelace, there is no reason to hide these sensors.

home-assistant/home-assistant#30699

frenck (changes requested):
Hi there @linkztream!

Thanks for the PR, I've reviewed it, and left a comment.
Could you please take a look? 👍

source/_integrations/spc.markdown
linkztream: The reason for the change was that the integration doesn't work unless the sensors are unhidden. The integration sets them to hidden, and just changing them in Lovelace reverts back to hidden at restart. At the very least we need to keep the information about the sensors not working unless they are manually unhidden. 

This was the only way I could make it work consistently.

frenck: @linkztream If you've clicked the link I've put in my last message, you would have found I have changed the integration in Home Assistant to not hide those sensors anymore.

Please remove the customization additions, as they are not needed anymore.

linkztream: You're absolutely right, my bad. Sorry for not reading properly! This was my first pull request ever, so got a little bit lost in it all.

With that change to the integration (and some better reading comprehension on my part) I completely agree with you.

frenck (approved):
Thanks, @linkztream! 👍

PR #11721, opened 2020-01-12
On->on & Off->off

akasma74:
just because they are not the same if used as state value.
stumbled upon it when discussing why this automation (https://community.home-assistant.io/t/integrating-rf-devices-with-sonoff-rf-bridge-on-tasmota-firmware/42484/112?u=ahmadk) doesn't work.

**Description:**

## Checklist:


frenck (approved):
Thanks, @akasma74! 👍

PR #11578, opened 2019-12-31
Entity_id isn't optional anymore at light.{turn_off/turn_on}

ppanagiotis:
**Description:**
After #29178 entity_id isn't optional anymore. You have to use
all as entity_id.

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#29178

## Checklist:

- [x ] Branch: `next` is for changes and new documentation that will go public with the next Home Assistant release. Fixes, changes and adjustments for the current release should be created against `current`.
- [x ] The documentation follows the [standards][standards].

[standards]: https://developers.home-assistant.io/docs/documentation_standards.html

PR #11749, opened 2020-01-15
Replace all 'Else targets all.' occurrences.

fhoekstra:
**Description:**

Please read on, I have 2 questions for further changes related to this PR.

I simply did a 'replace all' in all files:
Replace "Else targets all." with "Either `entity_id` or `area` must be specified. Use `entity_id:all` to target all."

There are 2 previous PRs, where @frenck 's comments/actions are contradictory: 
In #11578 a change was accepted where entity_id was made non-optional for the light integration.
In #11602 which was a similar change to the light docs, @frenck commented that entity_id is optional, since area can also be specified.

**1. Is this the way we want to do this?** If so, I will also apply it to the light documentation.
**2. Shall I also add an `area` parameter** to every integration that has an `entity_id` parameter?

## Checklist:

### DIFF
--- a/source/_integrations/cover.markdown
+++ b/source/_integrations/cover.markdown
@@ -35,7 +35,7 @@ Available services: `cover.open_cover`, `cover.close_cover`, `cover.stop_cover`,
 
 | Service data attribute | Optional | Description |
 | ---------------------- | -------- | ----------- |
-| `entity_id` | yes | String or list of strings that point at `entity_id`'s of covers. Else targets all.
+| `entity_id` | yes | String or list of strings that point at `entity_id`'s of covers. Either `entity_id` or `area` must be specified. Use `entity_id: all` to target all.
 
 ### Service `cover.set_cover_position`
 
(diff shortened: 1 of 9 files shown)


source/_integrations/cover.markdown
```diff
@@ -35,15 +35,15 @@ Available services: `cover.open_cover`, `cover.close_cover`, `cover.stop_cover`,
 
 | Service data attribute | Optional | Description |
 | ---------------------- | -------- | ----------- |
-| `entity_id` | yes | String or list of strings that point at `entity_id`'s of covers. Else targets all.
+| `entity_id` | yes | String or list of strings that point at `entity_id`'s of covers. Either `entity_id` or `area` must be specified. Use `entity_id: all` to target all.
```
frenck: The area is a different attribute...
So that should not be in here.

frenck (changes requested):

source/_integrations/cover.markdown
fhoekstra: Okay, how would you suggest to make it clear that either an entity_id or area must be provided? If I just leave this part out, they are both optional.
Actually, area isn't mentioned yet. Shall I put that parameter in the table for every integration?

frenck: `area_id` is the additional parameter allowed.

However, it is not clear on how to use that at this point, and is something we need to look into.

It does, however, show up, if you create an automation with the automation editor. Create a automation targeting an area, and you'll see it in the resulting generated YAML.

fhoekstra:
I removed that phrase from all the integration services. Let me know if there is anything else that must be done before we can merge this.

frenck (approved):
Thanks! 👍

fhoekstra:
🙂

PR #11760, opened 2020-01-15
Update ring.markdown

smccloud:
Added that it appears you need 0.104 for the UI method to work and that the YAML no longer works (I had it setup that way, broke after the issue with the footage being viewed came out).

**Description:**

YAML method no longer works and the UI method does not appear in 0.103.6.

## Checklist:

### DIFF
--- a/source/_integrations/ring.markdown
+++ b/source/_integrations/ring.markdown
@@ -31,31 +31,11 @@ This component does NOT allow for live viewing of your Ring camera within Home A
 
 ## Configuration
 
-Go to the integrations page in your config and click on new integration -> Ring.
+Go to the integrations page in your config and click on new integration -> Ring.  Please note, this requires 0.104 to appear.
 
 ## YAML configuration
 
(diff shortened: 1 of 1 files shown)


source/_integrations/ring.markdown
```diff
@@ -31,31 +31,11 @@ This component does NOT allow for live viewing of your Ring camera within Home A
...
-username:
-  description: The username for accessing your Ring account.
-  required: true
-  type: string
-password:
-  description: The password for accessing your Ring account.
-  required: true
-  type: string
-{% endconfiguration %}
+YAML configuration no longer works with the updates Ring has done to their backend.
```
balloob, line 38: This is not true. This still works.

frenck:
Thanks for the PR @smccloud!

Unfortunately, it is incorrect, the YAML configuration method is still active and valid.

As of the version number, our documentation always reflects the latest Home Assistant release and we really want to prevent cluttering the docs with changelogs.

Nevertheless, thanks for contributing, that is always appreciated 👍

PR #11778, opened 2020-01-16
Add notes about deprecated api_key

dshokouhi:
**Description:**

This is to go alongside #11777 since users can still use the `api_key` in 0.104 we will keep the references but note that it is deprecated.  All mentions of this api key have been changed to recommend service account instead.

Fixes https://github.com/home-assistant/home-assistant/issues/30863

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#30402
 PR already merged but this should've been caught alongside it :)

## Checklist:

### DIFF
--- a/source/_integrations/google_assistant.markdown
+++ b/source/_integrations/google_assistant.markdown
@@ -28,7 +28,7 @@ To use Google Assistant, your Home Assistant configuration has to be [externally
 
 </div>
 
-You need to create an API Key with the [Google Cloud API Console](https://console.cloud.google.com/apis/api/homegraph.googleapis.com/overview) which allows you to update devices without unlinking and relinking an account (see [below](#troubleshooting-the-request_sync-service)). If you don't provide one, the `google_assistant.request_sync` service is not exposed. It is recommended to set up this configuration key as it also allows the usage of the following command, "Ok Google, sync my devices". Once you have set up this component, you will need to call this service (or command) each time you add a new device in Home Assistant that you wish to control via the Google Assistant integration.
+You will need to create a service account [Create Service account key](https://console.cloud.google.com/apis/credentials/serviceaccountkey)  which allows you to update devices without unlinking and relinking an account (see [below](#troubleshooting-the-request_sync-service)). If you don't provide the service account, the `google_assistant.request_sync` service is not exposed. It is recommended to set up this configuration key as it also allows the usage of the following command, "Ok Google, sync my devices". Once you have set up this component, you will need to call this service (or command) each time you add a new device in Home Assistant that you wish to control via the Google Assistant integration.  See Step 2 after the note for more details.
 
 1. Create a new project in the [Actions on Google console](https://console.actions.google.com/).
     1. Add/Import a project and give it a name.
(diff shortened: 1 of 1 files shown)


Eklacin:
Shouldn't we add 
"Go to the Google API Console."
"Select your project and click Enable HomeGraph API." 
to Step 2 (like in Step3). 

Afaik this is necessary even for Service Accounts. Otherwise users might miss it.

elupus:
Yes homegraph is still needed.

dshokouhi:
@elupus since I did not submit the PR, I can close this....would you be able to update the docs?  We are having users asking how to fix the new warning they are seeing and there are no docs to go with the warning.  I was just trying to help out but I am not sure what is the correct order.

source/_integrations/google_assistant.markdown
```diff
@@ -75,7 +75,7 @@ If you've added Home Assistant to your phone's home screen, you have to first re
     6. For the Key type, select the JSON option.
     7. Click Create. A JSON file that contains your key downloads to your computer.
     8. Use the information in this file or the file directly to add to the `service_account` key in the configuration.
-3. If you didn't specify a service account and want to use the `google_assistant.request_sync` service, to update devices without unlinking and relinking, in Home Assistant, then enable HomeGraph API for your project:
+3. If you didn't specify a service account and want to use the `google_assistant.request_sync` service, to update devices without unlinking and relinking, in Home Assistant, then enable HomeGraph API for your project (Please note: this is now deprecated refer to Step 2 above, this feature will be removed in 0.105):
```
elupus: Just change the "didnt" to "did" and drop the note  about deprecated.

elupus (changes requested):

source/_integrations/google_assistant.markdown
dshokouhi: Ok will do, we also mention to note down the API key in step 3....should I remove that too (line 82)?

Eklacin: Shouldn't we simply move Step 3.1 and Step 3.2 to be Step 2.9 and Step 2.10 and remove the rest of Step 3.

elupus:
Both steps where it create credential and note down should be removed. Enable is still needed

source/_integrations/google_assistant.markdown
dshokouhi: Good suggestion! Updated now, I think this is better.  Let me know what you guys think.

Eklacin: Maybe we could even merge the entire block "You will need to create a service account..." with Step 2. So the docs for "Service Account" and "Resync" would be at the same place.

dshokouhi: The reason why it is in 2 places is because of how easy it is for people to miss.  We should keep them there so people know to perform those steps.  There are many steps here and people tend to skim over it so mentioning it more than once is useful IMO.

elupus (approved):

frenck (approved):
Thanks, @dshokouhi! 👍

Eklacin (approved):

PR #11796, opened 2020-01-18
WIP Condense Lovelace configuration for easier reading

iantrich:
This is what I use in lots of my custom Lovelace cards and I find it much easier to read through than what we are currently doing and wanted some feedback on what we can do differently. Right now, the Lovelace configurations are a mess and nearly impossible to read with lots of redundancy even on the same page at times with things like the actions. Feedback/Ideas very much welcome.

**Description:**

## Checklist:

### DIFF
--- a/source/_lovelace/iframe.markdown
+++ b/source/_lovelace/iframe.markdown
@@ -15,25 +15,12 @@ You can't embed sites using HTTP if you are using HTTPS for your Home Assistant.
   Windy weather radar as iframe.
 </p>
 
-{% configuration %}
-type:
-  required: true
-  description: iframe
-  type: string
-url:
-  required: true
-  description: Website URL.
-  type: string
-aspect_ratio:
-  required: false
-  description: Height-width-ratio.
-  type: string
-  default: "50%"
-title:
-  required: false
-  description: The card title.
-  type: string
-{% endconfiguration %}
+| Name         | Required | Description        | Type   | Default |
+| ------------ | -------- | ------------------ | ------ | ------- |
+| type         | true     | `iframe`           | string | none    |
+| url          | true     | Website URL        | string | none    |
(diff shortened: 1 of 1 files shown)


source/_lovelace/iframe.markdown
```diff
@@ -15,25 +15,12 @@ You can't embed sites using HTTP if you are using HTTPS for your Home Assistant.
...
-  type: string
-  default: "50%"
-title:
-  required: false
-  description: The card title.
-  type: string
-{% endconfiguration %}
+| Name         | Required | Description        | Type   | Default |
+| ------------ | -------- | ------------------ | ------ | ------- |
+| type         | true     | `iframe`           | string | none    |
```
bramkragten, line 20:
```suggestion
| type         | true     | Card type          | string | `iframe`|
```

bramkragten (commented):
Not sure, I do think our Lovelace documentation needs a massive update and it is too bulky, but a table might not be the best way forward as I think it is not so clear...

Especially our action config should be split out and linked to a general page, and maybe we can do that for more things.

iantrich:
Any card that has entities or elements suffers greatly right now. It is really hard to read those, especially picture elements.

Maybe an initial list of possible elements that links to the docs is the clean/simple solution?

I agree with the action stuff needing to be split out for sure.

frenck:
I don't think we should do this. This is working around an issue to make things more readable on a single place, instead of improving the configuration listing across all pages by changing the styling and/or plugin that processes it.

If we adjust this generally, all integrations and Lovelace configs benefit from that.

iantrich:
Changing said styling is beyond my skillset :) But I"m glad that it is in your head now and hopefully you have some ideas. In the meantime, I'll likely start extracting redundant pieces to their own subpages for easier maintenance and cleaner pages

PR #11813, opened 2020-01-19
Update installation.markdown

olsoma13:
I beat my head against this for a long time when first using the network-manager package. I figured this note would be helpful to other users!

**Description:**

## Checklist:

### DIFF
--- a/source/hassio/installation.markdown
+++ b/source/hassio/installation.markdown
@@ -117,6 +117,19 @@ Optional:
 
  - `network-manager`
 
+<div class='note'>
+   The default installation of `network-manager` will set a random Mac address over wifi on each restart. This will be troublesome in any situation where you would like to set a static IP. Edit `/etc/NetworkManager/NetworkManager.conf` and configure as follows:
+   ```[main]
+plugins=ifupdown,keyfile
+
+[ifupdown]
+managed=false
+
+[device]
+wifi.scan-rand-mac-address=no```
+
+</div>
+
 <div class='note warning'>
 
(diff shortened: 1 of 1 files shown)


source/hassio/installation.markdown
```diff
@@ -117,6 +117,19 @@ Optional:
 
  - `network-manager`
 
+<div class='note'>
+   The default installation of `network-manager` will set a random Mac address over wifi on each restart. This will be troublesome in any situation where you would like to set a static IP. Edit `/etc/NetworkManager/NetworkManager.conf` and configure as follows:
```
frenck, line 121: You need to add an empty newline between the HTML & markdown to allow the parser to switch context.

frenck, line 121: I'm not sure why this is relevant?

source/hassio/installation.markdown
```diff
@@ -117,6 +117,19 @@ Optional:
...
+   ```[main]
```
frenck, line 122:
```suggestion

   ```txt
```

frenck (changes requested):
Hi there @olsoma13!

I've left some comments to look at. Secondly, I'm not sure if we should add this. Mainly it will cause confusion and maybe some other things.

For example:

- We should not encourage WiFi use
- This is a system administration problem of the OS of your choice. There is no limitation on running Hass.io on a setup without this set.
- "troublesome in any situation where you would like to set a static IP" -> this is actually incorrect. A static IP, I think you are referring to DHCP reservation in this case, which is often confused with setting a static IP.

frenck:
PR is running stale, closing.

PR #11814, opened 2020-01-19
Prevent failure when reloading fail2ban

ndfred:
**Description:** fail2ban complains about a double `bantime` directive when reloading the service, suggest people change the value later rather than include two statements in the configuration file

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#11814

## Checklist:


frenck (approved):
Thanks, @ndfred! 👍

PR #11838, opened 2020-01-21
Remove standby_connection config option

bendavid:
Remove standby_connection config option from webostv integration since it is no longer needed and has been removed from client library.

Corresponds to https://github.com/home-assistant/home-assistant/pull/31042


klaasnicolaas (approved):
✅ Approved. Can be merged as soon as the parent PR gets merged 👍

PR #12122, opened 2020-02-18
Frontend release notes for 0.106

bramkragten:
Still need to add screenshots


source/_posts/2020-02-26-release-106.markdown
```diff
@@ -15,6 +15,63 @@ og_image: /images/blog/2020-02-0.105/components.png
 - Icons for scene/script.
 - brightness_step and brightness_step_pct added to light.turn_on. Also part of device actions.
 
+## All entities in entities config panel
+
+The entities config panel used to only contain entities that had unique ID's,
+because that are the entities you can change the name and entity id of from the UI.
+In this version we changed that, we now include all entities in the config panel.
+You still can only change the ones that have a unique ID but atleast now you have a clear overview of the entities that are available,
+and you can use the in the last release introduced related items to see where the entities are used!
```
cogneato: The entities config panel previously only contained entities which had unique ID's because these were the entities that could have their names and id's edited within the UI. In this version we now include *all* entities in the config panel. You are still only able to change those that have a unique ID, but now you have a clear overview of all the available entities and can also make use of the "related items" feature introduced in release 0.105.0 to see where they are being used!

frenck (approved):

PR #12126, opened 2020-02-19
update vizio integration name and add quality score

raman325:
## Proposed change
I updated the name of the vizio integration to reflect the fact that it doesn't just support TVs. I also added the appropriate quality score.

## Type of change

## Additional information

- Link to parent pull request in the codebase: home-assistant/home-assistant#31975

## Checklist

### DIFF
--- a/source/_integrations/vizio.markdown
+++ b/source/_integrations/vizio.markdown
@@ -1,5 +1,5 @@
 ---
-title: Vizio SmartCast TV
+title: Vizio SmartCast
 description: Instructions on how to integrate Vizio SmartCast TVs and sound bars into Home Assistant.
 logo: vizio-smartcast.png
 ha_category:
@@ -7,6 +7,7 @@ ha_category:
 ha_release: 0.49
 ha_iot_class: Local Polling
 ha_config_flow: true
+ha_quality_scale: platinum
 ha_codeowners:
   - '@raman325'
 ---


source/_integrations/vizio.markdown
```diff
@@ -1,12 +1,13 @@
...
 logo: vizio-smartcast.png
 ha_category:
   - Media Player
 ha_release: 0.49
 ha_iot_class: Local Polling
 ha_config_flow: true
+ha_quality_scale: platinum
 ha_codeowners:
   - '@raman325'
 ---
```
springstan, line 13: Update `required` to `true` for `device_class`, see https://github.com/home-assistant/home-assistant/pull/31975

raman325: That update was a mistake in the parent PR, will fix

raman325: Actually nevermind, it was not a mistake. The change was made because when a user uses config flow through the UI, that field is required to be filled out. In config import, that field is optional because a default is provided. So the documentation is correct.

springstan: Okay thanks for the clarification.

raman325: Alternatively, let me know if you would like me to revert the field to Optional in the integration code. The Optional vs Required decision has no impact to the UI config flow today because vol.In resolves to a drop down choice list in the UI and at least one value has to be selected. I switched it to required as a safety precaution, but it may not be necessary.

springstan: Well I think it is best to revert it back to optional since it does not impact on the UI and code wise it does not make a difference either. Required without default value is basically the same as optional with a default value, both are accessed via `dict[key]` and the respective value needs to be set for a successful setup.

raman325: Sounds good!

frenck:
These will be automatically synced on release 👍 
So no need to do it manually.

PR #12162, opened 2020-02-22
Change i.e. and e.g. to i.e.,, and e.g.,

arsaboo:
## Type of change

## Additional information

## Checklist


frenck:
For `e.g.` this works and is true, however, for `i.e.` it ain't that simple.

- `e.g.` (exmpli gratia): "for example", one of more examples. Often a list, or at least indicates one or more items.
- `i.e.` (id est): "that is", which, if followed by just one singular item, does not need a comma.

arsaboo:
`The consensus seems to be in favor of the comma in American usage; against it in British usage.` quoting from https://www.dailywritingtips.com/comma-after-i-e-and-e-g/

arsaboo:
We follow the same rule in academic writing. 

The AP Stylebook, whose “punctuation-pitch” leans generally to the side of “the fewer commas the better,” is pro-comma when it comes to i.e. and e.g. According to AP, both abbreviations are “always followed by a comma.”

I am fine either way.

frenck:
Depends on where you look, the Chicago Manual of Style, isn't conclusive about this either.

Microsoft took a different approach and simply forbids the use of `e.g.` and `i.e.` in their documentation:

- <https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/i/ie>
- <https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/eg>

Google's documentation style guide:

> i.e.
> Don't use. Instead, use phrases like that is. Too many people mix up e.g. and i.e.
<https://developers.google.com/style/word-list#letter-i>

> e.g.
> Don't use. Instead, use phrases like for example or such as. Too many people mix up e.g. and i.e.
<https://developers.google.com/style/word-list#letter-e>

@ the Daily Writing Tips:
> Style guides do not agree on whether or not a comma should follow both these abbreviations. They do all agree that a comma precedes i.e. when the i.e. phrase occurs in a running text (i.e., not enclosed in parenthesis).

> The consensus seems to be in favor of the comma in American usage; against it in British usage.


So....

giphy



Personally, never thought about it that much. I must say, I prefer both the Microsoft & Google recommendations, however, that is a bit much to achieve 😉 


Let's make it consistent, which your PR does. 👍

arsaboo:
consistency is what I was going for 😉

frenck:
PR looks good! Let's add a small expression to the `.textlintrc.json` file, to ensure we keep it consistent in the future.

e.g., `["e.g.", "e.g.,"]`

(☝️ The most confusing line I ever wrote 😟 )

.textlintrc.json
```diff
@@ -303,6 +303,7 @@
         ["colour", "color"],
         ["config\\b", "configuration"],
         ["DarkSky", "Dark Sky"],
+        ["e.g.", "e.g.,"],
```
frenck:
```suggestion
        ["e\\.g\\.", "e.g.,"],
```

frenck: It's a regular expression, so we need to escape the `.` to prevent it is being used in the expression as any character. (My bad though, I should have known when I wrote the previous comment).

Add the `i.e.` in as well please 👍

frenck (approved):
Nice! 👍 🎉

PR #12248, opened 2020-03-03
Update locked_out.md with a new solution

redbar0n:
I added a forgot password use case in which the problem was solved remarkably easy, in hope that it may help someone else.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/locked_out.md
+++ b/source/_docs/locked_out.md
@@ -10,6 +10,8 @@ or need to recover your data.
 
 ### Home Assistant (including Supervised)
 
+If you are still logged in to the web interface with your user, then you are in luck. Add a new user as administrator and give the new user a password you can remember. Then log out, and log in with this new user. You may then delete the old user account. But this way, your configuration will remain, and you don't have to do a new onboarding process.
+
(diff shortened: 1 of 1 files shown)


redbar0n:
Note:
In case it wasn't clear, the use case is for when you've forgotten/lost your password, but you still have at least one browser where you are still logged in. So you're not totally locked out (even though you are locked out in other browsers). But even though you are still logged in via one browser, you are not able to change your password because of the password lock. And the next time you log out of that browser (either manually or because of some login expiration date), then you will be completely locked out. So the use case is sufficiently similar to the others, and users in this situation will likely go to this same page in the documentation, as I did. But I make this commit note here since the site heading "Locked out!" is a truth with modifications in this particular case. Feel free to edit this commit to include any of this information if necessary (although I think this clarification should be superflous to other people than the repo admins).

frenck (approved):

redbar0n (commented):
Actually, my intention was "add a new user as administrator role" meaning that the new user should have admin privileges, so he can be admin and thus allowed to delete the old admin account. I see that this could have been clearer. So the original logged in user and the new user should both be admin.

PR #12372, opened 2020-03-14
correct note about trigger

akasma74:
Don't think the existing way of testing automations with trigger in templates makes sense.
I tried to show how to debug such automations.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/automation/templating.markdown
+++ b/source/_docs/automation/templating.markdown
@@ -76,7 +76,12 @@ Knowing how to access the [state object](/docs/configuration/state_object/) of a
 
 <div class='note'>
   
-  Be aware that if you reference a trigger state object in an automation action, attempting to test that automation by calling the `automation.trigger` service or by clicking EXECUTE in the More Info box for the automation will not work. This is because the trigger state object doesn't exist in those contexts. One way to test automations like these is to manually change the state of the trigger entity at Developer Tools > States.
+  Be aware that if you reference a `trigger` state object in templates of automation `action`, attempting to test that automation by calling the `automation.trigger` service or by clicking EXECUTE in the More Info box for the automation will not work. This is because the trigger state object doesn't exist in those contexts. One way to test automations like these is to manually check that the templates work as expected by pasting them in Developer Tools > Template together with your trigger's definition like:
+{%raw%
+{% set trigger={'to_state':{'state': 'heat'}} %}
+{% set option = trigger.to_state.state %}
+{{ 'on' if option == 'heat' else 'off' }}
+{%endraw%}
   
 </div>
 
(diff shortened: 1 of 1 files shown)


frenck:
I think both the existing and the proposed methods are both fine methods to use. 

maybe we should just add to it, instead of replacing it completely.

Furthermore, could you please fix the build?! Thanks!

akasma74:
> I think both the existing and the proposed methods are both fine methods to use.

Possibly.
When I read the existing method, I don't get the idea - could you explain?

> maybe we should just add to it, instead of replacing it completely.

I don't mind, really

> Furthermore, could you please fix the build?! Thanks!

fixed, sorry about that ;)

frenck:
> When I read the existing method, I don't get the idea - could you explain?

Home Assistant allows one to change the state of things directly in the state engine. This is helpful to test certain states/conditions.

You have removed parts of that now in the current description.

akasma74:
> > When I read the existing method, I don't get the idea - could you explain?
> 
> Home Assistant allows one to change the state of things directly in the state engine. This is helpful to test certain states/conditions.

yeah, but the bit I'm updating is about testing automations with templates and, namely `trigger` variable inside them.

with the code I'm proposing to add
```yaml
{% set option = trigger.to_state.state %}
{{ 'on' if option == 'heat' else 'off' }}
```
how that ability to change state would help?

do you neam they can set state of their `climate.thermostat` to `heat` and then change the template to
```yaml
{% set option = states('climate.thermostat') %}
{{ 'on' if option == 'heat' else 'off' }}
```
?

If that's the idea then yes, it will work that way but it's error-prone and more difficult so I personally wouldn't do that or wouldn't reccomend it to anyone.

got your answer that it does not, will reply separately

frenck:
It does not, I'm saying your addition for an alternative way to debug things, destroyed an already valid case of debugging things.

So please restore the previous, and add your additional.

akasma74:
> It does not, I'm saying your addition for an alternative way to debug things, destroyed an already valid case of debugging things.

Am I wrong in thinking that the "destroyed" valid case has nothing to do with debugging templates in automations and trigger state object?

If I'm not, the documentation will realy benefit from making it more clear and that's exactly what I'm trying to do.

> So please restore the previous, and add your additional.

I'm not against the deleted method, just want to make sure that the text is clearly understandable and is related to the section it's published in - Trigger State Object.
From what I can see now it's not like that for the reason I explained above.

frenck:
Well, the PR currently removed the text on changing states, so yes, this PR removes a current valid debug method.

It allows for debugging templates as a result, not the specific template as you are trying to document.

So again, please restore the previous part and feel free to add another.

akasma74:
> Well, the PR currently removed the text on changing states, so yes, this PR removes a current valid debug method.

Just to clarify, debugging method for automations *without* templates only?
Because I think it's clear that it doesn't help when debugging automations with `trigger` in templates.

> It allows for debugging templates as a result, not the specific template as you are trying to document.

I'm trying to document a method relevant to the section - Trigger State Object. So everithing in that section should be related to `trigger` in templates. And that method is not, it's for automations that **don't** use `trigger`.

> So again, please restore the previous part and feel free to add another.

I think the right way is to add another section and call it like "Testing automations with templates" and put part of the "destroyed" debug method (not relvant to `trigger`) first as a generic approach and then my proposed method with the `trigger` bit from the "destroyed" method.

How does that sound?

frenck (approved):
Thanks, @akasma74 👍

PR #12375, opened 2020-03-14
Templating: open external links in new tab

ThomDietrich:
Not sure if Kramdown is supported by the used rendering engine. Please test! Thanks

### DIFF
--- a/source/_docs/automation/templating.markdown
+++ b/source/_docs/automation/templating.markdown
@@ -58,9 +58,9 @@ There are a few very important rules to remember when writing automation templat
 1. You **must** use `data_template` in place of `data` when using templates in the `data` section of a service call.
 1. You **must** use `service_template` in place of `service` when using templates in the `service` section of a service call.
 1. You **must** surround single-line templates with double quotes (`"`) or single quotes (`'`).
-1. It is advised that you prepare for undefined variables by using `if ... is not none` or the [`default` filter](http://jinja.pocoo.org/docs/dev/templates/#default), or both.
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hello @klaasnicolaas,
PR is already against current and another meta change. 
On second thought. You or one of the other main contributors/maintainers should discuss how to deal with external facing links in general. For sure the links in this PR are not the only ones and maybe a more systematic and sustainable solution to this issue can be found. Food for thought. Best!

klaasnicolaas:
@frenck what are your thoughts on this PR?

frenck:
We don't use Kramdown.

image

We have discussed the opening of external links in a new window in the past, at the time we decided not to do that.

However, if we wanted to, it would be an easy adjustment to do that globally.

For reference: <https://github.com/home-assistant/home-assistant.io/blob/next/plugins/output_modder.rb#L18>

ThomDietrich:
Thanks for the answer and testing Franck.

> However, if we wanted to, it would be an easy adjustment to do that globally.

You say this and closed the PR. Are we going to track this topic in another issue or who is going to build a PR for this?

**Edit:**

> at the time we decided not to do that.

I'd suggest to reevaluate this now. The current behavior is not to norm and feels immature

frenck:
> You say this and closed the PR. 

Yes, this PR does not work as intended.

> Are we going to track this topic in another issue or who is going to build a PR for this?

I don't see a direct reason for this, but feel free to open up an issue to discuss this.

> I'd suggest to reevaluate this now. The current behavior is not to norm and feels immature

I don't agree with that call, especially since you didn't elaborate on any reasoning for it.

ThomDietrich:
Closing the PR was the right call.
I didn't give a reason as I thought it's basically clear and I like to keep discussions short and action-oriented.

That's an oversight on my side, sorry for that. I'll open a proper issue to start a discussion.

ThomDietrich:
I'm always willing to accept new facts.

According to this source https://css-tricks.com/use-target_blank/#article-header-id-2 it is in fact not correct to generally change the default behavior for "external" links.

I'll rest my case ;)

PR #12377, opened 2020-03-15
Update instructions regarding Docker on Mac and USB passthrough

backflip:
I was unable to see my Z-Wave stick in the VM. Based on the linked articles, there might be a workaround. However, I haven't tested it since I just dropped the whole Docker approach instead. Is anyone else able to clarify?

### DIFF
--- a/source/_docs/installation/docker.markdown
+++ b/source/_docs/installation/docker.markdown
@@ -267,3 +267,7 @@ or in a `docker-compose.yml` file:
       restart: always
       network_mode: host
 ```
+
+<div class='note'>
+On Mac, USB devices are <a href="https://github.com/docker/for-mac/issues/900">not passed through</a> by default. Follow the instructions in <a href="https://dev.to/rubberduck/using-usb-with-docker-for-mac-3fdd">Using USB with Docker for Mac</a> by Christopher McClellan if your device is not showing up.
+</div>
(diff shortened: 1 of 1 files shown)


source/_docs/installation/docker.markdown
```diff
@@ -267,3 +267,7 @@ or in a `docker-compose.yml` file:
       restart: always
       network_mode: host
 ```
+
+<div class='note'>
+On Mac, USB devices are <a href="https://github.com/docker/for-mac/issues/900">not passed through</a> by default. Follow the instructions in <a href="https://dev.to/rubberduck/using-usb-with-docker-for-mac-3fdd">Using USB with Docker for Mac</a> by Christopher McClellan if your device is not showing up.
```
frenck: Please use common Markdown for creating links.

frenck (changes requested):
Thanks for the PR @backflip!

I've left a small comment, could you please take a look? Thanks! 👍

source/_docs/installation/docker.markdown
backflip: @frenck Good point. Should I change the last paragraph of `Synology NAS` as well? I copied mine from there. What's important to note: In order for this to work, we'll need an empty line after `<div class='note'>` and before `</div>`.

frenck: Yes, it needs an empty line to separate HTML from Markdown, as is documented in our documentation standards.

Improvements to other parts of the documentation is, of course, always welcome.

backflip: Whoups, sorry, RTFM. :)
Fixed.

frenck (approved):
Thanks, @backflip! 👍

PR #12410, opened 2020-03-17
add filter float description

akasma74:
## Proposed change

Actually, it's a bit confusing to realise we have a HA global function float(value) and a Jinja filter float(value, default)
The current documentation does not mention that and even though the later is part of Jinja, I think it's worth to add its description as well so users can see the difference and choose which one to use.
Also I've done some minor wording corrections and made filters/regex section more consistent by removing `value`/`string` where it's unnecessary and referring to it as `input`.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/configuration/templating.markdown
+++ b/source/_docs/configuration/templating.markdown
@@ -55,9 +55,9 @@ Extensions allow templates to access all of the Home Assistant specific states a
 - Iterating `states.domain` will yield each state of that domain sorted alphabetically by entity ID.
 - `states.sensor.temperature` returns the state object for `sensor.temperature` (avoid when possible, see note below).
 - `states('device_tracker.paulus')` will return the state string (not the object) of the given entity or `unknown` if it doesn't exist.
-- `is_state('device_tracker.paulus', 'home')` will test if the given entity is the specified state.
-- `state_attr('device_tracker.paulus', 'battery')` will return the value of the attribute or None if it doesn't exist.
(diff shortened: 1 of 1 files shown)


source/_docs/configuration/templating.markdown
```diff
@@ -315,40 +315,41 @@ Closest to some entity:
 
 ### Formatting
 
-- `float` will format the output as float.
+- Global function `float` will attempt to convert given value to float.
```
fabaff, line 318: I don't think that we should start using `global` because without additional details it's not clear what it means. E.g., difference to `local` functions.

source/_docs/configuration/templating.markdown
```diff
@@ -315,40 +315,41 @@ Closest to some entity:
...
 
 ### Numeric functions and filters
 
-Some of these functions can also be used in a [filter](https://jinja.palletsprojects.com/en/master/templates/#id11). This means they can act as a normal function like this `sqrt(2)`, or as part of a filter like this `2|sqrt`.
+Some of these global functions can also be used as [filters](https://jinja.palletsprojects.com/en/master/templates/#other-operators). This means they can be used as a normal function like `sqrt(2)`, or as part of an expression like `2|sqrt`.
```
fabaff, line 322: Dito

source/_docs/configuration/templating.markdown
akasma74: Well, I wanted to highlight the fact that it's not a function of a particular class to avoid interpretations like "calling my_var.float() is fine".
Do you think it's important here or it's better to just drop "global"?

akasma74:
Adios field marshall

PR #12430, opened 2020-03-18
Needs to be fuel_types: instead of fuel_type:

plantoschka:
Error in example configuration.

This will lead to an invalid configuration.


frenck (approved):
Thanks, @plantoschka! Well spotted 👀

PR #12446, opened 2020-03-19
Update HEOS documentation style

klada:
- Use YAML instead of JSON in examples
- Add additional column in tables to clarify whether an attribute is required or not for a service call.

## Proposed change

The HEOS docs currently use JSON for examples and don't make it clear whether attributes in service calls are optional or not. This change has been suggested in #12373 

## Type of change

## Additional information

https://github.com/home-assistant/home-assistant.io/pull/12373

## Checklist

### DIFF
--- a/source/_integrations/heos.markdown
+++ b/source/_integrations/heos.markdown
@@ -47,17 +47,15 @@ A connection to a single device enables control for all devices on the network. 
 
 Use the sign-in service to sign the connected controller into a HEOS account so that it can retrieve and play HEOS favorites and playlists. An error message is logged if sign-in is unsuccessful. Example service data payload:
 
-```json
-{
-  "username": "example@example.com",
-  "password": "password"
-}
+```yaml
+username: "example@example.com"
+password: "password"
 ```
 
-| Attribute              | Description
-| ---------------------- | ---------------------------------------------------------|
-| `username`             | The username or email of the HEOS account. [Required]
-| `password`             | The password of the HEOS account. [Required]
+| Service data attribute | Optional | Description |
+| ---------------------- | -------- | ----------- |
+| `username`             | no       | The username or email of the HEOS account.
+| `password`             | no       | The password of the HEOS account.
 
 ### Service `heos.sign_out`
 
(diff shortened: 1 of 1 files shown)


klada:
@frenck We were talking about this in #12373

source/_integrations/heos.markdown
```diff
@@ -69,73 +67,65 @@ Use the sign-out service to sign the connected controller out of a HEOS account.
...
 ```
 
-| Attribute              | Description
-| ---------------------- | ---------------------------------------------------------|
-| `entity_id`            | `entity_id` of the player
-| `media_content_type`   | Set to the value `favorite`
-| `media_content_id`     | The nubmer (i.e., `1`) or name (i.e., `Thumbprint Radio`) of the HEOS favorite
+| Service data attribute | Optional | Description |
+| ---------------------- | -------- | ----------- |
+| `entity_id`            | no       | `entity_id` of the player
```
frenck: `entity_id` is optional, because alternatively an area could be passed. This applies to all others as well.

frenck (changes requested):
Thanks, @klada!

One small comment, could you take a look? 👍

source/_integrations/heos.markdown
klada: @frenck I cannot find any reference to this in any of the player docs. The official docs also states that `entity_id` is required: https://www.home-assistant.io/integrations/media_player/#service-media_playerplay_media
Are the official `media_player` docs wrong as well?

frenck: That would be incorrect as well in that case.

Let me guide you through the codebase:

The entity services schemas are created in the service register call, using the `make_entity_service_schema` method.

For example, the `play_media` is defined here: <https://github.com/home-assistant/core/blob/dev/homeassistant/components/media_player/__init__.py#L307>

This is the `make_entity_service_schema` function:

<https://github.com/home-assistant/core/blob/62f2ee5f60586402e90ac868354550571e96cb2a/homeassistant/helpers/config_validation.py#L790>

As you can see in that function, entity ID is declared optional:


`vol.Optional(ATTR_ENTITY_ID): comp_entity_ids,`

<https://github.com/home-assistant/core/blob/62f2ee5f60586402e90ac868354550571e96cb2a/homeassistant/helpers/config_validation.py#L798>

Conclusion: `entity_id` is optional.

klada: Okay, I see. I have updated the docs and tried to explain that either `entity_id` or `area_id` is required.

source/_integrations/heos.markdown
```diff
@@ -69,73 +67,69 @@ Use the sign-out service to sign the connected controller out of a HEOS account.
...
 
-| Attribute              | Description
-| ---------------------- | ---------------------------------------------------------|
-| `entity_id`            | `entity_id` of the player
-| `media_content_type`   | Set to the value `favorite`
-| `media_content_id`     | The nubmer (i.e., `1`) or name (i.e., `Thumbprint Radio`) of the HEOS favorite
+| Service data attribute | Optional | Description |
+| ---------------------- | -------- | ----------- |
+| `entity_id`            | yes      | Can be a single player entity or a list of player entities. **Either `entity_id` or `area_id` is required.** |
+| `area_id`              | yes      | A single area or a list of areas with media players. **Either `entity_id` or `area_id` is required.** |
```
frenck: We currently don't document `area_id` because it isn't used manually. Furthermore, it isn't "either", both or none can be provided.

Please just update it to what it was before, and set entity_id to be not required.

klada: Okay, I probably wanted too much there. I have removed `area_id` and the wrong note.

frenck (approved):
Thanks, @klada 👍

PR #12592, opened 2020-03-31
0.108.0 Beta release notes

frenck:
## Proposed change

Release blog for Home Assistant Core 0.108


zsarnett:
> The graph from the sensor card is now also available as a footer for the entities card.

This was added to the Entity card as well. Not sure if we want to make that know here

frenck:
@zsarnett Thanks, will ensure it is in. The current text is not in my local draft as is in this PR right now 😉 

Even better, I'll post the more final version later today, maybe you can suggest a change? 👍

source/_posts/2020-04-08-release-108.markdown
```diff
@@ -11,7 +11,969 @@ categories: Release-Notes
...
+- Selecting your default Lovelace dashboard has been moved to the user
+  profile page.
+
+- A lot of configuration pages are migrated to data tables, this means you can
+  now search in your automations, scenes, scripts, users and areas.
+
+- Lovelace cards in panel views can now also be edited with the UI editor.
+
+- The graph from the sensor card is now also available as a footer for the
+  entities card.
```
zsarnett:
```suggestion
- The graph from the sensor card is now also available as a footer for the
  entities and entity card.
```

PR #12618, opened 2020-04-01
Must be integer

Cooper-Dale:
better example with integer

### DIFF
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -79,7 +79,7 @@ Delays are useful for temporarily suspending your script and start it at a later
 # Waits however many seconds input_number.second_delay is set to
 - delay:
     # Supports milliseconds, seconds, minutes, hours, days
-    seconds: "{{ states('input_number.second_delay') }}"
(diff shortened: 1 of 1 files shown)


frenck:
Can you elaborate on the change? A template is always a string in the end...
Home Assistant will try to convert it, however, the `| int` will not make a difference in that case.

Cooper-Dale:
i was getting an error
`2020-04-01 18:17:45 ERROR (MainThread) [homeassistant.components.automation] Větrákuj: Error rendering Větrákuj delay template: expected int for dictionary value @ data['minutes']`
with input number defined like
```
input_number:
  vetrak_timeout:
    name: Timeout větráku v minutách
    initial: 5
    min: 1
    max: 30
    step: 1
    mode: slider
    unit_of_measurement: min
    icon: mdi:timeline-clock-outline
```
in automation action without `| int` as is in the original example
```
   action:
    ....
    - delay:
        minutes: "{{ states('input_number.vetrak_timeout') }}"
    .....
```
HA: 0.107.7

frenck (approved):
Check, thanks, @Cooper-Dale! 👍

PR #12981, opened 2020-04-15
You need to include the port in the URI

willidh74:
## Proposed change
The port was forgotten in the URI - so added ":8123" after the IP addresses

### DIFF
--- a/source/_integrations/spotify.markdown
+++ b/source/_integrations/spotify.markdown
@@ -31,8 +31,8 @@ To create the required Spotify application:
 - Select **Create An App**. Enter any name and description.
 - Once your application is created, view it and copy your **Client ID** and **Client Secret**, which are used in the Home Assistant configuration file.
 - Add a **Redirect URI** in one of the following forms:
-  - If you are not using SSL: `http://<your_home_assistant_url_or_local_ip>:8123/auth/external/callback`
-  - If you are using SSL: `https://<your_home_assistant_url_or_local_ip>:8123/auth/external/callback`
+  - If you are not using SSL: `http://<your_home_assistant_url_or_local_ip>:<port>/auth/external/callback`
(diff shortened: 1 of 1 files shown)


source/_integrations/spotify.markdown
```diff
@@ -31,8 +31,8 @@ To create the required Spotify application:
 - Select **Create An App**. Enter any name and description.
 - Once your application is created, view it and copy your **Client ID** and **Client Secret**, which are used in the Home Assistant configuration file.
 - Add a **Redirect URI** in one of the following forms:
-  - If you are not using SSL: `http://<your_home_assistant_url_or_local_ip>/auth/external/callback`
-  - If you are using SSL: `https://<your_home_assistant_url_or_local_ip>/auth/external/callback`
+  - If you are not using SSL: `http://<your_home_assistant_url_or_local_ip>:8123/auth/external/callback`
+  - If you are using SSL: `https://<your_home_assistant_url_or_local_ip>:8123/auth/external/callback`
```
frenck: That depends on your setup...
Mine doesn't run on 8123...

willidh74: Good point. Am new to HA.
So what would be the most suitable way to indicate that a port is most likely required, as the instructions were unclear?

frenck: Personally, I don't think it is needed, as the template says it needs your URL (of which the port is part of).

However, is we do this, lets make the port number a placeholder as well..

i.e.:
```
http://<your_home_assistant_url_or_local_ip>:<port>/auth/external/callback
```

willidh74: Sounds good - want to make the change or should I ?
What confused me is that it mentions an IP and for most people they wouldn't assume that a port is included in an IP address

frenck (approved):
Nice! Thanks, @willidh74! 🥇

PR #13134, opened 2020-04-22
Update Amcrest:markdown

jamesarbrown:
## Proposed change
Following new code mechanisms by @pnbrucker the scan_interval is now obsolete as the camera polling has been replaced by a new event mechanism. Also added a note that the new mechanism does not work with a certain block of FW.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/amcrest.markdown
+++ b/source/_integrations/amcrest.markdown
@@ -92,7 +92,7 @@ authentication:
   required: false
   type: string
   default: basic
-scan_interval:
+scan_interval: (Obsolete from 108.5)
   description: Defines the update interval of the sensor in seconds.
   required: false
   type: integer
@@ -142,6 +142,11 @@ Newer Amcrest firmware may not work, then `rtsp` is recommended instead.
 **Note:** If you set the `stream_source` option to `rtsp`,
 make sure to follow the steps mentioned at [FFmpeg](/integrations/ffmpeg/)
 documentation to install the `ffmpeg`.
+
+**Note:** From 108.5 a new motion detection mechanism has been implemented for Amcrest cameras
+to improve speed response, whilst lowering network demands. Some cameras have a 
+propreitary bug in the firmware (V2.420.0009.0.R.20151106) and will return a 500 Bad Response
+error and are incompatible with Motion Detection.
 
 ## Services
 
--- a/source/_integrations/amcrest.markdown
+++ b/source/_integrations/amcrest.markdown
@@ -92,8 +92,8 @@ authentication:
   required: false
   type: string
   default: basic
-scan_interval: (Obsolete from 108.5)
-  description: Defines the update interval of the sensor in seconds.
+scan_interval:
+  description: Defines the update interval of the sensor in seconds. This setting is fixed from HA version 108.5 and scan_interval has no effect
   required: false
   type: integer
   default: 10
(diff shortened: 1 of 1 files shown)


source/_integrations/amcrest.markdown
```diff
@@ -92,7 +92,7 @@ authentication:
   required: false
   type: string
   default: basic
-scan_interval:
+scan_interval: (Obsolete from 108.5)
```
frenck: This breaks the build

source/_integrations/amcrest.markdown
```diff
@@ -143,6 +143,11 @@ Newer Amcrest firmware may not work, then `rtsp` is recommended instead.
 make sure to follow the steps mentioned at [FFmpeg](/integrations/ffmpeg/)
 documentation to install the `ffmpeg`.
 
+**Note:** From 108.5 a new motion detection mechanism has been implemented for Amcrest cameras
+to improve speed response, whilst lowering network demands. Some cameras have a 
+propreitary bug in the firmware (V2.420.0009.0.R.20151106) and will return a 500 Bad Response
+error and are incompatible with Motion Detection.
```
frenck: Small typo. Furthermore, maybe we should use the infoboxes we have for this:

<https://developers.home-assistant.io/docs/documentation_create_page#html>

```suggestion
**Note:** From 108.5 a new motion detection mechanism has been implemented for Amcrest cameras
to improve speed response, whilst lowering network demands. Some cameras have a 
proprietary bug in the firmware (V2.420.0009.0.R.20151106) and will return a 500 Bad Response
error and are incompatible with Motion Detection.
```

frenck (changes requested):
Thanks for the PR @jamesarbrown!

Left a comment, could you please take a look? 👍

klaasnicolaas:
something for the alert website? @frenck

frenck:
I honestly don't know. This does not affect a Home Assistant version, but a specific firmware version from the vendor... Would be an endless alert? 🤷

source/_integrations/amcrest.markdown
jamesarbrown: Is there a deprecated format for this? 

Would this work?

```
scan_interval

    (integer)(Optional)(Before Version 108.5)

    Defines the update interval of the sensor in seconds.

    Default value:

    10

```

klaasnicolaas: Just put it in the description?

source/_integrations/amcrest.markdown
```diff
@@ -93,7 +93,7 @@ authentication:
   type: string
   default: basic
 scan_interval:
-  description: Defines the update interval of the sensor in seconds.
+  description: Defines the update interval of the sensor in seconds. This setting is fixed from HA version 108.5 and scan_interval has no effect
```
frenck, line 96: This is a bit weird? According to the code it is still there?

If it has no effect, a bug should be raised.

frenck (changes requested):

source/_integrations/amcrest.markdown
jamesarbrown: Hi Frenck,

Please see comment by @pnbruckner
https://github.com/home-assistant/core/issues/33875#issuecomment-617189318

But it seems although this was closed "wontfix"
https://github.com/tchellomello/python-amcrest/issues/155

That @pnbrucker has had a change of heart
https://github.com/home-assistant/core/pull/36955

So this markdown needs dropping.

James

pnbruckner:
For "the record", the `scan_interval` configuration option stopped having an effect when the integration was changed long ago from "platform based" sensors to "hub based." This was before my time. I made an attempt to "fix" that (see architecture issue [#194](https://github.com/home-assistant/architecture/issues/194)), but the approach was not approved.

PR #13196, opened 2020-04-28
Fix for changed location of Securing Debian Manual

AdriaanIO:
The location of the Securing Debian Manual on debian.org has been changed, causing the link to be broken.


frenck (approved):
Thanks, @AdriaanIO! 👍

PR #13216, opened 2020-04-29
Yeelight: Minimum brightness incorrect

pakwan:
The minimum brightness when using yeelight transitions is 1 and not 0. This can be seen in the python-yeelight docs (https://yeelight.readthedocs.io/en/stable/yeelight.html#flow-objects) on which this seems to be based. Entering 0 as brightness results in the following error on my machine:
Unable to set scene: {'code': -5000, 'message': 'general error'}

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/yeelight.markdown
+++ b/source/_integrations/yeelight.markdown
@@ -265,9 +265,9 @@ Possible transitions are `RGBTransition`, `HSVTransition`, `TemperatureTransitio
 
 Where the array values are as per the following:
 
-- RGBTransition: [red, green, blue, duration, brightness] with red / green / blue being an integer between 0 and 255, duration being in milliseconds (minimum of 50) and final brightness to transition to 0-100 (%).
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
In my opinion, this is a bug and should be fixed in Home Assistant Core.

Even if the device does not support it, it should still be handled correctly.

pakwan:
I think it's in the grey area between "bug" and "used incorrectly" if a user inputs a brightness of 0 in their service calls. But I agree with you that adding a fix (for example modifying brightness 0 to 1) in Home Assistant Core is the user-friendlier version. I'm not sure if you assume that I'm going to make that fix, but if you do I want to make clear that I'm not, as I have absolutely no experience with developing Home Assistant Core :-) Sry!

frenck (approved):
Double checked it, it is actually a correct PR, thanks @pakwan 👍

PR #13293, opened 2020-05-04
Added firewall setup

erwin-willems:
## Proposed change
Documentation change:
Added firewall information to homekit setup

### DIFF
--- a/source/_integrations/homekit.markdown
+++ b/source/_integrations/homekit.markdown
@@ -358,7 +358,13 @@ homekit:
 ```
 
 Restart your Home Assistant instance. This feature requires running an mDNS forwarder on your Docker host, e.g., `avahi-daemon` in reflector mode. This kind of setup most likely requires `safe_mode` during the bridge setup.
- 
+
+## Firewall
+If you have a firewall configured on your Home Assistant server, make sure you open the following ports:
+
+- UDP: 5353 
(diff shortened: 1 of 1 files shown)


fabaff (approved):

source/_integrations/homekit.markdown
```diff
@@ -358,7 +358,14 @@ homekit:
...
 
 Restart your Home Assistant instance. This feature requires running an mDNS forwarder on your Docker host, e.g., `avahi-daemon` in reflector mode. This kind of setup most likely requires `safe_mode` during the bridge setup.
- 
+
+## Firewall
+
+If you have a firewall configured on your Home Assistant system, make sure you open the following ports:
+
+- UDP: 5353 
+- TCP: 51827
```
frenck, line 367: This is not correct. The TCP port is not a fixed port.

erwin-willems: It is a configured port, see https://www.home-assistant.io/integrations/homekit/#configuration-variables

frenck: Yes, that is why it is not "fixed" as this example/comment does imply.
Furthermore, for 0.110, the HomeKit moves to the UI configuration flow, which allows for multiple HomeKit integration instances which makes for multiple ports even.

PR #13350, opened 2020-05-08
Home connect

DavidMStraub:
Hi @frenck, thanks for merging my PR #11336! However since it is already a couple of months old, I realized that a few minor things that were changed during the PR discussion in the Core repo need to be changed:

- The integration was renamed to `homeconnect` to `home_connect`
- The attributes of the power switch have been removed
- The elapsed time sensor has been removed
- The remaining time sensor has been converted to a timestamp sensor.

This PR reflects these changes in the docs.


frenck (approved):
Thanks, @DavidMStraub 👍

PR #13367, opened 2020-05-09
Revert "Cleanup old references of Supervised"

Kanga-Who:
Reverts home-assistant/home-assistant.io#13351


frenck:
This is not being reverted.

PR #13409, opened 2020-05-12
Added redirect information

sanathdw:
Added the following information 
- `http(s)://<your_home_assistant_url_or_local_ip>/api/spotify`
- `http(s)://<your_home_assistant_url_or_local_ip>/lovelace/<the id of the tab on HA where the card is stored>`

### DIFF
--- a/source/_integrations/spotify.markdown
+++ b/source/_integrations/spotify.markdown
@@ -33,6 +33,10 @@ To create the required Spotify application:
 - Add a **Redirect URI** in one of the following forms:
   - If you are not using SSL: `http://<your_home_assistant_url_or_local_ip>:<port>/auth/external/callback`
   - If you are using SSL: `https://<your_home_assistant_url_or_local_ip>:<port>/auth/external/callback`
+  
+  You will reuire the following two redirets as well
+  - `http(s)://<your_home_assistant_url_or_local_ip>/api/spotify`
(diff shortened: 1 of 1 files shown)


source/_integrations/spotify.markdown
```diff
@@ -33,6 +33,10 @@ To create the required Spotify application:
 - Add a **Redirect URI** in one of the following forms:
   - If you are not using SSL: `http://<your_home_assistant_url_or_local_ip>:<port>/auth/external/callback`
   - If you are using SSL: `https://<your_home_assistant_url_or_local_ip>:<port>/auth/external/callback`
+  
+  You will reuire the following two redirets as well
+  - `http(s)://<your_home_assistant_url_or_local_ip>/api/spotify`
+  - `http(s)://<your_home_assistant_url_or_local_ip>/lovelace/<the id of the tab on HA where the card is stored>`
```
frenck, line 39: This does not apply to this integration?

I think this applies to a custom integration maybe? As this integration doesn't do anything with Lovelace for example...

frenck (changes requested):

source/_integrations/spotify.markdown
ludeeus: That is for the custom spotify card, which already has it in its documentation.
<https://github.com/custom-cards/spotify-card#installation>

frenck:
As confirmed by @ludeeus, it is for custom integration and not related to this documentation.

Therefore, closing this PR. Nevertheless, thanks for contributing 👍

PR #10798, opened 2019-10-15
Cleanup of Cookbook examples Round 3

ghost:
**Description:**
Small cleanup of the cookbook examples like I have done in the past. 
https://github.com/home-assistant/home-assistant.io/pull/9091
https://github.com/home-assistant/home-assistant.io/pull/6645

Anything not updated in the past 6 months was removed.  As in the past, with the pace of Home Assistant these examples can be outdated or have breaking changes if not maintained current. 

**Pull request in home-assistant (if applicable):** home-assistant/home-assistant#<home-assistant PR number goes here>

## Checklist:

- [X ] Branch: `next` is for changes and new documentation that will go public with the next Home Assistant release. Fixes, changes and adjustments for the current release should be created against `current`.
- [X ] The documentation follows the [standards][standards].

[standards]: https://developers.home-assistant.io/docs/documentation_standards.html

PR #6645, opened 2018-10-09
Cleanup of Cookbook examples

ghost:
Cleanup of the cookbook examples.  I still use these occasionally for inspiration or to look for example code.  There are a number that are not maintained (last commit > 1 year), contain outdated examples of platforms or components, or reference depreciated items.

Given how quickly this project moves I think it is reasonable to expect 'cookbook' examples to be somewhat current to avoid giving examples with breaking changes in later versions.  

## Checklist:

- [X ] Branch: `next` is for changes and new documentation that will go public with the next [home-assistant](https://github.com/home-assistant/home-assistant) release. Fixes, changes and adjustments for the current release should be created against `current`.
- [X ] The documentation follow the [standards][standards].

[standards]: https://home-assistant.io/developers/documentation/standards/

PR #13595, opened 2020-05-26
Cleanup of Cookbook examples Round 4

ghost:
## Proposed change
Small cleanup of the cookbook examples like I have done in the past.
#10798
#9091
#6645

## Type of change
Removed any repositories that are greater than 6 months since the last update.  This is to keep the examples up to date with the fast moving project .


frenck (approved):
Thanks, @SilvrrGIT! 👍

PR #13648, opened 2020-06-02
Update code examples to remove comments

aidbish:
Removed comments from inside of code examples as this implies comments are allowed although they are lost as noted at bottom of the page. Just removes any additional confusion.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/automation/editor.markdown
+++ b/source/_docs/automation/editor.markdown
@@ -35,8 +35,8 @@ As "Service Data" we want a simple text that is shown as part of the notificatio
 
 The Automation Editor does not currently support **data_template** as a "Service Data" entry. For these actions, it will be necessary to edit as YAML. This can be achieved by clicking the menu (3 vertical dots) and selecting "Edit as YAML"
 
+Example of how to use data_template in the Automation Editor, using "Edit as YAML"
 ```yaml
-# Example of how to use data_template in the Automation Editor, using "Edit as YAML"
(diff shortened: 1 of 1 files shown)


frenck:
I don't think we should do this. They can get lost based on the method you use, however, this is not always the case. There is nothing wrong with adding code comments to documentation to clarify inline with what is happening.

I think the adjusted example you've made clearly shows it isn't getting clearer to read the example.

image

aidbish:
Hi @frenck thanks for the feedback
I was under the impressions that the Automations editor formats the code and removes comments regardless.
what other methods are there when it doesnt whilst using the automations editor?

i agree it may look as equally as not being clear but on the web page it shows it as part of a config which maybe copied and then their own comments added and code. Once saved the comments are removed leading to the discussions recently seen in the forums of HA mangling YAML and changing their files.

frenck:
The automations editor only writes to the `automations.yaml` file, however, automation can be placed in other files as well.

aidbish:
I agree about where they can be placed, but the documentation to be modified is about the automations editor only.  If i add automation to another place is it read by the editor if i use id:.

frenck:
It is not read by the automation editor, it is read by Home Assistant and added to the internal automations collection.

I really don't see a direct reason to make the documentation less clear.

I thank you for your thoughts but going to pass on this one. 👍

PR #13841, opened 2020-06-24
Update tplink.markdown

guitarmanusa:
## Proposed change
Since the deprecation of the 'scan_interval' setting TP  Link devices take approx 30s to respond to any automation.  User ljmerza identified a fix to allow users to define their own scan interval for specific devices @ <a href="https://github.com/home-assistant/core/issues/21768#issuecomment-470959458">#21768 (comment)</a>.

## Type of change
Adds documentation to address a common mismatch between expected and actual behavior that does not require a code change, only an additional automation to resolve.

## Additional information
This was reported in <a href="https://github.com/home-assistant/core/issues/21768">Issue 21768</a> which is marked as closed, however the fix action is not clear in the documentation.

## Checklist

### DIFF
--- a/source/_integrations/tplink.markdown
+++ b/source/_integrations/tplink.markdown
@@ -169,3 +169,24 @@ sensor:
         unit_of_measurement: 'kWh'
 ```
 {% endraw %}
+
+## Increase Polling Frequency
+
+Home Assistant deprecated the 'scan_interval' setting.  TP Link devices seem to be particularly affected by this change as many users report TP Link devices take upwards of 15-30 seconds to respond to automations.  If you find that the automatically determined polling frequency is too low  you can set an automation to poll specific devices at a user defined frequency with the following automation.
+
+{% raw %}
+```yaml
+# automations.yaml
+- id: update_tp_link
+  initial_state: 'on'
+  trigger:
+    - platform: time_pattern
+      seconds: '/5' # every 5 seconds
+  action:
+    - service: homeassistant.update_entity
+      data:
+        entity_id: 
+          - light.tp_link_light
+          - switch.tp_link_switch
+```
+{% endraw %}
--- a/source/_integrations/tplink.markdown
+++ b/source/_integrations/tplink.markdown
@@ -172,9 +172,8 @@ sensor:
 
 ## Increase Polling Frequency
 
-Home Assistant deprecated the 'scan_interval' setting.  TP Link devices seem to be particularly affected by this change as many users report TP Link devices take upwards of 15-30 seconds to respond to automations.  If you find that the automatically determined polling frequency is too low  you can set an automation to poll specific devices at a user defined frequency with the following automation.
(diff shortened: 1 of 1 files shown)


source/_integrations/tplink.markdown
```diff
@@ -169,3 +169,22 @@ sensor:
         unit_of_measurement: 'kWh'
 ```
 {% endraw %}
+
+## Increase Polling Frequency
+
+Home Assistant deprecated the 'scan_interval' setting. TP Link devices seem to be particularly affected by this change, as many users report TP Link devices take upwards of 15-30 seconds to respond to automations. If you find that the automatically determined polling frequency is too low, you can set an automation to poll specific devices at a user defined frequency with the following automation.
```
frenck, line 175: I think there is a misunderstanding going on here...

source/_integrations/tplink.markdown
```diff
@@ -169,3 +169,22 @@ sensor:
...
+  trigger:
+    - platform: time_pattern
+      seconds: '/5' # every 5 seconds
+  action:
+    - service: homeassistant.update_entity
+      data:
+        entity_id: 
+          - light.tp_link_light
+          - switch.tp_link_switch
+```
```
frenck, line 190: This can be a workaround, but I don't think this should be in the docs.
This is a case of a wrong use or wrong/bug/incorrect use in the integration itself.

Documenting workarounds is the worst we can do, we should either fix the issue or add documentation on why the limits are set.

frenck (changes requested):

source/_integrations/tplink.markdown
guitarmanusa: The scan_interval option was removed around v0.80ish.  Since then core decides the interval to poll entities automatically.  As described, unless you force Home Assistant to poll TP Link devices *in specific* any automation will be delayed until the next poll, sometimes up to 30 seconds.  This is absolutely contrary to the expected behavior of the devices.  I would expect that when I push a button on one TP Link device that triggers an automation on several other TP Link devices that those triggered devices would respond almost instantly, not 30 seconds later when the next poll comes around.

I'd be in favor of fixing the automatic polling interval, but it seems like this was intentionally removed due to some devices being 'hammered' for updates.

frenck: I see your point, but that is not how it works. Your wished description matches for devices that support push updated, however, this is a polling system. That is unfortunately not something in our control, but a limitation of the devices involved (created by a third-party manufacturer).

There is always a delay when using polling. Furthermore, setting high polling rates causes other issues. Hammering devices or services is never a good idea.

One is free to work around, like is suggested in this PR, however, it is not something we should recommend or imply.

If a more sane, default, polling rate is possible, I suggest opening a PR to change that polling rate in that case.

No matter the solution in all cases, adding this to the documentation isn't wished for.
Nevertheless, thanks for your contribution 👍

PR #13901, opened 2020-07-04
Add new automation configuration options

pnbruckner:
## Proposed change

Add descriptions of new automation configuration options.

## Type of change

## Additional information

- Link to parent pull request in the codebase: home-assistant/core#37479

## Checklist

### DIFF
--- a/source/_docs/automation.markdown
+++ b/source/_docs/automation.markdown
@@ -43,17 +43,50 @@ Actions are all about calling services. To explore the available services open t
 
 ### Automation initial state
 
-When you create a new automation, it will be enabled unless you explicitly add `initial_state: false` to it or turn it off manually via UI/another automation/developer tools. In case automations need to be always enabled or disabled when Home Assistant starts, then you can set the `initial_state` in your automations. Otherwise, the previous state will be restored.
+When you create a new automation, it will be enabled unless you explicitly add `initial_state: off` to it or turn it off manually via UI/another automation/developer tools. In case automations need to be always enabled or disabled when Home Assistant starts, then you can set the `initial_state` in your automations. Otherwise, the previous state will be restored.
 
 Please note that if for some reason Home Assistant cannot restore the previous state, it will result in the automation being enabled.
 
-```text
+```yaml
 automation:
 - alias: Automation Name
-  initial_state: false
+  initial_state: off
   trigger:
   ...
 ```
+
+### Automation Modes
+
+The automation's `mode` configuration option controls what happens when the automation is triggered while the actions are still running from a previous trigger. The default mode is `legacy`.
+
+Mode | Description
+-|-
+`legacy` | See [below](#legacy-mode).
+`error` | Raise an error. Previous run continues normally.
+`ignore` | Do not start a new run. Previous run continues normally.
+`parallel` | Start a new, independent run in parallel with previous runs which continue normally.
+`restart` | Start a new run after first stopping previous run.
+`queue` | Start a new run after all previous runs complete. Runs are guaranteed to execute in the order they were queued. The maximum number of queued up runs is controlled by the `queue_size` configuration option, which defaults to 10.
+
+#### Example Setting Automation Mode
+
+```yaml
+automation:
+- mode: queue
+  queue_size: 5
+  trigger:
+  ...
+```
+
+#### Legacy Mode
+
+<div class='note'>
+
+This mode is deprecated, and a warning to that effect will be issued at startup unless `mode: legacy` is specified.
+
+</div>
+
+This mode maintains the legacy script behavior. That is, the actions will run until a `delay` step, or a `wait_template` step (that actually waits) is executed, at which point the actions will suspend. If the automation is triggered while the actions are suspended, it will abort the delay/wait_template and continue immediately to the next step, or finish if there are no more steps.
 
 ### Deleting Automations
 
--- a/source/_docs/automation.markdown
+++ b/source/_docs/automation.markdown
@@ -72,10 +72,10 @@ Mode | Description
 
 ```yaml
 automation:
-- mode: queue
-  queue_size: 5
-  trigger:
-  ...
+  - mode: queue
+    queue_size: 5
+    trigger:
+    ...
 ```
 
 #### Legacy Mode
--- a/source/_docs/automation.markdown
+++ b/source/_docs/automation.markdown
@@ -43,7 +43,7 @@ Actions are all about calling services. To explore the available services open t
 
 ### Automation initial state
 
-When you create a new automation, it will be enabled unless you explicitly add `initial_state: off` to it or turn it off manually via UI/another automation/developer tools. In case automations need to be always enabled or disabled when Home Assistant starts, then you can set the `initial_state` in your automations. Otherwise, the previous state will be restored.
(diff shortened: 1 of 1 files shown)


source/_docs/automation.markdown
```diff
@@ -43,18 +43,51 @@ Actions are all about calling services. To explore the available services open t
...
+
+This mode is deprecated, and a warning to that effect will be issued at startup unless `mode: legacy` is specified.
+
+</div>
+
+This mode maintains the legacy script behavior. That is, the actions will run until a `delay` step, or a `wait_template` step (that actually waits) is executed, at which point the actions will suspend. If the automation is triggered while the actions are suspended, it will abort the delay/wait_template and continue immediately to the next step, or finish if there are no more steps.
+
 ### Deleting Automations
 
 When automations remain visible in the Home Assistant Dashboard, even after having deleted in the YAML file, you have to delete them in the UI.
```
balloob, line 93: lol wtf is this.

balloob (dismissed):

source/_docs/automation.markdown
frenck: It is caused by the ID... if you delete an automation in YAML it will remain visible in the UI as a removable entity..: that throws people off

source/_docs/automation.markdown
```diff
@@ -43,18 +43,51 @@ Actions are all about calling services. To explore the available services open t
...
+When you create a new automation, it will be enabled unless you explicitly add `initial_state: off` to it or turn it off manually via UI/another automation/developer tools. In case automations need to be always enabled or disabled when Home Assistant starts, then you can set the `initial_state` in your automations. Otherwise, the previous state will be restored.
 
 Please note that if for some reason Home Assistant cannot restore the previous state, it will result in the automation being enabled.
 
-```text
+```yaml
 automation:
 - alias: Automation Name
-  initial_state: false
+  initial_state: off
```
frenck: Please stick to YAML 1.2 specifications for the documentation.

```suggestion
  initial_state: false
```

source/_docs/automation.markdown
```diff
@@ -43,18 +43,51 @@ Actions are all about calling services. To explore the available services open t
 
 ### Automation initial state
 
-When you create a new automation, it will be enabled unless you explicitly add `initial_state: false` to it or turn it off manually via UI/another automation/developer tools. In case automations need to be always enabled or disabled when Home Assistant starts, then you can set the `initial_state` in your automations. Otherwise, the previous state will be restored.
+When you create a new automation, it will be enabled unless you explicitly add `initial_state: off` to it or turn it off manually via UI/another automation/developer tools. In case automations need to be always enabled or disabled when Home Assistant starts, then you can set the `initial_state` in your automations. Otherwise, the previous state will be restored.
```
frenck: Please stick to YAML 1.2 specifications for the documentation.

```suggestion
When you create a new automation, it will be enabled unless you explicitly add `initial_state: false` to it or turn it off manually via UI/another automation/developer tools. In case automations need to be always enabled or disabled when Home Assistant starts, then you can set the `initial_state` in your automations. Otherwise, the previous state will be restored.
```

source/_docs/automation.markdown
```diff
@@ -43,18 +43,51 @@ Actions are all about calling services. To explore the available services open t
...
+`queue` | Start a new run after all previous runs complete. Runs are guaranteed to execute in the order they were queued. The maximum number of queued up runs is controlled by the `queue_size` configuration option, which defaults to 10.
+
+#### Example Setting Automation Mode
+
+```yaml
+automation:
+- mode: queue
+  queue_size: 5
+  trigger:
+  ...
```
frenck: In our docs, we indent the lists, as it is easier to read.

```suggestion
automation:
  - mode: queue
    queue_size: 5
    trigger:
    ...
```

source/_docs/automation.markdown
```diff
@@ -43,18 +43,51 @@ Actions are all about calling services. To explore the available services open t
...
+```
+
+#### Legacy Mode
+
+<div class='note'>
+
+This mode is deprecated, and a warning to that effect will be issued at startup unless `mode: legacy` is specified.
```
frenck, line 85: So how does our automation editor in the UI handle this?
Considering our users that don't have experience editing YAML...

frenck (changes requested):

source/_docs/automation.markdown
pnbruckner: Of course I'll do what you want, but could you please explain why? `off` seems to be accepted by the current YAML implementation, and even `'off'` is accepted by [`config_validation.boolean`](https://github.com/home-assistant/core/blob/c3b5bf74371499122cfb9af5ab2b9c35ab59f425/homeassistant/helpers/config_validation.py#L135-L148). And `off` is more natural here; even the description of the option uses the term "off".

source/_docs/automation.markdown
pnbruckner: It doesn't, AFAIK. I don't use the Automation Editor, and I have no experience for working on the frontend. In the parent (code) PR @balloob said:

> btw, we should make sure that we update the automation and script editors. We should create an issue in the frontend for that.

& CC'd bramkragten

source/_docs/automation.markdown
frenck: Because:
A) Not future proof. It isn't valid with the YAML 1.2 specification.
B) It is confusing for users: As `off` and `'off'` are not the same.
C) All documentation follows this standard.

source/_docs/automation.markdown
frenck: Ok, but what if a user is fairly new to Home Assistant and never used YAML and just the frontend?

pnbruckner: Then like has always happened, they don't get to use the advanced feature until someone gets around to implementing it in the frontend. As I said, I can't help with that. And the default is legacy mode, which means it's still compatible with the frontend as-is in the interim.

source/_docs/automation.markdown
pnbruckner: > It is confusing for users: As off and 'off' are not the same.

I would argue `false` is confusing. The entity's state is either `'on'` or `'off'`, not `true` or `false`. For good or bad, a decision was apparently made long ago that a binary entity's state is `'on'` / `'off'`, not `'true`'/`'false'`. But the documentation for `initial_state` says `false`??? _That_ is confusing.

And, as I said, the YAML & config validation implementations work together to make `'on'` & `on`, and `'off'` & `off` equivalent here.

I guess we'll have to agree to disagree.

source/_docs/automation.markdown
frenck: Sure but they will be confronted with an log messages and warnings, visible in the frontend, which was not their doing and have to learn YAML in order to fix it. I think we should have come up with a solution for the frontend before having merged this.

source/_docs/automation.markdown
frenck: I guess so. Please change it so it matches the rest of the documentation as requested.

source/_docs/automation.markdown
pnbruckner: Not my decision. Whatever y'all want to do is fine by me. ~~Another option would be to automatically add `mode: legacy` to all automations & scripts, but I don't know how to do that either.~~ I just tried to improve a script implementation that had serious flaws.

frenck: @pnbruckner Don't get me wrong! You did an amazing job there m8. I've upgraded my production because of this 👍

frenck (approved):
👍 Looks good @pnbruckner!

PR #13906, opened 2020-07-05
Fix modbus switch example configuration

llytvynenko:
The error in the documentation prevents newcomers from an easy start.
Please see https://community.home-assistant.io/t/modbus-component-cant-setup-platform/151156 as an example of misconfiguration that might happen because of the error in the docs.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/switch.modbus.markdown
+++ b/source/_integrations/switch.modbus.markdown
@@ -24,6 +24,7 @@ switch:
       slave: 1
       coil: 13
     - name: Switch2
+      hub: hub1
       slave: 2
       coil: 14
   registers:


source/_integrations/switch.modbus.markdown
```diff
@@ -24,6 +24,7 @@ switch:
       slave: 1
       coil: 13
     - name: Switch2
+      hub: hub1
```
frenck, line 27: Hub isn't a required option, and actually should be removed from all examples in the first example.

frenck (changes requested):

source/_integrations/switch.modbus.markdown
llytvynenko: While hub might not be required in general, it is definitely required for `modbus`, I've had this issue - that's why I've created the pull request.
If you don't set hub name you will get [configuration error](https://community.home-assistant.io/t/modbus-component-cant-setup-platform/151156).

frenck: According to the integration code, `hub` is optional and has a default value: <https://github.com/home-assistant/core/blob/dev/homeassistant/components/modbus/__init__.py#L83>

The documentation should reflect the integration code. If the integration handles it incorrect, that should be fixed. The documentation should not be used as a workaround.

llytvynenko: I'm not sure I'll be able to jump into the code and figure what's wrong, hopefully someone will jump in:)
But I totally agree with your points, let's discard this pull request than!

PR #13955, opened 2020-07-08
Update screenshots to match latest appearance

ghost:
Simple update, images were compressed but is still slightly larger than the previous images.


frenck (approved):
Awesome, thanks! 👍

Issue #13207, opened 2020-04-28
Update Card Documentations Configuration Variables


zsarnett:
https://www.home-assistant.io/lovelace/alarm-panel/#configuration-variables

Current Config Variables look daunting and advanced. Find a better way to view variables while keeping them together with the editor vs YAML

I currently don't have a plan or example of how these may look. Open to suggestions.

cadihowley:
Would a table work? 

Something like:

|Configuration Variable| Type| Required?| Description| Options| Default|
|-|-|-|-|-|-|
|type| string|yes|alarm-panel| | |
|entity| string| yes|Entity ID of ``alarm_control_panel`` domain| | |
|name| string| no|Overwrites Friendly Name| | Current State of Alarm Entity |
|states| list| no|Controls which states to have available| <ul><li>arm_home</li><li> arm_away</li><li> arm_away</li><li> arm_night</li> <li>arm_custom_bypass </li> </ul>| arm_home, arm_away|
|theme| string| no|Set to any theme within ```themes.yaml```| | |

frenck:
We had tables in the past, they get to narrow really quick

zsarnett:
We are looking for a way to display the data to an Editor User. This is still catering to the YAML use. The YAML use will be in a new Section below the editor descriptions

cadihowley:
Ok, I understand what we’re trying to do now. I’ll work on that!

zsarnett:
This is currently being worked on

zsarnett:
Still have a lot of cards to do

klaasnicolaas:
Thanks for raising this. The dashboard card documentation has evolved significantly since this issue was opened.

Cards such as the alarm panel card now document the UI-based configuration first and keep the YAML-specific configuration in a separate **YAML configuration** section. This pattern is also used throughout the current dashboard card documentation.

The original goal of this issue has therefore been addressed, so I'm closing this as completed. Thanks!

PR #14008, opened 2020-07-15
Configuration Variable Redesign

zsarnett:
## Proposed change

Redesign of the Configuration Variable for Integrations, Lovelace cards, etc

image

image

## Type of change

- [x] Style changes

## Additional information

- Linked to: #13207

## Checklist

### DIFF
--- a/plugins/configuration.rb
+++ b/plugins/configuration.rb
@@ -91,7 +91,7 @@ module Jekyll
           isDefault = true
           defaultValue = converter.convert(attr['default'].to_s)
         elsif attr['type'].to_s.include? 'boolean'
-          # Is the type is a boolean, a default key is required
+          # If the type is a boolean, a default key is required
           raise ArgumentError, "Configuration key '#{key}' is a boolean type and"\
             " therefore, requires a default."
         end
--- a/plugins/configuration.rb
+++ b/plugins/configuration.rb
@@ -12,6 +12,8 @@ module Jekyll
       'device_class', 'icon', 'map', 'list', 'date', 'datetime'
     ]
 
+    MIN_DEFAULT_LENGTH = 30
+
     def initialize(tag_name, text, tokens)
       super
       @component, @platform = text.strip.split('.', 2)
@@ -83,6 +85,16 @@ module Jekyll
             unless ['list', 'template'].include? parent_type
         end
 
+        defaultValue = ""
+        isDefault = false
+        if attr.key? 'default' and not attr['default'].to_s.empty? 
+          isDefault = true
+          defaultValue = converter.convert(attr['default'].to_s)
+        elsif attr['type'].to_s.include? 'boolean'
+          # Is the type is a boolean, a default key is required
+          raise ArgumentError, "Configuration key '#{key}' is a boolean type and"\
+            " therefore, requires a default."
+        end
         
         if attr.key? 'required'
           # Check if required is a valid value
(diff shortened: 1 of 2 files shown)


frenck:
Nice! That is a good leap forward!

image


I do think the font is out of wack compared to the rest of the website...

The last line break should not be there I think, it is probably the cause of the additional margin at the bottom as well?

Lastly, what annoys me (in both this and the old version) is the default value. It feels so... dunno; dumped at the end?

Just an idea (which personally isn't fully satisfying my inner soul either):

image

zsarnett:
The problem is that a default value could be a sentence rather than just a value. In the case of the weather card. I do like that idea though. 

The font is the same. Different size and the opacity is turned down. I feel like it separates it good enough but not too much. I also feel like the rest of the site should be different. But I'm working on that 😉

I'll take a look at the last line break. It's already something I do in the nested so easy to remove from the main

frenck:
> The problem is that a default value could be a sentence rather than just a value. In the case of the weather card. I do like that idea though.

Well, since we are in a loop, we can determine the length of the default value and behave differently?
I'm kinda sure that at least 95% of the cases it contains a single word/number.

e.g., in case the default value is less than 20 chars, show it on top, else in the bottom.

zsarnett:
> The last line break should not be there I think, it is probably the cause of the additional margin at the bottom as well?

This margin is from the h2 in the documentation for Tradfi

zsarnett:
Also on the font. 

Would we be so crazy to change our main font presets from

```css
color: #222;
font-size: 18px;
```

to 

```css
color: #4f566b;
font-size: 16px;
```

### from: 
image


### to:
image


I find it incredibly unnecessary to have such large text as our base. and having the headers so large and bold and uppercased. I think we can do better and look more professional

zsarnett:
image

plugins/configuration.rb
```diff
@@ -145,7 +166,7 @@ def render(context)
 
       <<~MARKUP
         <div class="config-vars">
-          <h3><a class="title-link" name="configuration-variables" href="#configuration-variables"></a> Configuration Variables</h3>
+          <h4><a class="title-link" name="configuration-variables" href="#configuration-variables"></a> Configuration Variables</h4>
```
frenck: This is always a H3, as the configuration heading is always H2. For SEO structure it is important to have that in order.

plugins/configuration.rb
```diff
@@ -102,14 +127,10 @@ def render_config_vars(vars:, component:, platform:, converter:, classes: nil, p
...
-          markup << "<div class='config-vars-default'>\nDefault value: #{converter.convert(attr['default'].to_s)}</div>"
-        elsif attr['type'].to_s.include? 'boolean'
-          # Is the type is a boolean, a default key is required
-          raise ArgumentError, "Configuration key '#{key}' is a boolean type and"\
-            " therefore, requires a default."
+        if isDefault && defaultValue.length > MIN_DEFAULT_LENGTH
+          markup << "<div class='config-vars-default'>\nDefault: #{defaultValue}</div>"
         end
-
+
```
frenck:
```suggestion

```

plugins/configuration.rb
```diff
@@ -83,14 +85,37 @@ def render_config_vars(vars:, component:, platform:, converter:, classes: nil, p
             unless ['list', 'template'].include? parent_type
         end
 
+        defaultValue = ""
+        isDefault = false
+        if attr.key? 'default' and not attr['default'].to_s.empty? 
+          isDefault = true
+          defaultValue = converter.convert(attr['default'].to_s)
+        elsif attr['type'].to_s.include? 'boolean'
+          # Is the type is a boolean, a default key is required
```
frenck: Not your fault, but easy fix:

```suggestion
          # If the type is a boolean, a default key is required
```

frenck:
I agree that we can reduce the font size in general. It is huge!

Decreasing the contrast with the background by smoothing down the font color... Not sure if that is ideal for everybody. Mainly because of our dirty background (which I personally hate, looks like it was washed a thousand times).

If we can fix the h3/h4 and merge those 2 little other tiny things, then this should be ready to rock 🤘

frenck:
Some older screenshots of experiments, those are actually live embedded demos

image


image

frenck:
Some dark play and font test

image

image

frenck:
All unrelated, and all have too many things going on / not right. Was just some idea gathering from the past.

So sorry about that 😉

plugins/configuration.rb
zsarnett: We should decrease the size and boldness of each heading eventually

plugins/configuration.rb
```diff
@@ -149,7 +165,7 @@ def render(context)
 
       <<~MARKUP
         <div class="config-vars">
-          <h3><a class="title-link" name="configuration-variables" href="#configuration-variables"></a> Configuration Variables</h3>
+          <h4><a class="title-link" name="configuration-variables" href="#configuration-variables"></a> Configuration Variables</h4>
```
zsarnett:
```suggestion
          <h3><a class="title-link" name="configuration-variables" href="#configuration-variables"></a> Configuration Variables</h3>
```

PR #14010, opened 2020-07-15
Basic Configuration for Editors + Weather Card as example

zsarnett:
## Proposed change

Basic Configuration for Editors - This is a new automated block created by Ruby/Jekyll

This is very similar to the YAML configuration variables. The difference is the tone we send when rendering these variables. This way it doesn't look like I need to be a developer to understand the docs. This is for the more non-developers using editors.

image

## Type of change

- [x] Addition of configuration block / Style updates

## Additional information

- This PR is reliant on: #14008 (Once this PR is merged this PR will need to be based on the new code)
- This PR fixes or closes issue: #13207

## Checklist

### DIFF
--- a/plugins/configuration_basic.rb
+++ b/plugins/configuration_basic.rb
@@ -1,0 +1,74 @@
+module Jekyll
+  class ConfigurationBasicBlock < Liquid::Block
+    def initialize(tag_name, text, tokens)
+      super
+      @component, @platform = text.strip.split('.', 2)
+    end
+
+    def slug(key)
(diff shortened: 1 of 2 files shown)


zsarnett:
What do you think on having a picture of the editor. I am not sure I like it. I am thinking about just changing it so that it says `Card Settings` and then the basic list without the image

frenck:
Pictures are hard to keep up 2 date... with the current development pace.

Furthermore, another thing I'm thinking of, this is not limited to Lovelace cards, but this could also be used for things like: Integration setup dialogs?

zsarnett:
image

frenck (approved):

PR #14016, opened 2020-07-16
Update upb.markdown

bbrendon:
polling is only done when hass boots to get in sync. after that it's push.


frenck (approved):
Thanks, @bbrendon! 👍

bbrendon:
Gwww corrected me. It polls when HASS connects to the PIM. Which means reconnects cause syncs. Almost the same as what I said originally but just want to clarify for future reads.

PR #14047, opened 2020-07-19
Whole Site Font Changes

zsarnett:
## Proposed change

It's better if you just look at the Netlify Preview 😉 

## Type of change

- [x] Yes

## Checklist

### DIFF
--- a/sass/custom/_paulus.scss
+++ b/sass/custom/_paulus.scss
@@ -77,7 +77,6 @@ $primary-color: #049cdb;
     font-weight: 400;
     letter-spacing: -.012em;
     line-height: 40px;
-    color: #212121;
     padding: 20px 16px 12px;
     text-transform: capitalize;
     margin: 0;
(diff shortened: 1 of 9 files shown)


sass/custom/_paulus.scss
```diff
@@ -114,15 +141,79 @@ $primary-color: #049cdb;
...
+  .vision {
+    margin-bottom: 24px;
+
+    h2 {
+      margin-bottom: 0;
+    }
+
+    h3 {
+      margin-top: .75em;
+    }
```
bramkragten: Can we make the `font-weight` of headers less? Like 400 or something.

zsarnett: Definitely down to do that. I do wish we had a font that had more settings for boldness. Aka a 500 and 600 difference. 

I will play with it

bramkragten:
I think we should split this PR, one for the frontpage and one for the fonts etc.

zsarnett:
Most of this is just the frontpage. The font changes are very small changes.  Color, font size and font weight. But I can definitely split them

zsarnett:
This has now been split

zsarnett:
One day we will do this. maybe :) probably not

zsarnett:
Paulus stop blocking me 😄 This is better 😂

PR #14055, opened 2020-07-20
Home Assistant Home Page Changes

zsarnett:
## Proposed change

05296a101d364cf420111766f480b2b8

## Type of change

- [x] UI changes

## Additional information

- Relies on #14047 being merged: 

## Checklist


georgecm:
@zsarnett I had a look at your changes. My 2 cents:

**The good**
- Feels less "out-of-the-box Material"
- Layout is much cleaner
- Hierarchy is better
- Buttons are much more clear

**The not so good**
- I don't think the hero image serves the purpose (what is the purpose though? check below). I get that many people would want to see the demo and I think it is a good idea to have a very visual link to the demo but first, you need to understand what Home Assistant is, why it exists, what it can do etc. This way you lose the most valuable space on the page.
- Hierarchy could be bolder. Headings, spacing, colors can do the trick with no much work.
- Order of things could be different.

**More than this**
Being super new to this, but quite experienced in building/designing products, I think we need to answer a few important questions before pushing a new design in. 

- What are the product principles? Examples: Trust, Open, Friendly etc. 
- What is the goal of this page? Examples: To make people download, to trust us, to get hacky, to get the news...
- Who is the user? Example: noobie, experienced, DYI guy etc

PS: If this is too much for this channel let me know and I can move it elsewhere!

zsarnett:
bot be gone

PR #14086, opened 2020-07-26
allowlist_external_dirs does not exist.

pi3ch:
Replaced with whitelist_external_dirs


frenck:
This change is incorrect. The name of the variable has changed in Home Assistant 0.113.

PR #14148, opened 2020-08-04
Update ecovacs.markdown

ChalyFlavour:
I faced some login issues with my ecovacs credentials using this module. First I thought, i has some regular issues with the country/contintent, but in the end it was a simple special character ("-") in my password. After changing my password everything worked correctly.

This issue may be related to `def encrypt(text):`, (sucks/__init__.py, line 206).


frenck (approved):

frenck:
Thanks, @ChalyFlavour 👍

Issue #13933, opened 2020-07-06
Add Iot Class to the Simplisafe integration documentation


mike1237:
## The problem

Add Iot Class to the Simplisafe integration documentation.
It is missing, but included for Abode integration: e.g. Cloud Push.
https://github.com/home-assistant/home-assistant.io/blob/current/source/_integrations/simplisafe.markdown

frenck:
Feel free to open up a PR to add it! Thanks already 👍

mike1237:
Unfortunately, I'm not sure what class it would be considered! Not sure if this is Cloud Push or Cloud Polling.

The code for the integration appears to use a websocket, so I suspect Cloud Push, but I'm not positive.

`        for event_type in (EVENT_LOCK_LOCKED, EVENT_LOCK_UNLOCKED):
            self.websocket_events_to_listen_for.append(event_type)`

https://github.com/home-assistant/core/tree/dev/homeassistant/components/simplisafe

If anyone can confirm, I am happy to write the PR for the documentation update.

frenck:
<https://github.com/home-assistant/core/blob/dev/homeassistant/components/simplisafe/config_flow.py#L18>

PR #14370, opened 2020-09-01
Add iot class per #13933

ghost:
## Proposed change
Add iot class per #13933

### DIFF
--- a/source/_integrations/simplisafe.markdown
+++ b/source/_integrations/simplisafe.markdown
@@ -2,7 +2,7 @@
 title: SimpliSafe
 description: Instructions on how to integrate SimpliSafe into Home Assistant.
 ha_release: 0.81
-ha_iot_class: Cloud Poll
+ha_iot_class: Cloud Polling
 ha_category:
   - Alarm
   - Lock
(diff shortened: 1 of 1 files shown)


source/_integrations/simplisafe.markdown
```diff
@@ -2,6 +2,7 @@
 title: SimpliSafe
 description: Instructions on how to integrate SimpliSafe into Home Assistant.
 ha_release: 0.81
+ha_iot_class: Cloud Poll
```
davet2001: Cloud Polling

davet2001 (changes requested):
Other integrations use "Cloud Polling" rather than "Cloud Poll"

source/_integrations/simplisafe.markdown
ghost: Doh!

I think I got the update made.  Please confirm.

frenck (approved):
Thanks, @SilvrrGIT 👍

bachya:
FYI, this isn't totally true – it's technically Cloud Polling for some attributes, but Cloud Push for things like alarm state. Not the biggest deal.

frenck:
@bachya This has been taken from the configuration flow of SimpliSafe, which in that case if the place where it is wrongly specified.

Issue #14304, opened 2020-08-24
The Automation Trigger page should specify that you can use multiple entity_id properties for the same state trigger


melyux:
## The problem

The [Automation Trigger](https://www.home-assistant.io/docs/automation/trigger/) documentation page's section on state triggers is quite brief, and doesn't specify that you can use multiple entities under a list of `entity_id`, as can be seen in this example in [this blog post about version 0.113](https://www.home-assistant.io/blog/2020/07/22/release-113/#automations--scripts-chooser):

```yaml
    trigger:
      - platform: state
        entity_id:
          - binary_sensor.button1
          - binary_sensor.button2
          - binary_sensor.button3
```

frenck:
Feel free to help out, and hit the edit button that is shown on the top of every page.

Thanks already! 👍

ghost:
Although it doesn't specifically call it out as a separate item/note, there are examples of this on the trigger page. 

See time trigger
https://www.home-assistant.io/docs/automation/trigger/#time-trigger\

the last example under numeric state trigger. 
https://www.home-assistant.io/docs/automation/trigger/#numeric-state-trigger

and an alternate method under show for the entities in the second example under state trigger. 
https://www.home-assistant.io/docs/automation/trigger/#numeric-state-trigger

Would a section on triggering multiple items be helpful (see below)? Possibly below the existing information on https://www.home-assistant.io/docs/automation/trigger/#multiple-triggers





### Multiple Entity IDs for the same Trigger

It is possible to specify multiple Entity_IDs for the same Trigger. To do so add multiple Entity_IDs using a nested list or a comma separated list. The trigger will fire and start, [processing](#what-are-triggers) your automation each time the trigger is true for each Entity_ID listed.

```yaml
automation:
  trigger:
    # first option - nested list
    - platform: state
      entity_id: 
        - sensor.one
        - sensor.two
        - sensor.three
    # second option - comma separated list
    - platform: state
      entity_id: sensor.one, sensor.two, sensor.three

frenck:
Comma-separated should not be recommended. While it works, it adds an additional operation and is not a native collection format to YAML. Hence, I think it is better to leave out completely.

PR #14375, opened 2020-09-02
Add explanation of multiple entities for the same trigger

ghost:
Addressing Issue #14304

## Proposed change
Add explanation of multiple entities for the same trigger


frenck (approved):
Thanks, @SilvrrGIT 👍

PR #14447, opened 2020-09-09
Config block styling & text selection fixes

### LINKED ISSUE #14040: Documentation's new theme may be causing errors.
## The problem

The default `payload_on` and `payload_off` for an MQTT Binary Sensor are _uppercase_ `ON` and `OFF`. However, the new documentation theme appears to be forcing certain text into _lowercase_. As a result, it now shows that the **default states** are `on` and `off`.

Screenshot from 2020-07-19 11-37-17

So is the theme causing documentation errors or is there an upcoming breaking change to re-define the default payload values?

NOTE
The documentation may contain other instances of text forced to lowercase. I have mentioned the example of MQTT Binary Sensor because that's the first instance of the error I noticed.

### LINKED ISSUE #14414: Cannot copy name of configuration variables without pain
## The problem

This may be multiple different issues, but both preventing the same action. I reproduced both errors in Firefox 80.0 and Chromium 83.0.4103.116 (if not stated otherwise) on Debian Unstable (using X11 as backend) without page modifying add-ons. I found the issues originally on https://www.home-assistant.io/lovelace/history-graph/, but could reproduce the issue on multiple other pages and due to the consistent design, this may happen on all pages.

It is about the theming of the "configuration variables" / "options for entities" / ... tables. There I tried to first select a name of one variable and then copy the variable / option name. I failed to select the name due to multiple small design flaws (as I think off):
1. Every name has a anchor icon left of it when hovering over the name (which is great per se), but it's clickable area does overlap a large part of the name (e.g. the first three characters of `type`) preventing me to double-click the name to let the browser automatically only select the word I wish to select.
2. Also due to the anchor, it is not possible to begin selecting the word from the left to the right using the mouse.
3. (*Firefox only*) Selecting the word from the right to the left also seems impossible also due to the anchor overlapping a significant part of the name. Chromium seems to continue selecting texts below other elements (z-axis), but Firefox doesn't, maybe thinking that the text is not visible to the user so it does not need to be selectable.
4. When trying to prevent clicking on the anchor works (e.g. on longer names like `hours_to_show`) and using a double-click, both browsers always (in my testcases) select the name **and** its type, see:
  Screenshot_20200906_195215
  On other examples it may also include the `REQUIRED` marker:
  image
  A triple-click does select the whole "line" if not already done, which is expected:
  image
  This does happen because the website uses a padding to separate these words and so does not separate these words in a logical way like when using a common whitespace character (e.g. ` `), a grid, a flexbox, or using `display: inline-block` on the span. Then the browsers can detect that each item (name, type, …) are each a single entity.

It would be great if these could be fixed to improve the user experience while using the documentation.

frenck:
## Proposed change

This PR addresses 2 issues reported with our configuration blocks.

- Default values are text transformed to lower case. In some cases (like MQTT), this is not wished for, as the value should be `ON` not `on`. This text transformation has been removed from the default value. (#14040)

- Variable names in configuration blocks are hard to select/copy. A little old trick has been added to this PR to add spaces around the config variable name in the HTML. When double clicking, Chrome will now select just the config var. (#14414)

image

## Type of change

## Additional information

- This PR fixes or closes issue: fixes #14040 fixes #14414

## Checklist

### DIFF
--- a/plugins/configuration.rb
+++ b/plugins/configuration.rb
@@ -61,7 +61,7 @@ module Jekyll
 
       result << vars.map do |key, attr|
         markup = Array.new
-        markup << "<div class='config-vars-item'><div class='config-vars-label'><a name='#{slug(key)}' class='title-link' href='\##{slug(key)}'></a> <span class='config-vars-label-name'>#{key}</span>"
+        markup << "<div class='config-vars-item'><div class='config-vars-label'><a name='#{slug(key)}' class='title-link' href='\##{slug(key)}'></a> <span class='config-vars-label-name'> #{key} </span>"
 
         if attr.key? 'type'
 
@@ -115,7 +115,7 @@ module Jekyll
             shortDefaultValue = ", default: " + shortDefaultValue
           end
 
-          markup << "<span class='config-vars-required'>#{startSymbol}<span class='#{attr['required'].to_s}'>#{required_value(attr['required'])}</span>#{shortDefaultValue}#{endSymbol}</span>"
+          markup << "<span class='config-vars-required'>#{startSymbol}<span class='#{attr['required'].to_s}'>#{required_value(attr['required'])}</span><span class='default'>#{shortDefaultValue}</span>#{endSymbol}</span>"
         end
 
         markup << "</div><div class='config-vars-description-and-children'>"
--- a/plugins/configuration.rb
+++ b/plugins/configuration.rb
@@ -61,6 +61,7 @@ module Jekyll
 
       result << vars.map do |key, attr|
         markup = Array.new
+        # There are spaces around the "{key}", to improve double-click selection in Chrome.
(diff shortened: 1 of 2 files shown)


balloob (dismissed):

plugins/configuration.rb
```diff
@@ -61,7 +61,7 @@ def render_config_vars(vars:, component:, platform:, converter:, classes: nil, p
 
       result << vars.map do |key, attr|
         markup = Array.new
-        markup << "<div class='config-vars-item'><div class='config-vars-label'><a name='#{slug(key)}' class='title-link' href='\##{slug(key)}'></a> <span class='config-vars-label-name'>#{key}</span>"
+        markup << "<div class='config-vars-item'><div class='config-vars-label'><a name='#{slug(key)}' class='title-link' href='\##{slug(key)}'></a> <span class='config-vars-label-name'> #{key} </span>"
```
balloob, line 65: Should we add a comment?

frenck: Done 👍

PR #14556, opened 2020-09-18
Add one requirement at the spotify integration page

N1co93:
## Proposed change

Add one requirement at the spotify integration page. This requirement isn't documented at the moment and would be really helpful for most of the people. See issue https://github.com/home-assistant/core/issues/40194

## Type of change

## Additional information

- This PR fixes or closes issue: https://github.com/home-assistant/core/issues/40194

## Checklist

-  [x] This PR uses the correct branch, based on one of the following:

### DIFF
--- a/source/_integrations/spotify.markdown
+++ b/source/_integrations/spotify.markdown
@@ -18,7 +18,7 @@ The `spotify` media player integration allows you to control [Spotify](https://w
 
 - Spotify account
 - Spotify application, properly configured (see below).
-- `internal_url` and `external_url` set correctly. See [Setup basic information](https://www.home-assistant.io/docs/configuration/basic/)
+- `internal_url` and `external_url` set correctly. See [Setup basic information](/docs/configuration/basic/)
 
 <div class='note'>
   Controlling the Spotify integration (pause, play, next, etc.) requires a Premium account.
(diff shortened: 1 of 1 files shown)


source/_integrations/spotify.markdown
```diff
@@ -18,6 +18,7 @@ The `spotify` media player integration allows you to control [Spotify](https://w
 
 - Spotify account
 - Spotify application, properly configured (see below).
+- `internal_url` and `external_url` set correctly. See [Setup basic information](/docs/configuration/basic/)
```
frenck, line 21: . This implies you need to set both, or is a requirement. This is not true.

frenck (changes requested):

source/_integrations/spotify.markdown
N1co93: You are right.

Maybe "and/or" would be the best way? In some situation both values have to be set. 

What do you think about it?

frenck: I think in many cases no values should be set. The root cause lies in something else. If it was configured incorrectly, you should have gotten instruction from Home Assistant, which currently it does not.

I really think we should not add this.

N1co93: Okay, then I will close this pull request.

PR #14557, opened 2020-09-18
Add notes for device connection and battery powered devices

thecode:
## Type of change

- [x ] Adjusted missing or incorrect information in the current documentation (`current` branch).

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/shelly.markdown
+++ b/source/_integrations/shelly.markdown
@@ -20,7 +20,13 @@ To add a Shelly device to your installation, go to **Configuration** >> **Integr
 
 A device can also be added manually, click the button with `+` sign on the integrations page and from the list of integrations, select **Shelly** and follow the instructions shown.
 
+<div class="note">
+You need to configure and connect the device to same network using the Shelly cloud App or device Web page.
+Integration is communicating directly with the device; cloud connection is not needed.
+</div>
+
 ## Known issues and limitations
 
 - Only supports firmware 1.8 and later
 - Support relays, lights (with brightness), sensors and rollers
+- Support for battery powered devices is currently very limited
--- a/source/_integrations/shelly.markdown
+++ b/source/_integrations/shelly.markdown
@@ -16,12 +16,9 @@ Integrate [Shelly devices](https://shelly.cloud) into Home Assistant.
 
 ## Configuration
 
-To add a Shelly device to your installation, go to **Configuration** >> **Integrations** in the UI, click the + to add a new integration and search for Shelly. Devices will be automatically discovered when on the same network and listed on the integrations page. Click on "Configure" of the discovered device to add it to Home Assistant.
-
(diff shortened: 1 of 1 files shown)


source/_integrations/shelly.markdown
```diff
@@ -20,7 +20,13 @@ To add a Shelly device to your installation, go to **Configuration** >> **Integr
 
 A device can also be added manually, click the button with `+` sign on the integrations page and from the list of integrations, select **Shelly** and follow the instructions shown.
 
+<div class="note">
+You need to configure and connect the device to same network using the Shelly cloud App or device Web page.
+Integration is communicating directly with the device; cloud connection is not needed.
```
frenck: To the "same network" of what?

frenck (changes requested):

source/_integrations/shelly.markdown
thecode: I followed the previous definition which shows in the beginning of the documentation:  "Devices will be automatically discovered when on the same network",
Shelly devices and Home assistant needs to be on the same network so that CoAP messages are able to pass from Shelly devices to Home Assistant.

Is the following more clear:
"You need to configure and connect the device to same network of Home Assistant using the Shelly cloud App or device Web page."

frenck: Wait, so it is already written above? Why add a note?

thecode: I added the note based on feedback from users:
Users did not clearly understand that they need to first configure the device and connect to WIFI before trying to add them to HA.
The first sentence actually talks about auto discovery, while it is also possible to work using difference networks which are accessible from each other without being on the same network.

The main idea is to tell the user that the devices needs to be connected to network. Using the phrase "same network" may confuse some users, that is why I avoided it and I assume it is the same reason it is not written above.

Can you suggest a more clear way to emphasize this requirement?
If it is not clear I will remove it and leave only the cloud connection note.

frenck: I think we should avoid creating big notices/info/warning boxes in general, unless really important. 

IMHO, we can just add a line to the initial part in normal text or incorporate it into the existing text.

For example, rewriting the first section to:

> To add a Shelly device to your installation, make sure they are connected to your Wi-Fi network first. Next, go to **Configuration** >> **Integrations** in the UI.

> If the new device is on the same network as Home Assistant, it is discovered automatically. Clicking "Configure" on the discovered device, adds it to Home Assistant. If your device isn't discovered automatically, click the + to add a new integration and search for Shelly to add it manually.

thecode: I rewrite the first section similar to what you suggested (I keep some of the original text).

I also keep the note **"Integration is communicating directly with the device; cloud connection is not needed."** 
However this can also be added to the normal text if you think it is too "big"

balloob (approved):

frenck (approved):
Thanks, @thecode 👍

PR #14625, opened 2020-09-26
Restructure documentation to find content better

ajfriesen:
## Proposed change

I want to restructure the documentation a bit to better find topics and guide home assistant users.
- [ ] Separate pages with multiple topics into their own pages and link accordingly.
- [ ] Add missing links to pages
- [ ] Move from hassio to home assistant os in the docs
- [ ] Create separate docs for installation types (home assistant os, supervised and home assistant core) 
- [ ] Create a better overview in the sidebar in the docs
- [ ] Update tutorial 

### Why the change?

1. Already present docs should be easier to find
2. Separating content makes it easier to find and edit

## Type of change

## Additional information

## Checklist


ajfriesen:
In this commit 17a7808977f052147802e4fc51db26e0671cadd4 I moved the configuration from the docs collection to its own collection. Basically the same as the lovelace docs.

Benefits:
- Sidebar is autogenerated with all markdown files

I would like to know if this would be a better structure in general.
I would like to propose this to all docs, but before starting I would like to have your opinion on this @frenck. (I hope it is okay to ping you, since you did most recent changes on this repo)

```
_docs
├── authentication
├── automation
├── backend
├── configuration
├── frontend
├── hassio
├── installation
├── mqtt
├── scene
├── scripts
├── tools
├── z-wave
```

The /docs page could then be just a neat overview page for all the topics we want to cover in the docs.

I also want to add a folder and content for
- home-assistant-os
- home-assistant-core
- home-assistant-supervised
- home-assistant-container

like you have proposed in 
- https://github.com/home-assistant/architecture/blob/master/adr/0013-home-assistant-container.md
- https://github.com/home-assistant/architecture/blob/master/adr/0014-home-assistant-supervised.md
- https://github.com/home-assistant/architecture/blob/master/adr/0015-home-assistant-os.md
- https://github.com/home-assistant/architecture/blob/master/adr/0016-home-assistant-core.md

_config.yml
```diff
@@ -75,6 +75,8 @@ collections:
     output: true
   lovelace:
     output: true
+  configuration:
```
frenck, line 78: Why is this pulled out into a separate collection?
That makes no sense to me at this point.

source/_docs/installation/recommended-hardware.markdown
```diff
@@ -0,0 +1,25 @@
...
+
+This is a list of popular platforms and what to expect from them.
+
+Platform | Notes
+-- | --
+Raspberry Pi Zero/Pi 2 | **Only** use these for testing
+Raspberry Pi 3/3+/4 | This is a good starting point, and depending on the amount of devices you integrate this can be enough - use an [A2 class SD](https://amzn.to/2X0Z2di) card if possible.
+NUC i3 | This is if you need a little more power over a Pi
+NUC i5 | This will allow you to run multiple services without any issues, perfect for a homelab
+NUC i7/i9 | Pure power, you should not have *any* performance issues
```
frenck, line 25: I'm not sure why this is split out.

Making a lot of smaller pages, is really hard to follow, as everything is spread out everywhere. Instead, we should combine more (as we are doing already across the board) and provide navigation on the page itself.

source/_docs/installation/run-beta.markdown
```diff
@@ -0,0 +1,11 @@
...
+title: "Run the beta version on Home Assistant"
+description: "Instructions to run the Home Assistant beta."
+---
+
+If you would like to test next release before anyone else, you can install the beta version released every three weeks:
+
+1. Backup your installation, using the snapshot functionality Home Assistant offers.
+2. Check the [Home Assistant Beta release notes](https://rc.home-assistant.io/latest-release-notes/) for breaking changes. Be sure to check all release notes between the version you are running and the one you are upgrading to. Use the search function in your browser (`CTRL + f`) and search for **Breaking Changes**.
+3. Select _System_ tab from the _Supervisor_ menu, then select _Join Beta Channel_ under _Supervisor_, then select _Reload_.
+4. Select _Dashboard_ tab from the _Supervisor_ menu, and then select _Update_.
```
frenck, line 11: This now all of a sudden is a general installation instruction, which it is not. The documentation written here is specific for Supervisor enabled systems, which covers only 2 of the 4 installation methods we have.

source/_docs/installation/run-specific-version.markdown
```diff
@@ -0,0 +1,17 @@
...
+For this you would need to install the [Terminal & SSH add-on][ssh] or use the console
+that is available on your device by connecting a keyboard and screen.
+
+To install the Terminal & SSH add-on, choose **Supervisor**, which is located in the sidebar and then the add-on store.
+
+Use the web-based terminal or SSH to your Home Assistant system, or connect to the console, and run:
+
+```bash
+ha core update --version=0.XX.X
+```
```
frenck, line 15: Same as above, this only applies to supervisor enabled installation methods, and isn't running a beta running a specific version already?

frenck:
I see what you are trying to do, but we should avoid making changes all over the place and instead focus on little steps at the time. Making moves like you are starting here, take a lot of time and should not be stuffed into a single PR.

_config.yml
ajfriesen: When does it make sense to have a separate collection and when do we put it into the docs folder?
locelace and addons are a good example for this.
Also the hassio folder even though it is not a collection.

Is there a reason for that?

ajfriesen:
> I see what you are trying to do, but we should avoid making changes all over the place and instead focus on little steps at the time. Making moves like you are starting here, take a lot of time and should not be stuffed into a single PR.

Agree, this PR would be too big. I just used it to show the idea and start a discussion.

What would be a good approach in your opinion?

frenck:
I do not think that, see my other comments above

_config.yml
frenck: You've put it into a collection? Not me. I wonder why? Could you elaborate on that? Why is that needed?

PR #14855, opened 2020-10-03
Update nest documentation with Device Access program / SDM API information

allenporter:
## Proposed change

Add documentation for a new in progress integration to support the new Google Nest Device Access program.

This describes how to register for the program, pointers for how to configure a developer account, includes details configuration information, and instructions on how to authorize access to a particular device.

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/41689

## Checklist


frenck (changes requested):
We have a `nest` integration already, why are we introducing a new one here?

mretegan:
The previous `nest` integration documents the old API that was in place before Google bought Nest. This is about the new API.

allenporter:
Thank you for the review.  Great question, and that rationale is a big omission from the pull request on my part.  I had a similar question myself, and am happy to move to whatever approach makes sense.  This week is my first time contributing to Home Assistant, so I appreciate the guidance.

This question could also mean: *Please update the existing `nest` integration*, but i'll answer the question with the rationale and alternatives considered before assuming that.   I realize I probably should have consulted with @awarecan (code owner of `nest`) on approach here as well.

**Background**:
Compared to the existing nest API that is not accepting new users, the new API has some differences:
- Has a different data model focused on *traits* rather than *devices*
- Provides a fewer fields/device properties of the old API
- Supports fewer devices overall (e.g. no CO-2 sensor, no Protect).
- Uses different authentication endpoints, etc.
- Device Access Program also requires a US$5 fee,
Nothing too surprising here though, as different APIs obviously will have differences.  


**Approach: Separate integrations**:
This request proposes a new integration that follows the latest recommendations from the developer docs including a simple oauth setup flow with little customization, data polling libraries, etc.  This approach optimizes for simplicity within the integration itself, pushing the divergence between authentication, APIs, supported devices, up a level to the community.     This may have down sides that the *community* has to pay additional cost of knowing that two separate integrations exist (e.g. documentation overhead, etc), but leaves each standalone integration simpler.
**Assumption**: If the old Nest API is fully turned down, the old integration can be fully deleted.

**Alternative Approach: Update existing integration**:
An alternative can be add pivot points within the existing `nest` integration to support the new APIs.  This could appear to the end user as a separate configuration option, or set of configuration options.  Here are some considerations:

- **Configuration**: This path would require additional code/validation to (1) indicate that some configuration options are not supported in the new API and (2) disable functionality while it is developed, since we prefer to evolve integrations in smaller pull requests rather than a single change to support all functionality.  Over time, it may be difficult to tell which functionality is for the old API, and new API, but that complexity would remain within the integration itself.
- **Testing Plan**:  This is a good example of the power of unit tests, and presumably one big reason Home Assistant has a good culture around testing (observation as a newcomer).  This is a case where owning a device is not enough to do development on an integration, and instead you're really *integrating with an API*.

  - Automated test coverage: The existing API has test coverage for the configuration flow, though none of the existing platforms.  Before proceeding down the path of adding a split point, it would probably make sense to add more test coverage for the existing code to make sure that it does not break.  Increasing test coverage is a good thing to do anyway (more valuable to the community), though little value for an integration that does not accept new users.
  - Manual Testing: I own Nest devices, but am not able to use the API for the existing integration since it is restricted. Implication: To develop in `nest` would mean that you need to *both* be an authorized user of the old API (registered before a certain date) and have paid the device registration fee for the new API.
  - Manual Testing Alternative: more coordination in the community: e.g. I go find a friend that uses the old API and get them to test it out before every commit.


**Alternative Approach: Replace existing integration**:  The existing nest API is not accepting new integrations, but still works for anyone who was already authorized to use it.  Assumption: This is a non-starter as many members of the community would be upset by this given

**Alternative Approach: Custom Component**: Another alternative could be to develop this as an entirely separate custom component, so that the complexity of diverging Nest APIs does not get introduced into the Home Assistant core.  My impression is that Nest devices are popular enough that it is worth the effort to do the upfront design work to make this a core integration.  I suspect that doing more development in a silo will also make review more difficult later.  I am not pursuing this, but it is still a viable option.


**Python Library API**:
There was a similar decision point for the Python Library that talks to the API.  Should the existing [python-nest](https://github.com/jkoelker/python-nest/) library be extended to work with the Device Access program? 

When examining the existing `python-nest` library, it seemed the data models diverged a decent amount, and even slight differences between authentication models seemed to add up in making the code a little more complex to support both APIs.   Reviewing the Home Assistant docs on [Python Library: Modelling Data](https://developers.home-assistant.io/docs/api_lib_data_models) it seemed simplest to follow the guidance there to "expose the data from the API in the same structure as that the API offers it" and "API libraries should try to do as little as possible.".   That is good guidance, as following the documented best practice produced a very lean and simple library that is easy to follow.

A compatibility layer here could theoretically abstract away the API differences, but I think the APIs differ enough that leaning into the recommended approach makes more sense here. 
**Note:** diverging at the Python Library API is a separate decision point and  a single Home Assistant integration can use multiple APIs, or a single python package can support multiple APIs. 
**Assumption**: There will likely not be new investment in python-nest or new investment in improving the existing/old Nest API functionality with Home Assistant.
**Assumption**: Introduction of a new Nest API likely means there will soon be a deprecation plan for the old Nest API

**Some user discussion**:
There was some [discussion in the community](https://community.home-assistant.io/t/google-nest-offical-device-access-console-finally-released/229146/65) with some folks weighing in on this topic with reasons for both approaches from a user perspective.

frenck:
Lets put this discussion on the Core PR, as it is more relevant there.

allenporter:
This has been updated based on discussion in the original pull request.  The pull request https://github.com/home-assistant/core/pull/41689 now is an update to the existing nest integration.  The documentation is updated to reflect the new approach.

allenporter:
> We have a `nest` integration already, why are we introducing a new one here?

I believe I have addressed this requested change.   I am unsure how to mark this request as "addressed" in the UI.

frenck (approved):
Thanks, @allenporter 👍

Issue #14090, opened 2020-07-26
How do I get the uuid of a chromecast device?


marcus13371337:
## The problem
Previously (< 0.113.1), when adding a specific list of cast devices to the cast integration you were supposed to provide the ip-address to the device (by setting the host attribute in the configuration). Now when I booted up 0.113.1, I got the message that the host attribute is deprecated and supposed to be replaced by the uuid. As I have no idea how I can find the uuid of a cast device I went to the documentation for the cast integration, but I couldn't get any information on how I could find such uuid. Would it be possible to add some kind of explanation on how you could find this uuid?

Also, I noticed that the example in the [manual configuration](https://www.home-assistant.io/integrations/cast#manual-configuration) also refers to the deprecated host attribute

brandoncorry:
Also curious how to go about the same. I was initially confused by the example showing the deprecated host attribute as well. Hope someone can update the documentation and include a couple ways/best way to get UUIDs.

Smoetzak:
Any update on this? I am wondering as well.

OzGav:
Me three

IQndk:
Same. My integrations does not discover any of my 3 Chromecasts.
And none of the old configurations are working.

wwilsman:
Hey guys! I took a dive and figured this out! Home Assistant uses a python library, [PyChromecast](https://pypi.org/project/PyChromecast/) to interact with Chromecast devices. You can see in their example that you can query for Chromecasts by friendly name, then print the resulting UUIDs.

I’m no python expert, so I wonder if there is a one-liner we might be able to use to accomplish this. I ended up using the python REPL to print the UUIDs and copied them over to my secrets file. Here are the steps I took:

1. SSH into home assistant. (You can do this on your local machine if you prefer. I used an addon that comes with a terminal and did this all from the web interface)
2. Install PyChromecast: `$ pip install pychromecast`
3. Enter the python REPL: `$ python3`
4. Type the following (replacing the list of friendly names with your actual Chromecast names):

    ```python
    import pychromecast
    print(pychromecast.get_listed_chromecasts(friendly_names=["Living Room TV", "Bedroom TV", "Office Chromecast"]))
    ```

Since PyChromecast comes with HA, I wonder if we can skip step 2 by being in the correct docker container?

In any case, I hope this helps y'all along with anyone else finding this issue!

---

[...]

PR #14939, opened 2020-10-05
Add instructions on how to find UUID

OzGav:
## Type of change

## Additional information

- This PR fixes or closes issue: https://github.com/home-assistant/home-assistant.io/issues/14090

## Checklist

- [X ] This PR uses the correct branch, based on one of the following:
- [X ] The documentation follows the Home Assistant documentation [standards][].

### DIFF
--- a/source/_integrations/cast.markdown
+++ b/source/_integrations/cast.markdown
@@ -109,7 +109,7 @@ media_player:
   type: list
   keys:
     uuid:
-      description: UUID of a Cast device to add to Home Assistant. Use only if you don't want to add all available devices. The device won't be added until discovered through mDNS. In order to find the UUID for your device install pyChromecast on a computer or enter the Home Assistant Docker container and type the following changing the friendly names as required for your situation. python3 -c "import pychromecast; print(pychromecast.get_listed_chromecasts(friendly_names=["Living Room TV", "Bedroom TV", "Office Chromecast"]))"
+      description: UUID of a Cast device to add to Home Assistant. Use only if you don't want to add all available devices. The device won't be added until discovered through mDNS. In order to find the UUID for your device use a mDNS browser or advanced users can use the following python command - python3 -c "import pychromecast; print(pychromecast.get_listed_chromecasts(friendly_names=["Living Room TV", "Bedroom TV", "Office Chromecast"]))"
       required: false
       type: string
     ignore_cec:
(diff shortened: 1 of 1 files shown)


source/_integrations/cast.markdown
```diff
@@ -109,7 +109,7 @@ media_player:
   type: list
   keys:
     uuid:
-      description: UUID of a Cast device to add to Home Assistant. Use only if you don't want to add all available devices. The device won't be added until discovered through mDNS.
+      description: UUID of a Cast device to add to Home Assistant. Use only if you don't want to add all available devices. The device won't be added until discovered through mDNS. In order to find the UUID for your device install pyChromecast on a computer or enter the Home Assistant Docker container and type the following changing the friendly names as required for your situation. python3 -c "import pychromecast; print(pychromecast.get_listed_chromecasts(friendly_names=["Living Room TV", "Bedroom TV", "Office Chromecast"]))"
```
frenck: So how do users of the OS installation method do this? Or Supervised? And users running the container in Docker?

Why not simply use an mDNS browser?

image

frenck (changes requested):

source/_integrations/cast.markdown
OzGav: What I did, and what I tried to say in brief in my suggested change, is I entered the HA docker container and ran the pychromecast command there. In all the different installation options the user just has to get to where pychromecast is installed, I appreciate that is not always easy.

So that is one option. Your option is also good so perhaps we mention both of these options? Gives people choice as to what resources they have available to get the UUID?

OzGav: The only thing I notice is with your browser is that the UUID is not broken up into the segments (separates by hyphens) that mine were. I don’t know if that makes a difference when you put it into the config? If it does then might be worth mentioning it has to be formatted as such. (Five, number groups)

OzGav: Just following up. Did you want me to add the mDNS browser option?

frenck: I think we should find options that make it accessible for users of any level. Executing Python like that isn't that friendly.

So any other option should be preferred IMHO.

OzGav: OK I have changed my suggestion to first direct users to the mDNS browser option but kept a quick reference to using pychromecast. Thoughts?

frenck (approved):
Thanks, @OzGav 👍

PR #15163, opened 2020-10-10
Clarify differences between alternative install methods

lindhe:
## Proposed change

I recently tried out Home Assistant, and soon after installing using the Home Assistant Container method I understood that it does not support add-ons and a few other things that comes with the supervised version. Almost every guide I find online (both in the official documentation and elsewhere) seems to be referring to the supervised version, assuming add-on support etc. to be available. Looking in threads on your forum, discord, on Reddit, YouTube etc., I find that I'm far from alone in making this mistake. This PR aims to clarify for new users that the installation methods are not only different _ways of installing_, but actually results in _different feature sets_.

I have split this PR into three commits. The first commit (`5652422`) I think is fairly straight forward. It rephrases things in a way that I think it is a clear improvement and should be merged.

The other two commits (`2b94d77` and `28a704a`) I obviously also think are good improvements, but I understand if you disagree with them. They are a bit opinionated, in the sense that they are not only clarifying what the text already says about installation methods, but also changes the meaning of the text slightly. They add the perspective that the Core installation (which is what you get with the "Home Assistant Container" method) may not be for everyone, since it does not have support for add-ons (as far as I can tell).

I have tried to explain the rationale for each commit in their respective commit messages.

If you disagree with changes I've made, let's discuss it and see if we can reach a consensus on how to clarify for new users what the difference is between the install methods.

## Type of change

## Checklist

### DIFF
--- a/source/_docs/installation.markdown
+++ b/source/_docs/installation.markdown
@@ -53,7 +53,7 @@ The only installation method that allows you to use Home Assistant Add-ons is us
 **Method**|**You have**|**Recommended for**
 :-----|:-----|:-----
 [Home Assistant OS](/hassio/installation/)|[A supported platform](/hassio/installation/)|Anybody
-[Home Assistant Container](/docs/installation/docker/)|Docker|Anybody already running Docker
+[Home Assistant Container](/docs/installation/docker/)|Docker|Anybody already running Docker and only want to run Home Assistant Core (i.e., no supervisor or add-ons available).
 
 ## Alternative installs
 
--- a/source/_docs/installation.markdown
+++ b/source/_docs/installation.markdown
@@ -42,7 +42,7 @@ NUC i7/i9 | Pure power, you should not have *any* performance issues
 
 ## Recommended
 
-These install options are fully supported by Home Assistant's documentation. For example, if an integration requires that you install something to make it work on one of these methods then the integration page will document the steps required.
+Using the Home Assistant OS method is the fully supported installation option by Home Assistant's documentation. For example, if an integration requires that you install something to make it work then the integration page will document the steps required. The Home Assistant Container method, on the other hand, may have limited documentation available, for example if you want to install and use add-ons.
 
 <div class='note'>
 
--- a/source/_docs/installation.markdown
+++ b/source/_docs/installation.markdown
@@ -46,7 +46,7 @@ These install options are fully supported by Home Assistant's documentation. For
 
 <div class='note'>
 
-The only installation methods that allow you to use Home Assistant Add-ons is using the Home Assistant image. All other methods only install the base Home Assistant packages, however the software from the add-ons may still usually be installed manually like any other program.
+The only installation method that allows you to use Home Assistant Add-ons is using the Home Assistant OS method. All other methods (including Home Assistant Container) only install the base Home Assistant packages, however the software from the add-ons may still usually be installed manually like any other program.
 
 </div>
 


source/_docs/installation.markdown
```diff
@@ -42,18 +42,18 @@ NUC i7/i9 | Pure power, you should not have *any* performance issues
 
 ## Recommended
 
-These install options are fully supported by Home Assistant's documentation. For example, if an integration requires that you install something to make it work on one of these methods then the integration page will document the steps required.
+Using the Home Assistant OS method is the fully supported installation option by Home Assistant's documentation. For example, if an integration requires that you install something to make it work then the integration page will document the steps required. The Home Assistant Container method, on the other hand, may have limited documentation available, for example if you want to install and use add-ons.
```
frenck, line 45: This is an incorrect change. It is not limited to Home Assistant OS.
Also, the mention of the Container method having limited documentation is incorrect IMHO. Especially if you'd consider that the OS installation method runs the exact same containers.

source/_docs/installation.markdown
```diff
@@ -42,18 +42,18 @@ NUC i7/i9 | Pure power, you should not have *any* performance issues
...
 
 <div class='note'>
 
-The only installation methods that allow you to use Home Assistant Add-ons is using the Home Assistant image. All other methods only install the base Home Assistant packages, however the software from the add-ons may still usually be installed manually like any other program.
+The only installation method that allows you to use Home Assistant Add-ons is using the Home Assistant OS method. All other methods (including Home Assistant Container) only install the base Home Assistant packages, however the software from the add-ons may still usually be installed manually like any other program.
```
frenck, line 49: Incorrect, Supervised is another method that provides add-ons. Add-ons are not limited to the OS method.

frenck (changes requested):
Thanks for the PR @lindhe!

Left some comments, could you take a look? Thanks! 👍

source/_docs/installation.markdown
lindhe: Yeah, this was the change I was most unsure about. In that case, let's skip this change. But I really do feel that it's very hard indeed to find documentation that describes for example how to install add-ons manually unless the user have installed using the HA OS method. Perhaps I have simply failed to find that documentation.

Do you understand what I'm after, and do you know of any such documentation that we can perhaps refer to? Maybe we can keep the text as is, but add a brief note that informs about the different approaches when it comes to installing add-ons?

source/_docs/installation.markdown
lindhe: Ah, right. I think I'm getting confused between OS and Supervised. I meant to only distinguish between the OS and Container methods, since they are the ones in the "Recommended" section.

Would it be correct to say something along the lines of this?
> "While the Home Assistant OS method allows you to install and use Home Assistant Add-ons directly from the Web GUI, the Home Assistant Container method does not."

source/_docs/installation.markdown
frenck: >  documentation that describes for example how to install add-ons manually

That is the whole point, there are no add-ons. If you want to install additional software, it is not up to us to document that. If you want to install X, please go to the website of X on how to install it.

source/_docs/installation.markdown
frenck: You can read more about this in ADR-0012, we have 4 supported installation methods.

source/_docs/installation.markdown
lindhe: But it seems like Home Assistant (sometimes?) integrates with these additional software. Of course I can install e.g. NodeRed next to Home Assistant, but I am under the impression that Home Assistant would be able to integrate the NodeRed UI into Home Assistant if I had installed it as an add-on.

I agree that it's not up to you to describe how to install other software, but there has to be something going on that allows one version of Home Assistant to support add-ons while the other version does not. And I would expect that to be documented, but I cannot find that. At least apart from "this installation method magically has support for add-ons, but this similar method does not".

source/_docs/installation.markdown
lindhe: Sorry, I'm at a loss here. What's ADR-0012? I find nothing relevant when searching here on GitHub, on https://www.home-assistant.io/ or on Google. 

**EDIT:** I think i found it! Searching for literally `ADR-0012` gave no results, but this ought to be it. https://github.com/home-assistant/architecture/blob/master/adr/0012-define-supported-installation-method.md

lindhe: If I changed this part to the following, would that be a reasonable improvement to the text?

> The only installation methods that allows you to use Home Assistant Add-ons are using the Home Assistant OS or Home Assistant Supervised methods. All other methods (including Home Assistant Container) only install the base Home Assistant packages, however the software from the add-ons may still usually be installed manually like any other program.

lindhe:
I'm looking at this part of the paragraph now:
> however the software from the add-ons may still usually be installed manually like any other program.

I interpreted that as if I would be able to install add-ons, just not "automatically bundled using the HA Web GUI" so to say. That the add-ons would still be possible to install, just manually. But perhaps you mean that many add-ons are programs of their own right, and may be installed next to (but completely separate from) Home Assistant?

PR #15192, opened 2020-10-11
Add compensation integration

Petro31:
Adds the compensation integration

## Proposed change

Adds the compensation integration documentation

## Type of change

- [x] Added documentation for a new integration I'm adding to Home Assistant (`next` branch).

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/41675

## Checklist

### DIFF
--- a/source/_integrations/compensation.markdown
+++ b/source/_integrations/compensation.markdown
@@ -5,7 +5,7 @@ ha_category:
   - Utility
   - Sensor
 ha_iot_class: Calculated
-ha_release: "2021.4.0"
+ha_release: "2021.5"
 ha_quality_scale: internal
 ha_codeowners:
   - '@petro31'
--- a/source/_integrations/compensation.markdown
+++ b/source/_integrations/compensation.markdown
@@ -22,14 +22,14 @@ To enable the compensation sensor, add the following lines to your `configuratio
 # Example configuration.yaml entry
 compensation:
   media_player_db_volume:
-    entity_id: media_player.yamaha_receiver
+    source: media_player.yamaha_receiver
     data_points:
       - [0.2, -80.0]
       - [1.0, 0.0]
 ```
 
 {% configuration %}
-entity_id:
+source:
   description: The entity to monitor.
   required: true
   type: string
@@ -37,10 +37,9 @@ data_points:
   description: "The collection of data point conversions with the format `[uncompensated_value, compensated_value]`.  e.g., `[1.0, 2.1]`. The number of required data points is equal to the polynomial `degree` + 1. For example, a linear compensation (with `degree: 1`) requires at least 2 data points."
   required: true
   type: list
-name:
-  description: Name of the sensor to use in the frontend.
+unique_id:
+  description: An ID that uniquely identifies this sensor. Set this to a unique value to allow customization through the UI.
   required: false
-  default: Compensation
   type: string
 attribute:
   description: Attribute to monitor.
--- a/source/_integrations/compensation.markdown
+++ b/source/_integrations/compensation.markdown
@@ -5,7 +5,7 @@ ha_category:
   - Utility
   - Sensor
 ha_iot_class: Calculated
-ha_release: '0.117'
+ha_release: "2021.4.0"
 ha_quality_scale: internal
 ha_codeowners:
   - '@petro31'
--- a/source/_integrations/compensation.markdown
+++ b/source/_integrations/compensation.markdown
@@ -12,7 +12,7 @@ ha_codeowners:
 ha_domain: compensation
 ---
 
-The Compensation integration consumes the state from other sensors. It exports the compensated value as state and the following values as attributes: `entity_id` and `coefficients`.
+The Compensation integration consumes the state from other sensors. It exports the compensated value as state and the following values as attributes: `entity_id` and `coefficients`.  A single polynomial, linear by default, is fit to all data points provided.
 
 ## Configuration
 
--- a/source/_integrations/compensation.markdown
+++ b/source/_integrations/compensation.markdown
@@ -37,7 +37,7 @@ entity_id:
   required: true
   type: string
 data_points:
-  description: The collection of data point conversions with the format `uncompensated_value -> compensated_value`.  e.g. `1.0 -> 2.1`. The number of required data points is equal to the polynomial `degree` + 1. For example, a linear compensation (with `degree: 1`) requires at least 2 data points.
+  description: "The collection of data point conversions with the format `uncompensated_value -> compensated_value`.  e.g. `1.0 -> 2.1`. The number of required data points is equal to the polynomial `degree` + 1. For example, a linear compensation (with `degree: 1`) requires at least 2 data points."
   required: true
   type: list
 name:
--- a/source/_integrations/compensation.markdown
+++ b/source/_integrations/compensation.markdown
@@ -12,7 +12,7 @@ ha_codeowners:
 ha_domain: compensation
 ---
 
-The `compensation` sensor platform consumes the state from other sensors. It exports the `compensated` value as state and the following values as attributes: `entity_id` and `coefficients`.
+The Compensation integration consumes the state from other sensors. It exports the compensated value as state and the following values as attributes: `entity_id` and `coefficients`.
 
 ## Configuration
 
--- a/source/_integrations/compensation.markdown
+++ b/source/_integrations/compensation.markdown
@@ -1,0 +1,66 @@
+---
+title: Compensation
+description: Instructions on how to integrate compensation sensors into Home Assistant.
+ha_category:
+  - Utility
+  - Sensor
+ha_iot_class: Local Push
+ha_release: '0.117'
+ha_quality_scale: internal
+ha_codeowners:
+  - '@petro31'
+ha_domain: compensation
+---
+
+The `compensation` sensor platform consumes the state from other sensors. It exports the `compensated` value as state and the following values as attributes: `entity_id` and `coefficients`.
+
+## Configuration
+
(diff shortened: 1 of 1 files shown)


source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,66 @@
...
+      - 1.0 -> 0.0
+```
+
+{% configuration %}
+entity_id:
+  description: The entity to monitor.
+  required: true
+  type: string
+data_points:
+  description: The collection of data point conversions with the format `uncompensated_value -> compensated_value`.  e.g. `1.0 -> 2.1`. The number of required data points is equal to the polynomial `degree` + 1. For example, a linear compensation (with `degree: 1`) requires at least 2 data points.
```
klaasnicolaas:
```suggestion
  description: "The collection of data point conversions with the format `uncompensated_value -> compensated_value`.  e.g. `1.0 -> 2.1`. The number of required data points is equal to the polynomial `degree` + 1. For example, a linear compensation (with `degree: 1`) requires at least 2 data points."
```

klaasnicolaas (dismissed):

source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,66 @@
+---
+title: Compensation
+description: Instructions on how to integrate compensation sensors into Home Assistant.
+ha_category:
+  - Utility
+  - Sensor
+ha_iot_class: Local Push
```
frenck: This integration calculates its states based on existing data.

```suggestion
ha_iot_class: Calculated
```

source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,66 @@
...
+ha_release: '0.117'
+ha_quality_scale: internal
+ha_codeowners:
+  - '@petro31'
+ha_domain: compensation
+---
+
+The `compensation` sensor platform consumes the state from other sensors. It exports the `compensated` value as state and the following values as attributes: `entity_id` and `coefficients`.
```
frenck:
```suggestion
The Compensation integration consumes the state from other sensors. It exports the compensated value as state and the following values as attributes: `entity_id` and `coefficients`.
```

source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,66 @@
...
+To enable the compensation sensor, add the following lines to your `configuration.yaml`:
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: compensation
+    name: Yamaha Zone 1 Volume
+    entity_id: media_player.yamaha_receiver
+    attribute: volume_level
+    unit_of_measurement: dB
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,66 @@
...
+The `compensation` sensor platform consumes the state from other sensors. It exports the `compensated` value as state and the following values as attributes: `entity_id` and `coefficients`.
+
+## Configuration
+
+To enable the compensation sensor, add the following lines to your `configuration.yaml`:
+
+```yaml
+# Example configuration.yaml entry
+sensor:
+  - platform: compensation
```
frenck: We no longer accept YAML configuration for integrations on a platform level.

For more information about this decision see:

- ADR-0007: <https://github.com/home-assistant/architecture/blob/master/adr/0007-integration-config-yaml-structure.md#decision>

source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,66 @@
...
+  description: Name of the sensor to use in the frontend.
+  required: false
+  default: Compensation
+  type: string
+attribute:
+  description: Attribute to monitor.
+  required: false
+  type: string
+degree:
+  description: The degree of a polynomial. e.g. Linear compensation (y = x + 3) has 1 degree, Quadratic compensation (y = x<sup>2</sup> + x + 3) has 2 degrees, etc.
```
frenck:
```suggestion
  description: "The degree of a polynomial. e.g., Linear compensation (`y = x + 3`) has 1 degree, Quadratic compensation (`y = x<sup>2</sup> + x + 3`) has 2 degrees, etc."
```

source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,66 @@
...
+      - 1.0 -> 0.0
+```
+
+{% configuration %}
+entity_id:
+  description: The entity to monitor.
+  required: true
+  type: string
+data_points:
+  description: "The collection of data point conversions with the format `uncompensated_value -> compensated_value`.  e.g. `1.0 -> 2.1`. The number of required data points is equal to the polynomial `degree` + 1. For example, a linear compensation (with `degree: 1`) requires at least 2 data points."
```
frenck:
```suggestion
  description: "The collection of data point conversions with the format `uncompensated_value -> compensated_value`.  e.g., `1.0 -> 2.1`. The number of required data points is equal to the polynomial `degree` + 1. For example, a linear compensation (with `degree: 1`) requires at least 2 data points."
```

frenck (changes requested):
Thanks for the PR @Petro31.

I've reviewed it, and left some comments. Could you please take a look?

Thanks 👍

source/_integrations/compensation.markdown
Petro31: Yep, I assumed that was 'local push'.  Makes more sense as calculated.

source/_integrations/compensation.markdown
Petro31: Ok, I'll probably need to update the other PR to handle that.

source/_integrations/compensation.markdown
Petro31: will the `` kill the x<sup>2</sup>?  That's why I avoided it in the first place.

Petro31: `test<sup>test</sup>` vs test<sup>test</sup>

Petro31: Looks like it does

source/_integrations/compensation.markdown
Petro31: fixed

source/_integrations/compensation.markdown
Petro31: fixed

source/_integrations/compensation.markdown
Petro31: I left the equations as non-code to ensure the superscript stays.

MatthewFlamm:
I think it is worth clearly stating that this integration does not interpolate.  It fits a single line (or higher order polynomial) to all the data points.  I think this will cause confusion as it does with the similar esphome usage.  For example:

```yaml
compensation:
  media_player_db_volume:
    entity_id: media_player.yamaha_receiver
    data_points:
      - 0.2 -> -80.0
      - 0.5 -> -20.0
      - 1.0 -> -10.0
```

A value of 0.75 will not transform to -15.0.  In fact a value of 0.5 won't transform to -20 in this case either!

Petro31:
> I think it is worth clearly stating that this integration does not interpolate. It fits a single line (or higher order polynomial) to all the data points. I think this will cause confusion as it does with the similar esphome usage. For example:
> 
> ```yaml
> # Example configuration.yaml entry
> compensation:
>   media_player_db_volume:
>     entity_id: media_player.yamaha_receiver
>     data_points:
>       - 0.2 -> -80.0
>       - 0.5 -> -20.0
>       - 1.0 -> -10.0
> ```
> 
> A value of 0.75 will not transform to -15.0. In fact a value of 0.5 won't transform to -20 in this case either!

@MatthewFlamm It's funny you should mention that.  I was planning on adding non-linear interpolation at some point down the road.  Does the ESPHome version do interpolation?

MatthewFlamm:
I am only aware of the `calibrate_linear` (and `calibrate_polynomial`) filter in esphome that works like this integration.  I think an interpolate integration would also be useful, but this seems off topic.  My suggestion here is to add something like "A single polynomial, linear by default, is fit to all data points provided."

Petro31:
> I am only aware of the `calibrate_linear` (and `calibrate_polynomial`) filter in esphome that works like this integration. I think an interpolate integration would also be useful, but this seems off topic. My suggestion here is to add something like "A single polynomial, linear by default, is fit to all data points provided."

I'm not the greatest tech writer, mind if I use that line?

MatthewFlamm:
>  I'm not the greatest tech writer, mind if I use that line?

You can use it.

frenck (dismissed):

frenck (dismissed):

source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,63 @@
+---
+title: Compensation
+description: Instructions on how to integrate compensation sensors into Home Assistant.
+ha_category:
+  - Utility
+  - Sensor
+ha_iot_class: Calculated
+ha_release: '0.117'
```
frenck:
```suggestion
ha_release: "2021.4.0"
```

source/_integrations/compensation.markdown
```diff
@@ -0,0 +1,62 @@
+---
+title: Compensation
+description: Instructions on how to integrate compensation sensors into Home Assistant.
+ha_category:
+  - Utility
+  - Sensor
+ha_iot_class: Calculated
+ha_release: "2021.4.0"
```
frenck:
```suggestion
ha_release: "2021.5"
```

frenck (approved):
🎉 @Petro31 👍

PR #15415, opened 2020-10-26
using device_class instead of sensor_class

eddyliao:
using device_class instead of sensor_class


frenck (approved):
Thanks, @eddyliao 👍 

Happy Hacktoberfest 🎉

PR #15439, opened 2020-10-27
Problem with '' in the API

exetico:
## Proposed change
Google don't like the ' in the XGM_RAW statement, so we need to use " instead. 

But I'm not sure, if we should to with \", not that the official documentation ask us to use `" ... 'something' ..."` instead of `'... "something"...'`  - So what do you think?

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/imap.markdown
+++ b/source/_integrations/imap.markdown
@@ -93,7 +93,7 @@ sensor:
     password: YOUR_PASSWORD
     search: FROM <sender@email.com>, SUBJECT <subject here>
     # Or use X-GM-RAW search-term like this, to find unread emails from the last 7 days in your inbox
-    # search: "X-GM-RAW 'in: inbox newer_than:7d is:unread'"
+    # search: "X-GM-RAW \"in: inbox newer_than:7d is:unread\""
 
 # Example configuration.yaml entry for Office 365
 sensor:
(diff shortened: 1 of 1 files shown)


source/_integrations/imap.markdown
```diff
@@ -93,7 +93,7 @@ sensor:
     password: YOUR_PASSWORD
     search: FROM <sender@email.com>, SUBJECT <subject here>
     # Or use X-GM-RAW search-term like this, to find unread emails from the last 7 days in your inbox
-    # search: "X-GM-RAW 'in: inbox newer_than:7d is:unread'"
+    # search: "X-GM-RAW \"in: inbox newer_than:7d is:unread\""
```
frenck: Maybe change the quotes on our end to remove the need for escaping entirely?
```suggestion
    # search: 'X-GM-RAW "in: inbox newer_than:7d is:unread"'
```

exetico: Sure, if we can ignore the [standards](https://developers.home-assistant.io/docs/documenting/standards/#single-vs-double-quotation-marks), it's totally fine with me. That's the only reason why i ended up using " \" :-)

frenck: Yeah, I think it is fine in this case.

frenck (approved):
Thanks, @exetico 👍 

Happy Hacktoberfest 🎉

PR #15453, opened 2020-10-28
Update Docs - same subnet only

ryoku-cha:
## Proposed change

Added a Note to that WOL usually only works if Home Assistant and the Target Device is on the same subnet.
Because WOL works on Layer 2 and is not routable. Some Routers/Firewalls offer "IP Helpers" to forward the packet o a different subnet.

Added a Second Note to mention a possible solution/workaround for Users of OPNsense, where it is possible to use the REST API to send a WOL Packet from the Firewall.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/wake_on_lan.markdown
+++ b/source/_integrations/wake_on_lan.markdown
@@ -51,10 +51,6 @@ This usually only works if the Target Device is connected to the same network. R
 The Service to Route the packet is most likely named "IP Helper" which may support WakeOnLan, but not all Routers support this.
 </div>
 
-<div class='note'>
-On OPNsense you can work around this by installing the os-wol Package. WOL can then be triggered via [REST API](https://wiki.opnsense.org/development/api/plugins/wol.html) from Home Assistant, by using the [RESTful Command](https://www.home-assistant.io/integrations/rest_command) Integration.
-</div>
-
 ## Switch
 
 The `wake_on_lan` (WOL) switch platform allows you to turn on a [WOL](https://en.wikipedia.org/wiki/Wake-on-LAN) enabled computer.
--- a/source/_integrations/wake_on_lan.markdown
+++ b/source/_integrations/wake_on_lan.markdown
@@ -46,6 +46,15 @@ Sample service data:
 }
 ```
 
+<div class='note'>
+This usually only works if the Target Device is connected to the same network. Routing the WakeOnLan packet to a different subnet requires special configuration on your router or may not be possible.
+The Service to Route the packet is most likely named "IP Helper" which may support WakeOnLan, but not all Routers support this.
+</div>
+
+<div class='note'>
+On OPNsense you can work around this by installing the os-wol Package. WOL can then be triggered via [REST API](https://wiki.opnsense.org/development/api/plugins/wol.html) from Home Assistant, by using the [RESTful Command](https://www.home-assistant.io/integrations/rest_command) Integration.
+</div>
+
 ## Switch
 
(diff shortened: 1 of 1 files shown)


ryoku-cha:
With a directed broadcast (by supplying the broadcast_address) it might be possible to send the broadcast to another subnet.
Requirements for that seem to be a corresponding Static ARP Entry and Firewall Rules. Not sure what exactly. I will probably try to find out more and extend the page, as soon as i found out how this works.
(https://forum.netgate.com/topic/80892/wake-on-lan-to-different-vlan-s/21)

Another way is to use the mentioned OPNsense WOL Plugin. I implemented this and it works.
That needs some refinement, after that i think it would be a good idea to extend this page with an example on how to implement this.

ryoku-cha:
this morning i tried the directed broadcast
- created a static arp entry on my firewall (OPNsense) on a free IP in the TV Network with mac addr ff:ff:ff:ff:ff:ff (broadcast)
- added a firewall rule to allow HA to send any Packet o the defined IP
- changed the turn on action for the TV to WakeOnLan and restarted HA

Turning on the TV with that configuration did not work. The Firewall Log shows a UDP Packet on Port 9 which is sent from HA. But nothing happens (sending a WOL Packet from the Firewall via the other mentioned Solution works).

Not sure what is missing or if that is even possible with my setup.
If i find out what else is necessary i will extend the Documentation with the required steps.

source/_integrations/wake_on_lan.markdown
```diff
@@ -46,6 +46,15 @@ Sample service data:
...
 
+<div class='note'>
+This usually only works if the Target Device is connected to the same network. Routing the WakeOnLan packet to a different subnet requires special configuration on your router or may not be possible.
+The Service to Route the packet is most likely named "IP Helper" which may support WakeOnLan, but not all Routers support this.
+</div>
+
+<div class='note'>
+On OPNsense you can work around this by installing the os-wol Package. WOL can then be triggered via [REST API](https://wiki.opnsense.org/development/api/plugins/wol.html) from Home Assistant, by using the [RESTful Command](https://www.home-assistant.io/integrations/rest_command) Integration.
+</div>
+
```
frenck: This is not the OPNsense documentation.

We want to prevent documenting third-party workaround/solutions like this. The main reasoning is to prevent extensive documentation on all routers/firewalls out there (which would be a lot, hard to maintain and hard to read).

```suggestion
```

source/_integrations/wake_on_lan.markdown
```diff
@@ -46,6 +46,15 @@ Sample service data:
 }
 ```
 
+<div class='note'>
+This usually only works if the Target Device is connected to the same network. Routing the WakeOnLan packet to a different subnet requires special configuration on your router or may not be possible.
+The Service to Route the packet is most likely named "IP Helper" which may support WakeOnLan, but not all Routers support this.
```
frenck:
```suggestion
This usually only works if the Target Device is connected to the same network. Routing the WakeOnLan packet to a different subnet requires a special configuration on your router or may not be possible.
The Service to Route the packet is most likely named "IP Helper" which may support WakeOnLan, but not all Routers support this.
```

frenck (changes requested):

ryoku-cha:
@frenck thanks for suggesting the changes.
I completely agree about the OPNsense Block. The official Home Assistant Documentation is not the right place to document something like that. 👍

frenck (approved):
Thanks, @ryoku-cha 👍

PR #15463, opened 2020-10-29
Added KP105 smart plug as supported Kasa device

markuspalme:
Added KP105 smart plug to list of supported Kasa devices. I have tested the device with my local setup - it’s working perfectly and I think that information is helpful for other users.


frenck (approved):
Thanks, @markuspalme 👍 

Next time you do a PR, could you please use the PR template presented? Thanks 👯

PR #15545, opened 2020-11-06
Added "OS" in the title

ssm2017:
This page title is not clear so i have added the "OS" specification because the links on this page are speaking about the os installation version


frenck:
Incorrect, this page is about Home Assistant itself, not the OS.

Also, please use the PR template next time. Thanks 👍

PR #15546, opened 2020-11-06
Updated the installation icon link

ssm2017:
i have changed the link on this icon because it is sending to the "os" version of home-assistant so it can be confusing people using the menu on the right that is speaking about all the versions :)


frenck:
This is done on purpose. Nevertheless, thanks for your suggestion.

PR #15547, opened 2020-11-06
Update switch.template.markdown

DrakeSeven:
Explains what booleans do in a value_template.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/switch.template.markdown
+++ b/source/_integrations/switch.template.markdown
@@ -55,7 +55,7 @@ switch:
         required: false
         type: string
       value_template:
-        description: Defines a template to set the state of the switch. If not defined, the switch will optimistically assume all commands are successful.
+        description: Defines a template to set the state of the switch. If the template returns `true`, the device is `on`. If the template returns any other value, the device will be `off`. If not defined, the switch will optimistically assume all commands are successful. 
         required: false
         type: template
         default: optimistic


source/_integrations/switch.template.markdown
```diff
@@ -55,7 +55,7 @@ switch:
         required: false
         type: string
       value_template:
-        description: Defines a template to set the state of the switch. If not defined, the switch will optimistically assume all commands are successful.
+        description: Defines a template to set the state of the switch. If the template returns `true`, the device is `on`. If the template returns any other value, the device will be `off`. If not defined, the switch will optimistically assume all commands are successful.
```
frenck, line 58: This is not complete, the template could also return `on` to make it show up as turned on. Additionally, it is case insensitive as well.

DrakeSeven: Interesting.
Based on your comment I have an educated guess:
Defines a template to set the state of the switch **to the value that the template returns**.

frenck: I don't think that is the right wording either, mainly because the state isn't set to the value the template returns.

DrakeSeven: So the question is: What value corresponds to what state?
Since the docs for template switch have the sample code:
`value_template: "{{ is_state('sensor.skylight', 'on') }}"`
I thought it evaluates either to `true` or `false`.
So my hypothesis was that `value_template: true` sets the state of the switch to `on`.

frenck: That is correct.

These as well: 

- `value_template: "{{ is_state('sensor.skylight', 'on') }}"`
- `value_template: "{{ states('sensor.skylight') }}"`
- `value_template: "{{ true }}"`
- `value_template: "{{ 'on' }}"`
- `value_template: "{{ 'true' }}"`
- `value_template: "{{ 'ON' }}"`
- `value_template: "{{ 'TRUE' }}"`
- `value_template: "{{ 'oN' }}"`
- `value_template: "{{ 'trUE' }}"`

DrakeSeven: Interesting.
For the first 2 List items, I guess it depends on what the value is at runtime.
But the rest is what I was looking for. Is this list exclusive?
May I ask you where you found it?

frenck: Well, it is all the values Home Assistant considers true/false in general.

However, we recommend using true/false.

DrakeSeven: Ok, so my change request was incorrect insofar that there are multiple possible values besides "true" that also set the state of the switch to "on".
I came across several docs that have a `value_template` and in case of "Template Cover" it lists valid values.
However it doesn't explain:
1. value_template: true -> state of the template displays "on"
2. value_template: true -> state of the template displays a string that this template maps to true
3. value_template: [value] -> state of the switch displays "[value]"

frenck: ? I'm not sure if I follow. The current documentation states: 

`Defines a template to set the state of the switch.`

So if you want it to be "on" set it to `on`, if you want it to be off, set it to `off`. 👍 
The others are additionally accepted due to the way YAML works (`on` in YAML == `true`) and how it is handled in a similar way in Home Assistant to keep things working as expected.

I'm kinda confused about the problem you are trying to address, to be honest.

DrakeSeven: Now it makes sense to me. Including your answer
"_[...] the state isn't set to the value the template returns_".
Thank's for your effort to clarify it!

frenck: No problem! Thanks for willing to contribute 👍

PR #15696, opened 2020-11-20
Add extra Naba Casa instructions for nest integrations

mikemc35:
Added the URI syntax for Nabu Casa external link for the URI Redirect. This was not obvious until I troubleshooted an error that exposed the correct URI.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/nest.markdown
+++ b/source/_integrations/nest.markdown
@@ -67,6 +67,10 @@ steps are:
 
 Additionally, Home Assistant must be configured with a URL (e.g., external exposed [`http`](/integrations/http/), Nabu Casa, etc). When setting up the OAuth credentials, make sure the Home Assistant URL is in the list of *Authorized redirect URIs*, so the redirect back to Home Assistant can get an OAuth authorization code.
 
+<div class='note'>
+If you use Nabu Casa for Home Assistant external access, use the https://xxxxxxxxxxx.ui.nabu.casa/auth/external/callback syntax for the Authorized redirect URIs after all parts of the Google Device Access are complete.
+</div>
+
 Follow all of the instructions in [Device Access: Quick Start Guide](https://developers.google.com/nest/device-access/get-started) carefully as it is easy to make a configuration mistake that is difficult to debug. It is recommended to exercise the entire guide, including the command to test out the API, to make sure that it is working before configuring Home Assistant.
 
 When you get to the steps about configuring events make sure to follow guide under [Events](https://developers.google.com/nest/device-access/api/events) that configures the [Pub/Sub subscription](https://console.cloud.google.com/cloudpubsub/subscription/list) from the Google Cloud console. Make sure to use the *topic name* from the device access console and a unique subscription ID in the cloud console. Note the message retention is how long messages will queue while offline, so keep that short (e.g., under an hour) to avoid a potentially large backlog of updates.


allenporter (approved):

allenporter:
@mikemc35 cheers!

source/_integrations/nest.markdown
```diff
@@ -67,6 +67,10 @@ steps are:
 
 Additionally, Home Assistant must be configured with a URL (e.g., external exposed [`http`](/integrations/http/), Nabu Casa, etc). When setting up the OAuth credentials, make sure the Home Assistant URL is in the list of *Authorized redirect URIs*, so the redirect back to Home Assistant can get an OAuth authorization code.
 
+<div class='note'>
+If you use Nabu Casa for Home Assistant external access, use the https://xxxxxxxxxxx.ui.nabu.casa/auth/external/callback syntax for the Authorized redirect URIs after all parts of the Google Device Access are complete.
```
frenck, line 71: I'm confused by this note, what makes the nabu casa remote UI URL any different from any other URL?

frenck (changes requested):

source/_integrations/nest.markdown
mikemc35: It's just calling out the specific syntax for the Nabu Casa URI since there's no guidance on any specifics past that fact you need to use a Nabu Casa URI. I'm guessing that there are quite a few other who use Nabu Casa with HA like me, and possibly several more just to get this integration going.

It was only through a 400 error I got on the Google side that I figured out the https://xxxxxxxxxxx.ui.nabu.casa/auth/external/callback syntax was the correct one. And that was after a lot of confusion using https://xxxxxxxxxxx.ui.nabu.casa/ and https://xxxxxxxxxxx.ui.nabu.casa/lovelace/default_view (both of which did not appear to work).

I had a couple folks who pinged me after I asked some questions on the Discord, and it looks like it helped them, so I thought I'd try adding it here to hopefully save some frustration for others who use (or will sign up for) Nabu Casa for this integration.

frenck: But, my point is, the same applies to `http://xxxxxxxx.duckdns.org/auth/external/callback`...

allenporter: @mikemc35 So maybe you want to remove mention of nabu casa and say, you must include the full path of the URL such as "https://example.com/auth/external/callback"?

frenck: I think it makes more sense to mention it should match the URL of what you use in your browser to access Home Assistant. It can be `homeassistant.local` or even an IP.

allenporter: I thought this advice is related to the config in Google's side about the authorized url list, and it needs to be a public url, and the same one that Google needs to ping back to.  My impression is that oath doesn't work for a local url.  Is that right?

frenck: OAuth does not rely on direct communication from the third party. So, it doesn't need to be public accessible. If this integration requires a public URL for OAuth, it isn't OAuth but a custom flavor.

allenporter: It would make sense that it would only need to propagate url parameters via redirect so it doesn't care if it is publicly exposed or not. (e.g. https://developers.google.com/identity/protocols/oauth2/web-server#creatingcred gives a localhost example)

For some reason I must have misremembered something required a public url. To be safe I'd like to go try out the flow again end to end with a local url before declaring that

mikemc35: Another thought...if it's just the `/auth/external/callback` part of the syntax that's crucial to this working for _any_ Oauth URI direct, then that is the important part that's missing here.

So (if true), the note would mention something about tacking on the `/auth/external/callback` part onto the main external URI (for example, `https://xxxxxxxxxxx.ui.nabu.casa/auth/external/callback` for Nabu Casa or `https://myexternalfacingurl.com/auth/external/callback` for some other generic external facing URI). Does that make sense?

frenck: Yes, that is documented in many other integrations as well. E.g., Spotify, Somfy, Ring ...

allenporter: @mikemc35 what do you think of this? https://github.com/home-assistant/home-assistant.io/compare/current...allenporter:nest_no_devices

mikemc35: @allenporter Yep that looks good to me. Apologies for not seeing the rest of the thread--I think github timed out on me without me realizing. Thanks!

allenporter: Would you like to incorporate any of that here? or want me to send that whole update?

mikemc35: @allenporter If that works for both you and @frenck, what you have looks good to me.

frenck: I'm going to close this PR.

The reasoning for this: the current wording is not complete (as discussed above), but mainly, we have made a big change to Home Assistant removing this whole thing. As of the next release, the actual browser URL will automatically be used by Home Assistant; Removing the whole internal/external discussion.

PR #15705, opened 2020-11-20
Add Colored Flash Example to deCONZ

D34DC3N73R:
## Proposed change
Update documentation with an example of colored flashing of hue bulbs via `deconz.configure` as a workaround since the `light.turn_on` service with `flash:long|short` and `color` does not work with deCONZ.

## Type of change

## Additional information

- Additional discussion: https://community.home-assistant.io/t/update-deconz-integration-to-support-flash-with-color/246904

## Checklist

### DIFF
--- a/source/_integrations/deconz.markdown
+++ b/source/_integrations/deconz.markdown
@@ -275,12 +275,16 @@ Note: Requires `on: true` to change color while the bulb is off. If `on:true` is
 {% raw %}
 
 ```yaml
-script:
-  hue_lamp_flash:
-    alias: Hue Lamp Flash
-    sequence:
+automation:
+  - alias: Flash Hue Bulb with Doorbell Motion
+    trigger:
+    - platform: state
+      entity_id: binary_sensor.doorbell_motion
+      to: 'on'
+    action:
     - service: deconz.configure
       data:
+        entity: light.hue_lamp
         field: /state
         data:
           'on': true
@@ -288,13 +292,13 @@ script:
           sat: 255
           bri: 255
           alert: breathe
-      entity_id: light.hue_lamp
-    - delay: '15'
+    - delay: 00:00:15
     - service: deconz.configure
       data:
+        entity: light.hue_lamp
+        field: /state
         data:
           'on': false
-      entity_id: light.hue_lamp
     mode: single
 ```
 
--- a/source/_integrations/deconz.markdown
+++ b/source/_integrations/deconz.markdown
@@ -268,6 +268,37 @@ automation:
 ```
 
 {% endraw %}
+
+#### Colored Flashing - RGB Philips Hue bulb using deconz.configure
+Note: Requires `on: true` to change color while the bulb is off. If `on:true` is specified, the bulb remains on after flashing is complete. The previous color is not saved.
+
+{% raw %}
+
+```yaml
+script:
+  hue_lamp_flash:
+    alias: Hue Lamp Flash
+    sequence:
+    - service: deconz.configure
+      data:
+        field: /state
+        data:
+          'on': true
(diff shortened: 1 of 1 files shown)


Kane610 (dismissed):
Awesome! Thanks for contributing. Looks good to me

D34DC3N73R:
Ok so, I think the original script example had some errors. It should use `entity` instead of `entity_id` if I'm not mistaken, and `entity` should be in the first `data:` of the service call. I thought the nested `data:` lines were an error too but it turns out it's required so `field` can be included. I couldn't get the script to run correctly at first so I changed the example to an automation. 

The replacement automation example works too. But IMO, this updated script may be a better example with less extraneous detail.  

```
script:
  hue_lamp_flash:
    alias: Hue Lamp Flash
    sequence:
    - service: deconz.configure
      data:
        field: /state
        data:
          'on': true
          hue: 65535
          sat: 255
          bri: 255
          alert: breathe
        entity: light.hue_lamp
    - delay: '15'
    - service: deconz.configure
      data:
        field: /state
        data:
          'on': false
        entity: light.hue_lamp
    mode: single
```

Sorry for the runaround. Let me know if you'd like this to be submitted as another commit, or even as a separate PR.

source/_integrations/deconz.markdown
```diff
@@ -267,6 +267,41 @@ automation:
     mode: restart
 ```
 
+{% endraw %}
+
+#### Colored Flashing - RGB Philips Hue bulb using deconz.configure
```
frenck, line 272: New line after heading.

```suggestion
#### Colored Flashing - RGB Philips Hue bulb using deconz.configure

```

source/_integrations/deconz.markdown
```diff
@@ -267,6 +267,41 @@ automation:
...
+Note: Requires `on: true` to change color while the bulb is off. If `on:true` is specified, the bulb remains on after flashing is complete. The previous color is not saved.
+
+{% raw %}
```
frenck: Why is this escaped? Doesn't seem needed?

source/_integrations/deconz.markdown
```diff
@@ -267,6 +267,41 @@ automation:
...
+          bri: 255
+          alert: breathe
+    - delay: 00:00:15
+    - service: deconz.configure
+      data:
+        entity: light.hue_lamp
+        field: /state
+        data:
+          'on': false
+    mode: single
```
frenck: move this line to the top of the automation

frenck (changes requested):
This is not documenting flashing but a breathing flash effect, which is kinda Philips specific. I think we need to be clear on the difference in the description.

frenck:
removed the review request, as my original suggestion in the review comment wasn't resolved.

D34DC3N73R:
I updated the Note: line to include Philips Hue by name. 
alert: breathe is what hue bridge uses with service light.turn_on and flash, so I think it makes sense to keep that description for users looking for a replacement.

source/_integrations/deconz.markdown
frenck: Why is this marked resolved? The template is still escaped without reason. Please remove it.

D34DC3N73R: If you're referring to the {% raw %} {% endraw %} tags, I just followed the format of the other published examples.

frenck: Ok, but it isn't needed here, so please remove those.

D34DC3N73R: Ok, removed the tags in the latest commit.

frenck (approved):

frenck:
Thanks 👍

PR #15712, opened 2020-11-21
Update Spotify URL and port instructions with more detail

freekeys:
## Type of change

## Additional information

## Checklist

- [ x] The documentation follows the Home Assistant documentation [standards][].

### DIFF
--- a/source/_integrations/spotify.markdown
+++ b/source/_integrations/spotify.markdown
@@ -36,7 +36,7 @@ The Spotify media player integration allows you to control [Spotify](https://www
 - Click **Save** after adding the URI.
 
 <div class='note'>
-  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses.
+  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses. You can get the IP & port for this address from the **Configuration > General** panel in Home Assistant. You should copy the Internal URL here into the External URL field. Note that you should also use this URL to access Home Assistant when you follow the **Configuration** steps below.
 </div>  
 
 ## Configuration
--- a/source/_integrations/spotify.markdown
+++ b/source/_integrations/spotify.markdown
@@ -36,7 +36,7 @@ The Spotify media player integration allows you to control [Spotify](https://www
 - Click **Save** after adding the URI.
 
 <div class='note'>
-  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses. You can get the IP & port for this address from the **Configuration > General** panel in Home Assistant. You should copy the Internal URL here into the External URL field. Note that you should also use this URL to access Home Assistant when you follow the **Configuration** steps below.
+  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses. You can get the IP & port for this address from the **Configuration > General** panel in Home Assistant. You should copy the Internal URL here into the External URL field. 
 </div>  
 
 ## Configuration
@@ -63,6 +63,7 @@ client_secret:
 
 ## Activating the Spotify integration
 
+- Access Home Assistant using the URL that you used in the **Redirect URI** step above 
 - Go to the **Configuration** page in the Home Assistant frontend.
 - Click on **Integrations**.
 - Click the orange **+** on the bottom-right.
--- a/source/_integrations/spotify.markdown
+++ b/source/_integrations/spotify.markdown
@@ -36,7 +36,7 @@ The Spotify media player integration allows you to control [Spotify](https://www
 - Click **Save** after adding the URI.
 
 <div class='note'>
-  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses. You can get the IP & port for this address from the **Configuration > General** panel in Home Assistant.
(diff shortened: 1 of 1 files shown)


source/_integrations/spotify.markdown
```diff
@@ -36,7 +36,7 @@ The Spotify media player integration allows you to control [Spotify](https://www
 - Click **Save** after adding the URI.
 
 <div class='note'>
-  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses.
+  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses. You can get the IP & port for this address from the **Configuration > General** panel in Home Assistant. You should copy the Internal URL here into the External URL field.
```
frenck: This is incorrect. Home Assistant will automatically determine the URL to use (this can be either internal, external, IP based, default hostname based or Nabu Casa remote UI URLs).
There is no reason to do what you have described here.

frenck (changes requested):

source/_integrations/spotify.markdown
freekeys: Okay I've updated the file to remove that part (weird that people seemed to think that helped, must have been something else instead that actually did it). Are the other 2 parts still valid?

frenck: Unfortunately, not really. The internal & external fields can be completely empty and still valid (e.g., when using IP, hostname or the Nabu Casa remote UI) and it only shows up for people that have enabled advanced mode.

freekeys: Ah okay, maybe if I rephrase to say that it might be there? A lot of people seemed to be unsure where to find it so it feels like it's worth pointing out that this is a place they can possibly find it? Ports etc are quite mysterious to newbies on this. You know better than me though of course who this would actually show up for - perhaps we could just specify that?

frenck: I'm not sure what the confusion might be here? The URL is in the omnibar/urlbar of your browser all the time?!

freekeys: Ha yeah but the nabu casa URL wouldn't work for me with port 8123. The local IP one did. Also the address in the browser doesn't show 8123 so not everyone knows that's the port (especially newer users who have come straight into nabu casa and not used duckdns etc)

frenck: But I don't get the whole port story. Just use the URL from the browser? Whatever is in there is what to use...

freekeys: Ah so do you not need the port? I initially tried my browser url (nabu.casa) and added :8123. .. but this didn't work. I had to switch to my local ip:8123

frenck: You could need the port, yes... but again... It is the same URL as you use in your browser...

freekeys: So the nabu casa url + :8123 should work?

frenck: wut? Where did I say that? Does it work in your browser? If so, sure... But I'm pretty sure that doesn't work.

source/_integrations/spotify.markdown
```diff
@@ -36,7 +36,7 @@ The Spotify media player integration allows you to control [Spotify](https://www
 - Click **Save** after adding the URI.
 
 <div class='note'>
-  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses.
+  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses. You can get the IP & port for this address from the **Configuration > General** panel in Home Assistant.
```
frenck: Let's revert this, as it is simply incorrect.

```suggestion
  Your Home Assistant instance does not need to be exposed to the internet. It works just fine with local IP addresses.
```


The other change below is valid and a nice addition 👍

source/_integrations/spotify.markdown
freekeys: I think the docs imply that you can use your URL, which for me for example is my nabu.casa URL. So I tried that with 8123 and it didn't work. So I then tried my local IP and that worked. Wondering if that should be clarified in the docs as I think "home assistant URL" could often be mistaken as the nabu.casa. 

And then secondly I think it should be made clear how to find the port (and that for most people it will be 8123) - I knew this but can see from forums not all do.

frenck: So; you can use the Nabu Casa URL, that works and is tested.

Furthermore, the port has nothing to do with this. The URL has something to do with this, which has to match your browser and is what is documented.

Considering this conversation, it seems like you are guessing as well. I suggest we remove this part completely and merge this PR with just the last added bit (which is a valuable addition IMHO).

freekeys: Okay cool, happy to defer to your view! So the last bit you mean is point 1 in the configuration part?

freekeys: And yeah I'm totally guessing! It took me a while to get working, hence trying to do my bit to clarify docs if possible (haven't done so before so apologies if not thought through enough!)

frenck: Yes, I meant the last point of this PR that add the note on using the redirect URL as configured @ Spotify.

freekeys (commented):
Okay I think that's good to go now

frenck (approved):
Thanks 👍

PR #15751, opened 2020-11-25
Clarify the use of choose in advanced actions

patatman:
Signed-off-by: Patatman <git@jeursen.nl>

## Proposed change

I found it hard to find documentation about the `if/else` logic in the actions part of automations. I thought it wasn't possible, until I found the `choose` action. 
I hope to clarify the possibility by adding this "advanced actions" section. 

I was actually hesitant to make a new doc, or to add it to the existing docs. 
For example moving the [choose-a-group-of-actions](https://www.home-assistant.io/docs/scripts/#choose-a-group-of-actions) to the [actions page](https://www.home-assistant.io/docs/automation/action/)

I ended up creating a new doc, since this can be used for more advanced action usage in the future. 

## Type of change

## Checklist


source/_docs/automation/advanced-actions.markdown
```diff
@@ -0,0 +1,55 @@
+---
+title: "Advanced Actions"
+description: "Automations can have conditionals"
+---
+
+Sometimes to keep things organized in one automation it might be needed to add logic to the action.
```
frenck: Why is this a new page? And why is the choose action an advanced action?

As a matter of fact, we don't even document actions for automation, as it refers to the extensive documentation on the script syntax for the actions.

So this page feels a bit added at random. If any place, I think we should adjust the current automation action page; or, figure out the problem with the current script actions page and fix that.

frenck (changes requested):

source/_docs/automation/advanced-actions.markdown
patatman: It's a new page because I was hesitant if it could be included in the normal actions page(like I mentioned in the PR).
Since it's more than just a basic flow (e.g. something triggers > action), and for beginners who don't have experience with programming/automations it might seem advanced. 

> As a matter of fact, we don't even document actions for automation, as it refers to the extensive documentation on the script syntax for the actions.

And I think this is the reason a lot of users are missing these cool features(myself included). If I'm creating a `automation` I don't expect to lookup documentation stuff related to `scripts`. 

The main thing I was trying to accomplish is to have a clear documentation in regards to `if/else` statements within automations. I couldn't find this myself, maybe because it's named `choose` and not an actual `if/else` statement, so my search query was wrong. To help others in the future, I wanted to help clarify the docs.

frenck: > And I think this is the reason a lot of users are missing these cool features(myself included). If I'm creating a automation I don't expect to lookup documentation stuff related to scripts.

We should explain that in that case. The new page isn't helping with that imho.

> The main thing I was trying to accomplish is to have a clear documentation in regards to if/else statements within automations. I couldn't find this myself, maybe because it's named choose and not an actual if/else statement, so my search query was wrong. To help others in the future, I wanted to help clarify the docs.

But that is fixing one single case, on hidden away additional page. I think that is not the right thing to do.

patatman: >We should explain that in that case. The new page isn't helping with that imho.

Agreed, this would add more clutter.
I'll see if I can rework it, and clarify that the `scripts` section can also be used in `automations`

Do you want me to create a new PR when done? Or Should I use this one?

frenck: Either way it is fine 👍

patatman:
I've changed my previous config to be included into the action page.
I'm not sure if the example YAML can be more simplified? I'm using the new `device` actions, which makes it odd compared to the other examples on the page.
What are your thoughts on this? @frenck

patatman:
I already changed the initial commit, have you had a look after my commit 051a78a ?

PR #15829, opened 2020-12-05
Update zha.markdown

Jpsy:
## Proposed change
* HA restart required to detect new modules (i.e. USB)
* Add details on where to find serial device paths.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/zha.markdown
+++ b/source/_integrations/zha.markdown
@@ -84,6 +84,8 @@ Some other Zigbee coordinator hardware may not support a firmware that is capabl
 
 ## Configuration - GUI
 
+Connect your radio module and restart Home Assistant.
+
 From the Home Assistant front page go to **Configuration** and then select **Integrations** from the list.
 
 Use the plus button in the bottom right to add a new integration called **ZHA**.
@@ -118,7 +120,8 @@ radio type. In the pop-up:
 
 Most devices need at the very least the serial device path, like `/dev/ttyUSB0`, but it is recommended to use
 device path from `/dev/serial/by-id` folder,
-e.g., `/dev/serial/by-id/usb-Silicon_Labs_HubZ_Smart_Home_Controller_C0F003D3-if01-port0`
+e.g., `/dev/serial/by-id/usb-Silicon_Labs_HubZ_Smart_Home_Controller_C0F003D3-if01-port0`  
+A list of available device paths can be found in **Supervisor** > **System** > **Host System** > **dot menu** > **Hardware**.
 
 Press `Submit`. The success dialog will appear or an error will be displayed in the popup. An error is likely if Home Assistant can't access the USB device or your device is not up to date. Refer to [Troubleshooting](#troubleshooting) below for more information.
 


source/_integrations/zha.markdown
```diff
@@ -84,6 +84,8 @@ Some other Zigbee coordinator hardware may not support a firmware that is capabl
 
 ## Configuration - GUI
 
+Connect your radio module and restart Home Assistant.
```
frenck, line 87: That should not be needed?

source/_integrations/zha.markdown
```diff
@@ -118,7 +120,8 @@ radio type. In the pop-up:
 
 Most devices need at the very least the serial device path, like `/dev/ttyUSB0`, but it is recommended to use
 device path from `/dev/serial/by-id` folder,
-e.g., `/dev/serial/by-id/usb-Silicon_Labs_HubZ_Smart_Home_Controller_C0F003D3-if01-port0`
+e.g., `/dev/serial/by-id/usb-Silicon_Labs_HubZ_Smart_Home_Controller_C0F003D3-if01-port0`  
+A list of available device paths can be found in **Supervisor** > **System** > **Host System** > **dot menu** > **Hardware**.
```
frenck, line 124: ZHA automatically detects the hardware and provides a list?

Jpsy: Yes, you can choose from a dropdown if the radio is recognized correctly. But then you don't even get to that step. 
If you get to that step, HA did not recognize your module and you only see a text field, not a dropdown.

frenck: That sounds like a bug to me. Has that been reported?

Jpsy: No idea if this is a bug. I will try to reproduce the situation and take a screenshot.
Seems to happen when you plug in a Zigbee USB radio stick and start to add ZHA integration right away without first restarting HA. 
That is the reason for the other part of my edit: Restart HA first after plugging in a USB radio.

frenck: Yes, which is not "normal" or wished for behavior. Therefore we should report it as a bug, instead of documenting workarounds.

Jpsy:
I am sorry, but I do not see any bug according to the docs.

_Press Submit and the integration will try to detect radio type automatically. **If unsuccessful, you will get a new pop-up asking for a radio type.** ...:_

This is exactly what happened.
The only problem is that the radio was not recognized in the first place. This is likely due to the fact that I plugged the USB radio (Conbee II) in and did not restart HA. After a restart the automatic detection worked perfectly. Without detection the flow of dialogs was correct too.

frenck:
>  This is likely due to the fact that I plugged the USB radio (Conbee II) in and did not restart HA.

The point is... a restart to detect a USB device is not correct behavior, it should have been picked up.

source/_integrations/zha.markdown
Hedda: Restart should not be needed with Home Assistant OS, but if not using Home Assistant OS as the installation method...

source/_integrations/zha.markdown
Hedda: Restart should not be needed with Home Assistant OS, but if not using Home Assistant OS as the installation method...

Hedda:
+1 for this likely being a bug. Restart of Home Assistant should not be needed with Home Assistant OS as the installation method.

https://www.home-assistant.io/docs/installation/

Hedda:
You could try to install/update to latest pyusb (libusb) with the command

python3 -m pip install pyusb>=1.1.0

https://pypi.org/project/pyusb/

source/_integrations/zha.markdown
Adminiuga: For whatever reason, after plugging a new USB stick into Running Home Assistant OS installation for the very 1st time, it does show as a serial port in the Supervisor Hardware tab, however accessing the serial port from the HA Core instance fails. Exactly same steps work after the host reboot.

source/_integrations/zha.markdown
Adminiuga: ZHA automatically detects the hardware, **BUT** user have to pick the right serial port from the provided list of serial ports first. List of the serial is provided to the user. 
Given the number of Adapters ZHA and different adapters could be using different ports (USB or hardware serial) it is not feasible to automatically detect the port and type of the hardware.

frenck: @Adminiuga The problem described in this case; is that the list of serial ports is not updating / stays empty after plugging in the device (while Home Assistant is running).

Adminiuga: And I think this is consistent with what i'm seeing with Home Assistant OS. The newly plugged USB serial device is not available inside homeassistant container, not unless the host is restarted at least once.

ZHA configuration flow gets list of ports here https://github.com/home-assistant/core/blob/20c3fdfe1454dbdc845d1c6c33d934caac294ae8/homeassistant/components/zha/config_flow.py#L42 and it is updated every time the ZHA configuration flow is started. So for some reason the fresh device is not exposed for HA Core container immediately.

Adminiuga: I was able to reproduce the issue with Home Assistant OS (latest beta 1.0.0b3, HassOS: 4.19, Supervisor: 2020.12.6) running on virtual box.
Plugging a HUSBZB stick does show new USB ports in Supervisor -> Hardware, but Core container does not have those devices, not until  you restart the core container.

Adminiuga: Essentially USB stick must be plugged in when Core container is started, otherwise the serial devices won't be available inside the core container

Jpsy: I see the same results on RPi 4, HA OS 5.6 dev, Supervisor: 2020.12.6. 
(My HA version is currently still on 0.117.6 due to a KNX bug in 0.118.)

frenck: Just reproduced it myself as well. I've also seen issues have been raised, so hopefully, this can be addressed in the future.

However, this means, right now, this is how it is. So let's merge this in until happier days arrive 🎉 👍

frenck (approved):

source/_integrations/zha.markdown
Hedda: @frenck So there is already an issue/bug-report open for this? Does anyone have link to open issue(s) for reference?

Update: See the link now -> https://github.com/home-assistant/supervisor/issues/2358

Adminiuga: This is targeted to be resolved in Q2-Q3 2021 per Pascal

PR #16101, opened 2021-01-04
Tweak ping device_tracker docs

ronaldevers:
## Proposed change

I've clarified the ping integration docs a little bit.  I could not figure out from the `device_tracker` example and the text that `hostname` (under `hosts`) was meant to be the device name. Also, the type is wrong, `hosts` is not a list but a map.

So based on the current docs I was trying to do this in my config which would not work:

```
device_tracker:
  - platform: ping
    hosts:
      - hostname: 192.168.2.10
      - hostname: 192.168.2.11
```

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/ping.markdown
+++ b/source/_integrations/ping.markdown
@@ -100,4 +100,4 @@ count:
   type: integer
 {% endconfiguration %}
 
-See the [device tracker integration page](/integrations/device_tracker/) for instructions how to configure the people to be tracked.
+See the [person integration page](/integrations/person/) for instructions on how to configure the people to be tracked.
--- a/source/_integrations/ping.markdown
+++ b/source/_integrations/ping.markdown
@@ -86,14 +86,14 @@ To use this presence detection in your installation, add the following to your `
 device_tracker:
   - platform: ping
     hosts:
-      hostname: 192.168.2.10
+      device_name_1: 192.168.2.10
 ```
 
 {% configuration %}
 hosts:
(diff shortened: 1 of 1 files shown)


source/_integrations/ping.markdown
```diff
@@ -86,14 +86,15 @@ To use this presence detection in your installation, add the following to your `
 device_tracker:
   - platform: ping
     hosts:
-      hostname: 192.168.2.10
+      hostname1: 192.168.2.10
+      hostname2: 192.168.2.11
```
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters that they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

```suggestion
```

frenck (changes requested):

source/_integrations/ping.markdown
ronaldevers: Reading the docs again is says this is a map of device name to hostname or IP address. So maybe instead we make it:

```
hosts:
    device_name_1: 192.168.2.10
```

What do you think?

ronaldevers: Updated and added one more commit to fix the link to the people-tracking documentation.

frenck (approved):
Thanks, @ronaldevers 👍 

Happy New Year 🥂

PR #16121, opened 2021-01-07
Update aquostv.markdown with Sharp LC-70LE747

Fietspomp86:
Added Sharp LC-70LE747E to the list

## Additional information
Volume control doesn't work.


frenck (approved):
Thanks, @Fietspomp86 👍

PR #16131, opened 2021-01-09
Add an nest event example automation

TheJefe:
## Proposed change

I struggled a bit getting started with automations for this event
because I didn't know the event_type would actually be a `nest_event`.
This made me think that others might want to get a bit of a short cut
working with these events, so I am sharing my automation here to help
the next person along.

## Type of change

Adding additional documentation to the existing nest integration

## Additional information

I followed https://github.com/home-assistant/home-assistant.io/pull/15724 , and thought it could use just this little bit of additional information that makes it actionable to use in automations

## Checklist

### DIFF
--- a/source/_integrations/nest.markdown
+++ b/source/_integrations/nest.markdown
@@ -337,6 +337,8 @@ action:
     data:
       title: motion detected
       message: front door motion detected
+      data:
+        image: /api/camera_proxy/camera.front_door
 mode: single
 ```
 
--- a/source/_integrations/nest.markdown
+++ b/source/_integrations/nest.markdown
@@ -301,26 +301,7 @@ See [Automating Home Assistant](/getting-started/automation/) for the getting st
 
 ### Example
 
-The events for emitted to Home Assistant look like this..
-
-```json
-{
-    "event_type": "nest_event",
-    "data": {
-        "device_id": "37d6639fa63c348310d930825bb9e430",
-        "type": "camera_motion"
-    },
-    "origin": "LOCAL",
-    "time_fired": "2021-01-03T17:59:32.509459+00:00",
-    "context": {
-        "id": "7cf91c62a74c36ea6de0634c18138384",
-        "parent_id": null,
-        "user_id": null
-    }
-}
-```
-
-So an example automation looks like this..
+This automation will trigger when a `nest_event` event type with a type of `camera_motion` is received from the specified `device_id`.
 
 ```yaml
 alias: motion alert
@@ -341,6 +322,8 @@ action:
         image: /api/camera_proxy/camera.front_door
 mode: single
 ```
+
+The action in this section uses the [Android Companion App](https://companion.home-assistant.io/docs/notifications/notifications-basic/) and the camera proxy to send an notification with a snapshot from the camera.
 
 # Legacy Works With Nest API
 
--- a/source/_integrations/nest.markdown
+++ b/source/_integrations/nest.markdown
@@ -305,14 +305,12 @@ This automation will trigger when a `nest_event` event type with a type of `came
 
 ```yaml
 alias: motion alert
-description: ''
 trigger:
   - platform: event
     event_type: nest_event
     event_data:
-      device_id: 37d6639fa63c348310d930825bb9e430
+      device_id: YOUR_DEVICE_ID
       type: camera_motion
-condition: []
 action:
   - service: notify.mobile_app_pixel_2
     data:
@@ -320,7 +318,6 @@ action:
       message: front door motion detected
       data:
         image: /api/camera_proxy/camera.front_door
-mode: single
 ```
 
 The action in this section uses the [Android Companion App](https://companion.home-assistant.io/docs/notifications/notifications-basic/) and the camera proxy to send an notification with a snapshot from the camera.
--- a/source/_integrations/nest.markdown
+++ b/source/_integrations/nest.markdown
@@ -299,6 +299,46 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
 
 See [Automating Home Assistant](/getting-started/automation/) for the getting started guide on automations or the [Automation](/docs/automation/) documentation for full details.
 
+### Example
+
+The events for emitted to Home Assistant look like this..
+
+```json
+{
+    "event_type": "nest_event",
+    "data": {
+        "device_id": "37d6639fa63c348310d930825bb9e430",
+        "type": "camera_motion"
+    },
+    "origin": "LOCAL",
+    "time_fired": "2021-01-03T17:59:32.509459+00:00",
+    "context": {
+        "id": "7cf91c62a74c36ea6de0634c18138384",
+        "parent_id": null,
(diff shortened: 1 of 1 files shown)


source/_integrations/nest.markdown
```diff
@@ -299,6 +299,46 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
...
+    event_type: nest_event
+    event_data:
+      device_id: 37d6639fa63c348310d930825bb9e430
+      type: camera_motion
+condition: []
+action:
+  - service: notify.mobile_app_pixel_2
+    data:
+      title: motion detected
+      message: front door motion detected
```
allenporter, line 318: You can also include an event image for a fun example:
```
      data:
        image: /api/camera_proxy/camera.front_door
```

source/_integrations/nest.markdown
```diff
@@ -299,6 +299,46 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
...
+    "time_fired": "2021-01-03T17:59:32.509459+00:00",
+    "context": {
+        "id": "7cf91c62a74c36ea6de0634c18138384",
+        "parent_id": null,
+        "user_id": null
+    }
+}
+```
+
+So an example automation looks like this..
```
allenporter: Lets give more context that this is for the Android Companion app so users don't get confused about what device types are supported
https://companion.home-assistant.io/docs/notifications/notifications-basic

source/_integrations/nest.markdown
```diff
@@ -299,6 +299,46 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
 
 See [Automating Home Assistant](/getting-started/automation/) for the getting started guide on automations or the [Automation](/docs/automation/) documentation for full details.
 
+### Example
+
+The events for emitted to Home Assistant look like this..
```
allenporter: I would lean towards just including the automation details, since those are more user friendly.  I feel like folks familiar with events may be able to figure this out.

allenporter (commented):
Also as an FYI I have this PR pending:
https://github.com/home-assistant/home-assistant.io/pull/16043

source/_integrations/nest.markdown
TheJefe: Oh wow! I didn't even know I could do that! :D , thanks for the feedback, I've included this change in a new commit

source/_integrations/nest.markdown
TheJefe: Good call, I've added more details about the trigger and the actions

source/_integrations/nest.markdown
TheJefe: Yeah, that makes sense. I've removed that section

source/_integrations/nest.markdown
```diff
@@ -299,6 +299,31 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
 
 See [Automating Home Assistant](/getting-started/automation/) for the getting started guide on automations or the [Automation](/docs/automation/) documentation for full details.
 
+### Example
+
+This automation will trigger when a `nest_event` event type with a type of `camera_motion` is received from the specified `device_id`.
+
+```yaml
+alias: motion alert
+description: ''
```
frenck: Let's either add an actual description or remove the option.

source/_integrations/nest.markdown
```diff
@@ -299,6 +299,31 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
...
+trigger:
+  - platform: event
+    event_type: nest_event
+    event_data:
+      device_id: 37d6639fa63c348310d930825bb9e430
```
frenck: Please use a placeholder; alternatively; add a YAML comment on that it needs to be replaced?

```suggestion
      device_id: YOUR_DEVICE_ID
```

source/_integrations/nest.markdown
```diff
@@ -299,6 +299,31 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
...
+      type: camera_motion
+condition: []
```
frenck: Lets either have conditions or drop the option.

source/_integrations/nest.markdown
```diff
@@ -299,6 +299,31 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
...
+action:
+  - service: notify.mobile_app_pixel_2
+    data:
+      title: motion detected
+      message: front door motion detected
+      data:
+        image: /api/camera_proxy/camera.front_door
+mode: single
```
frenck: Any specific reason this needs to be `single`?

frenck (changes requested):
I wonder, if we should use a Blueprint instead?

allenporter:
> I wonder, if we should use a Blueprint instead?

My impression is that Blueprints don't yet support device events, but once it does that this would be a great fit.

source/_integrations/nest.markdown
TheJefe: :+1: works for me, I've applied this change in a new commit. Thanks for your feedback!

source/_integrations/nest.markdown
TheJefe: :+1: Removed

source/_integrations/nest.markdown
TheJefe: :+1: removed

source/_integrations/nest.markdown
TheJefe: Nope, removed in a new commit

TheJefe:
All feedback addressed, let me know if there is anything else I can do. Thank you for spending your time reviewing this :bow:

frenck (approved):
Thanks, @TheJefe 👍

source/_integrations/nest.markdown
```diff
@@ -299,6 +299,28 @@ All Google Nest Cam models and the Google Nest Hello Video Doorbell support [Dev
...
 See [Automating Home Assistant](/getting-started/automation/) for the getting started guide on automations or the [Automation](/docs/automation/) documentation for full details.
 
+### Example
+
+This automation will trigger when a `nest_event` event type with a type of `camera_motion` is received from the specified `device_id`.
+
+```yaml
+alias: motion alert
+trigger:
+  - platform: event
```
allenporter, line 309: I looked closer and realized: I think we should encourage use of device triggers instead of event triggers.  The examples for `nest` are very similar, but i think my impression is device triggers are the preferred approach and they work better in the UI anyway.
 
```
trigger:
  - platform: device
    device_id: YOUR_DEVICE_ID
    domain: nest
    type: doorbell_chime
```

frenck: Yes, however, device triggers are generally for use with the UI, not in bare YAML (but could be done, but we don't have it anywhere).

allenporter: OK, so i guess the yaml for the action is the part that is still relevant.  But then maybe just blueprint support for device triggers is the answer...

PR #16516, opened 2021-02-11
Getting Started follow up changes

agners:
## Type of change

## Additional information

## Checklist

### DIFF
--- a/sass/custom/_getting_started.scss
+++ b/sass/custom/_getting_started.scss
@@ -3,10 +3,6 @@
         padding-bottom: 16px;
     }
 
-    em {
-        font-weight: bold;
-    }
-
     .installations {
 
         h2 {
--- a/sass/custom/_getting_started.scss
+++ b/sass/custom/_getting_started.scss
@@ -3,6 +3,10 @@
         padding-bottom: 16px;
     }
 
+    em {
+        font-weight: bold;
+    }
+
     .installations {
 
         h2 {
--- a/source/installation/index.html
+++ b/source/installation/index.html
@@ -10,8 +10,8 @@ toc: true
 Home Assistant can be installed in 4 flavors:
 </p>
 <ul>
-    <li><em>Home Assistant Operating System</em>: Minimal Operating System optimized to power Home Assistant. Comes with the Supervisor pre-installed. Recommended for new users.</li>
-    <li><em>Home Assistant Supervised</em>: Container based installation with our Supervisor to manage Home Assistant Core and Add-Ons.</li>
+    <li><em>Home Assistant Operating System</em>: Minimal Operating System optimized to power Home Assistant. Comes with Supervisor to manage Home Assistant Core and add-ons pre-installed. Recommended installation type.</li>
+    <li><em>Home Assistant Supervised</em>: Manual installation of the Supervisor.</li>
     <li><em>Home Assistant Container</em>: Standalone container based installation of Home Assistant Core.</li>
     <li><em>Home Assistant Core</em>: Manual installation using a Python virtual environment.</li>
 </ul>
@@ -19,7 +19,7 @@ Home Assistant can be installed in 4 flavors:
 The list below shows which flavor is available on what device or platform.
 </p>
 <p>
-If you are using the <em>Home Assistant Blue</em>, the Home Assistant Operating System is already preinstalled and you can skip the installation step.
+If you are using the <a href="/blue">Home Assistant Blue</a>, the Home Assistant Operating System is already preinstalled. You can skip this installation step.
 </p>
 <p>
 We recommend a dedicated system to run Home Assistant. If you are unsure what to select, using a <em>Raspberry Pi</em> and the <em>Home Assistant Operating System</em> is a good start.
@@ -136,7 +136,7 @@ We recommend a dedicated system to run Home Assistant. If you are unsure what to
                 <div>
                     <b>macOS</b>
                     <ul>
-                        <li>Home Assistant Operating System (using Virtual Machine)</li>
+                        <li>Home Assistant Operating System (VM)</li>
                         <li>Home Assistant Core</li>
                     </ul>
                 </div>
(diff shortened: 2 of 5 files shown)


sass/custom/_getting_started.scss
```diff
@@ -3,6 +3,10 @@
         padding-bottom: 16px;
     }
 
+    em {
+        font-weight: bold;
```
frenck: Why not just use `<b>` in that case?

source/installation/index.html
```diff
@@ -6,6 +6,25 @@
 toc: true
 ---
 
+<p>
+Home Assistant can be installed in 4 flavors:
+</p>
+<ul>
+    <li><em>Home Assistant Operating System</em>: Minimal Operating System optimized to power Home Assistant. Comes with Supervisor pre-installed. Recommended for new users.</li>
```
frenck:
```suggestion
    <li><em>Home Assistant Operating System</em>: Minimal Operating System optimized to power Home Assistant. Comes with the Supervisor pre-installed. Recommended for new users.</li>
```

Additionally, is it recommended for new users? As in, this is recommended in general, right?

source/installation/index.html
```diff
@@ -6,6 +6,25 @@
...
+    <li><em>Home Assistant Supervised</em>: Container based installation with our Supervisor to manage Home Assistant Core and Add-Ons.</li>
```
frenck:
```suggestion
    <li><em>Home Assistant Supervised</em>: Container based installation with our Supervisor to manage Home Assistant Core and add-ons.</li>
```

frenck: I think we should avoid "Container-based installation". A container is an important driver, but we want to avoid confusion with the actual container-based installation type.

In general, this is considered a highly advanced installation type, so I think this should move lower down the list.

source/installation/index.html
```diff
@@ -6,6 +6,25 @@
...
+    <li><em>Home Assistant Container</em>: Standalone container based installation of Home Assistant Core.</li>
```
frenck: Should this mention the term "Docker"?

source/installation/index.html
```diff
@@ -6,6 +6,25 @@
...
+    <li><em>Home Assistant Core</em>: Manual installation using a Python virtual environment.</li>
+</ul>
+<p>
+The list below shows which flavor is available on what device or platform.
+</p>
+<p>
+If you are using the <em>Home Assistant Blue</em>, the Home Assistant Operating System is already preinstalled and you can skip the installation step.
```
frenck: Let's make sure we link it to the Blue page.

source/installation/index.html
```diff
@@ -98,7 +117,7 @@ <h2>Windows</h2>
                 <div>
                     <b>Windows</b>
                     <ul>
-                        <li>Home Assistant Operating System</li>
+                        <li>Home Assistant Operating System (using Virtual Machine)</li>
```
frenck: This makes this item a bit long (especially on mobile). I think this can be emphasized on the page where this leads to. I would simply shorten to VM in these items?

```suggestion
                        <li>Home Assistant Operating System (VM)</li>
```

sass/custom/_getting_started.scss
agners: Didn't made the font bold, not sure why.

But isn't emphasizing preferred over direct format tags such as `b` these days?

source/installation/index.html
agners: Agreed, recommend it in general make sense. Will change that.

source/installation/index.html
agners: It does requires a container engine to be useful, so I think mentioning the term container does make sense. Maybe we can put it to the end to weaken it?

I agree its highly advanced and we should not encourage people using it. But I kinda like to keep the order, as it is logically more to less...

How about adding some words to make it clear that this is an advanced installation? E.g. `Manual`: 

```suggestion
    <li><em>Home Assistant Supervised</em>: Manual installation of the Supervisor to manage Home Assistant Core and add-ons (using containers).</li>
```

sass/custom/_getting_started.scss
frenck: dunno, feels weird to introduce another property, but that is me :)

I'm not the CSS expert here, this is more @ludeeus domain.

source/installation/index.html
frenck: I really want to ditch containers, we have deliberately not put it into the ADRs as well. There has been a long history of misunderstand of this installation method in our community.

It requires an OS and a whole bunch of requirements to be useful as well 😉 

"Hass.io in Docker" is a term often used, that is just super wrong (which drives this).

> This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System.

That is the tekst from the ADR. Maybe we can incorporate that?

source/installation/index.html
ludeeus: I don't think so, Docker is only one of the runtimes that can be used, which is explained on the installation page for the container installation.

sass/custom/_getting_started.scss
ludeeus: I think `<b>` should be used here, well not here but in the intro text/list.
I don't see a reason it should not work, it works when I do it.

source/installation/index.html
```diff
@@ -6,6 +6,25 @@
 toc: true
 ---
 
+<p>
```
ludeeus: The entire intro should be wrapped in a div with the `intro` class

source/installation/index.html
frenck: Yeah, I agree on that... However, covers 95% of the use cases. I think it is good, was just suggesting on for example:

```suggestion
    <li><em>Home Assistant Container</em>: Standalone container-based installation of Home Assistant Core (ex: Docker).</li>
```

Or something.
Purely so people recognize is more quickly directly from the index.

frenck (approved):
Thanks, @agners 👍

PR #16903, opened 2021-03-07
Update info about UI configuration of integrations

kongo09:
Make explicit that it is standard behaviour that integrationgs - once configured - don't necessarily have a way to change their parameters in the UI.
Follow-up of discussion with @frenck in issue https://github.com/home-assistant/core/issues/47561

## Proposed change

Make documentation more explicit. With more and more integrations getting a UI configuration, it gets increasingly confusing for beginner users to understand what is configured where.

## Type of change

## Additional information

- This PR fixes or closes issue: #47561

## Checklist

### DIFF
--- a/source/_docs/configuration/devices.markdown
+++ b/source/_docs/configuration/devices.markdown
@@ -12,7 +12,9 @@ Classification for the available integrations:
 - [IoT class](/blog/2016/02/12/classifying-the-internet-of-things): Classifier for the device's behavior.
 - [Quality scale](/docs/quality_scale/): Representation of the integration's quality.
 
-Usually every entity needs its own entry in the `configuration.yaml` file. There are two styles for multiple entries:
+Integrations that are automatically discovered, added and configured via the UI typically don't need an extra entry in the `configuration.yaml` file. Required parameters are directly entered during setup in the UI. Note, that once configured, it is standard behaviour that configuration paramters of integrations cannot be changed anymore. To change the paramters, delete the integration from the UI, let the system discover it again or add it manually back.
+
+Entites of integrations without an UI configuration usually need their own entry in the `configuration.yaml` file. There are two styles for multiple entries:
 
 ## Style 1: Collect every entity under the "parent"
 
--- a/source/_docs/configuration/devices.markdown
+++ b/source/_docs/configuration/devices.markdown
@@ -12,7 +12,7 @@ Classification for the available integrations:
 - [IoT class](/blog/2016/02/12/classifying-the-internet-of-things): Classifier for the device's behavior.
 - [Quality scale](/docs/quality_scale/): Representation of the integration's quality.
 
-Integrations that are automatically discovered, added and configured via the UI typically don't need an extra entry in the `configuration.yaml` file. Required parameters are directly entered during setup in the UI. Note, that once configured, it is standard behaviour that configuration paramters of integrations cannot be changed anymore. To change the paramters, delete the integration from the UI, let the system discover it again or add it manually back.
+Integrations that are automatically discovered, added and configured via the UI typically don't need an extra entry in the `configuration.yaml` file. Required parameters are directly entered during setup in the UI. Note, that once configured, it is standard behavior that configuration paramters of integrations cannot be changed anymore. To change the paramters, delete the integration from the UI, let the system discover it again or add it manually back.
 
 Entites of integrations without an UI configuration usually need their own entry in the `configuration.yaml` file. There are two styles for multiple entries:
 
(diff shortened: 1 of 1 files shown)


kongo09:
also closes #47393
https://github.com/home-assistant/core/issues/47393

homeassistant:
Hi @kongo09, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#16903).

Thanks!

source/_docs/configuration/devices.markdown
```diff
@@ -12,7 +12,9 @@ Classification for the available integrations:
 - [IoT class](/blog/2016/02/12/classifying-the-internet-of-things): Classifier for the device's behavior.
 - [Quality scale](/docs/quality_scale/): Representation of the integration's quality.
 
-Usually every entity needs its own entry in the `configuration.yaml` file. There are two styles for multiple entries:
+Integrations that are automatically discovered, added and configured via the UI typically don't need an extra entry in the `configuration.yaml` file. Required parameters are directly entered during setup in the UI. Note, that once configured, it is standard behavior that configuration paramters of integrations cannot be changed anymore. To change the paramters, delete the integration from the UI, let the system discover it again or add it manually back.
```
frenck: This statement is partly true.
Some UI integrations still might need additional configuration in configuration.yaml, this would be explained in the integration documentation.

> are automatically discovered, added and configured via the UI

It also applies to integration that are not discovered, but still managed via the UI.

> Required parameters are directly entered during setup in the UI. Note, that once configured, it is standard behavior that configuration paramters of integrations cannot be changed anymore.

This is not universal. for all integrations. Parameters is spelled incorrectly.

frenck (changes requested):

source/_docs/configuration/devices.markdown
kongo09: I've adapted the text to be more precise. The length of the text indicates to me that the system behavior is not simple and easy to follow for the user.

kongo09:
Anything else required?

source/_docs/configuration/devices.markdown
```diff
@@ -12,7 +12,11 @@ Classification for the available integrations:
 - [IoT class](/blog/2016/02/12/classifying-the-internet-of-things): Classifier for the device's behavior.
 - [Quality scale](/docs/quality_scale/): Representation of the integration's quality.
 
-Usually every entity needs its own entry in the `configuration.yaml` file. There are two styles for multiple entries:
+Integrations that are automatically discovered, or manually added and configured via the UI, typically don't need an extra entry in the `configuration.yaml` file. Required parameters are directly entered during setup in the UI. However, please be aware that some UI integrations still might need additional configuration in the `configuration.yaml` file. These cases are explained in the respective integration documentation.
+
+Note, that for integrations configured in the UI, it is standard behavior that configuration parameters cannot be changed anymore. If the integration doesn't offer an UI option for change, delete the integration from the UI, let the system discover it again or add it back manually.
```
frenck, line 17: This information is incorrect. You are describing an import method here, which is only used for deprecation periods (and instructions for these are provided via the release notes).

I suggest removing this section.

```suggestion
```

frenck (changes requested):

source/_docs/configuration/devices.markdown
kongo09: Sorry. I'm only a user trying to contribute by making things clear that are totally confusing from a user perspective. As a user, I have no idea about what an import method is. And I shouldn't really have to. Point is: something apparently is standard behaviour that from a user perspective hides data that in the past (and for many integrations in the present) was visible and configurable. Deleting this paragraph would continue to leave everyone in the dard.

My documentation principle would be to always document current state. Not future desired state as this doesn't exist yet nor past state as this is irrelevant and confusing. Should the current state be one that is supposed to go away, than the documentation needs to evolve with the changing state. Which is what it needs to do anyway all the time.

Please accept the request, close it or modify it as you see fit.

frenck: > As a user, I have no idea about what an import method is. And I shouldn't really have to.

That is correct, hence it doesn't belong here and should be removed.

> My documentation principle would be to always document current state. 

We have the same policy, and thus this is not a normal pattern, it should not be part of the docs.

> Please accept the request, close it or modify it as you see fit.

Ok that is a bit weird thing to say. I made a suggestion and its open for discussion, this feels like you slamming the door?

If so, close the PR. I'm happy to accept contributions, and are absolutely thankful to get those, however, they must be correct.

kongo09: Sorry, I didn't mean to be rude. I guess I'm just lost.

As a user, I got stuck over a thing. You suggested to improve documentation, which is a fair request. I tried but I realize that I still haven't understood it well enough. Which makes the point that the current documentation doesn't work.

I'll close the PR.

PR #16962, opened 2021-03-12
Update systemmonitor.markdown

bessarabov:
There are a lot of users who are using the exact yaml from the page. 

They are checking disk space in /home, when in fact they need to check the space in / For many users the numbers are the same, so they don't understand that they are using it slightly incorrect. 

Here is a GitHub search showing some of that incorrect use cases: https://github.com/search?q=systemmonitor+disk_use_percent+arg+%2Fhome&type=code

I think that the code sample should show the code that suits best for the simplest use case.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/systemmonitor.markdown
+++ b/source/_integrations/systemmonitor.markdown
@@ -23,7 +23,7 @@ sensor:
   - platform: systemmonitor
     resources:
       - type: disk_use_percent
-        arg: /
+        arg: /config
       - type: memory_free
 ```
 
--- a/source/_integrations/systemmonitor.markdown
+++ b/source/_integrations/systemmonitor.markdown
@@ -23,7 +23,7 @@ sensor:
   - platform: systemmonitor
     resources:
       - type: disk_use_percent
-        arg: /home
+        arg: /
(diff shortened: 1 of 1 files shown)


source/_integrations/systemmonitor.markdown
```diff
@@ -23,7 +23,7 @@ sensor:
   - platform: systemmonitor
     resources:
       - type: disk_use_percent
-        arg: /home
+        arg: /
```
frenck: `/` is often incorrect as well, especially inside containers (which is the largest part of our user base). If we want something that works for the largest user group, I would suggest on changing it to `/config`.

```suggestion
        arg: /config
```

frenck (changes requested):

source/_integrations/systemmonitor.markdown
agners: `/` does work with OS since `/` is on the same data disk as `/config`. But for Supervised/Core in Container installation `/config` is better, and works in all cases, so I agree using `/config` is more likely what users want.

frenck (approved):

PR #17063, opened 2021-03-19
Add integration doc for Leviosa_Shades

altersis:
## Type of change

- [X] Added documentation for a new integration I'm adding to Home Assistant (`next` branch).
  - [X] I've opened up a PR to add logo's and icons in [Brands repository](https://github.com/home-assistant/brands).

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/48048
- Link to parent pull request in the Brands repository: https://github.com/home-assistant/brands/pull/2337

## Checklist


altersis:
Hey there. I'm sorry but I don't understand what is happening here. I originally created #17062, using ```current``` as the source branch and ```next``` as the target, then was asked to rebase it on ```dev```. I decided to close #17062, as only one file was added, and I believe what I did produces the same result. the probot is asking for a rebase, but wouldn't that takes us to the original problem?

klaasnicolaas:
Probot never asks you to rebase to `dev` 🤷‍♂️ but the problem is that `next` and `current` are 2 different branches. `current` is for changes independent of the releases, while `next` is always linked to a pull request in the core repository.

Wanting to merge a `current` branch (by means of a PR) in `next` then causes problems. But now it has gone well! 🚀

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,55 @@
...
+
+## Configuration
+
+To Add Leviosa Shades to your Home Assistant instance, please configure your Leviosa Shades and Zones with the app before doing it in Home Assistant. Once done, take the following steps:
+
+- Browse to your Home Assistant instance.
+- In the sidebar click on [Configuration](https://my.home-assistant.io/redirect/config).
+- From the configuration menu select [Integrations](https://my.home-assistant.io/redirect/integrations).
+- In the bottom right, click on the [Add Integration](https://my.home-assistant.io/redirect/config_flow_start?domain=leviosa_shades) button.
+- From the list, search and select *Leviosa Motor Shades Zone*.
```
klaasnicolaas: This part can be replaced with: `{% include integrations/config_flow.md %}`

Maybe also line 27.

klaasnicolaas (changes requested):

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+{% include integrations/config_flow.md %}
+- Each Leviosa Zone manages up to 6 groups of shades.
+- If you have more than one Leviosa Zone in your Network, Home Assistant will display a list, with one IP address for each Leviosa Zone discovered in your network, in the same order as they were discovered. Once a Leviosa Zone is configured in Home Assistant, it will not appear in this list. Refer to the next section to find the IP address of each zone group.
+- If you have only one Leviosa Zone, then you'll be taken directly to the screen to enter the Leviosa Zone name and shade group names.
+- You can name the shade groups as you wish. They do not have to match the names already in the Leviosa App.
+
+After completing the steps above, the {{ name }} integration will be immediately available for use. A `cover` entity for each shade group will be ready for you. An additional `cover` entity is created with the name of the zone (prefixed `all_`) to move all groups in a Leviosa Zone at the same time.
+
+## Retrieving the IP Address of the Leviosa Zone (WiFi bridge)
```
klaasnicolaas: Maybe keeping these titles shorter will also make it a bit more readable.

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+This is needed when multiple Leviosa Zones are detected, to identify each Leviosa Zone:
+
+- Open the Leviosa App.
+- Press/hold on any group name. After 1 second, a new screen will appear.
+- Select the ‘Name’ tab at the top.
+- The screen will now show the IP address of the Leviosa Zone that shade group is on.
+
+## Services that allow you to move your blinds to intermediate positions
```
klaasnicolaas: same

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,47 @@
...
+ha_codeowners:
+  - '@altersis'
+ha_config_flow: true
+ha_platforms:
+  - cover
+---
+
+This integration allows you to control [Leviosa Zone hubs](https://leviosashades.com/products/leviosa-zone) from [Leviosa Motor Shades](https://leviosashades.com/).
+
+{% include integrations/config_flow.md %}
```
klaasnicolaas, line 18:
```suggestion
{% include integrations/config_flow.md %}

```

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,47 @@
...
+- Each Leviosa Zone manages up to 6 groups of shades.
+- If you have more than one Leviosa Zone in your Network, Home Assistant will display a list, with one IP address for each Leviosa Zone discovered in your network, in the same order as they were discovered. Once a Leviosa Zone is configured in Home Assistant, it will not appear in this list. Refer to the next section to find the IP address of each zone group.
+- If you have only one Leviosa Zone, then you'll be taken directly to the screen to enter the Leviosa Zone name and shade group names.
+- You can name the shade groups as you wish. They do not have to match the names already in the Leviosa App.
+
+After completing the steps above, the {{ name }} integration will be immediately available for use. A `cover` entity for each shade group will be ready for you. An additional `cover` entity is created with the name of the zone (prefixed `all_`) to move all groups in a Leviosa Zone at the same time.
```
klaasnicolaas:
```suggestion
A `cover` entity for each shade group will be ready for you. An additional `cover` entity is created with the name of the zone (prefixed `all_`) to move all groups in a Leviosa Zone at the same time.
```

Maybe move this to after the config_flow block

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,47 @@
...
+  - '@altersis'
+ha_config_flow: true
+ha_platforms:
+  - cover
+---
+
+This integration allows you to control [Leviosa Zone hubs](https://leviosashades.com/products/leviosa-zone) from [Leviosa Motor Shades](https://leviosashades.com/).
+
+{% include integrations/config_flow.md %}
+- Each Leviosa Zone manages up to 6 groups of shades.
```
klaasnicolaas, line 22: Possibly a title above this makes it a bit clearer, what you read underneath.

Considering it's a bit of a "keep in mind" piece 😄

klaasnicolaas (changes requested):

altersis:
Changes applied as requested. Thanks for your help!

altersis (commented):
Thanks for your help! I hope all is ok now.

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+{% include integrations/config_flow.md %}
+
+## Things to keep in mind
+
+- Each Leviosa Zone manages up to 6 groups of shades.
+- If you have more than one Leviosa Zone in your Network, Home Assistant will display a list, with one IP address for each Leviosa Zone discovered, in the same order as they were discovered. Once a Leviosa Zone is configured in Home Assistant, it will not appear in this list. Refer to the next section to find the IP address of each zone group.
+- If you have only one Leviosa Zone, then you'll be taken directly to the screen to enter the Leviosa Zone name and shade group names.
+- You can name the shade groups as you wish. They do not have to match the names already in the Leviosa App.
+- A `cover` entity for each shade group will be ready for you. An additional `cover` entity is created with the name of the zone (prefixed `all_`) to move all groups in a Leviosa Zone at the same time.
```
frenck, line 26: The additional `all_` group/entity should not be created. It is not the responsibility of integration to provide that.

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,50 @@
...
+When you have multiple Leviosa Zones, you'll need to know which IP address is associated with each Leviosa Zone hub. The easiest way to do this is to:
+
+- Open the Leviosa App.
+- Press and hold on any shade group name. After 1 second, a new screen will appear.
+- Select the ‘Name’ tab at the top.
+- The screen will show the IP address of the Leviosa Zone that shade group is on.
+
+## Services to move your blinds to intermediate positions
+
+When you want your blinds to go to intermediate positions defined by you, set these intermediate positions with the [Leviosa remote controller](https://cdn.shopify.com/s/files/1/1346/0347/files/Leviosa_Shades_Programming_instructions.pdf?). These intermediate positions are stored in each Leviosa shade motor, and two services in Home Assistant tell shades in each group to move there. If you do not program intermediate points, the shade will fully travel to the upper or lower limit when these services are used. Once the intermediate positions are set, you can:
```
frenck:
```suggestion
When you want your blinds to go to intermediate positions defined by you, set these intermediate positions with the [Leviosa remote controller](https://cdn.shopify.com/s/files/1/1346/0347/files/Leviosa_Shades_Programming_instructions.pdf?). These intermediate positions are stored in each Leviosa shade motor, and two services in Home Assistant tell shades in each group to move there. If you do not program intermediate points, the shade willfully travels to the upper or lower limit when these services are used. Once the intermediate positions are set, you can:
```

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+- Use `leviosa_shades.next_up_pos` to move up to the next intermediate position.
+- Use `leviosa_shades.next_down_pos` to move down to the next intermediate position.
+
+Both services need the `entity_id` of the shade group that you want to move
```
frenck: That seems weird? This should be an entity service, which can accept a device or area as well. This comment makes me think it is not implemented correctly.

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+## Troubleshooting
+
+- You cannot use the Home Assistant UI to move your blinds to arbitrary positions, only to completely open/close them. Use the services explained above to move to your pre-set intermedate positions.
```
frenck:
```suggestion
- You cannot use the Home Assistant UI to move your blinds to arbitrary positions, only to completely open/close them. Use the services explained above to move to your pre-set intermediate positions.
```

source/_integrations/leviosa_shades.markdown
```diff
@@ -0,0 +1,50 @@
...
+- The services above will only fully open or close the blinds if you do not set intermediate positions using a Leviosa remote.
+- If Home Assistant cannot find your Leviosa Zone hubs, double check that you can ping your Home Assistant instance as well as your Leviosa Zones from the same node in your network, and the all nodes are on the same subnet. If you have a network switch that can filter multicast packets, ensure it is not filtering SSDP traffic. Home Assistant can only discover your Leviosa Zone hubs if the SSDP advertisements can reach the Home Assistant instance.
```
frenck:
```suggestion
- If Home Assistant cannot find your Leviosa Zone hubs, double-check that you can ping your Home Assistant instance as well as your Leviosa Zones from the same node in your network, and all nodes are on the same subnet. If you have a network switch that can filter multicast packets, ensure it is not filtering SSDP traffic. Home Assistant can only discover your Leviosa Zone hubs if the SSDP advertisements can reach the Home Assistant instance.
```

source/_integrations/leviosa_shades.markdown
frenck: Interesting though, Home Assistant does support positional opening/closing.

<https://developers.home-assistant.io/docs/core/entity/cover#set-cover-position>

frenck (changes requested):

source/_integrations/leviosa_shades.markdown
altersis: Hi! thank you for taking a look at this. I would like to clarify that the Leviosa Zone can handle 6 shade groups, plus a special group (called Group 0 in the Leviosa Zone hub) that allows moving all Blind groups up or down. The additional shade group 'all' is just a way to expose this Group 0 functionality to the user, otherwise there would not be a way to use it. If after this explanation this appears incorrect, could you please help me to understand how the Group 0 should be exposed?

frenck: I think it should not be exposed? They can be grouped in Home Assistant if one wishes to do so.

source/_integrations/leviosa_shades.markdown
altersis: Hi. Thanks again for taking a look at this. I'll try to explain in more detail: The service is in fact an entity service. In the Leviosa Zone one channel typically represents all the shades you have in one room.  Imagine that we have a living room with five shades, and all of them go up or down at the same time. All five shades would be set by the installer under channel one of the Leviosa Zone; that channel is represented in HASS using one entity. When configuring the Leviosa Zone through the UI, the user would enter the name 'Family room' for channel one. Calling either service with 'cover.family_room' will make all shades move to the next up/down position (these positions are programmed into the motor, so HASS can only tell the channel to go up/down to the next  position programmed).  This is why, at least in my eyes there is no need to pass an area to the service. Makes sense?

frenck: Sure but if it uses an entity service, this statement is untrue, thus should be reworded or removed.

source/_integrations/leviosa_shades.markdown
altersis: I see your point, and I agree that all shade groups could be grouped; that would produce a different  user experience. I'll explain: Group 0 commands all shades to move at exactly the same time, using only one RF command over the air. All motors react to this command immediately, as they all belong to Group 0 in addition to their configured group. If I understand HASS grouping correctly, HASS would send a command to each Group in the Leviosa Zone in rapid succession, which would start raising the shades at different times, possibly with a different order each time. This happens because the Zone would send multiple RF commands over the air, one for each  shade group (RF transmission is slow, so it takes about 100 ms per command). I'm sure this behavior would lower the 'wow' factor that is expected from all blinds reacting at exactly the same time. Hope this clarifies why the 'all_' shade group is important for these blinds.

source/_integrations/leviosa_shades.markdown
altersis: You say that because it appears to imply that HASS groups would not work? if that is the case I know they would work and would gladly change the wording to indicate that you can pass a HASS group or one or more cover entities. Just want to make sure that I address the point of your comment.

frenck: No I say that because this line:

> Both services need the `entity_id` of the shade group that you want to move

Is simply not true when using entity services. An area or device can be provided alternatively to those service calls as part of their target. This line should be removed.

altersis: Thanks for your patience. I think I understand where the problem is, and agree to remove the line. I can now see that the nomenclature used by the product (Leviosa Zone, which controls one or more Groups of Shades) can create some confusion with the Hass concepts of Zone and Group. Will correct the document in a moment.

frenck (approved):

Not sure about the group thing, but that is something for the code PR to further find out.

frenck:
Closing PR as the parent PR has been closed.

PR #17178, opened 2021-03-30
Add note about Yeelight removing control

robbiet480:
What the note says.


robbiet480:
Closing as we issued an alert instead

PR #17317, opened 2021-04-07
Readd sentence about URL change in devolo Home Control

Shutgun:
## Proposed change

With #16508 we lost a sentence about changing the URL provided with the advanced settings. This PR readds it.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/devolo_home_control.markdown
+++ b/source/_integrations/devolo_home_control.markdown
@@ -29,6 +29,8 @@ devolo Home Control is a Z-Wave ecosystem with a Z-Wave to IP gateway in the cen
 
 {% include integrations/config_flow.md %}
 
+Please do not change the URL provided in the advanced mode unless you know what you are doing.
+
 ## Switches
 
 The integration provides support for the following Z-Wave devices:


source/_integrations/devolo_home_control.markdown
```diff
@@ -29,6 +29,8 @@ devolo Home Control is a Z-Wave ecosystem with a Z-Wave to IP gateway in the cen
 
 {% include integrations/config_flow.md %}
 
+Please do not change the URL provided in the advanced mode unless you know what you are doing.
```
frenck, line 32: Why is that URL even there?

frenck (changes requested):

source/_integrations/devolo_home_control.markdown
Shutgun: Hi @frenck ,

I hear two flavours in you question and don't know, which one you are aiming at, so I'm simply gonna answer them both.

Why does an integration with IoT class "Local Push" need a publicly available URL? 
We ask the user for his credentials. However, the credentials he knows don't fit the credentials the devolo central unit uses for local connections. They need to be translated and we ask "my devolo" (the devolo cloud) for that. If the central unit is discovered via Zeroconf, my devolo is not used for anything afterwards. If the central unit is not discoverable via Zeroconf, the integration can seamlessly work as "Cloud Push", too.

Why is that URL configurable?
There are different stages of my devolo, representing different development states. The normal user will be on production and therefore use the predefined URL. However, I know about devolo employees using this integration but being on a different stage with a different URL. In our first shot we wanted to hide that URL in a configuration YAML and merge it with the UI configuration, but that also violated ADR0010, so it became an advanced setting. 

Still the devolo support was contacted by people messing up the URL. They can't help, as the integration is unofficial, so I wanted to add that sentence.

frenck: > However, I know about devolo employees using this integration but being on a different stage with a different URL. 

Looking at all history, that has not been made clear. This configuration should not be user-facing ​at any time and thus not part of the configuration YAML nor part configuration flow.

This needs to be removed completely, making this documentation PR obsolete.

Shutgun: The history is a bit hidden here: https://github.com/home-assistant/core/pull/33181#discussion_r411104710

Can you point me in any other direction how to make the backend URL configurable? I would hate to loose those users as from Home Assistants point of view, they are normal users and for me they detect changes on devolos side I need to adapt the integration to before they hit the majority of users.

frenck: I guess we have different opinions on the org about this. IMHO, it doesn't belong here at all.
If an option isn't valuable for the larger part of the users, it should not be added.

Going to merge in this change though, as it might prevent confusion with the current state at least.

frenck (approved):

PR #17520, opened 2021-04-19
Note motion sensor states are not re-triggered

deviantintegral:
## Proposed change

Some devices update their status periodically, such as motion sensors re-detecting motion. Relying on that behaviour is a common pattern in automations, and group entities eating those changes can be confusing.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/group.markdown
+++ b/source/_integrations/group.markdown
@@ -100,5 +100,9 @@ and `off`, the group state will be `on` or `off`.
 It is possible to create a group that the system cannot calculate a group state.
 Groups with entities from unsupported domains will always have an unknown state.
 
+Devices like motion sensors that re-trigger their state change periodically will not be
+bubbled up to the group. For automations with timeouts that need to re-trigger, like motion
+sensor lights, instead create multiple triggers in the automation.
+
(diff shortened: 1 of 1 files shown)


source/_integrations/group.markdown
```diff
@@ -100,5 +100,9 @@ and `off`, the group state will be `on` or `off`.
 It is possible to create a group that the system cannot calculate a group state.
 Groups with entities from unsupported domains will always have an unknown state.
 
+Devices like motion sensors that re-trigger their state change periodically will not be
+bubbled up to the group. For automations with timeouts that need to re-trigger, like motion
+sensor lights, instead create multiple triggers in the automation.
```
frenck, line 105: This is not specific to motion sensors, or groups at all.
If the state doesn't change, it doesn't trigger...

That applies to everything in Home Assistant, I think, therefore, this is incorrect and might cause confusion on being an exception.

deviantintegral: Yeah, looks like I misunderstood this. I just discovered this is actually an "feature" of the motion sensor themselves, where they bind to the on / off zigbee attribute instead of the occupancy attribute. Thanks for the review.

PR #17722, opened 2021-05-06
Encourage users to configure a prometheus metric namespace

knyar:
## Proposed change

Default empty namespace makes it harder to distinguish Home Assistant metrics from other metrics reported to the same Prometheus server.

Setting up a non-empty namespace as the default will break metric continuity for most users, which will be very disruptive. However, encouraging users to configure a non-empty namespace will make experience better for new users.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/prometheus.markdown
+++ b/source/_integrations/prometheus.markdown
@@ -19,6 +19,7 @@ To use the `prometheus` integration in your installation, add the following to y
 ```yaml
 # Example configuration.yaml entry
 prometheus:
+  namespace: hass
 ```
 
 {% configuration %}


knyar:
cc  @mweinelt

mweinelt:
Agreed, the only thing I have configured for prometheus is the namespace, as it keeps things easily findable and prevents unneccesary cluttering of the prometheus search view.

mweinelt (approved):

source/_integrations/prometheus.markdown
```diff
@@ -19,6 +19,7 @@ To use the `prometheus` integration in your installation, add the following to y
 ```yaml
 prometheus:
+  namespace: hass
```
frenck, line 22: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample.

I this should have a sane default, I suggest we adjust the integration for this (although that will be quite a breaking change).

Alternatively, a more extensive example can be defined below.

frenck (changes requested):

knyar:
> I this should have a sane default, I suggest we adjust the integration for this (although that will be quite a breaking change).

Sounds good! This was actually what @mweinelt originally proposed in https://github.com/home-assistant/architecture/issues/391#issuecomment-657249045.

I will prepare a separate PR to introduce a default namespace (or combine it with https://github.com/home-assistant/core/pull/50156 that has a few other naming changes).

frenck:
I really think a separate example might be better, considering the breaking nature

knyar:
We already have a [full example](https://www.home-assistant.io/integrations/prometheus/#full-example) documented that covers `namespace`, but I think most users close the page before reaching that.

PR #17736, opened 2021-05-07
Document hard coded limitations

blakadder:
## Proposed change

Add warning for limited use using the UI.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/number.mqtt.markdown
+++ b/source/_integrations/number.mqtt.markdown
@@ -10,6 +10,13 @@ ha_domain: mqtt
 
 The `mqtt` Number platform allows you to integrate devices that might expose configuration options through MQTT into Home Assistant as a Number. Every time a message under the `topic` in the configuration is received, the number entity will be updated in Home Assisant and vice-versa, keeping the device and Home Assistant in-sync.
 
+
+<div class='note warning'>
+
+MQTT Number entities have a hard coded values: min = 1, max = 100 and step = 1, meaning you can only set values below 1 and above 100 using the UI with increments of 1 only.
+
+</div>
+
 ## Configuration
 
 To enable MQTT Number in your installation, add the following to your `configuration.yaml` file:


source/_integrations/number.mqtt.markdown
```diff
@@ -10,6 +10,13 @@ ha_domain: mqtt
 
 The `mqtt` Number platform allows you to integrate devices that might expose configuration options through MQTT into Home Assistant as a Number. Every time a message under the `topic` in the configuration is received, the number entity will be updated in Home Assisant and vice-versa, keeping the device and Home Assistant in-sync.
 
+
+<div class='note warning'>
+
+MQTT Number entities have a hard coded values: min = 1, max = 100 and step = 1, meaning you can only set values below 1 and above 100 using the UI with increments of 1 only.
```
frenck, line 16: Isn't this a bug?

frenck (changes requested):

source/_integrations/number.mqtt.markdown
blakadder: Possibly but Discord chat wasn't that helpful in revealing if it is a bug or a design choice.

MQTT number in general has a few design limitations, like not having templating for values or ability to set min,max,step or unit of measurement like input_number

blakadder: after some soul searching i came to realize it is a feature as described in https://developers.home-assistant.io/docs/core/entity/number 0 and 100 are the defaults and since the MQTT integration has no way of specifying them its locked to those values

frenck: Right, but if it causes issues, it is a bug. I just find it weird to document this way. Especially considering you could now also have a problem of invalid values (e.g., if an entity actually can only accept 0-50, but you enter 70).

I think we should go the bug route first on this one.

blakadder: Gonna go the bug fixing route then

PR #18039, opened 2021-06-01
fix out of date rainbird docs

ex-nerd:
## Proposed change

- Fix a couple missing times that reference `seconds` instead of `minutes` (users following these instructions as they were will get errors; see home-assistant/core#36214)
- Bring example automation yml format up to date.

## Type of change

## Additional information

N/A

## Checklist

### DIFF
--- a/source/_integrations/rainbird.markdown
+++ b/source/_integrations/rainbird.markdown
@@ -75,7 +75,8 @@ More complex configuration using all possible features could look like this exam
 rainbird:
   - host: IP_ADDRESS_OF_MODULE
     password: YOUR_PASSWORD
-    trigger_time: 6
+    trigger_time:
+      minutes: 6
     zones:
       1:
         friendly_name: My zone 1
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @ex-nerd, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#18039).

Thanks!

source/_integrations/rainbird.markdown
```diff
@@ -126,12 +128,11 @@ The service can be used as part of an automation script. For example:
...
     action:
-      service: rainbird.start_irrigation
-      target:
-        entity_id: switch.sprinkler_1
-      data:
-        duration: 5
+      - service: rainbird.start_irrigation
+        data:
+          entity_id: switch.sprinkler_1
+          duration: 5
```
frenck, line 137: This is actually incorrect / a bug in the rainbird integration.
The service here is registered incorrectly in the code base.

I'm not sure if we should "workaround" in documentation or open an issue (or better) a PR to fix it.

ex-nerd: Oh interesting.  This is the values generated by the automation config UI and it seemed to work (though I'll admit I didn't let it run the full 5 minutes and stopped it with the rainbird app)

frenck (approved):
Thanks, @ex-nerd 👍

source/_integrations/rainbird.markdown
frenck: Ooh this change  you are sending in works...
but the issue is that is should work with a target :)

PR #18133, opened 2021-06-08
Add note to clarify requirement for reverse proxy

Icexist:
## Proposed change

HA after 2021.7 will require a trusted reverse proxy IP, but there isn't anywhere that I've seen in the documentation that explicitly states this requirement and how to achieve it.

I see it in the 2021.6 release notes, but the documentation could be clearer.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/http.markdown
+++ b/source/_integrations/http.markdown
@@ -105,6 +105,21 @@ http:
 
 The [Set up encryption using Let's Encrypt](/blog/2015/12/13/setup-encryption-using-lets-encrypt/) blog post gives you details about the encryption of your traffic using free certificates from [Let's Encrypt](https://letsencrypt.org/).
 
+<div class='note'>
+
+  When using a reverse proxy, you will need to enable the `use_x_forwarded_for` and `trusted_proxies` options. Requests from reverse proxies will be blocked after 2021.7 if these options are not set.
+  
+  ```yaml
+  http:
+    use_x_forwarded_for: true
+    trusted_proxies:
+      - 172.30.33.0/24
+      - 127.0.0.1
+      - ::1
+  ```
+
+</div>
+
 ## APIs
 
 On top of the `http` integration is a [REST API](https://developers.home-assistant.io/docs/api/rest), [Python API](https://developers.home-assistant.io/docs/api_lib_index) and [WebSocket API](https://developers.home-assistant.io/docs/api/websocket) available.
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @Icexist, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#18133).

Thanks!

source/_integrations/http.markdown
```diff
@@ -105,6 +105,21 @@ http:
 
 The [Set up encryption using Let's Encrypt](/blog/2015/12/13/setup-encryption-using-lets-encrypt/) blog post gives you details about the encryption of your traffic using free certificates from [Let's Encrypt](https://letsencrypt.org/).
 
+<div class='note'>
```
frenck: This doesn't have to be a note block. Just normal text in a chapter will do.

source/_integrations/http.markdown
```diff
@@ -105,6 +105,21 @@ http:
...
+
+  When using a reverse proxy, you will need to enable the `use_x_forwarded_for` and `trusted_proxies` options. Requests from reverse proxies will be blocked after 2021.7 if these options are not set.
```
frenck: Please remove the version reference, we don't document history, but the current state. For changelogs the release notes are available.

source/_integrations/http.markdown
```diff
@@ -105,6 +105,21 @@ http:
...
+  
+  ```yaml
+  http:
+    use_x_forwarded_for: true
+    trusted_proxies:
+      - 172.30.33.0/24
+      - 127.0.0.1
+      - ::1
```
frenck: Considering users copy & paste in general, I think this is kinda risky.
Accepting whole IP blocks and the local loopback interfaces seems way too wide IMHO.

Do you really have a cluster of reverse proxies? Do you expect others to have that?

frenck (changes requested):

source/_integrations/http.markdown
Icexist: For my setup, the nginx addon is actually at 172.30.33.4, but I included the whole /24 block for easy copy-paste. How could we specify the correct IP? Not sure that all users who would use a reverse proxy (especially an addon) will know the correct IP to state.

Removing the `::1` and `127.0.0.1` seems like a good idea.

frenck: Not setting the correct IP, can become a security risk. For example, when using trusted network authentication, another entity on the network could send in a fake X-Forwarded-For header to get authenticated as any user. This example allows that whole range to do just that.

Icexist: Removing `::1` and `127.0.0.1` and leaving the `172.30.33.0/24` matches the example directly above. I also added a comment that clarifies that the proxy server's IP should be added to the list.

frenck (approved):
Thanks, @Icexist 👍

PR #18404, opened 2021-07-06
Added example of custom sensor for time trigger

robvangeloven:
Time trigger refers to using a sensor with "device_class: timestamp" however I could find no examples of how to do this. I've added an example how such a sensor could be implemented

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -589,6 +589,9 @@ automation:
 ```
 
 An example of a custom [template sensor](/integrations/template/) with the "timestamp" device class that adds 15 minutes to an input_datetime:
+
+{% raw %}
+
 ```yaml
 sensor:
 - unique_id: "phone_next_alarm"
@@ -604,6 +607,9 @@ sensor:
     
     {{ timeaware.isoformat() }}
 ```
+
+
+{% endraw %}
 
 ### Multiple Times
 
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -588,6 +588,23 @@ automation:
           entity_id: light.bedroom
 ```
 
+An example of a custom [template sensor](/integrations/template/) with the "timestamp" device class that adds 15 minutes to an input_datetime:
+```yaml
+sensor:
+- unique_id: "phone_next_alarm"
+  name: "Phone Next Alarm"
+  device_class: timestamp
+  state: >-
+    # This will create a datetime object that has no timezone:
+    {% set time = strptime(now().date() ~ states("input_datetime.phone_next_alarm"), "%Y-%m-%d%H:%M:%S") %}    
+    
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @robvangeloven, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#18404).

Thanks!

source/_docs/automation/trigger.markdown
```diff
@@ -588,6 +588,23 @@ automation:
...
+  device_class: timestamp
+  state: >-
+    # This will create a datetime object that has no timezone:
+    {% set time = strptime(now().date() ~ states("input_datetime.phone_next_alarm"), "%Y-%m-%d%H:%M:%S") %}    
+    
+    # We need a timezone however, else Home Assistant can't compare this sensor with the current time.
+    # By using the astimezone() function we now create a timezone aware datetime object based on the local timezone:
+    {% set timeaware = time.astimezone(now().tzinfo) + timedelta(minutes = 15) %}
+    
+    {{ timeaware.isoformat() }}
```
frenck, line 608: So, a couple of things with this change:

- This is not an automation trigger, which is what this page is about.
- There is not need to make a template sensor to trigger automation earlier.
- The datetime handling is too complex and can be simplified (especially with the date parsing features being released tomorrow).
- CI builds are not passing and the YAML format doesn't match our documentation guidelines.

frenck (changes requested):

source/_docs/automation/trigger.markdown
robvangeloven: - It's an example of a sensor of the device class timestamp, which has no examples or documentation I could find. That can be used to trigger this automation. However I understand your point that it might be put somewhere better. 
- I actually have quite a few automations running based on a time offset from a single datetime input, stuff like wake-up light before my alarm goes off. Also, isn't that why there's a timedelta() included now? Also: I can't work with features that aren't released yet ;-)
- I agree that it's complex, but there is no documentation I could find except in the warning in configuration/templating that it should output ISO 8601: _(e.g., as a sensor entity with device_class: timestamp), you have to ensure that it is the ISO 8601 format (meaning it has the “T” separator between the date and time portion)._ implementing it any other way gave me errors or non-working automation triggers
- The CI builds not passing isn't good. I thought I formatted it according to the documentation guidelines, could you point to what's wrong with the yaml?

So is it worthwhile to create an example somewhere else in the documentation how to create a device_class: timestamp sensor?

frenck: Shouldn't that be part of the template documentation instead? (which I think it already is, as the device_class is documented and linked up with the matching platform documentation).

robvangeloven: Agreed, it feels a bit out of place here. And yes, the new as_timestamp feature makes this a whole lot better! I'll close this PR

PR #18423, opened 2021-07-08
Add description for removing holidays by name

matthewgottlieb:
## Proposed change

The workday integration loads region-specific holidays.  Some of these holidays may not be days off from work, and it is tedious to remove each of these holidays by date, year after year.  The Python module [holidays](https://pypi.org/project/holidays/) supports removal by partial string match, the parent PR exposes that functionality to the integration.  This PR documents the new functionality.

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/52700

## Checklist

### DIFF
--- a/source/_integrations/workday.markdown
+++ b/source/_integrations/workday.markdown
@@ -66,7 +66,7 @@ add_holidays:
   required: false
   type: list
 remove_holidays:
-  description: "Remove holidays (treat holiday as workday). Needs to formatted as `YYYY-MM-DD`."
+  description: "Remove holidays (treat holiday as workday). Can be formatted as `YYYY-MM-DD` or by name for a partial string match (e.g. Thanksgiving)."
   required: false
   type: list
 {% endconfiguration %}
@@ -125,6 +125,20 @@ binary_sensor:
     remove_holidays:
       - '2020-11-26'
       - '2020-12-25'
+```
+
+This example excludes Saturdays, Sundays and holidays. Two holidays are removed by name: Thanksgiving and Christmas Day.
+
+```yaml
+# Example 4 configuration.yaml entry
+binary_sensor:
+  - platform: workday
+    country: US
+    workdays: [mon, tue, wed, thu, fri]
+    excludes: [sat, sun, holiday]
+    remove_holidays:
+      - 'Thanksgiving'
+      - 'Christmas Day'
 ```
 
 ## Automation example


frenck (approved):

source/_integrations/workday.markdown
```diff
@@ -66,7 +66,7 @@ add_holidays:
   required: false
   type: list
 remove_holidays:
-  description: "Remove holidays (treat holiday as workday). Needs to formatted as `YYYY-MM-DD`."
+  description: "Remove holidays (treat holiday as workday). Can be formatted as `YYYY-MM-DD` or by name for a partial string match (e.g. Thanksgiving)."
```
TomBrien, line 69: We should point users to where to find the correctly formatted strings to use. There is no nicely maintained list so I think the best compromise is to point to [the source](https://github.com/dr-prodigy/python-holidays/tree/master/holidays/countries) where they can look in the `.py` file for their country.

TomBrien (changes requested):

source/_integrations/workday.markdown
frenck: I don't agree on sending our end-users to Python source files.

TomBrien: ~What do you suggest as an alternate. `Holidays` is super picky on format and the library doesn't offer a clear list anywhere I could find~

So I think this is the upstream source for the holidays: https://en.wikipedia.org/wiki/List_of_holidays_by_country so perhaps that?

TomBrien: Although I note that for example King's Day in the Netherlands is not translated in the Wikipedia page (presumably is in the Dutch version) but is in `Holidays`

matthewgottlieb: I have to agree that directing users to Python source files isn't the most elegant.  I'd thought the partial string match would be forgiving enough for the majority of cases and debug logs would be available for the rest, but a more accessible reference would be ideal.

I'd planned a future PR (once I got the hang of contributing) to add additional attributes to the workday integration, potentially including the list of "known holidays".  Would this be better?

TomBrien: That sounds good although not sure attributes is the way to do it. Migrating to a config flow and allowing a picker in the UI seems like the best solution long term. Looking back I think the partial matching is going to work well enough and this is already more robust than date-format removal

TomBrien (approved):

PR #18454, opened 2021-07-10
MQTT Fan docs - sorted config and small text corrections

jbouwh:
## Proposed change

MQTT Fan docs update:

- Small text corrections/improvements
- Sorted configuration attributes

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/fan.mqtt.markdown
+++ b/source/_integrations/fan.mqtt.markdown
@@ -270,8 +270,6 @@ In this section you find some real-life examples of how to use this fan.
 The example below shows a full configuration for a MQTT fan using percentage and preset modes.
 There are 10 speeds within the speed range, so  `percentage_step` = 100 / 10 steps = 10.0 %.
 
-{% raw %}
-
 ```yaml
 # Example using percentage based speeds with preset modes configuration.yaml
 fan:
@@ -299,8 +297,6 @@ fan:
     speed_range_min: 1
     speed_range_max: 10
 ```
-
-{% endraw %}
 
 ### Configuration using command templates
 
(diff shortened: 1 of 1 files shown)


source/_integrations/fan.mqtt.markdown
```diff
@@ -270,6 +270,8 @@ In this section you find some real-life examples of how to use this fan.
 The example below shows a full configuration for a MQTT fan using percentage and preset modes.
 There are 10 speeds within the speed range, so  `percentage_step` = 100 / 10 steps = 10.0 %.
 
+{% raw %}
```
frenck: There is no liquid syntax in the YAML example below, is there a reason this was added?

frenck (changes requested):

source/_integrations/fan.mqtt.markdown
jbouwh: It was added to give an example with command templates, that might be referred elsewere. The only thing missing was a correct index.
About `liquid syntax`, my knowlegde lacks a bit. What should be different to get a `liquid syntax`?
If your statement is that this example should not be in here since we have https://www.home-assistant.io/docs/configuration/templating/ then, let me know. I could remove the example too.

frenck: Jekyll (the static site generate) uses liquid syntax, which might be triggered by Jinja2 syntax (as they share similarities). This when using Jinja2 (e.g., templates) we need to use the raw tags. However, this example does not have templating, thus I don't see why the raw is needed here.

jbouwh: Ah, without the raw tag, the example below is not showing correct. The first example could do without raw tags. I added them for consistency. I will remove those.

frenck (approved):
Thanks, @jbouwh 👍

PR #18696, opened 2021-08-02
More numeric_state attribute/value_template docs

slovdahl:
## Proposed change

Adds a bit more documentation for the `attribute` and `value_template` properties for the `numeric_state` trigger. 

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -173,7 +173,7 @@ automation:
 
 ## Numeric state trigger
 
-Fires when the numeric value of an entity's state (or attribute's value if using the `attribute` property) **crosses** (and only when crossing) a given threshold. On state change of a specified entity, attempts to parse the state as a number and fires if the value is changing from above to below or from below to above the given threshold.
+Fires when the numeric value of an entity's state (or attribute's value if using the `attribute` property, or the calculated value if using the `value_template` property) **crosses** (and only when crossing) a given threshold. On state change of a specified entity, attempts to parse the state as a number and fires if the value is changing from above to below or from below to above the given threshold.
 
 {% raw %}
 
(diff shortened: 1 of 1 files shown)


source/_docs/automation/trigger.markdown
```diff
@@ -182,13 +182,13 @@ automation:
   trigger:
     - platform: numeric_state
       entity_id: sensor.temperature
-      # Optional
-      value_template: "{{ state.attributes.battery }}"
+      # If given, will trigger when the value of the given attribute for the given entity changes..
+      attribute: attribute_name
+      # ..or alternatively, will trigger when the value given by this evaluated template changes.
+      value_template: "{{ state.attributes.value - 5 }}"
```
frenck, line 188: Actually, this one should be removed.

> the use of templates should be avoided if a pure YAML version is available.

<https://developers.home-assistant.io/docs/documenting/yaml-style-guide#templates>

source/_docs/automation/trigger.markdown
```diff
@@ -198,6 +198,36 @@ automation:
...
+automation:
+  trigger:
+    - platform: numeric_state
+      entity_id: climate.kitchen
+      value_template: "{{ state.attributes.current_temperature - state.attributes.temperature_set_point }}"
+      above: 3
+```
+
+{% endraw %}
+
```
frenck, line 230: This should be removed.

---

PS: Never use `state.attributes.x` in your automation or templates.

image

frenck (changes requested):

source/_docs/automation/trigger.markdown
slovdahl: Hm, but don't they serve different purposes? The `value_template` property can do much more than the `attribute` property. If only a pure attribute is wanted, one should of course use `attribute`, but `attribute` doesn't work for the more advanced case I added below this one (calculating a difference between two attributes of the same entity).

Though I can agree that using a trivial example like this one (`- 5`) doesn't make sense, because `attribute` plus adjusting `above` / `below` works too.

Or am I missing something here?

source/_docs/automation/trigger.markdown
slovdahl: Is `state` as bad as `states`? 🤔 I'm far from an expert here, so please correct me if I misunderstand anything. Wouldn't the `state` variable always be available in this specific case? Because it's bound to the entity in the trigger.

Presuming it is bad, what's the alternative? Is `state_attr` even available in this context? Reading https://www.home-assistant.io/docs/configuration/templating/#limited-templates hints that it might not be.

source/_docs/automation/trigger.markdown
frenck: yes, they do, but this part of the documentation is about triggers, not about templates. We should avoid templating when possible. In this case, it isn't needed to talk about a template to explain what a trigger is.

source/_docs/automation/trigger.markdown
frenck: This is not a limited template.

source/_docs/automation/trigger.markdown
slovdahl: Ok, makes sense to some degree. So where should `value_template` be documented then? 🤔

frenck: Oh darn, you are right... I'm sorry this is the page listing all triggers types 🤦 
This is actually the correct place.

frenck (approved):
Thanks for the PR @slovdahl 👍 

Sorry, I was confused about the page being edited here. Thanks for the addition 👍

slovdahl:
> Thanks for the PR @slovdahl 👍
> 
> Sorry, I was confused about the page being edited here. Thanks for the addition 👍

No problem, thanks for the review 👍😊

PR #18865, opened 2021-08-09
Update zha.markdown with note about changing Zigbee channel

Hedda:
## Proposed change

Update zha.markdown with a note about how changing Zigbee channel will require the user to re-pair devices in ZHA.

## Type of change

## Additional information

- Link to parent pull request in the codebase: NA
- Link to parent pull request in the Brands repository: NA
- This PR fixes or closes issue: NA

## Checklist

### DIFF
--- a/source/_integrations/zha.markdown
+++ b/source/_integrations/zha.markdown
@@ -211,7 +211,8 @@ Note: `cluster_id: 25` may also be `cluster_id: 0x0019`. The two are synonymous.
 
 ### Defining Zigbee channel to use
 
-ZHA prefers to use Zigbee channel 15 by default. You can change this using YAML configuration, however, note that changing the channel requires repairing all devices.
+ZHA prefers to use Zigbee channel 15 by default. You can change this using YAML configuration, but this only works
+if there's no existing network. To change the channel for an existing channel, radio has to be factory reset and a new network to be formed. This requires re-pairing of all the devices.
 
 ```yaml
 zha:
--- a/source/_integrations/zha.markdown
+++ b/source/_integrations/zha.markdown
@@ -222,7 +222,7 @@ zha:
       channels: [15, 20, 25]  # Channel mask
 ```
 
-Tip: Reference this for channel selection [Zigbee and WiFi coexistence (using non-overlapping frequency-bands)](https://support.metageek.com/hc/en-us/articles/203845040-ZigBee-and-WiFi-Coexistence).
+This is a good reference for channel selection [Zigbee and Wifi Coexistance](https://support.metageek.com/hc/en-us/articles/203845040-ZigBee-and-WiFi-Coexistence)
 
 ### Modifying the device type
 
--- a/source/_integrations/zha.markdown
+++ b/source/_integrations/zha.markdown
@@ -211,7 +211,7 @@ Note: `cluster_id: 25` may also be `cluster_id: 0x0019`. The two are synonymous.
 
 ### Defining Zigbee channel to use
 
-Zigbee network channel changes 15 will be used by default by ZHA but the Zigbee network channel can be changed via YAML, however note that changing Zigbee network channel changes requires repairing all already devices paired to ZHA.
+ZHA prefers to use Zigbee channel 15 by default. You can change this using YAML configuration, however, note that changing the channel requires repairing all devices.
 
 ```yaml
 zha:
--- a/source/_integrations/zha.markdown
+++ b/source/_integrations/zha.markdown
@@ -212,7 +212,7 @@ Note: `cluster_id: 25` may also be `cluster_id: 0x0019`. The two are synonymous.
 ### Defining Zigbee channel to use
 
 ZHA prefers to use Zigbee channel 15 by default. You can change this using YAML configuration, but this only works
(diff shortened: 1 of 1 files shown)


martin3000 (dismissed):

Hedda:
Not stale.

Adminiuga:
The information in this PR is incorrect.  Zha does not change the channel after the network was formed. It only tells what channel to prefer for network formation for "out of the box" radios which don't have any network.

Hedda:
> The information in this PR is incorrect. Zha does not change the channel after the network was formed. It only tells what channel to prefer for network formation for "out of the box" radios which don't have any network.

So would you prefer changing that to something like this instead:

_**Zigbee network channel is set to prefer channel 15 by default by ZHA but you can change this via YAML, however note that changing prefered Zigbee network channel changes requires repairing all already devices paired to ZHA if you want them to try use that new channel instead.**_

source/_integrations/zha.markdown
```diff
@@ -211,6 +211,8 @@ Note: `cluster_id: 25` may also be `cluster_id: 0x0019`. The two are synonymous.
 
 ### Defining Zigbee channel to use
 
+Zigbee network channel changes 15 will be used by default by ZHA but the Zigbee network channel can be changed via YAML, however note that changing Zigbee network channel changes requires repairing all already devices paired to ZHA.
```
frenck:
```suggestion
ZHA prefers to use Zigbee channel 15 by default. You can change this using YAML configuration, however, note that changing the channel requires repairing all devices.
```

source/_integrations/zha.markdown
```diff
@@ -219,7 +221,7 @@ zha:
       channels: [15, 20, 25]  # Channel mask
 ```
 
-This is a good reference for channel selection [Zigbee and Wifi Coexistance](https://support.metageek.com/hc/en-us/articles/203845040-ZigBee-and-WiFi-Coexistence)
+Tip: Reference this for channel selection [Zigbee and WiFi coexistence (using non-overlapping frequency-bands)](https://support.metageek.com/hc/en-us/articles/203845040-ZigBee-and-WiFi-Coexistence).
```
frenck: I don't think this is an improvement, and quite an unneeded change. Please revert it.

frenck (changes requested):

source/_integrations/zha.markdown
martin3000: "repairing all already devices" ? what is this?

Adminiuga: Because currently it requires killing the current network and forming a new one. But here's the challenge: re-forming the network is radio specific and now we run into the same issue when this is now not ZHA integration specific, but radio specific and shouldn't really belong to ZHA integration documentation. 

Right now channel configuration is taken into consideration only for new sticks. if ZHA was configured at least once and already formed the network on the stick, then channel configuration is ignored.

frenck: So, shall we just close this instead?

martin3000:
One day I recognized that my zigbee was on a bad channel. So I wanted to change the channel. I read the chapter  "DEFINING ZIGBEE CHANNEL TO USE", implemented it in my configuration.yaml and it did not work. And I wondered why. So it is necessary to mention that defining a channel only works when you start the first time. Maybe this note should also go into the installation instructions. Because if you use a zigbee channel with a frequency the nearby hospital uses for its devices, you are lost.

source/_integrations/zha.markdown
```diff
@@ -211,6 +211,8 @@ Note: `cluster_id: 25` may also be `cluster_id: 0x0019`. The two are synonymous.
 
 ### Defining Zigbee channel to use
 
+ZHA prefers to use Zigbee channel 15 by default. You can change this using YAML configuration, however, note that changing the channel requires repairing all devices.
```
Adminiuga:
```suggestion
ZHA prefers to use Zigbee channel 15 by default. You can change this using YAML configuration, but this only works
if there's no existing network. To change the channel for an existing channel, radio has to be factory reset and a new network to be formed. This requires re-pairing of all the devices.
```

source/_integrations/zha.markdown
frenck:
```suggestion
This is a good reference for channel selection [Zigbee and Wifi Coexistance](https://support.metageek.com/hc/en-us/articles/203845040-ZigBee-and-WiFi-Coexistence)
```

frenck (dismissed):

source/_integrations/zha.markdown
```diff
@@ -211,6 +211,9 @@ Note: `cluster_id: 25` may also be `cluster_id: 0x0019`. The two are synonymous.
 
 ### Defining Zigbee channel to use
 
+ZHA prefers to use Zigbee channel 15 by default. You can change this using YAML configuration, but this only works
+if there's no existing network. To change the channel for an existing channel, radio has to be factory reset and a new network to be formed. This requires re-pairing of all the devices.
```
Adminiuga: Doh
```suggestion
if there's no existing network. To change the channel for an existing network, radio has to be factory reset and a new network to be formed. This requires re-pairing of all the devices.
```

frenck (approved):

PR #18950, opened 2021-08-16
add more specifics to Boolean Warning

Mariusthvdb:
because quoting is only needed when in fact using boolean values

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -422,7 +422,7 @@ The `for` template(s) will be evaluated when an entity changes as specified.
 
 <div class='note warning'>
 
-Use quotes around your values for `from` and `to` to avoid the YAML parser from interpreting values as booleans.
+Use quotes around boolean values (`Y`, `true`, `Yes`, `ON` and `N`, `FALSE`, `No`, `off`) for `from` and `to` to avoid the YAML parser from interpreting values as booleans. See [Booleans](https://www.home-assistant.io/docs/configuration/yaml/#booleans).
 
 </div>
 


source/_docs/automation/trigger.markdown
```diff
@@ -422,7 +422,7 @@ The `for` template(s) will be evaluated when an entity changes as specified.
 
 <div class='note warning'>
 
-Use quotes around your values for `from` and `to` to avoid the YAML parser from interpreting values as booleans.
+Use quotes around boolean values (`Y`, `true`, `Yes`, `ON` and `N`, `FALSE`, `No`, `off`) for `from` and `to` to avoid the YAML parser from interpreting values as booleans. See [Booleans](https://www.home-assistant.io/docs/configuration/yaml/#booleans).
```
frenck, line 425: The Home Assistant state engine, only works with strings for states. Thus using native boolean values doesn't make sense / is actually wrong.

frenck (changes requested):

source/_docs/automation/trigger.markdown
Mariusthvdb: yes, and thats why they need to be quoted? 
I am not sure I follow what you are saying Frenck, do you want me to use a different terminology? Or do you want me to add that sentence, which I think would be a perfect (and educational) addition.

(I was merely triggered by the fact this warning is displayed under an example using quotes on a non boolean string, thus being confusing itself..)

frenck: > yes, and thats why they need to be quoted?

Yes

> I am not sure I follow what you are saying Frenck, do you want me to use a different terminology?

I'm raising the issue that this change is incorrect IMHO, should not be merged; and giving you the opportunity to discuss it.

Mariusthvdb: right. thanks, right now it says quotes are (always) needed, to prevent the parser to interpret values as booleans. However, this is _only_ necessary when using 'boolean' values. Any other, even home/not_home, do not require quotes.

Since I only took the additional text from another HA documentation page (and I linked that), I am not sure what you see as incorrect at all.

added to that, the examples this Warning block is added to, are using `to: "heating"` and `to: "home"`. Neither of these require quotes, so are confusing to say the least.

frenck: > right. thanks, right now it says quotes are (always) needed, to prevent the parser to interpret values as booleans. However, this is only necessary when using 'boolean' values.

Yes, however, our YAML standards and best practices recommend always quote strings.

Mariusthvdb: well, I'll leave it with you, but the current text simply is incorrect, as it specifically states quoting is  necessary `to avoid the YAML parser from interpreting values as booleans`. If you want users to follow the style guide you should simply state that. 
no matter what, quote to: and from: because its in our style guide.

I wont start a discussion on the used YAML standards and best practices.... this seems to boil down to:

_you need to quote boolean strings to prevent them from being treated as booleans in the to: and from: options, and because of that, you should always quote._

Nothing goes amiss if you don't quote, (except on the boolean strings) but still you have to. ;-)

PR #19040, opened 2021-08-22
remove "military" comment on time format

enapupe:
## Proposed change
24 hour format isn't really military format, it's standard time format for most of the world and default for ISO 8601

## Type of change

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -578,7 +578,7 @@ A string that represents a time to fire on each day. Can be specified as `HH:MM`
 automation:
   - trigger:
     - platform: time
-      # Military time format. This trigger will fire at 3:32 PM
+      # ISO 8601 time format. This trigger will fire at 3:32 PM
       at: "15:32:00"
 ```
 


homeassistant:
Hi @enapupe, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#19040).

Thanks!

frenck:
This change is actually incorrect, as ISO 8601 covers more than time alone. As a matter of fact, this field isn't ISO 8601 compatible.

"military time" is a common term, "24-hour clock" would be another common term.

See: <https://en.wikipedia.org/wiki/24-hour_clock>

Therefore the current wording is debatable, but not wrong. The provided alternative in this PR, however, is ambiguous.

Thanks for willing to contribute 👍

enapupe:
From your wiki link:
>  This system is the most commonly used time notation in the world today,[1] and is used by the international standard ISO 8601.[3]

Anyway, I just thought that military time was the wrong word to use there because it's mostly known as "24h-time digits without punctuation", like 1200 (twelve hundred) means 12:00.

ISO 8601 isn't about just a single date/time format, it defines how date and time information should be formatted and time alone is one of its scopes.

> ISO 8601 uses the 24-hour clock system. As of ISO 8601-1:2019, the basic format is T[hh][mm][ss] and the extended format is T[hh]:[mm]:[ss]. Earlier versions omitted the T (representing time) in both formats.
> So a time might appear as either "T134730" in the basic format or "T13:47:30" in the extended format. ISO 8601-1:2019 allows the T to be omitted in the extended format, as in "13:47:30", but only allows the T to be omitted in the basic format when there is no risk of ambiguity with date expressions.

This means essentially that this field **is** ISO 8601 compatible but **is not** military time format compatible, unless we can actually have `1200` there representing `12:00:00`.

frenck:
`is used by`, which is not the same as equal. As said, ISO 8601 is a wider format (which we use in Home Assistant as well), however, this field is not compliant.

The current wording is debatable, but as said, the suggestion change isn't valid because it isn't compliment with the iso format.

PR #19289, opened 2021-09-10
Add Warning on Exposing PII

ghost:
Add warning on exposing personal identifiable information which is visible to anyone when exposed for remote access.

## Proposed change

Reviewing a Shodan.io search for Home Assistant Instances, I noticed that the "location name:" is exposed (example: https://imgur.com/a/rRovDl7).  If one reviews 1 or 2 pages you will see some Personal Identifiable Information exposed.  There are a few addresses utilized and also some surnames utilized.  Given that through a Shodan search or a IP lookup one can determine the location it is relatively easy to find the location when someone has a unique surname.

On its on this information isn't particularly useful however combined with a security breach of ones instance it would allow a attacker to possibly unlock/open a door and have the location information.  Combined with security/alarm controls exposed via HA this issue gets worse.  

This is a long shot but something I think is worth a warning in the docs.

### DIFF
--- a/source/getting-started/onboarding.markdown
+++ b/source/getting-started/onboarding.markdown
@@ -14,6 +14,12 @@ Set your username and password.
 
 Next, you can enter a name for your home and set your location and unit system. Click "DETECT" to find your location and set your time zone and unit system based on that location. If you'd rather not send your location, you can set these values manually.
 
+<div class='note warning'>
+
+Avoid adding any personal identifiable information (address, surname, etc.) when naming your home.  If you decide to expose your instance to the internet this information is visiable to anyone.
+
+</div>
+
 <p class='img'>
 <img src='/images/getting-started/location.png' />
 Set your location, time zone, and unit system.


source/getting-started/onboarding.markdown
```diff
@@ -14,6 +14,12 @@ Set your username and password.
 
 Next, you can enter a name for your home and set your location and unit system. Click "DETECT" to find your location and set your time zone and unit system based on that location. If you'd rather not send your location, you can set these values manually.
 
+<div class='note warning'>
```
ludeeus, line 17: But default nothing is exposed on the internet.
This is better suited for <https://www.home-assistant.io/docs/configuration/remote/> or even <https://www.home-assistant.io/docs/configuration/securing/>

ghost: I thought about placing it in one of those locations but it seemed better to me to place it in the location where the information would be entered.

I have no issue with the dev team moving the warning to where they see fit but I think it should be somewhere in the docs.

ludeeus:
With <https://github.com/home-assistant/core/pull/56097> now merged, there is no longer a need for this. Thanks anyway :+1:

PR #19431, opened 2021-09-22
Advice using reserved or static IP with Zigbee gateways and bridges for ZHA

Hedda:
## Proposed change

Advice using reserved or static IP with Zigbee gateways and bridges for ZHA. This is also recommended for Tube's Zigbee Gateways:

https://github.com/tube0013/tube_gateways/blob/main/README.md#getting-up-and-running

"_Using a Reserved or static IP Address is strongly advised. Reserve the address in the router or use a static ESPHome build with static ip_"

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/zha.markdown
+++ b/source/_integrations/zha.markdown
@@ -147,12 +147,12 @@ Press `Submit`. The success dialog will appear or an error will be displayed in 
 
 ### ZiGate or Sonoff ZBBridge Devices
 
-If you are use ZiGate or Sonoff ZBBridge you have to use some special usb_path configuration:
+If you are using ZiGate Ethernet/WiFi Gateway, Sonoff ZBBridge, or other network-attached Zigbee adapter you will have to use some special usb_path configuration. It is also strongly advised that you use a reserved or static IP Address is strongly by reserving the IP address in your router and/or configure a static IP address in your Zigbee gateway/bridge.
 
-- ZiGate USB TTL or DIN: `/dev/ttyUSB0` or `auto` to auto discover the zigate
-- PiZigate : `pizigate:/dev/ttyS0`
-- Wifi Zigate : `socket://[IP]:[PORT]` for example `socket://192.168.1.10:9999`
-- Sonoff ZBBridge : `socket://[IP]:[PORT]` for example `socket://192.168.1.11:8888`
+- ZiGate USB, USB-TTL or DIN: `/dev/ttyUSB0` or `auto` to auto discover the zigate
+- ZiGate PiZigate : `pizigate:/dev/ttyS0`
+- ZiGate Ethernet and WiFi Gateways : `socket://[IP]:[PORT]` for example `socket://192.168.1.10:9999`
+- ITead Sonoff ZBBridge : `socket://[IP]:[PORT]` for example `socket://192.168.1.11:8888`
 
 ### Discovery via USB or Zeroconf
 


source/_integrations/zha.markdown
```diff
@@ -147,12 +147,12 @@ Press `Submit`. The success dialog will appear or an error will be displayed in
 
 ### ZiGate or Sonoff ZBBridge Devices
 
-If you are use ZiGate or Sonoff ZBBridge you have to use some special usb_path configuration:
+If you are using ZiGate Ethernet/WiFi Gateway, Sonoff ZBBridge, or other network-attached Zigbee adapter you will have to use some special usb_path configuration. It is also strongly advised that you use a reserved or static IP Address is strongly by reserving the IP address in your router and/or configure a static IP address in your Zigbee gateway/bridge.
```
frenck, line 150: What makes ZHA different compared to any other device/solution that requires this to be highlighted?
IMHO, this is not specific to this integration, additionally, most people won't have a clue what you are talking about in this regard.

frenck (changes requested):

source/_integrations/zha.markdown
Hedda: It is different because Zigbee bridges for ZHA like these  are really only used as network-attached remote serial port proxy adapters, like ser2net, and streaming serial communication like this is extremely sensitive to communication loss. 

So you can not compare this connection to something like MQTT or other protocols that are designed to handle communication loss.

frenck: I don't see how I cannot compare this. Additionally, if a connection fails to a remote, the integration should help to user to update the connection settings by triggering the flow.

That said, I still don't get how this is different from e.g., a Hue hub, or a remote MQTT server or my deCONZ machine, or remote Z-WaveJS2MQTT that is integrated without MQTT but over Z-Wave JS directly...

Therefore, as this is not a special or different case than any other, I don't think this belong in this part of the documentation at all.

Hedda: > That said, I still don't get **how this is different from e.g., a Hue hub, or a remote MQTT server or my deCONZ machine, or remote Z-WaveJS2MQTT that is integrated without MQTT** but over Z-Wave JS directly...

It is very different from all those you mentioned there because with the ZHA integration the hub/gateway software that communicates directly via the microcontroller-unit's serial/UART port is not translated to the other transport protocol.

ZHA connect directly via serial port to the Zigbee, and as far as it knows is does not connect to a remote "hub" but a local serial port. ZHA does not use a client-server model and it is therefore you can not run the ZHA integration backend with a Zigbee adapter on a another remote device  like you can with a Hue hub, or a remote MQTT server or my deCONZ machine.

frenck: > It is very different from all those you mentioned there because with the ZHA integration the hub/gateway software that communicates directly via the microcontroller-unit's serial/UART port is not translated to the other transport protocol.

The protocol itself is not part of the IP communication. Those are the same for all sketched. My above answer stays the same.

Hedda: Please instead compare direct serial connection versus other integration that converts to other APIs and/or protocols:

* Zigbee MCU <-> Serial/UART connection <-> ZHA integration

vs. 

* Zigbee MCU <-> Serial/UART connection<-> Zigbee2MQTT <-> MQTT Broker <-> MQTT Client
* Zigbee MCU <-> Serial/UART connection <-> Philips Hue Hub <-> Philips Hue API <-> Philips Hue API integration
* Zigbee MCU <-> Serial/UART connection <-> deCONZ <-> deCONZ API <-> deCONZ API integration

As you can maybe see, all those convert Zigbee serial protocol to some other API or protocol for their integrations.

Yes same applies to Z-WaveJS2MQTT that is integrated without MQTT but over Z-Wave JS via websocket integration:

* Z-Wave MCU <-> Serial/UART connection <-> node-zwave-js <-> Z-Wave JS Server <-> Z-Wave JS integration

I think it be more fair to compare the ZHA integration with something like RFLink which also uses serial connection:

* 433MHz radio and Arduino with RFLink gateway firmware<-> Serial/UART connection <-> RFLink integration

https://www.home-assistant.io/integrations/rflink/ -> https://www.home-assistant.io/integrations/rflink/#tcp-mode

I believe that if you connect your RFLink Gateway (like Nodo RFLink Gateway) via ser2net then you get the same issues.

There is however a huge difference in how much and type of communication there is between ZHA and Zigbee MCU verses how much communication there is between RFLink and a RFLink Gateway.

frenck: See my above comments.

Hedda: It is still not correct. Connections via serial stream proxies are direct serial connections as servers like ser2net is only a tunnel.

Please consider getting a second opinion.

frenck: I don't need to consider a second opinion. THe integration needs to handle IP changes correct and guide the user in the UI in those cases. This adds a lot of details that might confuse the user and it no different from other situations (e.g., DSMR, and all others I mentioned).

If ZHA doesn't handle the change correct or doesn't guide the user correct on the IP change of the gateway/bridge connecting too, that should be filed as a bug. I'm not willing to accept this addition.

Hedda: > THe integration needs to handle IP changes correct and guide the user in the UI in those cases.

Well the ZHA integration does not handle IP changes for Zigbee coordinators remotely connected via socat today. So bug?

frenck: Well, yes, that should be handled correctly. Network changes happen and should be taken into account.

Adminiuga: Zha integration handles up changes for tubezb coordinator. If user went all the way to setup zha coordinator to use ipserial tunnel, he may as well be smart enough to reserve an IP address without hand holding.

Landrash: I agree with @frenck .
The handling of a IP change is not unique to this integration.

frenck:
Alright closing this as per above review comments.

PR #19476, opened 2021-09-26
Update trigger.markdown

finity69x2:
Add warnings to be considered when using the "for:" option.

## Proposed change

The use of the "for:" option in automation triggers doesn't survive server restarts or automation reloading. This results in unexpected behavior if the user relies on the trigger to occur at the time specified. This is a big reliability concern when used and the user is unaware of this behavior.

If the underlying behavior can't be corrected then at a minimum there should be a warning in the documentation pointing out this limitation especially for time critical automation triggers.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -607,11 +607,9 @@ Templates that do not contain an entity will be rendered once per minute.
 
 <div class='note warning'>
 
-Use of the 'for:' option will not survive Home Assistant server restarts or upon re-loading automations. If this option is used in an automation trigger and the referenced entity has changed to the desired state but the 'for:' time hasn't been reached and Home Assistant is restarted or automations are re-loaded the trigger will never occur.
-  
-For this reason it is not recopmmended to use the 'for:' option for automation triggers that are critical that they actually occur at the time interval specified.
-  
-In those cases it is more reliable to use the automation to set an input_datetime to the desired time and then use that input_datetime as the automastion trigger to perform the desired actions.
+Use of the `for` option will not survive Home Assistant restart or the reload of automations. During restart or reload, automations that were awaiting `for` the trigger to pass, are reset.
+
+If for your use case this is undesired, you could consider using the automation to set an [`input_datetime`](/integrations/input_datetime) to the desired time and then use that [`input_datetime`](/integrations/input_datetime) as an automation trigger to perform the desired actions at the set time.
 
 </div>
 
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -459,11 +459,9 @@ Use quotes around your values for `from` and `to` to avoid the YAML parser from 
 
 <div class='note warning'>
 
-Use of the 'for:' option will not survive Home Assistant server restarts or upon re-loading automations. If this option is used in an automation trigger and the referenced entity has changed to the desired state but the 'for:' time hasn't been reached and Home Assistant is restarted or automations are re-loaded the trigger will never occur.
-  
-For this reason it is not recopmmended to use the 'for:' option for automation triggers that are critical that they actually occur at the time interval specified.
-  
-In those cases it is more reliable to use the automation to set an input_datetime to the desired time and then use that input_datetime as the automastion trigger to perform the desired actions.
+Use of the `for` option will not survive Home Assistant restart or the reload of automations. During restart or reload, automations that were awaiting `for` the trigger to pass, are reset.
+
+If for your use case this is undesired, you could consider using the automation to set an [`input_datetime`](/integrations/input_datetime) to the desired time and then use that [`input_datetime`](/integrations/input_datetime) as an automation trigger to perform the desired actions at the set time.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


source/_docs/automation/trigger.markdown
```diff
@@ -291,6 +291,16 @@ automation:
 
 The `for` template(s) will be evaluated when an entity changes as specified.
 
+<div class='note warning'>
```
frenck, line 295: Its not just this but many other things.

I'm not sure if this needs such an extensive note. It almost looks like you want to say: Don't use `for`. But that would not be our projects recommendation at all.

frenck (changes requested):

source/_docs/automation/trigger.markdown
finity69x2: It's not just a blanket "don't use it" but instead a warning to not use it for critical automation triggers that need to survive restarts/re-loads. I don't see a problem using it as long as the user knows its limitations and can code resiliency into their automations.

I'm not sure how to make it more concise and still convey the message that using the option is not as reliable as one would think from the current documentation. Right now there is no inking at all in the documentation that the time trigger will fail under certain circumstances.

However, I actually am typically recommending that people not use any time based functions (including for:) that are longer than just a few minutes in automations or scripts when the topic gets brought up in the forums for this reason.

And as for the "many other things" this is just the first documentation change to get a feel for what will be accepted before adding warnings to all the other time based functions to give appropriately similar warnings.

frenck: Still think this warning needs to be limited more.

finity69x2: feel free to make a suggestion that limits it but still correctly describes the concern and offers a better alternative.

I can't see a way to do that.

source/_docs/automation/trigger.markdown
```diff
@@ -291,6 +291,16 @@ automation:
...
+
+Use of the 'for:' option will not survive Home Assistant server restarts or upon re-loading automations. If this option is used in an automation trigger and the referenced entity has changed to the desired state but the 'for:' time hasn't been reached and Home Assistant is restarted or automations are re-loaded the trigger will never occur.
+  
+For this reason it is not recopmmended to use the 'for:' option for automation triggers that are critical that they actually occur at the time interval specified.
+  
+In those cases it is more reliable to use the automation to set an input_datetime to the desired time and then use that input_datetime as the automastion trigger to perform the desired actions.
```
frenck: How about:

```suggestion
Use of the `for` option will not survive Home Assistant restart or the reload of automations. During restart or reload, automations that were awaiting `for` the trigger to pass, are reset.

If for your use case this is undesired, you could consider using the automation to set an [`input_datetime`](/integrations/input_datetime) to the desired time and then use that [`input_datetime`](/integrations/input_datetime) as an automation trigger to perform the desired actions at the set time.
```

source/_docs/automation/trigger.markdown
```diff
@@ -426,6 +434,16 @@ Use quotes around your values for `from` and `to` to avoid the YAML parser from
 
 </div>
 
+<div class='note warning'>
+
+Use of the 'for:' option will not survive Home Assistant server restarts or upon re-loading automations. If this option is used in an automation trigger and the referenced entity has changed to the desired state but the 'for:' time hasn't been reached and Home Assistant is restarted or automations are re-loaded the trigger will never occur.
+  
+For this reason it is not recopmmended to use the 'for:' option for automation triggers that are critical that they actually occur at the time interval specified.
+  
+In those cases it is more reliable to use the automation to set an input_datetime to the desired time and then use that input_datetime as the automastion trigger to perform the desired actions.
```
frenck:
```suggestion
Use of the `for` option will not survive Home Assistant restart or the reload of automations. During restart or reload, automations that were awaiting `for` the trigger to pass, are reset.

If for your use case this is undesired, you could consider using the automation to set an [`input_datetime`](/integrations/input_datetime) to the desired time and then use that [`input_datetime`](/integrations/input_datetime) as an automation trigger to perform the desired actions at the set time.
```

source/_docs/automation/trigger.markdown
```diff
@@ -566,6 +584,16 @@ The `for` template(s) will be evaluated when the `value_template` becomes 'true'
 
 Templates that do not contain an entity will be rendered once per minute.
 
+<div class='note warning'>
+
+Use of the 'for:' option will not survive Home Assistant server restarts or upon re-loading automations. If this option is used in an automation trigger and the referenced entity has changed to the desired state but the 'for:' time hasn't been reached and Home Assistant is restarted or automations are re-loaded the trigger will never occur.
+  
+For this reason it is not recopmmended to use the 'for:' option for automation triggers that are critical that they actually occur at the time interval specified.
+  
+In those cases it is more reliable to use the automation to set an input_datetime to the desired time and then use that input_datetime as the automastion trigger to perform the desired actions.
```
frenck:
```suggestion
Use of the `for` option will not survive Home Assistant restart or the reload of automations. During restart or reload, automations that were awaiting `for` the trigger to pass, are reset.

If for your use case this is undesired, you could consider using the automation to set an [`input_datetime`](/integrations/input_datetime) to the desired time and then use that [`input_datetime`](/integrations/input_datetime) as an automation trigger to perform the desired actions at the set time.
```

frenck (approved):

PR #19606, opened 2021-10-05
Added step to delete apt cache after update/install to save space

swedishmike:
This isn't a change to the documenation - this is to make the container built from the Dockerfile as small as possible by deleting the update/download cache. Hence not sure how to fill out the below?

## Type of change

## Additional information

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,8 @@ RUN \
   && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
       ack \
   && echo "en_US UTF-8" > /etc/locale.gen \
-  && locale-gen en_US.UTF-8
+  && locale-gen en_US.UTF-8 \
+  &&  rm -rf /var/lib/apt/lists/*
 
 # Install the specific version of bundler we need
 COPY Gemfile.lock ./


Dockerfile
```diff
@@ -16,8 +16,9 @@ RUN \
   && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
       ack \
   && echo "en_US UTF-8" > /etc/locale.gen \
-  && locale-gen en_US.UTF-8
+  && locale-gen en_US.UTF-8 \
+  &&  rm -rf /var/lib/apt/lists/*
```
frenck, line 20: This is used for a local dev container, not for distribution.

What is the advantage of doing this? This will A) add time to building up the dev container, and B) will criple the dev environment in case someone likes to install/use custom tools.

frenck (changes requested):

Dockerfile
swedishmike: The advantage is to save space on the built image/container. Depending on how often the upstream container is updated this can be quite a lot at times.

The time added is, at least according to my tests, minimal. This is output from building with the original Dockerfile and the updated one, deleting the built image in between:

```
real	0m14.148s
user	0m0.325s
sys	0m0.331s

real	0m13.970s
user	0m0.347s
sys	0m0.344s

```

In my experience this does not cripple the container - you can still add more programs etc after it. It is just the downloaded update packages and packages for installation that gets deleted. All of those have already been installed so are no longer needed.

frenck: > All of those have already been installed so are no longer needed.

This is where it goes wrong. That depends on the developer and personal needs.

IMHO, in these kinds of environments (that are not distributed across the globe) optimizations like these should not be done.

swedishmike: Ok - I see your point. People like myself can adapt the Dockerfile if need be, like for myself. I'll close this PR.

PR #19744, opened 2021-10-13
explain how the light switch platfom generates entity IDs

activescott:
## Proposed change

It took me forever to figure out that the new entity's ID was generated based on the name rather than the previous entity ID. When they're the same I guess this is intuitive but in my case they weren't. I hope this clarification helps others realize how to use this feature more quickly.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/light.switch.markdown
+++ b/source/_integrations/light.switch.markdown
@@ -23,7 +23,15 @@ To enable this platform in your installation, add the following to your
 light:
   - platform: switch
     name: Christmas Tree Lights
-    entity_id: switch.christmas_tree_lights
+    entity_id: switch.christmas_tree_switch
+```
+
+This will create a new light entity. The entity ID for the generated light source will be based on the name defined here. So the ID of the entity defined above will be `light.christmas_tree_lights`. The new entity can then be used with a light platform service as shown in the following example:
+
+```yaml
+service: light.toggle
+target:
+  entity_id: light.christmas_tree_lights
 ```
 
 {% configuration %}


homeassistant:
Hi @activescott, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#19744).

Thanks!

source/_integrations/light.switch.markdown
```diff
@@ -23,7 +23,15 @@ To enable this platform in your installation, add the following to your
 light:
   - platform: switch
     name: Christmas Tree Lights
-    entity_id: switch.christmas_tree_lights
+    entity_id: switch.christmas_tree_switch
+```
+
+This will create a new light entity. The entity ID for the generated light source will be based on the name defined here. So the ID of the entity defined above will be `light.christmas_tree_lights`. The new entity can then be used with a light platform service as shown in the following example:
```
frenck, line 29: This is how entity IDs are generated in general, why is this documented exception? It is no guarantee it will get that name btw...

activescott: I'm open to suggestions, but my general feedback is that it should be more intuitive to understand what to do after creating this light. This is the content that would have helped me know what to do.

I added one of these "light switch platform" entities for a number of switches and assumed it just didn't do anything for /months/ after I had created them. Then I stumbled across a very uniquely named ID that got me curious if it had been there all along. Again, open to solving for that in a different way, but I wanted to at least propose something that would have helped me.

If there is _no guarantee it will get that name_ it isn't even deterministic apparently, so how could it be intuitive? Maybe another approach would be a general topic about entity ID generation and then linking to it from here (and potentially other places where relevant)?

frenck: > If there is no guarantee it will get that name it isn't even deterministic apparently, so how could it be intuitive?

It isn't. All entities (not just the ones from this integration) are based on slugs of their names, if available. If not, it will try some others until it finds one that is available.

One can have multiple Christmas tree lights for example 😉 

As this applies to every entity, I'm not sure / having trouble understanding why it is added here.

activescott: Fair enough. As I explained above, I struggled to know what to do next with this integration. That's why it's here. I think your point that it _just happened to be with this integration_ and could have happened anywhere is fair one, but I imagine this general issue isn't intuitive for others either so I was trying to clear it up. Again, open to other approaches and not defending this particular PR/change, I was just taking a stab at making things better. Some other ideas:

Maybe a brief general statement like _A new entity with a `light.*` ID will be available in home assistant after configuring this platform_? Again open to ideas.

frenck: A new light entity will be available for every light platform you configure, which is the same for all light platforms (or climate platform, switch platform).

activescott: I don't know what else to say. I struggled to figure this one out and generally I think hass could be more intuitive or easier to use. This is what would have helped me, but I defer to your judgement 🤷‍♂️ Let me know how I can help you.

frenck:
The stale notification brought this one up again. Thought about it a bit; but going to friendly decline this addition. Nevertheless, thanks for willing to contribute 👍

activescott:
Thank you for your considerate reply @frenck

PR #19826, opened 2021-10-17
updating utility_meter example to use new template format

cturra:
## Proposed change
updating `utility_meter` example to use new template format introduced in 2021.4. i also took the opportunity to demonstrate setting the appropriate `device_class` .

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/utility_meter.markdown
+++ b/source/_integrations/utility_meter.markdown
@@ -231,17 +231,25 @@ Additionally, you can add template sensors to compute daily and monthly total us
 {% raw %}
 
 ```yaml
-sensor:
-  - platform: template
-    sensors:
-      daily_energy:
-        friendly_name: Daily Energy
-        unit_of_measurement: kWh
-        value_template: "{{ states('sensor.daily_energy_offpeak')|float + states('sensor.daily_energy_peak')|float }}"
-      monthly_energy:
-        friendly_name: Monthly Energy
-        unit_of_measurement: kWh
-        value_template: "{{ states('sensor.monthly_energy_offpeak')|float + states('sensor.monthly_energy_peak')|float }}"
+template:
+  - sensor:
+    - name: 'Daily Energy Total'
+      device_class: energy
+      unit_of_measurement: kWh
+      state: >
+        {% set offpeak = states('sensor.daily_energy_offpeak') | float %}
+        {% set peak = states('sensor.daily_energy_peak') | float %}
+
+        {{ (offpeak + peak) }}
+
+    - name: 'Monthly Energy Total'
+      device_class: energy
+      unit_of_measurement: kWh
+      state: >
+        {% set offpeak = states('sensor.monthly_energy_offpeak') | float %}
+        {% set peak = states('sensor.monthly_energy_peak') | float %}
+
+        {{ (offpeak + peak) }}
 ```
 
 {% endraw %}
--- a/source/_integrations/utility_meter.markdown
+++ b/source/_integrations/utility_meter.markdown
@@ -237,19 +237,19 @@ template:
       device_class: energy
       unit_of_measurement: kWh
       state: >
-        {% set offpeak = states('sensor.daily_energy_offpeak') | float %}
-        {% set peak = states('sensor.daily_energy_peak') | float %}
+        {% set offpeak = states('sensor.daily_energy_offpeak') %}
+        {% set peak = states('sensor.daily_energy_peak') %}
 
-        {{ (offpeak + peak) }}
+        {{ (offpeak + peak) | float }}
 
     - name: 'Monthly Energy Total'
       device_class: energy
       unit_of_measurement: kWh
       state: >
-        {% set offpeak = states('sensor.monthly_energy_offpeak') | float %}
-        {% set peak = states('sensor.monthly_energy_peak') | float %}
+        {% set offpeak = states('sensor.monthly_energy_offpeak') %}
+        {% set peak = states('sensor.monthly_energy_peak') %}
 
-        {{ (offpeak + peak) }}
+        {{ (offpeak + peak) | float }}
 ```
 
 {% endraw %}
--- a/source/_integrations/utility_meter.markdown
+++ b/source/_integrations/utility_meter.markdown
@@ -228,7 +228,7 @@ utility_meter:
 
 Additionally, you can add template sensors to compute daily and monthly total usage. Important note, in these examples,
 we use the `is_number()` [function](/docs/configuration/templating/#numeric-functions-and-filters) to verify the values
-returned from the sensors are numeric. If this evalutes to false no value it returned.
+returned from the sensors are numeric. If this evalutes to false, `None` is returned.
 
 {% raw %}
 
@@ -241,6 +241,8 @@ template:
       state: >
         {% if is_number(states('sensor.daily_energy_offpeak')) and is_number(states('sensor.daily_energy_peak')) %}
           {{ (states('sensor.daily_energy_offpeak') + states('sensor.daily_energy_peak')) | float }}
+        {% else %}
+          None
         {% endif %}
 
     - name: 'Monthly Energy Total'
@@ -249,6 +251,8 @@ template:
       state: >
         {% if is_number(states('sensor.monthly_energy_offpeak')) and is_number(states('sensor.monthly_energy_peak')) %}
           {{ (states('sensor.monthly_energy_offpeak') + states('sensor.monthly_energy_peak')) | float }}
+        {% else %}
+          None
         {% endif %}
 ```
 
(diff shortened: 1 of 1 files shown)


dgomes:
Thank you for your contribution!

I would also advise to change the `float` filter in accordance to https://www.home-assistant.io/docs/configuration/templating/#numeric-functions-and-filters

dgomes:
OK.... my comment was towards address a common situation in which `peak` or `offpeak` values are unavailable and the cast to float results in `0` and not `unavailable`

cturra:
thanks for the feedback @dgomes and i apologize for not getting this right away. hopefully my last commit gets closer to what you're looking for. :metal:

dgomes (dismissed):

cturra:
anything outstanding on this pr to get it across the line? happy to help if there are any action items required for me.

source/_integrations/utility_meter.markdown
```diff
@@ -212,22 +212,30 @@ utility_meter:
     cycle: monthly
 ```
 
-Additionally, you can add template sensors to compute daily and monthly total usage.
+Additionally, you can add template sensors to compute daily and monthly total usage. Important note, in these examples,
+we use a [float filter](https://www.home-assistant.io/docs/configuration/templating/#numeric-functions-and-filters) to
```
frenck:
```suggestion
we use a [float filter](/docs/configuration/templating/#numeric-functions-and-filters) to
```

source/_integrations/utility_meter.markdown
```diff
@@ -212,22 +212,30 @@ utility_meter:
...
-        friendly_name: Monthly Energy
-        unit_of_measurement: kWh
-        value_template: "{{ states('sensor.monthly_energy_offpeak')|float + states('sensor.monthly_energy_peak')|float }}"
+template:
+  - sensor:
+    - name: 'Daily Energy Total'
+      device_class: energy
+      unit_of_measurement: kWh
+      state: >
+        {% set offpeak = float(states('sensor.daily_energy_offpeak'), default='Invalid Number') %}
```
frenck: This is not correct usage of the float default.

The float default can provide an alternative float value in case the value can't be parsed as a float, and is not means to provide an alternative string.

The state of this sensor should be returning an int or float value, otherwise, it should become unknown.

If you want to check if a value is numeric, you can use the `is_number` template method.

frenck (changes requested):

source/_integrations/utility_meter.markdown
cturra: thanks for this feedback @frenck and that makes a lot of sense to me. i want to take a step back here though -- i feel like we're making this example too complicated for a newer user. the intent of my change was to simply update to the new template format. if you both see value in continuing down this path, i'm happy to help, but i'm not convinced it's the best. sorry for being difficult, but this has been on my mind for a bit. /cc: @dgomes

frenck: Well, the fact is, the current example is wrong, so we can't merge it in this state.

cturra: absolutely! no argument there and sorry if i miscommunication my intentions, but that was not what i meant. i was proposing we backed up to the original proposal of simply updating the template format. to clean the slate, i'm happy to close off this pr and propose a new one.. or leave it.

dgomes:
The current example is actually very bad and the cause of some issue reports where one of the source sensors is unavailable causing havoc in the energy panel. Even-though complex I think it is necessary to document it officially, can't we default to "unavailable" @frenck ?

frenck:
> can't we default to "unavailable" @frenck ?

Unavailable is a state where the devices are unreachable, which is something that should not be set from here IMHO.

I think the best approach is to check if the current value is a number else, return no value.

cturra:
... or, should we come up with a different example that still demonstrates something related to the utility-meter integration and is more new  user friendly?

dgomes:
defaulting to 0 is bad (it will cause havoc with statistics) 

I only see as an option adding `unavailable` to the template (not state = unavailable)

cturra:
thanks for your patience with my response here gents. i've refactored the example to confirm `peak` and  `offpeak` values are numeric. if that's true, then i add the values, if not, i return the string: `Invalid Number`

source/_integrations/utility_meter.markdown
cturra: i've refactored the example to use the `is_number()` function in my [last commit](https://github.com/home-assistant/home-assistant.io/pull/19826/commits/3a00e2bd0818f7bd5e20c4e944bd7c974c75a313).

source/_integrations/utility_meter.markdown
```diff
@@ -225,17 +225,21 @@ template:
...
       device_class: energy
       unit_of_measurement: kWh
       state: >
-        {% set offpeak = float(states('sensor.monthly_energy_offpeak'), default='Invalid Number') %}
-        {% set peak = float(states('sensor.monthly_energy_peak'), default='Invalid Number') %}
-        {{ float((offpeak + peak), default='Invalid Number') }}
+        {% if is_number(states('sensor.monthly_energy_offpeak')) and is_number(states('sensor.monthly_energy_peak')) %}
+          {{ (states('sensor.monthly_energy_offpeak') + states('sensor.monthly_energy_peak')) | float }}
+        {% else %}
+          Invalid Number
```
frenck: "Invalid Number" is not a valid state for a sensor with a kWh unit of measurement.

```suggestion
```

source/_integrations/utility_meter.markdown
```diff
@@ -225,17 +225,21 @@ template:
       device_class: energy
       unit_of_measurement: kWh
       state: >
-        {% set offpeak = float(states('sensor.daily_energy_offpeak'), default='Invalid Number') %}
-        {% set peak = float(states('sensor.daily_energy_peak'), default='Invalid Number') %}
-        {{ float((offpeak + peak), default='Invalid Number') }}
+        {% if is_number(states('sensor.daily_energy_offpeak')) and is_number(states('sensor.daily_energy_peak')) %}
+          {{ (states('sensor.daily_energy_offpeak') + states('sensor.daily_energy_peak')) | float }}
+        {% else %}
+          Invalid Number
```
frenck:
```suggestion
```

frenck (changes requested):

source/_integrations/utility_meter.markdown
cturra: setting to `unknown` instead, which i believe is consistent with what home assistant does today. if i've got that wrong, please let me know.

frenck: Yes, that is wrong, `unknown` is a result of the state engine getting `None` value.

cturra: alright. i think it's time for me to bow-out of this pull request. i'm trying to help, but clearly, playing wack-a-mole isn't going to get this done. sorry for not being able to get this across the line.

frenck: ? I have the exact suggestion that was needed, instead, you ignored it and did something else and are now bow-out because playing "wack-a-mole"?

Interesting.

Thanks for trying though.

../Frenck

cturra: maybe i misread - were you suggesting i set the value to `None` instead?

frenck: Nowhere. None isn't a value.

Hence the removal of the else:

image


It returns nothing, thus None.

cturra: i misunderstood the review (github interface). my bad - update incoming that matches.

cturra: oh. i think i found it!

> I think the best approach is to check if the current value is a number else, return no value.

would this be acceptable then? 

```
      state: >
        {% if is_number(states('sensor.daily_energy_offpeak')) and is_number(states('sensor.daily_energy_peak')) %}
          {{ (states('sensor.daily_energy_offpeak') + states('sensor.daily_energy_peak')) | float }}
        {% endif %}
```

frenck: Yes, I think that would return None. As it is processed as a native type. I'm going to test that out. Hold on.

frenck: I guess we are both wrong. It will now return an empty string 😢 

So what happens, is that the result of a Jinja2 template is always a string. That string is evaluated by Python, to find a type of the value that is returned. An empty string, is a... string :(

So is "unknown", however, that is a result of the state engine when there is no state (None).

So, apparently, the trick is to return `None` explicitly. When that is evaluated by Python, it recognizes it, and will actually be a `None` type value.

```suggestion
        {% else %}
          None
```

In some way, thanks! Because this was a refresh course for me even (even though I wrote part of this handling 😟 )

cturra: thanks for sharing that - it'll definitely be useful to understand for me also.

i think because of all the back-and-forth, github is grumpy with these comments being against an outdated commit, so i cannot approve the change through the ui. will follow up with a commit on my end to reflect this suggestion.

frenck (dismissed):
Nice 👍

source/_integrations/utility_meter.markdown
frenck: 👍  Nice!

frenck:
Rebased to PR onto the latest `current` branch to deal with the build error.

frenck (approved):

PR #19933, opened 2021-10-24
Update unit_of_measurement description

joshmcrty:
When a `sensor` has a `unit_of_measurement` attribute, its value is formatted in Lovelace based on the Number Format setting in the user's profile. This updates the description to mention that.

## Type of change

## Additional information

## Checklist


source/_integrations/template.markdown
```diff
@@ -112,7 +112,7 @@ sensor:
       required: true
       type: template
     unit_of_measurement:
-      description: "Defines the units of measurement of the sensor, if any. This will also influence the graphical presentation in the history visualization as a continuous value. Sensors with missing `unit_of_measurement` are showing as discrete values."
+      description: "Defines the units of measurement of the sensor, if any. This will also display the value based on the user profile Number Format setting and influence the graphical presentation in the history visualization as a continuous value. Sensors with missing `unit_of_measurement` are showing as discrete values."
```
frenck: I don't think this is fully correct. Isn't state class part of this as well?

frenck (changes requested):

source/_integrations/template.markdown
joshmcrty: @frenck There is a check for `unit_of_measurement` in [`compute_state_display.ts`](https://github.com/home-assistant/frontend/blob/dev/src/common/entity/compute_state_display.ts#L23) that conditionally applies number formatting. Wherever `computeStateDisplay()` is called to display a template sensor value, having a `unit_of_measurement` will determine whether the value is formatted as a number.

frenck: That sounds incorrect, it should take state class into account as well, so that is bug imho.

joshmcrty: @frenck a sensor value is, by definition, numeric if it has a `unit_of_measurement` and should therefore have number formatting applied, regardless of `state_class`.

The concept of `state_class` was added to core after I added the number formatting feature to the frontend last year. I could also submit a frontend PR that also checks for `state_class` and applies number formatting if it is present. Is that the bug you are thinking of?

frenck: State class is also part of defining if it contains a number or not. This is also used by the frontend to e.g., detect if a graph is shown for a sensor.

IMHO it is a bug that state class is ignored for number formatting in the frontend.

joshmcrty: @bramkragten @spacegaier care to weigh in on this discussion as it relates to the frontend implementation of number formatting? There are a lot of areas in the frontend where `unit_of_measurement` is used to determine whether a value should use number formatting, but nowhere is `state_class` checked that I can find.

I can work on a frontend PR that adds checking for `state_class` throughout the frontend codebase for number formatting purposes if that's what should be done, let me know! But is there ever a case where a sensor _should_ have a `state_class` defined without also having a `unit_of_measurement`? Right now it can technically be done using template sensors which is what this documentation PR is addressing, but perhaps that should throw a warning/error and we continue to reliably use `unit_of_measurement` to determine if a value is numeric?

frenck: > But is there ever a case where a sensor should have a state_class defined without also having a unit_of_measurement

Yes, multiple. Power factor for example.

joshmcrty: My PR to update the frontend to check for `state_class` was merged; I'll update this PR to include that in the documentation as well when I have some free time later this week.

joshmcrty: @frenck I have updated this PR with additional description for `state_class`. Let me know if this needs further changes. Thanks!

frenck:
@joshmcrty Could you look at the CI/Build error?

joshmcrty:
@frenck Build issues are fixed now.

source/_integrations/template.markdown
```diff
@@ -112,12 +112,12 @@ sensor:
...
       required: false
       type: string
       default: None
     state_class:
-      description: The [state_class](https://developers.home-assistant.io/docs/core/entity/sensor#available-state-classes) of the sensor.
+      description: "The [state_class](https://developers.home-assistant.io/docs/core/entity/sensor#available-state-classes) of the sensor. Only possible value currently is `measurement`. Set this if your template sensor represents a measurement of the current value (so not a daily aggregate etc). This will also display the value based on the user profile Number Format setting and influence the graphical presentation in the history visualization as a continuous value."
```
emontnemery: This reverts changes from #20243, please fix that.
Maybe mention that if neither `state_class` nor `unit_of_measurement` is defined, the history view will display discrete values.

source/_integrations/template.markdown
emontnemery: This is not correct, if state_class is defined the sensor will also not be shown as discrete values:
> Sensors with missing `unit_of_measurement` are showing as discrete values.

emontnemery (changes requested):

source/_integrations/template.markdown
joshmcrty: I revised the description so that `unit_of_measurement` and `state_class` mention how they affect Number Format and history visualization without the previous description of `state_class`.

The legacy style sensor can not use `state_class` (throws an error when checking the configuration) so I did not mention it in that description of the `unit_of_measurement`. I revised that description to remove the last sentence that explicitly states it will be shown as discrete values; now it is consistent with the descriptions in the current method to configure template sensors.

emontnemery (approved):

frenck (approved):

PR #19941, opened 2021-10-24
Fix value template for detecting lost connections

selmaohneh:
The current sample did not work. The sensor stayed unavailable, the template editor showed errors. Here is a working version.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/homematic.markdown
+++ b/source/_integrations/homematic.markdown
@@ -493,7 +493,7 @@ binary_sensor:
           - sensor.office_voltage
           - sensor.time
         value_template: >-
-          {{ as_timestamp(now()) - as_timestamp(state_attr('sensor.office_voltage', 'last_changed')) < 600 }}
+          {{(as_timestamp(now())-as_timestamp(states.sensor.office_voltage.last_changed)) < 600}}
 
 automation:
   - alias: "Homematic Reconnect"


homeassistant:
Hi @selmaohneh, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#19941).

Thanks!

source/_integrations/homematic.markdown
```diff
@@ -493,7 +493,7 @@ binary_sensor:
           - sensor.office_voltage
           - sensor.time
         value_template: >-
-          {{ as_timestamp(now()) - as_timestamp(state_attr('sensor.office_voltage', 'last_changed')) < 600 }}
+          {{(as_timestamp(now())-as_timestamp(states.sensor.office_voltage.last_changed)) < 600}}
```
frenck, line 496: We should never access `states.X` directly, this is bad practice and something we highly discourage. If that is needed, we should at least have the proper checks in the template, to prevent accessing a non-existing entity state object.

Also, you've adjusted the example to no longer match our styling guide. Could you please add back the spacing?

frenck (changes requested):

selmaohneh:
Thanks for the review!

Just to prove my point:

image

Accessing `last_changed` seems to be only possible via `states.X`. The templating documentation does this as an example, as well:

image

So is last_changed some kind of special attribute? I guess so since it as not listed as other attributes but on the right side in the developer tools:

image

I want to understand that - adding the needed checks would just fix the symptoms.  I think allowing `state_attr` to work with `last_changed` would be the prefered fix.

Thoughts @frenck ?

frenck:
My thoughts are in my previous review, we should not access the states object directly.

source/_integrations/homematic.markdown
frenck, line 496: This suggestion will do:

```suggestion
          {{ (now() - states.sensor.office_voltage.last_changed).seconds < 600 }}
```

frenck (changes requested):

PR #19970, opened 2021-10-25
edimax update template sensor example

SNoof85:
## Proposed change
Update template sensor example

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/edimax.markdown
+++ b/source/_integrations/edimax.markdown
@@ -50,17 +50,15 @@ Starting with [version 2 of the firmware](https://www.edimax.com/edimax/download
 {% raw %}
 
 ```yaml
-  - platform: template
-    sensors:
-      edimax_current_power:
-        friendly_name: Edimax Current power consumption
-        unit_of_measurement: "W"
-        value_template: "{{ state_attr('switch.edimax_smart_plug',  'current_power_w') | replace('None', 0) }}"
-
-      edimax_total_power:
-        friendly_name: Edimax Accumulated daily power consumption
-        unit_of_measurement: "kWh"
-        value_template: "{{ state_attr('switch.edimax_smart_plug',  'today_energy_kwh') | replace('None', 0) }}"
+template:
+  - sensor:
+    - name: edimax_current_power
+      unit_of_measurement: "W"
+      state: "{{ state_attr('switch.edimax_smart_plug',  'current_power_w') | replace('None', 0) }}
+      
+    - name: edimax_total_power
+      unit_of_measurement: "kWh"
+      state: "{{ state_attr('switch.edimax_smart_plug',  'today_energy_kwh') | replace('None', 0) }}"
 ```
 
 {% endraw %}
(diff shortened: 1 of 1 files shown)


source/_integrations/edimax.markdown
```diff
@@ -50,17 +50,15 @@ Starting with [version 2 of the firmware](https://www.edimax.com/edimax/download
...
-
-      edimax_total_power:
-        friendly_name: Edimax Accumulated daily power consumption
-        unit_of_measurement: "kWh"
-        value_template: "{{ state_attr('switch.edimax_smart_plug',  'today_energy_kwh') | replace('None', 0) }}"
+template:
+  - sensor:
+    - name: edimax_current_power
+      unit_of_measurement: "W"
+      state: "{{ state_attr('switch.edimax_smart_plug',  'current_power_w') | replace('None', 0) }}"
```
frenck: Double-space in the middle.

`None` is weird, that is not a valid state?

source/_integrations/edimax.markdown
```diff
@@ -50,17 +50,15 @@ Starting with [version 2 of the firmware](https://www.edimax.com/edimax/download
...
-        unit_of_measurement: "W"
-        value_template: "{{ state_attr('switch.edimax_smart_plug',  'current_power_w') | replace('None', 0) }}"
-
-      edimax_total_power:
-        friendly_name: Edimax Accumulated daily power consumption
-        unit_of_measurement: "kWh"
-        value_template: "{{ state_attr('switch.edimax_smart_plug',  'today_energy_kwh') | replace('None', 0) }}"
+template:
+  - sensor:
+    - name: edimax_current_power
```
frenck: Please use normal names

frenck (changes requested):

source/_integrations/edimax.markdown
SNoof85: I don't know the integration, I just adapt what was already done. But as far as I can understand this I guess `|float(default=0)` would be better. What do you think ?

frenck: I dunno, I think if it is truly a possible `None`-type value, it would be:

`{{ state_attr('switch.edimax_smart_plug', 'current_power_w') | default(0, true) }}`

SNoof85: https://github.com/home-assistant/core/blob/7024a5d7d95d27e6a818aa09ee08e2d07965c334/homeassistant/components/edimax/switch.py#L41
Could be None then ?

No ?

SNoof85: https://github.com/home-assistant/core/blob/7024a5d7d95d27e6a818aa09ee08e2d07965c334/homeassistant/components/edimax/switch.py#L92
Looks like it can be None.

`|float(default=0)` on `None` is not correct ?

frenck: That would work as well. But attributes are type native, so we get back a number of None. So instead of casting it, we can just default for None values.

SNoof85: OK, as the attributes looks to be `None` possible, let's do like you proposed then ;)

frenck (approved):
Thanks, @SNoof85 👍

PR #20067, opened 2021-10-29
Update nest.markdown to define the expiry period

chrisns:
## Proposed change
the expiry period for a subscription by default is 31 days, which means if the integration is offline for any reason, when you bring it back it won't work as the subscription will have expired

updating the docs to make it clear this should be set to not expire
image

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/nest.markdown
+++ b/source/_integrations/nest.markdown
@@ -184,7 +184,7 @@ and typically looks like `projects/sdm-prod/topics/EXAMPLE`.
 
 1. Lower the message retention duration to be something short (e.g., 10 minutes or under an hour) to avoid a large backlog of updates when Home Assistant is turned off.
 
-1. Select **Never expire** as the *Expiry Period*. To prevent the subscription you're creating being removed if for example your Home Assistant or the integration is offline for a while.
+1. Select **Never expire** as the *Expiry Period* to prevent the subscription you're creating being removed if for example your Home Assistant or the integration is offline for a while.
 
 1. Leave the rest of the defaults and click **Create**.
 
--- a/source/_integrations/nest.markdown
+++ b/source/_integrations/nest.markdown
@@ -184,6 +184,8 @@ and typically looks like `projects/sdm-prod/topics/EXAMPLE`.
 
 1. Lower the message retention duration to be something short (e.g., 10 minutes or under an hour) to avoid a large backlog of updates when Home Assistant is turned off.
 
+1. Select **Never expire** as the *Expiry Period*
+
 1. Leave the rest of the defaults and click **Create**.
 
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @chrisns, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#20067).

Thanks!

allenporter:
Hi, thanks for the update. I think it would make sense to explain why in the docs they should do this since it may not be obvious. (Also message expiration vs subscription expiration might be confusing)

I would like to move to have the integration automatically create the subscription in the future to resolve this type of issue.

chrisns:
> Hi, thanks for the update. I think it would make sense to explain why in the docs they should do this since it may not be obvious. (Also message expiration vs subscription expiration might be confusing)

Is that better?

> I would like to move to have the integration automatically create the subscription in the future to resolve this type of issue.

Sounds helpful but also a bit dangerous to have Home Assistant (or just the BAU service account credentials) able to manage cloud infra (not a BAU activity, done once) as well as just consume what has been created.

Granted In this particular case there **should** be no financial exposure, but it's a precedent to manage carefully before its overused by other integrations.

source/_integrations/nest.markdown
```diff
@@ -184,6 +184,8 @@ and typically looks like `projects/sdm-prod/topics/EXAMPLE`.
 
 1. Lower the message retention duration to be something short (e.g., 10 minutes or under an hour) to avoid a large backlog of updates when Home Assistant is turned off.
 
+1. Select **Never expire** as the *Expiry Period*. To prevent the subscription you're creating being removed if for example your Home Assistant or the integration is offline for a while.
```
allenporter:
```suggestion
1. Select **Never expire** as the *Expiry Period* to prevent the subscription you're creating being removed if for example your Home Assistant or the integration is offline for a while.
```

allenporter:
> > Hi, thanks for the update. I think it would make sense to explain why in the docs they should do this since it may not be obvious. (Also message expiration vs subscription expiration might be confusing)
> 
> Is that better?
> 
> > I would like to move to have the integration automatically create the subscription in the future to resolve this type of issue.
> 
> Sounds helpful but also a bit dangerous to have Home Assistant (or just the BAU service account credentials) able to manage cloud infra (not a BAU activity, done once) as well as just consume what has been created.
> 
> Granted In this particular case there **should** be no financial exposure, but it's a precedent to manage carefully before its overused by other integrations.

Agreed, it can be done during an interactive config flow, with the user in the loop. This is a great point to think through risks and mitigations in more detail, so thanks for that.

allenporter (approved):

allenporter:
Thank you @chrisns 👍🏼

PR #20085, opened 2021-10-30
Document new automation.stop service

RobertMe:
## Proposed change

Adds documentation for new `automation.stop` service.

## Type of change

## Additional information

- Link to parent pull request in the codebase: home-assistant/core#58758

## Checklist

### DIFF
--- a/source/_docs/automation/services.markdown
+++ b/source/_docs/automation/services.markdown
@@ -39,6 +39,14 @@ Service data attribute | Optional | Description
 `entity_id` | no | Entity ID of automation to trigger. Can be a list. `none` or `all` are also accepted.
 `skip_condition` | yes | Whether or not the condition will be skipped (defaults to true).
 
+## Service {% my developer_call_service service="automation.stop" %}
+
+This service will stop any currently active actions of the automation.
+
+Service data attribute | Optional | Description
+-|-|-
+`entity_id` | no | Entity ID of automation to turn on. Can be a list. `none` or `all` are also accepted.
+
 ## Service {% my developer_call_service service="automation.reload" %}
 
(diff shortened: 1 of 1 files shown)


source/_docs/automation/services.markdown
```diff
@@ -39,6 +39,14 @@ Service data attribute | Optional | Description
 `entity_id` | no | Entity ID of automation to trigger. Can be a list. `none` or `all` are also accepted.
 `skip_condition` | yes | Whether or not the condition will be skipped (defaults to true).
 
+## Service {% my developer_call_service service="automation.stop" %}
+
+This service will stop any currently active actions of the automation.
+
+Service data attribute | Optional | Description
+-|-|-
+`entity_id` | no | Entity ID of automation to turn on. Can be a list. `none` or `all` are also accepted.
```
frenck, line 48: `none`?

RobertMe: I copy pasted it from the others. I didn't know about both of them (`none` and `all`). I tested `all` and it does work (at least with just a single automation running) so there is some "behind the scenes" magic for it which I presume handles `none` as well.
But I don't know what the purpose of `none` would be for `turn_on` and `turn_off` either.

So if you want me to change it I will do so. But I guess some magicall `none` will work. Wether it has a real purpose is up for discussion, but I think the same can be said for `none` when used with `turn_on`/`turn_off` which also document this.

And maybe this `none` derives from a feature where executing the service should be conditional, from before `choose` existed? As you could then use a template and return either an entity id or `none`.
```yaml
actions:
  - service: automation.turn_on
    target:
      entity_id: "{{ 'automation.xyz' if true else 'none' }}"
```
But that's obviously a wild guess why `none` would exist in general.

PR #20102, opened 2021-10-31
Additional hint in th recorder docs to statistics

klacol:
## Proposed change
When a user adds a new integration, he/she has to be careful, to not let the database get too big in size. Often, all the sensor values are recorded automatically. That is ok, normal recorder behavior.

My thing is, that I am always very unsure to change something in the recorder (e.g. in the exclude list) because I do not want to affect the data, that is shown in the energy dashboard. That data shall not be deleted (purged), because the historical view is the essence of the energy dashboard. And there is not way to reimport the energy data (the statistics) manually.

So, a user will perhaps struggle, to activate the recorder and that would leed to an ever growing database. I propose to make a note on the docs for the recorder and make clear, the the recorder 

* neither controls, which statistics are stored in the database
* nor purges the existing statistics

I know, it it not good practice to document "what not is", but in this cas, it would have helped me. And perhaps many others.

## Type of change

- [x ] Adjusted missing or incorrect information in the current documentation (`current` branch).

## Checklist

  - x made a change to the existing documentation and used the `current` branch.

### DIFF
--- a/source/_integrations/recorder.markdown
+++ b/source/_integrations/recorder.markdown
@@ -17,6 +17,12 @@ This integration constantly saves data. If you use the default configuration, th
 
 </div>
 
+<div class='note'>
+
+This integration does not affect the data, that is calculated and stored in the database by the [`statistics`](/integrations/statistics/) integration.
+
+</div>
+
 Home Assistant uses [SQLAlchemy](https://www.sqlalchemy.org/), which is an Object Relational Mapper (ORM). This means that you can use **any** SQL backend for the recorder that is supported by SQLAlchemy, like [MySQL](https://www.mysql.com/), [MariaDB](https://mariadb.org/), [PostgreSQL](https://www.postgresql.org/), or [MS SQL Server](https://www.microsoft.com/en-us/sql-server/).
 
 The default database engine is [SQLite](https://www.sqlite.org/) which does not require any configuration. The database is stored in your Home Assistant configuration directory ('/config/') and is named `home-assistant_v2.db`.


source/_integrations/recorder.markdown
```diff
@@ -17,6 +17,12 @@ This integration constantly saves data. If you use the default configuration, th
 
 </div>
 
+<div class='note'>
+
+This integration does not affect the data, that is calculated and stored in the database by the [`statistics`](/integrations/statistics/) integration.
```
frenck, line 22: This is not true, if the data isn't recorded, we cannot generate statistics either.

frenck (changes requested):

source/_integrations/recorder.markdown
klacol: Ok, thanks for this clarification. Then, the recorder integration is related to the statistics (a very importan relation). That should then be mentioned in the docs. What about such a text?

> This integration does affect the data, that is calculated and stored in the database by the statistics. The statistics  integration can only generate statistics for data that is recorded.

IMHO there must be a time, after the statistics where generated, that some parts of the states and the events could be cleaned (purged) without affecting the statistics. But how is that timing? What is needed to calculate/recalculate the statistics?

frenck: statics integration is something different?

I think we (both) are mixing up long-term statistics feature vs statistics integration. Anyways, if there are specifics for the statistics integration, it needs to be documented on that page specifically (and not somewhere else).

klacol: Ah, I was perhaps irritated by this page:
https://data.home-assistant.io/docs/statistics/

It describes the `long-term statistics feature` and not the statistics integration that is documented [here](https://www.home-assistant.io/integrations/statistics). (Perhaps the name could be not the same)

So something like this?

```suggestion
This integration does affect the data, that is calculated and stored in the database by the [`long-term statistics`](https://data.home-assistant.io/docs/statistics) feature. The long-term statistics feature will only generate statistics for data that is recorded.
```

It can also be written in the docs of the  `long-term statistics feature`, but I would suggest to support a normal user with the information, that purging/excluding sensor data with the recorder can have effects on the long-term statistics.

PR #20399, opened 2021-11-22
Add clarification regarding templates in Lovelace cards

eikowagenknecht:
## Proposed change
Small addition to the docs.

## Type of change

## Checklist

### DIFF
--- a/source/_docs/scripts/service-calls.markdown
+++ b/source/_docs/scripts/service-calls.markdown
@@ -77,6 +77,8 @@ entity_id: switch.ac
 
 {% endraw %}
 
+Please note that templates can not be used in most standard Lovelace cards (the exception being the Markdown card).
+
 ### Using the Services Developer Tool
 
 You can use the Services Developer Tool to test data to pass in a service call.


source/_docs/scripts/service-calls.markdown
```diff
@@ -77,6 +77,8 @@ entity_id: switch.ac
 
 {% endraw %}
 
+Please note that templates can not be used in most standard Lovelace cards (the exception being the Markdown card).
```
frenck, line 80: This is a page about scripts, I'm not sure why we should add a warning about Lovelace here?

eikowagenknecht:
@frenck I see your point, maybe this is the wrong place.

Since this affects multiple contexts (templating, calls with templating examples, Lovelace cards), the documentation could be:

1. On a page regarding templates (https://www.home-assistant.io/docs/configuration/templating/)
2. On a page regarding Lovelace cards in general (https://www.home-assistant.io/lovelace/cards/)
3. On all the cards pages that don't support templates (all but the Markdown card, e.g. https://www.home-assistant.io/lovelace/alarm-panel/)
4. On all the cards that describe templating use that looks like it would work inside a Lovelace card but actually doesn't (e.g. https://www.home-assistant.io/docs/scripts/service-calls/)

If this behaviour ("templating doesn't work in Lovelace cards") is currently documented somewhere, I couldn't find it.

Thinking of it, I kind of like adding "This card does not support templates." to all cards but the Markdown card, but it would also be quite redundant.

Where would you see this best placed?

frenck:
> If this behaviour ("templating doesn't work in Lovelace cards") is currently documented somewhere, I couldn't find it.

Automations syntax can't be used in Lovelace cards either... I can't use filters like the recorder either in lovelace state cards? I can't use .... on Lovelace either...

My point is, do you want to add this warning to about every page we have that is not about Lovelace?

This page has nothing to do with Lovelace, just like around ~1500 other pages we have... I don't see why it should be added here in this case (or anywhere else for that matter).

eikowagenknecht:
I get that you don't want to include "feature xyz doesn't work here" on some thousand pages when it is obvious to an experienced Home Assistant user.

But I still think a clear description where templates can be used and where they can't be used would benefit the documentation (maybe on the "Templating" site), because for new users it is not quite that obvious as can be seen in questions or misunderstandings about this in the forums.

I'm not familiar enough with Home Assistant (yet) to create a better PR to thoroughly cover all use cases of templating so I'll just close this one now.

PR #20499, opened 2021-11-29
Remove "bash" highlighting for help output docs

eikowagenknecht:
## Proposed change
The bash type highlighting highlighted every "for", "reboot" etc. which doesn't make sense for this kind of output.
Also rearranged the "backups" config entry to match the colsole output.

## Type of change
The bash type highlighting highlighted every "for", "reboot" etc.

## Additional information

## Checklist

### DIFF
--- a/source/_includes/common-tasks/commandline.md
+++ b/source/_includes/common-tasks/commandline.md
@@ -58,7 +58,7 @@ Replace x.y.z with the desired version like `--version {{current_version}}`
 
 You can get a better description of the CLI capabilities by typing `ha help`:
 
-```bash
+```txt
 The Home Assistant CLI is a small and simple command line utility that allows
 you to control and configure different aspects of Home Assistant
 
@@ -69,6 +69,7 @@ Available Commands:
   addons         Install, update, remove and configure Home Assistant add-ons
   audio          Audio device handling.
   authentication Authentication for Home Assistant users.
+  backups        Create, restore and remove backups
   banner         Prints the CLI Home Assistant banner along with some useful information
   cli            Get information, update or configure the Home Assistant cli backend
   core           Provides control of the Home Assistant Core
@@ -84,7 +85,6 @@ Available Commands:
   observer       Get information, update or configure the Home Assistant observer
   os             Operating System specific for updating, info and configuration imports
   resolution     Resolution center of Supervisor, show issues and suggest solutions
-  backups        Create, restore and remove backups
   supervisor     Monitor, control and configure the Home Assistant Supervisor
 
 Flags:


source/_includes/common-tasks/commandline.md
```diff
@@ -58,7 +58,7 @@ Replace x.y.z with the desired version like `--version {{current_version}}`
 
 You can get a better description of the CLI capabilities by typing `ha help`:
 
-```bash
+```
```
frenck: There must be a language flag set.

frenck (changes requested):

source/_includes/common-tasks/commandline.md
eikowagenknecht: Why is that? For example in the github docs (https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) I read "You can add an *optional* language identifier. I coudn't find a markdown guide related to code formatting for home assistant. If there is one, might you point me to it? https://developers.home-assistant.io/docs/documenting/standards doesn't seem to cover this.

frenck: This is not GitHub pages or docs. So their documentation does not apply here.
Or CI guards it, to prevent it is begin forgotten (which happened a lot).

In this case, just setting it to `txt` might be fitting.

eikowagenknecht:
I see the language fences are a requirement of the automatically run checks here. Since there are use cases where no code style is clearly preferrable (as with documentation output here) to avoid miscoloring, I suggest adding "{ allowEmpty: true }" as documented here: https://github.com/remarkjs/remark-lint/blob/main/packages/remark-lint-fenced-code-flag/readme.md

frenck:
> to avoid miscoloring, I suggest adding

No, we should not do that, see my comment above.

source/_includes/common-tasks/commandline.md
eikowagenknecht: Of course I understand that this is not GitHub pages, I referenced it only as a (widely used) example. But I understand why you want to leave this check in the CI. Changing it to "txt" seems to work fine.

source/_includes/common-tasks/commandline.md
```diff
@@ -33,7 +33,7 @@ ha supervisor update
 
 ### Host
 
-```bash
+```txt
```
frenck: These are valid shell commands and should therefore not be `txt` but `bash`. Please revert this change.

eikowagenknecht: Done. Reason was that the highlighting is kind of weird with the bash markdown style, but I see that this is probably better addressed somewhere else (in the markdown plugin).

frenck (approved):
Thanks, @eikowagenknecht 👍

PR #20601, opened 2021-12-07
Update ondilo_ico

Alex9779:
## Proposed change

Update ondilo_ico integration documentation
add information how to change the polling interval respecting rate limits

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/ondilo_ico.markdown
+++ b/source/_integrations/ondilo_ico.markdown
@@ -40,7 +40,7 @@ Ondilo has rate limits for using their API too, see [here](https://interop.ondil
 [homeassistant.update_entity](https://www.home-assistant.io/integrations/homeassistant/#service-homeassistantupdate_entity) service just accepts entities as input while this integration always gets all data for all pools you have even if you only update just one entity! So it is enough to configure an automation with just one entity to update everything!
 </div>
 
-Configuring the automation with all entities (normally 6 per pool) is not nessecary and will result in multiple queries which then may result in hitting the rate limit and getting inconsistent updates.
+Configuring the automation with all entities (normally 6 per pool) is not necessary and will result in multiple queries which then may result in hitting the rate limit and getting inconsistent updates.
 
 ## Known limitations
 
--- a/source/_integrations/ondilo_ico.markdown
+++ b/source/_integrations/ondilo_ico.markdown
@@ -28,6 +28,20 @@ Sensors will be created for all those data.
 
 {% include integrations/config_flow.md %}
 
+## Polling
+
+The default interval to update all data is 1 hour.
+
+If you want to update in shorter intervals, because an Ondilo ICO measures only once an hour, which may result in a difference of the time a measure has been taken and the time the Home Assistant sensor is updated, you disable automatic polling in the configuration screen of the integration and setup an automation using [homeassistant.update_entity](https://www.home-assistant.io/integrations/homeassistant/#service-homeassistantupdate_entity) service.
+
+Ondilo has rate limits for using their API too, see [here](https://interop.ondilo.com/docs/api/customer/v1/), which are 5 queries a second and 30 queries an hour. So the minimal polling interval would be 2 minutes to get all the data consistently.
+
+<div class='note warning'>
+[homeassistant.update_entity](https://www.home-assistant.io/integrations/homeassistant/#service-homeassistantupdate_entity) service just accepts entities as input while this integration always gets all data for all pools you have even if you only update just one entity! So it is enough to configure an automation with just one entity to update everything!
+</div>
+
+Configuring the automation with all entities (normally 6 per pool) is not nessecary and will result in multiple queries which then may result in hitting the rate limit and getting inconsistent updates.
+
 ## Known limitations
 
 - Recommendations are not yet supported.


source/_integrations/ondilo_ico.markdown
```diff
@@ -28,6 +28,20 @@ Sensors will be created for all those data.
 
 {% include integrations/config_flow.md %}
 
+## Polling
+
+The default interval to update all data is 1 hour.
+
+If you want to update in shorter intervals, because an Ondilo ICO measures only once an hour, which may result in a difference of the time a measure has been taken and the time the Home Assistant sensor is updated, you disable automatic polling in the configuration screen of the integration and setup an automation using [homeassistant.update_entity](https://www.home-assistant.io/integrations/homeassistant/#service-homeassistantupdate_entity) service.
```
frenck, line 35: This is not integration specific documentation, and thus should not be documented here.

Alex9779:
Ok well, it is just a little more explanation than [here](https://www.home-assistant.io/integrations/google_travel_time/#updating-sensors-on-demand-using-automation).
See [here](https://github.com/home-assistant/developers.home-assistant/pull/1149) and [here](https://github.com/home-assistant/core/pull/60983), I struggled to get this info anywhere into the docs and I don't find a general page explaining how to configure custom polling other than this short paragraph in the annoucing blog entry [here](https://www.home-assistant.io/blog/2021/06/02/release-20216/#disable-polling-updates-on-any-integration) and in the meanwhile at some other integrations...

What shall I do, remove the explanation and provide a demo automation like in the other integration? I really do not know what you expect, now it seems I provide too much information but which I don't find anywhere else...

frenck:
Honestly, I don't think we should document this stuff at all.

Home Assistant should provide sane defaults by default, displaying/adding such things will lead to people playing with it, or trying to bump polling massively. In the past, this has caused problems for us many times. For example, API provides closing off Home Assistant completely, resulting in nobody being able to use it again.

Is the default for this integration, not sane?

Alex9779:
This was the start of my journey: https://github.com/home-assistant/core/issues/60825
As I explained there I think it is not sane...



> Honestly, I don't think we should document this stuff at all.

Maybe then you should think about changing that [here](https://www.home-assistant.io/docs/configuration/platform_options/#scan-interval)
Having the blog post in mind this seems outdated, maybe this should be updated?

frenck:
> Maybe then you should think about changing that here

That is actually deprecated and no longer accepted/used. It is for the legacy platforms and not part of new or updated integrations for the same reasoning as before.

> Having the blog post in mind this seems outdated, maybe this should be updated?

I'm not following that one.

Anyways, do you think this integration does or does not provide a sane default?

Alex9779:
> Anyways, do you think this integration does or does not provide a sane default?

The device measures once an hour, actually about ~61-64 minutes.
If you pull the results from the server in an one hour interval you can get serious differences of the time the measure has been taken and the time the sensor in Home Assistant is updated so statistics on those values are some kind of unusable when you want to see what actions (like adding chlorine or taking a bath) had which results on the measurements. That is because Home Assistant logs the time for the measurement the sensor has been updated while the API provides a time stamp when the measurement has been taken. This information cannot be pushed to Home Assistant.

frenck:
So, it's not a sane default or, is it?

Alex9779:
No, it is not...
I recommend 5 minutes to have some spare queries left on that rate limit of 30 queries per hour but to have the data in Home Assistant at more accurate timestamps...
I can reopen the initial issue or provide a PR with changing that, the code owner didn't respond on any comment yet...

frenck:
> No, it is not...

So... If something is not working correctly, out of the box, please raise an issue or open a contribution to improve on the situation from the source. Documenting a workaround for these things is not a good solution.

Alex9779:
I didn't see it as a workaround as this method of changing polling intervals is recommended since 2021.06, if you don't want to document it in detail, why did you announce it? And maybe the interval was reasonable for the creator but not for me... Maybe for others it is not reasonable too, though I know this integration is not used very often as this is a very special and expensive device. But there are other polling integrations out there and looking at some of them your principles are hard to pin, some have what I tried to add, allowing to change the polling interval with a config entry, others explain how to change the polling interval with less information than I put in.

Anyways, this was my last try on this topic and honestly this annoys me right now, I got my solution, I tried to help others who might also stumble over this, I should have stopped this when I got my solution and restrict my contributions to correct typos...

Alex9779:
Don't get me wrong, still a proud user of Home Assistant, proud supporter of Nabu Casa and patreon of you...

frenck:
> I didn't see it as a workaround as this method of changing polling intervals is recommended since 2021.06

So, the reason I started asking question, is because of the details in this PR about rate limits and such. Those are generally great indicators to spot incorrect, or, opportunities for improving current polling frequencies.

Asking some additional questions turns out, that might be the case. As you confirmed the current defaults are probably not what they should be to be useful.

We should always strive for the best out-of-the-box experience for anyone. I hope you can agree that documenting this, to improve an experience (by using a higher poll frequency to make it useful), is not a good improvement for that matter. A new user should have an acceptable experience on first use, not when applying scripts like suggested.

We are (and I am) striving for a better and greater good for everyone. Effortlessly, out-of-the-box.

> And maybe the interval was reasonable for the creator but not for me... Maybe for others it is not reasonable too, though I know this integration is not used very often as this is a very special and expensive device.

Sure, but that is something we can discuss/look for; we should not assume it's blocking / cannot be changed and just document it instead 🤷 

> Anyways, this was my last try on this topic and honestly this annoys me right now, I got my solution, I tried to help others who might also stumble over this

That is appreciated, just that I think (based on your answers) we should solve the root cause instead...

> Don't get me wrong, still a proud user of Home Assistant, proud supporter of Nabu Casa and patreon of you...

Thank you! Good to hear and appreciated, however not related to this issue and doesn't change the general review comments ❤️

Alex9779:
> however not related to this issue and doesn't change the general review comments ❤️

I am aware of this, just wanted to point that out that I am not just a frustrated user using stuff, mocking and not trying to contribute and help out in ways possible to me...

So to break this down, I re-open my initial issue and prepare a PR with an update just for the default polling interval with some clarifying comments...

PR #20982, opened 2022-01-02
Added info what to do when no internet access is available

VivantSenior:
## Proposed change

Added information that TOTP won't work when no internet access or local NTP server is available and what to do when user has setup TOTP and is locked out from HA.

## Type of change

## Additional information

## Checklist


homeassistant:
Hi @VivantSenior, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#20982).

Thanks!

source/_docs/authentication/multi-factor-auth.markdown
```diff
@@ -54,7 +54,7 @@ Once TOTP is enabled, Home Assistant requires the latest code from your phone ap
 
 <div class='note'>
 
-TOTP is _time based_ so it relies on your Home Assistant clock being accurate. If the verification keeps failing, make sure the clock on Home Assistant is correct.
+TOTP is _time based_ so it relies on your Home Assistant clock being accurate. If the verification keeps failing, make sure the clock on Home Assistant is correct. In orther the clock to be accured Home Assistant needs to have internet access or local [NTP server](https://en.wikipedia.org/wiki/Network_Time_Protocol) available. If you will enable TOTP when no internet access is available and you are locked out from your account, to restore access you would need to remove or rename `auth_module.totp` file and restart your Home Assistant instance. The file is typically stored in `/supervisor/homeassistant/.storage` folder. You can do that by using previously setup Samba share add-on or login via SSH to you Home Assistant ([Home Assistant OS SSH login](https://developers.home-assistant.io/docs/operating-system/debugging/#home-assistant-operating-system)).
```
frenck, line 57: This information is not fully correct and depends on your setup.

The current documentation, imho, is already correct; the time needs to be set correctly.

Whether or not this needs to be set up when starting your system, or if your system has an RTC, is a bit out of scope here.

VivantSenior: Hi,

I've just described my issue and the way I was able to get it "fixed".

My HA instance is connected to the internet only few times in a given month (to pull updates and send analytics). It was connected during initial setup (including TOTP) - that why I was able to setup TOTP successfully.

I agree that the docs are correct but can be somewhat extended because for someone the information that "the time needs to be set up correctly" is not the same as "the system needs to have constant access to the internet".

Best,
Viv

frenck: It depends on your system. A correct time is needed, how that is restored by your system depends on your hardware. Most computers have an RTC to handle this, which doesn't require the internet at all.

Other setups get the time from the parent system (eg VMs), or use local NTPs even.

The problem with the suggested edit, is that is occurs that an internet connection is always needed to do MFA, which is thus incorrect.

VivantSenior: I have a RPi 4B (as 35% of HA users according to the Analytics) which by defult doesn't have RTC. So when I run my HA instance on RPi without RTC hat, MFA won't work correctly because of incorrect time. Right?

You are right, I can clarify that MFA require correct time gathered from any source like internet / local NTP server / RTC extension / parent system. If no source is available the MFA won't work.

frenck: >  If no source is available the MFA won't work.

That is what the documentation already states. It needs a correct time.

VivantSenior: I agree, but average user doesn't know what are the sources of time and won't connect the dots that e.g. no internet means no current time and that means incorrectly working MFA.

frenck: Right, I definitely agree on that one.

I think we should add part of resetting TOPT to the "I'm locked out" page, and link to it from here:

<https://www.home-assistant.io/docs/locked_out/>

That will allow for some more in-depth explanation (it's kinda squeezed in here in this paragraph now if you know what I mean).

PR #20985, opened 2022-01-02
Add docs for UniFi Protect number platform

AngellusMortis:
## Type of change

- [X] Added documentation for a new feature I'm adding to Home Assistant (`next` branch).

## Additional information

- This PR fixes or closes issue: https://github.com/home-assistant/core/pull/63220

## Checklist

### DIFF
--- a/source/_integrations/unifiprotect.markdown
+++ b/source/_integrations/unifiprotect.markdown
@@ -7,6 +7,7 @@ ha_category:
   - Camera
   - Light
   - Media Player
+  - Number
   - Switch
 ha_release: 2022.2
 ha_iot_class: Local Push
@@ -22,6 +23,7 @@ ha_platforms:
   - camera
   - light
   - media_player
+  - number
   - switch
 ---
 
@@ -38,7 +40,7 @@ This Integration supports all UniFi OS Consoles that can run UniFi Protect. Curr
 * UniFi Dream Machine Pro (**UDMP**)
 * UniFi Cloud Key Gen2 Plus (**CKGP**) firmware version v2.0.24+
 
-CKGP with Firmware v1.x **do NOT run UniFiOS**, you must upgrade to firmware v2.0.24 or newer.
+CKGP with Firmware v1.x **do NOT run UniFi OS**, you must upgrade to firmware v2.0.24 or newer.
 
 ### Software Support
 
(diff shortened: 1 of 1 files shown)


source/_integrations/unifiprotect.markdown
```diff
@@ -94,6 +96,11 @@ Each UniFi Protect camera will get the following entities added:
 
 * **Camera** - A camera entity for each camera channel and RTSP(S) combination found for each camera (up to 6). Only the highest resolution RTSPS camera entity will be enabled by default.
 * **Media Player** - If your camera has a speaker, you will get a media player entity that allows you to play audio to your camera's speaker. Any audio file URI that is playable by FFmpeg will be able to be played to your speaker, including via the [TTS Say Service](https://www.home-assistant.io/integrations/tts/#service-say).
+* **Number** - The following number entities will be added for each camera:
```
frenck, line 99: I really do think the documentation structure doesn't make sense.

People look at the information they can have access to and don't look them up by the platform.

source/_integrations/unifiprotect.markdown
```diff
@@ -110,7 +117,10 @@ Each UniFi Protect camera will get the following entities added:
 Each UniFi Protect floodlight will get the following entities added:
 
 * **Light** - A light entity will be added for each floodlight device. The light entity will let you control turning on or off your light as well as adjust the brightness of your floodlight.
-* **Switch** - The following switch entities will be added for each camera:
+* **Number** - The following number entities will be added for each floodlight:
```
frenck, line 120: Same as above

source/_integrations/unifiprotect.markdown
AngellusMortis: Okay.... do you have a suggestion on how to make it better?

frenck: The problem is that the main chapter has devices/types that also are in this list here as platforms.

Now you end up with sensors that provide numbers and sensors that have cameras, and cameras having media players... the confusion here is that an end-user will have no clue in context between the difference of an actual device vs an entity platform.

I think it is too complex, to begin with. For example: `#### Entities`
While Home Assistant indeed has entities, and those drive the system, from a user perspective we work with devices & services. Entities are part of that system.

I think it makes more sense to talk in term of functionality.

For example; this integration supports UniFi protect cameras (and just that).
You could extend it with a more functional description of what can be controlled.

AngellusMortis: I think I get what you mean. Should I refactor this PR for these changes (and the other ones you mentioned in previous reviews)? Or should I let this one go as is and do it all in a follow up?

frenck: Let do a refactor afterward 👍

frenck (approved):

PR #20992, opened 2022-01-02
Alarmdecoder chime status template example incorrect

TD22057:
## Proposed change
 I used the example from the alarmdecoder to create a template sensor and it generates an error because there is no target entity in the example.  This PR adds that so the example will work.


frenck (approved):
Thanks, @TD22057 👍

PR #21047, opened 2022-01-05
Added workaround for using analog pins with binary_sensors

christianschneider89:
See https://github.com/home-assistant/core/issues/63371

### DIFF
--- a/source/_integrations/firmata.markdown
+++ b/source/_integrations/firmata.markdown
@@ -144,7 +144,7 @@ binary_sensor:
     pin:
       description: The digital or analog pin number on the board. For analog pins the corresponding digital number must be used. Please use the pinout that corresponds to your board. For example, on the Arduino Uno, pin A2 corresponds to the digital pin number 16.
       required: true
-      type: integer
+      type: [integer, string]
     pin_mode:
       description: The digital or analog pin input mode. Supported modes are `INPUT` and `PULLUP`. Check your board specifications to see which pins have optional internal pullups available.
       required: true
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @christianschneider89, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#21047).

Thanks!

source/_integrations/firmata.markdown
```diff
@@ -142,9 +142,9 @@ binary_sensor:
       required: true
       type: string
     pin:
-      description: The digital or analog pin number on the board.
+      description: The digital or analog pin number on the board. For analog pins the corresponding digital number must be used. Please use the pinout that corresponds to your board. For example, on the Arduino Uno, pin A2 corresponds to the digital pin number 16.
       required: true
-      type: [integer, string]
+      type: integer
```
frenck: This is unfortunately incorrect/not true, the integration accepts a string.

See: https://github.com/home-assistant/core/blob/5fa6ef2849a9412e97526c98187ea68ed8b204f8/homeassistant/components/firmata/__init__.py#L81

And: <https://github.com/home-assistant/core/blob/5fa6ef2849a9412e97526c98187ea68ed8b204f8/homeassistant/components/firmata/__init__.py#L50>

frenck (changes requested):

source/_integrations/firmata.markdown
christianschneider89: Ok, I reverted this commit.

DaAwesomeP: Yes, it *should* accept the form `A#` for pins (and that worked originally). This pull documents a less-intuitive workaround, not the intended method. See https://github.com/home-assistant/core/issues/63371

frenck (approved):
Thanks, @christianschneider89 👍

PR #21222, opened 2022-01-17
Add RainMachine docs for `restrict_watering` and `unrestrict_watering` services

bachya:
## Proposed change

This PR adds docs for two new forthcoming RainMachine services.

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/64219

## Checklist

### DIFF
--- a/source/_integrations/rainmachine.markdown
+++ b/source/_integrations/rainmachine.markdown
@@ -49,7 +49,7 @@ Services accept either device IDs or entity IDs, depending on the nature of the 
 
 ### `rainmachine.pause_watering`
 
-Pause all watering activities for a number of seconds. After the pause is complete, the previous watering activities will resume.
+Pause all watering activities for a number of seconds. After the pause is complete, the previous watering activities will resume. Note that controllers can only be paused for a maximum of 12 hours.
 
 | Service Data Attribute | Optional | Description                    |
 | ---------------------- | -------- | ------------------------------ |
--- a/source/_integrations/rainmachine.markdown
+++ b/source/_integrations/rainmachine.markdown
@@ -39,8 +39,10 @@ Services accept either device IDs or entity IDs, depending on the nature of the 
 - Services that require a device ID as a target:
   - `rainmachine.pause_watering`
   - `rainmachine.push_weather_data`
+  - `rainmachine.restrict_watering`
   - `rainmachine.stop_all`
   - `rainmachine.unpause_watering`
+  - `rainmachine.unrestrict_watering`
 - Services that require an entity ID as a target (note that the correct entity ID type must be provided, such as a program for a program-related service)
   - `rainmachine.start_program`
   - `rainmachine.start_zone`
--- a/source/_integrations/rainmachine.markdown
+++ b/source/_integrations/rainmachine.markdown
@@ -49,7 +49,7 @@ Services accept either device IDs or entity IDs, depending on the nature of the 
 
 ### `rainmachine.pause_watering`
 
-Pause all watering activities for a number of seconds.
+Pause all watering activities for a number of seconds. After the pause is complete, the previous watering activities will resume.
 
 | Service Data Attribute | Optional | Description                    |
 | ---------------------- | -------- | ------------------------------ |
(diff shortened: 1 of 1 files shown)


source/_integrations/rainmachine.markdown
```diff
@@ -49,7 +49,7 @@ Services accept either device IDs or entity IDs, depending on the nature of the
 
 ### `rainmachine.pause_watering`
 
-Pause all watering activities for a number of seconds.
+Pause all watering activities for a number of seconds. After the pause is complete, the previous watering activities will resume.
```
mrand: Can the 12 hour maximum be mentioned either here, or in the table, so the users know up front, rather than potentially stumbling upon it later and having to redo something?

source/_integrations/rainmachine.markdown
```diff
@@ -81,6 +81,14 @@ See details of RainMachine API here:
 | `pressure`             | no       | Barametric Pressure (kPa)                                                                                             |
 | `dewpoint`             | no       | Dew Point (°C)                                                                                                        |
 
+### `rainmachine.restrict_watering`
+
+Restrict any and all watering activities from staring for a time period.
```
mrand, line 88: Thanks for adding this!  

Further up, it says "Services accept either device IDs or entity IDs, depending on the nature of the service", but restrict_watering doesn't appear in either list.

Speaking of which, isn't clear to me exactly what it's referring to when it says device ID.  Is it the name that a user gives the device?  And is that requirement absolute?  I ask because [issue #101](https://github.com/bachya/regenmaschine/issues/101) shows a user calling 'pause_watering' with an entity_id, yet it's listed here in the documentation as requiring a device ID.  If it confuses someone using HA for years (me), I assume it would also confuse a new user.  An example service call might go a long ways here.

Thanks again!

mrand: One other thing:  I would blindly assume that this is hh:mm:ss, but I'm not certain since the internal RainMachine webpage is days, hours, and minutes.

bachya: >Further up, it says "Services accept either device IDs or entity IDs, depending on the nature of the service", but restrict_watering doesn't appear in either list.

Great catch!

>Speaking of which, isn't clear to me exactly what it's referring to when it says device ID. Is it the name that a user gives the device? And is that requirement absolute? I ask because issue #101 shows a user calling 'pause_watering' with an entity_id, yet it's listed here in the documentation as requiring a device ID. If it confuses someone using HA for years (me), I assume it would also confuse a new user. An example service call might go a long ways here.

`device_id` represents an entry in the [Device Registry](https://developers.home-assistant.io/docs/device_registry_index/). In the case of RainMachine, the "device" is the controller (and it implements a series of entities, like switches, sensors, etc.) – some services only make sense in the case of the controller itself (e.g., you pause all watering activities on the device, not on an entity). So, we use a [selector](https://www.home-assistant.io/docs/blueprint/selectors/#device-selector) to allow the user to pick the correct device via the service UI.

I do agree that it's a tad confusing when to use which. I think the "standard" these days is to have people fill out the UI to create a service, then use its YAML browser to see the format. I think any clarifications would be better applied somewhere upstream, since this applies to any integration that makes use of devices.

>One other thing: I would blindly assume that this is hh:mm:ss, but I'm not certain since the internal RainMachine webpage is days, hours, and minutes.

In the code, I allow the service to submit a `duration` value that matches any [time period](https://github.com/home-assistant/core/blob/4cd222e70aa241c4d9e644581f3e2e49942081de/homeassistant/helpers/config_validation.py#L465) – this could take several forms:

_(note that all of these examples equate to 3 days)_

## As an `HH:MM:SS` string:

```yaml
duration: "72:00:00"
```

## As a number of seconds:

```yaml
duration: 259200
```

## As a "time dictionary" (can contain any of the following keys)

```yaml
duration:
  days: 3
  hours: 0
  minutes: 0
  seconds: 0
  milliseconds: 0
```

Using the `HH:MM:SS` string seemed like the simplest example, but I'm open to suggestions. Again, if there's confusion on this, it might make sense to consider an upstream clarification (since many integrations do this same thing).

mrand: I didn't mean to question using hh:mm:ss - just that I wasn't sure if it was indeed hh:mm:ss (vs. days:hours:mins).

As for *my* device_id confusion, thanks for expanding on that!

bdraco (approved):
Looks good.  Checked grammar as well. 👍

PR #21230, opened 2022-01-17
Change recommendation for 700 series z-wave controllers due to buggy fw

blhoward2:
## Proposed change

Currently, we recommend people purchase 700 series controllers. There is currently a nasty bug affecting a number of people that causes the stick to become unresponsive. Essentially, it fails to send out Acks when it is receiving a lot of messages, which cause all of those devices to then begin flooding the controller with repeat messages, which makes the problem even worse as it cascades. We've reported this to SiLabs (from Z-Wave JS) weeks ago, but it doesn't seem like it'll be a quick fix. In the interim we should change our recommendation. A lot of people are downgrading from the 700 sticks temporarily. 

https://github.com/zwave-js/node-zwave-js/issues/3906

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/z-wave/controllers.markdown
+++ b/source/_docs/z-wave/controllers.markdown
@@ -24,7 +24,11 @@ You need to have a compatible Z-Wave stick or module installed. The following de
   - Aeotec Z-Pi 7 (700 series)
   - ZWave.me Razberry Board (500 series)
 
-If you are just starting out, you should purchase a [Z-Wave Plus](https://z-wavealliance.org/z-wave_plus_certification/) controller. At the moment, **we do not recommend purchasing 700 series controllers due to a bug in the current firmware for those devices**.
+If you are just starting out, you should purchase a [Z-Wave Plus](https://z-wavealliance.org/z-wave_plus_certification/) controller.
+
+<p class='note'>
+At the moment, **we do not recommend purchasing 700 series controllers due to a bug in the current firmware for those devices**.
+</p>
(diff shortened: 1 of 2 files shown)


frenck:
Duplicate of: <https://github.com/home-assistant/home-assistant.io/pull/21209>

blhoward2:
I didn't see that one was closed. I do think we should change our recommendation here and @AlCalzone agrees. This current bug makes the network unusable so it isn't a small thing. We don't have to mention the bug if you don't want to do so, we could just recommend people purchase 500 sticks for now and leave it at that. It's just that we're presently recommending people purchase a setup that very well may not work.

frenck:
> I do think we should change our recommendation here

We don't document bugs. If there is a big issue affecting a lot of people, putting out an alert can be considered an alternative.

So this PR contains 2 things, 1 change the recommendation, and 2 it adds an alert/warning/note for a firmware issue. I think the first is fine, the second isn't and should be an alert imho.

I do not agree on merging this into our docs as is.

blhoward2:
It doesn't have to document a bug. See my edit to my comment. What if we just recommend 500 series sticks and leave it at that? Just to remove the recommendation RE 700 series.

frenck:
Sorry, I was editing my previous answer, please revisit.

blhoward2:
@frenck How about now?

frenck (approved):
Thanks, @blhoward2 👍

PR #21431, opened 2022-02-02
Remove the other trailing comma and minor copyedit (mqtt/discovery)

### LINKED ISSUE #21143: In JSON Message a ','  (Comma) after last element should be removed (Documentation MQTT Discovery)
### Feedback

The last comma in the section **Use object_id to influence the entity id** should be removed 
{
  "name":"My Super Device",
  "object_id":"device1",
  "state_topic": "homeassistant/sensor/device1/state",    <---- The last comma here should be removed
 }

Because this command failed to work for me:
mosquitto_pub -h 192.168.x.y -p 1883 -u myuser -P mypwd -t "homeassistant/sensor/device1/config" -m '{"name":"My Super Device","object_id":"device1","state_topic": "homeassistant/sensor/device1/state",}'

and this one did work for me:
mosquitto_pub -h 192.168.x.y -p 1883 -u myuser -P mypwd -t "homeassistant/sensor/device1/config" -m '{"name":"My Super Device","object_id":"device1","state_topic": "homeassistant/sensor/device1/state"}'

Apart from that I was very much puzzled by the way the _entity_id_ is grabbed from the "name" instead of the _object-id_ in the topic for the configuration. Also in the example above there is already **sensor/device1/config** 
It is stated indeed that the topic in the topic-path is not used in the _entity_id_ but apparently once it is mentioned in the json-message it IS used as the _entity_id_   My remark here is when both are named _object_id_  (The one in the Topic-path and the one in the JSON-Message) . Why does it not suffice to have only the <object_id> defined in the path and use that value also as the _entity_id_ , it is very confusing as it is now and doing it in that way the need to have a separate element in the JSON-Message would no longer exist. The only way I can think of is one uses e.g. two sensors mounted on one device (E.g. Temperature + Humidity) and one wants to use a object_id in the path that is the same.  However the example concerning this in this same page at **SENSORS WITH MULTIPLE VALUES** stipulates that 'sensorBedroomT' and 'sensorBedroomH' are to be used as the <object_id> in the Topic-Path, so one for each sensor on the device.

Long story I know but I think one or the other can be left out OR just rename "object_id" to "entity_id" in the JSON-Message 
or at least have "entity_id" it as an alternative to "object_id" in the JSON-Message, like so:

 {
  "name":"My Super Device",
  "entity_id":"device1",
  "state_topic": "homeassistant/sensor/device1/state"
 }
If the _entity_id_ is already taken then HA-Server should  just not accept the config-message or see as an update of the existing values for "name" and "state_topic" in this case, I would prefer the latter. 

In that light the use of <node_id> could be illuminated with an example like the one of the mentioned. That was extremely helpful to get my mind around that, to see that the Topic for the state is the same in both cases. Now <node_id> seems to be something of olden days and the details of use is left out of this documentation. 
Thanks anyway. I am impressed with the work, just trying to see how to integrate a Mesh-System with Home-Assistant, and thought to start with Discovery to make the first steps with that. Possibly leading to another "Integration" in the Eco-system. 

### URL

https://www.home-assistant.io/docs/mqtt/discovery/

### Version

2021.12.8 ? (It is not stated , but it is the documentation of 11 Jan 2022)

### Additional information

_No response_

mrand:
## Proposed change
Remove another trailing comma in JSON that causes error.

Also some minor copyediting.

## Type of change

## Additional information

- This PR fixes or closes issue: fixes #21143 

## Checklist

### DIFF
--- a/source/_docs/mqtt/discovery.markdown
+++ b/source/_docs/mqtt/discovery.markdown
@@ -339,7 +339,7 @@ A motion detection device which can be represented by a [binary sensor](/integra
 - State topic: `homeassistant/binary_sensor/garden/state`
 - Payload:  `{"name": "garden", "device_class": "motion", "state_topic": "homeassistant/binary_sensor/garden/state"}`
 
-To create a new sensor manually. For more details please refer to the [MQTT testing section](/docs/mqtt/testing/).
+To create a new sensor manually.
 
 ```bash
 mosquitto_pub -h 127.0.0.1 -p 1883 -t "homeassistant/binary_sensor/garden/config" -m '{"name": "garden", "device_class": "motion", "state_topic": "homeassistant/binary_sensor/garden/state"}'
@@ -356,6 +356,8 @@ Delete the sensor by sending an empty message.
  ```bash
 mosquitto_pub -h 127.0.0.1 -p 1883 -t "homeassistant/binary_sensor/garden/config" -m ''
 ```
+
+For more details please refer to the [MQTT testing section](/docs/mqtt/testing/).
 
 ### Sensors with multiple values
 
@@ -453,7 +455,7 @@ Setting up a climate integration (heat only):
 {
   "mode":"off",
   "target_temp":"21.50",
-  "current_temp":"23.60",
+  "current_temp":"23.60"
 }
 ```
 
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @mrand 👍

PR #21436, opened 2022-02-02
Fix merge conflict resolution in flux_led docs

bdraco:
Fix merge conflict between `current` and `next` for #21435


bdraco:
These are getting a bit large and probably need a cleanup pass, but I figure fixing the merge conflict is more important and that can be done later

frenck (approved):
Thanks, @bdraco 👍

PR #21693, opened 2022-02-19
Improve glossary entries of device and entity

Hypfer:
## Proposed change

[Someone on reddit reported](https://www.reddit.com/r/homeassistant/comments/svg8s8/comment/hxj8lav/) that they were confused about the difference between devices and entities.

This PR attempts to clarify that.

Sidenote:

Clicking on the edit button at the bottom [of the glossary docs page](https://www.home-assistant.io/docs/glossary/) does not take you to the correct yaml file.

Maybe `source/_docs/glossary.markdown` should contain a comment pointing to the right file?

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_data/glossary.yml
+++ b/source/_data/glossary.yml
@@ -15,7 +15,7 @@
 - topic: Customize
   description: "[Customization](/docs/configuration/customizing-devices/) allows you to overwrite the default parameters of your devices in the configuration."
 - topic: Device
-  description: "A device is usually a physical unit which can do or observe something."
+  description: "A device is a named collection of entities that all represent the same physical/logical unit, which can do or observe something. An example for a device would be a smart plug named "Coffee Machine" which provides a `switch` entity plus one or more `sensor` entities for power monitoring or similar."
 - topic: Device tracker
   description: "[Device trackers](/integrations/device_tracker) are used to track the presence, or location, of a device."
 - topic: Discovery
@@ -23,7 +23,7 @@
 - topic: Domain
   description: "Entities and services belong to a domain, which is the first part of the entity or service, before the `.`. For example `light.kitchen` is an entity in the `light` domain, while `homeassistant.turn_on` is the `turn_on` service for the `homeassistant` domain."
 - topic: Entity
-  description: "An [entity](/docs/configuration/platform_options/) is the representation of a function of a single device, unit, or web service. There may be multiple entities for a single device, unit, or web service, or there may be only one."
+  description: "Entities are things with a unique ID, a name and some kind of state that may be any primitive type such as a number or a string. As the core building block of state in Home Assistant, an [entity](/docs/configuration/platform_options/) is the representation of a functionality of a single device, unit, or web service. There may be multiple entities for a single device, unit, or web service."
 - topic: Event
   description: "An [event](/docs/configuration/events/) is when something happens."
 - topic: Frontend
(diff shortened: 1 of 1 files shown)


source/_data/glossary.yml
```diff
@@ -15,15 +15,15 @@
...
 - topic: Discovery
   description: "[Discovery](/integrations/discovery/) is the automatic setup of zeroconf/mDNS and uPnP devices after they are discovered."
 - topic: Domain
   description: "Entities and services belong to a domain, which is the first part of the entity or service, before the `.`. For example `light.kitchen` is an entity in the `light` domain, while `homeassistant.turn_on` is the `turn_on` service for the `homeassistant` domain."
 - topic: Entity
-  description: "An [entity](/docs/configuration/platform_options/) is the representation of a function of a single device, unit, or web service. There may be multiple entities for a single device, unit, or web service, or there may be only one."
+  description: "Entities are things with a unique ID, a name and some kind of state that may be any primitive type such as a number or a string. As the core building block of state in Home Assistant, an [entity](/docs/configuration/platform_options/) is the representation of a functionality of a single device, unit, or web service. There may be multiple entities for a single device, unit, or web service."
```
frenck: Hmm right, this description is more from the era when we didn't have the concept of "devices" in Home Assistant. So indeed needs an update.

Some points:

- Entity IDs don't have to have a unique ID, also, that is more an implementation detail I guess.
- I don't think we need to tell about data types (numbers, strings, etc)
- Let's avoid terms like "things" and "some kind of".
- The link to the platform options should be dropped. It doesn't apply to all anymore, and additionally, cannot be applied to most.
- "There may be multiple entities for a single device, unit, or web service." -> We should change this to refer back to the "Device" item.
- "web service" should become "service", as we call it "devices & services" nowadays.

source/_data/glossary.yml
```diff
@@ -15,15 +15,15 @@
...
 - topic: Discovery
   description: "[Discovery](/integrations/discovery/) is the automatic setup of zeroconf/mDNS and uPnP devices after they are discovered."
 - topic: Domain
   description: "Entities and services belong to a domain, which is the first part of the entity or service, before the `.`. For example `light.kitchen` is an entity in the `light` domain, while `homeassistant.turn_on` is the `turn_on` service for the `homeassistant` domain."
 - topic: Entity
-  description: "An [entity](/docs/configuration/platform_options/) is the representation of a function of a single device, unit, or web service. There may be multiple entities for a single device, unit, or web service, or there may be only one."
+  description: "An entity is the representation of a single functionality of a device or service inside of Home Assistant. Every entity has an ID and a state, which may change over time. Futhermore, entities may feature attributes that provide context or other metadata. An example of an entity of the `sensor` domain would be a temperature sensor with the ID `sensor.temperature_outdoor`, state `23.75` and attribute `unit_of_measurement: °C`."
```
frenck: I think we should drop:

>  Every entity has an ID and a state, which may change over time.

As it doesn't add to explaining what an entity is. Additionally, it might impose the risk people think that is a problem without a solution?

> Futhermore, entities may feature attributes that provide context or other metadata.

I guess a bit the same as the previous line. Attributes are something, we in general, try to get rid of as much as possible. Attributes belonging to e.g. sensor, like unit of measurement, are interesting, but not something someone generally works with.

Maybe something like this? What do you think?

```suggestion
  description: "An entity is the representation of a single control or data point of a device or service inside Home Assistant. A single device or service can thus provide multiple entities to be able to monitor and control all features a device provides. For example, a combined temperature and humidity sensor, in general, provides two `sensor` entities. One for the temperature (e.g., `sensor.temperature` with state `21.0` and unit `°C`) and one for the humity (e.g., `sensor.humidity` with state `65.4` and unit `%`)."
```

frenck (approved):
Alright, thanks @Hypfer 👍

PR #21721, opened 2022-02-21
Rename manual alarm docs

avee87:
## Proposed change

Currently these integrations are called "Manual" and "Manual MQTT" which doesn't really tell what they are about and also isn't searchable. Adding "Alarm" to their names to make it clearer. Also, moving docs to /integrations/alarm_control_panel.manual similar to mqtt and template versions.

Next branch because it needs to be in sync with updated documentation links in core repo - see parent PR.

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/66979

## Checklist

### DIFF
--- a/source/_integrations/alarm_control_panel.manual.markdown
+++ b/source/_integrations/alarm_control_panel.manual.markdown
@@ -1,5 +1,5 @@
 ---
-title: Manual
+title: Manual Alarm
 description: Instructions on how to integrate manual alarms into Home Assistant.
 ha_category:
   - Alarm
--- a/source/_integrations/alarm_control_panel.manual.markdown
+++ b/source/_integrations/alarm_control_panel.manual.markdown
@@ -1,5 +1,5 @@
 ---
-title: Manual Alarm
+title: Manual Alarm Control Panel
 description: Instructions on how to integrate manual alarms into Home Assistant.
 ha_category:
   - Alarm
--- a/source/_integrations/alarm_control_panel.manual_mqtt.markdown
+++ b/source/_integrations/alarm_control_panel.manual_mqtt.markdown
@@ -1,5 +1,5 @@
 ---
-title: Manual MQTT Alarm
+title: Manual MQTT Alarm Control Panel
 description: Instructions on how to integrate manual alarms into Home Assistant with MQTT support.
 ha_category:
   - Alarm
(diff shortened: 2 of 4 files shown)


source/_integrations/alarm_control_panel.manual.markdown
```diff
@@ -1,5 +1,5 @@
 ---
-title: Manual
+title: Manual Alarm Control Panel
```
frenck, line 2: There is no need to change titles, they are synced from our core codebase.

source/_integrations/alarm_control_panel.manual_mqtt.markdown
```diff
@@ -1,5 +1,5 @@
 ---
```
frenck, line 1: This file rename no longer matches the domain it belongs to.

source/_integrations/alarm_control_panel.manual.markdown
```diff
@@ -1,5 +1,5 @@
 ---
```
frenck, line 1: This file rename is incorrect. Integrations should be listed by their domain key.

frenck (changes requested):

source/_integrations/alarm_control_panel.manual_mqtt.markdown
avee87: There is already a precedent of `alarm_control_panel.mqtt` and `alarm_control_panel.template` which to be honest make a lot of sense to me as these are different internal implementations of the same thing.

frenck: > There is already a precedent

That is legacy.

avee87: Does it mean that all template docs should live in the same file because they all technically use the same integration? That would be very hard to read and understand, especially since there are so many unrelated template integrations.

But okay, I can move files back and just change the name...

frenck: > Does it mean that all template docs should live in the same file because they all technically use the same integration? 

Yes.

However, that integration hasn't done that in docs at this point, as the template integration is currently moving from platform to integration level configuration itself. We haven't squashed everything in the docs yet as the code is still in progress.

source/_integrations/alarm_control_panel.manual.markdown
avee87: Is there an automated process for this? So only the core PR is enough?

avee87: If so, I guess I'll just close this PR

PR #21874, opened 2022-03-02
Add workday breaking change to 2022.3 release notes

TomBrien:
## Proposed change

Add  details of breaking change to workday binary sensor to release notes.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_posts/2022-03-02-release-20223.markdown
+++ b/source/_posts/2022-03-02-release-20223.markdown
@@ -818,7 +818,7 @@ The stale restored sensor can be safely removed.
 
 {% details "Workday" %}
 
-Workday now requires a 2-character [ISO 3166-1] country code instead of the country's name.
+Workday now requires a 2-character [ISO 3166-1] country code instead of the country's name. States, provinces and consituent countries must now be listed using the `province` setting. Please see the [documentation](/integrations/workday) for examples.
 
 [ISO 3166-1]: https://en.wikipedia.org/wiki/ISO_3166-1
 
--- a/source/_posts/2022-03-02-release-20223.markdown
+++ b/source/_posts/2022-03-02-release-20223.markdown
@@ -818,7 +818,7 @@ The stale restored sensor can be safely removed.
 
 {% details "Workday" %}
 
-Work day now requires a 2-character [ISO 3166-1] country code instead of the country's name.
+Workday now requires a 2-character [ISO 3166-1] country code instead of the country's name.
 
 [ISO 3166-1]: https://en.wikipedia.org/wiki/ISO_3166-1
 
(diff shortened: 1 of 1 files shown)


source/_posts/2022-03-02-release-20223.markdown
```diff
@@ -816,6 +816,14 @@ The stale restored sensor can be safely removed.
 
 {% enddetails %}
 
+{% details "Workday" %}
+
+Work day now requires a 2-character [ISO 3166-1] country code instead of the country's name.
+
+[ISO 3166-1]: https://en.wikipedia.org/wiki/ISO_3166-1
+
```
frenck, line 824: Got a reference/commit related to this?

source/_posts/2022-03-02-release-20223.markdown
```diff
@@ -816,6 +816,14 @@ The stale restored sensor can be safely removed.
 
 {% enddetails %}
 
+{% details "Workday" %}
+
+Work day now requires a 2-character [ISO 3166-1] country code instead of the country's name.
```
frenck:
```suggestion
Workday now requires a 2-character [ISO 3166-1] country code instead of the country's name.
```

source/_posts/2022-03-02-release-20223.markdown
TomBrien: It's an inadvertent result of changes to the checking logic in the integration, I don't think there is a clear commit for it since it's the result of the library bump and actually we're failing cases that should work. For example "UK" used and "Wales" (constituent country of UK) both used to work in 2022.2 and work fine with `holidays` itself however we throw them as invalid and now require "GB". We should fix the checking logic but figured it would be best to get this info up.

TomBrien: Probably this is the closest thing we could reference https://github.com/dr-prodigy/python-holidays/pull/608

frenck: If it is a bug, we should fix it... not document it.

TomBrien: I don't think it is truly a bug and we could be more permissive when checking for sure (our province key is actually not needed at all). Part of the problem is terminology around country and province. 

Routes forward I see:
- Accept the current logic is reasonable, add the breaking change note. I think this is best as it forces stricter compliance to the library which does not clearly document which subdivisions can be called from the base class (`holidays.Wales()` is fine `holidays.Fl()` is not, I don't consider the table in their readme, which we point to, as clear on this)
- Update integration to be more permissive. This should allow the integration to work just like before but is potentially less versatile. I did do this in: https://github.com/home-assistant/core/pull/67508, but changed my mind, should we re open (needs work though)? 
- Do the above and then remove the redundant `province` key in the configuration and possibly change `country` to something more generic. This is a bigger breaking change that I don't think is the best option.

Let me know preferred route forward

frenck (approved):

PR #22021, opened 2022-03-14
Adding complex example for event-based trigger using an icon template

parautenbach:
Thanks to the assistance provided here: https://github.com/home-assistant/core/issues/66459.

## Proposed change
Adding complex example for event-based trigger using an icon template.

## Type of change

## Additional information
- This PR fixes or closes issue: fixes #66459

## Checklist


source/_integrations/template.markdown
```diff
@@ -561,6 +561,28 @@ template:
...
+        event_type: YOUR_EVENT
+      - platform: state
+        entity_id: binary_sensor.doorbell_rang
+        to: ~
+    binary_sensor:
+      name: doorbell_rang
+      icon: "{{ 'mdi:bell-ring-outline' if is_state('binary_sensor.doorbell_rang', 'on') else 'mdi:bell-outline' }}"
+      state: "{{ trigger.platform == 'event' or states('binary_sensor.doorbell_rang') }}"
+      auto_off:
+        seconds: 5
```
frenck: Wouldn't this be better?

```suggestion
template:
  - trigger:
      - platform: event
        event_type: YOUR_EVENT
      - platform: state
        entity_id: binary_sensor.doorbell_rang
        to: "off"
    binary_sensor:
      name: doorbell_rang
      icon: "{{ (trigger.platform == 'event') | iif('mdi:bell-ring-outline', 'mdi:bell-outline') }}"
      state: "{{ trigger.platform == 'event' }}"
      auto_off:
        seconds: 5
```

It triggers the exact changes that are relevant. The event and the auto turn off, instead of triggering all state changes. It also removes the need to query its own state (at that can already be determined as the state template does. Lastly, it only triggers 1 update for the template entity when the doorbell rings (icon & state, in that case, are rendered in a single update).

frenck (changes requested):
This feels a bit "meh" to me, as it is self referencing, but I don't see a better way around it either (but do have an improvement suggestion below).

source/_integrations/template.markdown
parautenbach: Thanks for the suggestion! It reads a lot easier to me – and I can see it being more efficient.

parautenbach:
> This feels a bit "meh" to me, as it is self referencing, but I don't see a better way around it either (but do have an improvement suggestion below).

I agree. The solution was not obvious (to me) and it's not pretty, but the only alternative I saw was to create an event-based sensor with the auto off, and then a second template sensor that references the former sensor in both it's value and icon templates. 

Thanks for the suggestion, btw.

source/_integrations/template.markdown
```diff
@@ -561,6 +561,28 @@ template:
 
 {% endraw %}
 
+A more advanced use case could be to set the icon based on the sensor's own state like above, but when triggered by an event. This example demonstrates a binary sensor that turns on momentarily, such as when a doorbell button is pressed. When the button gets pressed, the first trigger will be triggered and the binary sensor's state will be updated. This in turn will trigger the second trigger, which will cause the icon template to be evaluated. The platform check in the state template is to avoid the update from going into an infinite loop of updating.
```
frenck: This needs to be adjusted to the new situation :)

parautenbach: Bussy week... :-) 

I updated it now. 

I dropped the infinite loop part, because it doesn't seem relevant at all now.

Mariusthvdb:
seeing this becoming the first section in the docs where self referencing is actually encouraged, wouldn't it be useful to make that explicit?
I mean there are many other cases where self-referencing could be used (and we were explicitly told to not do that)/ It seems it would need some explicit notes from HA side whether this is in fact the advised way of configuring template sensors.
Also there is some congruency with https://github.com/home-assistant/core/pull/65201 where `this.entity_id` is introduced in template entities. 
Wouldn't that be a better way of 'self_referencing'?

frenck:
> seeing this becoming the first section in the docs where self referencing is actually encouraged,

Its not, as it has been removed from the example.

frenck (dismissed):
Thanks, @parautenbach 👍

frenck:
Hmm can we resolve the merge conflict to we can get this beauty in?

parautenbach:
> Hmm can we resolve the merge conflict to we can get this beauty in?

Thanks for picking it up @frenck. I checked before the 2022.5 release whether it's been merged but not again after, so I didn't spot the merge conflicts. I hope I've resolved it correctly.

frenck (approved):
Thanks for the rebase! (and the PR to begin with of course)

../Frenck

PR #22037, opened 2022-03-15
Update filter.markdown

wora55:
I improved on the definition of the unique_id and included the tag into the example.
This seems to make sense because I needed more then one day to find out that without this field there was an autocreated id, that was not visible to me

## Proposed change
Extend description as proposed

## Type of change

## Additional information
Improved explanation of handling. Asking around among other users of HA showed they did not know this feature either. And I only located it by finding an older error message and improvement request. The problem here is, that the term 'unique_id' is used for the object resulting from the filtering, usually this would be 'entity_id' but that has historically already been used of the object of the filter-operation.

From reading this documentation I was not able to conclude that I have to define this 'unique_id' since I was looking for some expression containing 'entity'.

I hope the new phrasing makes it easier for the next guy or girl

## Checklist

- [ X] This PR uses the correct branch, based on one of the following:
- [ X] The documentation follows the Home Assistant documentation [standards][].

### DIFF
--- a/source/_integrations/filter.markdown
+++ b/source/_integrations/filter.markdown
@@ -31,6 +31,7 @@ To enable Filter Sensors in your installation, add the following to your `config
 sensor:
   - platform: filter
     name: "filtered realistic humidity"
+    unique_id: sensor.realistic_humidity_filtered
     entity_id: sensor.realistic_humidity
     filters:
       - filter: outlier
@@ -65,7 +66,7 @@ name:
   required: false
   type: string
 unique_id:
-  description: An ID that uniquely identifies the filter sensor. Set this to a unique value to allow customization through the UI.
+  description: An ID that uniquely identifies the filter sensor. Set this to a unique value to allow customization through the UI. If you do not define it, it will be automatically derived from the ´name´ by HA
   required: false
   type: string
 filters:


homeassistant:
Hi @wora55, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#22037).

Thanks!

source/_integrations/filter.markdown
```diff
@@ -31,6 +31,7 @@ To enable Filter Sensors in your installation, add the following to your `config
 sensor:
   - platform: filter
     name: "filtered realistic humidity"
+    unique_id: sensor.realistic_humidity_filtered
```
dgomes, line 34: to avoid confusion I would not use a unique_id that resembles a entity name

source/_integrations/filter.markdown
```diff
@@ -65,7 +66,7 @@ name:
   required: false
   type: string
 unique_id:
-  description: An ID that uniquely identifies the filter sensor. Set this to a unique value to allow customization through the UI.
+  description: An ID that uniquely identifies the filter sensor. Set this to a unique value to allow customization through the UI. If you do not define it, it will be automatically derived from the ´name´ by HA
```
frenck, line 69: This is not correct, the unique ID will not be derived from the name.

source/_integrations/filter.markdown
frenck: We try to keep all configuration samples minimal. Thus, no optional requirement in the default sample. This helps a user to get started quickly by copy-&-paste the sample without worrying about optional parameters which they most likely not need. If required, insert a full configuration sample later that covers special setups or alike.

frenck (changes requested):

source/_integrations/filter.markdown
wora55: dgomes said:
> to avoid confusion I would not use a unique_id that resembles a entity name

Sounds right at first thought, but my situation was exactly the opposite problem. I was asked by some integration for an entity_id and after much searching found out that unique_id was the entity_id the other input form asked for (and I did not have it, of course). After understanding the logic behind it, I understand the reasoning as well. But I am still a newbie with HA so I also see the problem for beginners that results from the naming of this parameters.

source/_integrations/filter.markdown
wora55: Unsatisfying answer. There probably is an internal unique Identifier. How is that created? (Do not say "No", but rather give the/a correct explanation) 
Makes it easier to correct my proposal

frenck: It isn't created when not set 🤷 
This line added is simply not true.

source/_integrations/filter.markdown
frenck: > unique_id was the entity_id 

unique IDs and entity IDs are not related to each other in any way. One is an internal unique identifier used in Home Assistant registries, while the entity ID is an identifier for end-user consumption.

frenck:
I'm closing this PR, as the addition is incorrect. Nevertheless, thanks for willing to contribute @wora55 👍 

../Frenck

wora55:
Franck Nijhof schrieb am Dienstag, 15. März 2022 18:30:39 (+01:00):



@frenck requested changes on this pull request.

In source/_integrations/filter.markdown:

> @@ -65,7 +66,7 @@ name:
   required: false
   type: string
 unique_id:
-  description: An ID that uniquely identifies the filter sensor. Set this to a unique value to allow customization through the UI.
+  description: An ID that uniquely identifies the filter sensor. Set this to a unique value to allow customization through the UI. If you do not define it, it will be automatically derived from the ´name´ by HA


This is not correct, the unique ID will not be derived from the name.

Would the phrase 'an ID' instead of 'the unique ID'  be ok?



In source/_integrations/filter.markdown:

> @@ -31,6 +31,7 @@ To enable Filter Sensors in your installation, add the following to your `config
 sensor:
   - platform: filter
     name: "filtered realistic humidity"
+    unique_id: sensor.realistic_humidity_filtered


An additional configuration sample would be fine as well. But in this case your line of reasoning (keep it minimal) was not helpful because the naming of the parameters was not consistent with other examples I saw.



—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.
You are receiving this because you were mentioned.


-- 
Wolfgang Radke
Schubertstr. 6f
76344 Eggenstein

source/_integrations/filter.markdown
wora55: OK. Killer argument. I withdraw that part of my change :-(

PR #22083, opened 2022-03-20
Add note that UDP port 20002 needs to be open

guysoft:
Had to run for UFW:
```
sudo ufw allow 20002/udp
```

You can see its being used here in the lib this integration imports:
https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/bridge.py#L112

## Proposed change

I could not understand why my switcher is not detected. 

## Type of change

## Additional information

## Checklist

- [ x ] The documentation follows the Home Assistant documentation [standards][].

### DIFF
--- a/source/_integrations/switcher_kis.markdown
+++ b/source/_integrations/switcher_kis.markdown
@@ -23,7 +23,7 @@ Supported devices:
 - Switcher Power Plug
 - Switcher Touch (from firmware 1.51)
 - Switcher V2 (Espressif chipset - from firmware 3.21)
-- Switcher V2 (Qualcomm chipset - from firmware 72.32)
+- Switcher V2 (Qualcomm chipset - from firmware 72.32) Note: make sure port 20002 with udp is open.
 - Switcher V4
 
 If you completed the integration setup but are still unable to control the device, please make sure your device's firmware is up-to-date.


homeassistant:
Hi @guysoft, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#22083).

Thanks!

source/_integrations/switcher_kis.markdown
```diff
@@ -23,7 +23,7 @@ Supported devices:
 - Switcher Power Plug
 - Switcher Touch (from firmware 1.51)
 - Switcher V2 (Espressif chipset - from firmware 3.21)
-- Switcher V2 (Qualcomm chipset - from firmware 72.32)
+- Switcher V2 (Qualcomm chipset - from firmware 72.32) Note: make sure port 20002 with udp is open.
```
frenck, line 26: Does this need intructions?

guysoft: I had to open a different switcher library in order to see that it uses a non-standard port my firewall blocked.
I wasted an hour or so after reading thr homesssistant instructions because it was not noted.

guysoft:
Missed that.  Signed :)

source/_integrations/switcher_kis.markdown
frenck: Oh wait, so this is no opening a port on the device, but on firewall, you have created on the Home Assistant end

Sorry, but that is really not clear on this note. It actually looks like you need to change something on the device. Additionally, we do not provide a manual for managing your firewall.

thecode: Port 20002 applies to all devices, not only to `Switcher V2`, this note is confusing, it also make a standard user think he needs to do something, while the standard configuration not using IOT vlan does not need any action from the user.

However, I don't mind adding this as the last part of the documentation, see an example for [webOS TV notes](https://www.home-assistant.io/integrations/webostv/#notes)

@frenck do you think we should add it? I know that in general we don't add instructions for advanced installations in integrations docs.

frenck: I think its fine to add a generic note at the bottom of the page, but in the current setting it is even more confusing IMHO.

guysoft: webOS TV notes looks like a better way to add it. And thanks for confirming that the port is the same for all devices. shall i close this pr and open a new one with the note at the bottom?

thecode: @guysoft you can copy the example from [webOS TV notes](https://www.home-assistant.io/integrations/webostv/#notes) and if you are already there also add that **port 9957 TCP** also needs to be open (this probably works for you since you don't have a firewall from HA side to the device, but some others might have firewall in both directions.

thecode: > webOS TV notes looks like a better way to add it. And thanks for confirming that the port is the same for all devices. shall i close this pr and open a new one with the note at the bottom?

both options are good

guysoft:
See  #22086

source/_integrations/switcher_kis.markdown
guysoft: Done and done, see https://github.com/home-assistant/home-assistant.io/pull/22086

PR #22114, opened 2022-03-22
Revert "Add mqtt notify service and update example page"

jbouwh:
Reverts home-assistant/home-assistant.io#21344

Reason is that the parent PR is to be reverted.

Parent PR:
https://github.com/home-assistant/core/pull/68505


MartinHjelmare (approved):
Thanks!

frenck (approved):

PR #22194, opened 2022-03-30
Revert "Correct MQTT scene docs"

frenck:
Reverts home-assistant/home-assistant.io#22193

as per <https://github.com/home-assistant/home-assistant.io/pull/22193#pullrequestreview-925981260>


emontnemery:
It was my mistake, we don't need to revert the change to `current`

PR #22293, opened 2022-04-07
Updated google_assistant integration troubleshooting

aaronpowell:
This update is based on a discussion on the Discord channel and some testing done by user `katty`.

I was experiencing the same problem and resolved it by following these additional steps.

### DIFF
--- a/source/_integrations/google_assistant.markdown
+++ b/source/_integrations/google_assistant.markdown
@@ -337,3 +337,7 @@ If you're having trouble with *Account linking failed* after you unlinked your s
 ### Failed linking - Could not update the setting. Please check your connection
 
 If you're having trouble linking your account, with the error message `Could not update the setting. Please check your connection` after logging into your Home Assistant instance, try setting `expose_by_default: false` then exposing a single simple device (light or switch preferably). It is also worth checking if any home ad blocker is disabled if you are having issues.
+
+### Failed linking - `Could not reach [test] <your project>. Please try again.`
+
+If you're unable to link a new project, or relink an existing project, there may be something cached in the Google Actions project that needs refreshing (we're unsure the exact underlying cause). One way to force this refresh is to complete more of the details around the project under **Deploy** -> **Directory Information**. Fill this information out and then redeploy the test instance, as per Step 3 above.


homeassistant:
Hi @aaronpowell, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#22293).

Thanks!

kittywitch:
It would be worthwhile to completely redo the tutorial now, considering Google Actions Console has been different from what is shown and stipulated for over a year now. Not to mention poor advice within the tutorial, including hitting "Add Scope" twice, when there is already a default empty field, so you end up with an extra empty field (which, admittedly is removed upon save).

The structure of the tutorial in general is also poor. Talking about the configuration JSON steps but not mentioning that requires the creation of the Service Account Token Creator to begin with, while that stuff is in a completely different section.

The whole matter of steps 2.3 and 2.4 are now completely obsolete:
```
If asked, leave options as they default No, I only want to allow account creation on my website and select Next.
Then if asked, for the Linking type select OAuth and Authorization Code. Click Next
```

Whereas the requirement for doing a new integration involving the Deploy->Directory information likely ultimately replaces said steps, somewhat.

Admittedly, this would require taking new screenshots, which would likely be cumbersome for the poor soul who decides to go about redoing said tutorial. (I understand if you don't want to be said poor soul. :p)

Dare I say, it would also be worth it to include mobile screenshots involving the initial linking process and the login page, for more transparency of what it looks like when done correctly, and done incorrectly (such as without providing the Deploy->Directory information).

Omnipius:
It looks like @aaronpowell 's merge does not yet cover that the Deploy -> Directory Information step is now required. Account linking now does not work without this part completed. [See forum topic here.](https://community.home-assistant.io/t/can-no-longer-link-google-assistant/409495/25) Should that be opened as a separate issue?

aaronpowell:
I didn't realise it was now required, when did that change (is there any relevant Google docs/announcement on that)?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

klaasnicolaas:
duplicate of #22349 ?

Issue #21804, opened 2022-02-25
How to configure meter in Consumption path (1)?


GianlucaCollot:
### Feedback

Hi, 
I am using the Fronius integration that seems well written and works well for production and consumption.
Unfortunately, if the meter is in the consumption path, it does not provide the value of Energy "returned to grid".
So there is noway to configure this value in the energy dashboard.
Could you add to the documentation this case and explain possible workarounds?

Thank you in advance

Gianluca

### URL

https://www.home-assistant.io/integrations/fronius/

### Version

2022.2.9

farmio:
Hi 👋!

I don't know which entities are even available with a consumption path meter. If you have power_flow entities you could do something like this
```yaml
# positive values are from grid
from_grid: {{ max(states('sensor.power_grid_fronius_power_flow_0_http_192_168_68_10') | float, 0) }}
# negative values are to grid
to_grid: {{ max((0 - states('sensor.power_grid_fronius_power_flow_0_http_192_168_68_10') | float), 0) }}
```
and use these as source for a Riemann sum integration. 

There is a fantastic forum thread explaining such calculations regarding batteries - I'm sure you can get some inspiration from it: https://community.home-assistant.io/t/howto-fronius-integration-with-battery-into-energy-dashboard/376329

When you've got it working it would be awesome if you'd add a high-level explanation to the documentation (I think more detailed information fits the forums better since template stuff can change/update in the future).

GianlucaCollot:
Thank you,I prefer to use the values already calculated by the meter
I have available 2 energy values: 

- Consumed Energy (from meter in consumption path):  `sensor.energy_real_consumed_fronius_meter_0_http_192_168_1_xxx`
- Produced Energy (from the inverter): `sensor.energy_total_fronius_inverter_1_http_192_168_1_xxx`

So I built a Template entity with the difference:
```
 - name: "energy_to_grid"
            # friendly_name: 'Energia immessa in rete (neg se consumata)'
            unit_of_measurement: Wh
            state: >-
               {{ (states("sensor.energy_total_fronius_inverter_1_http_192_168_1_103") | float) - (states("sensor.energy_real_consumed_fronius_meter_0_http_192_168_1_103") | float) | int }}
            device_class: energy
            state_class: total
```
I am waiting to see the statistics update and confirm this solution works.
I will eventually update the documetnation for meter in the consumption path.

farmio:
[...]

PR #22320, opened 2022-04-09
Updates Fronius docs adding consumption path

luixal:
## Proposed change
Documentation improvement.

Adds config steps for Froinus meter installed in consumption path. I've reproduced and tested myself this configuration from 
https://github.com/home-assistant/home-assistant.io/issues/21804

Documentation gives credit to real author (not me).

## Type of change

Documentation

## Additional information

Doesn't edit previous content.

- This PR fixes or closes issue: documents solution in closed issue #21804 

## Checklist

### DIFF
--- a/source/_integrations/fronius.markdown
+++ b/source/_integrations/fronius.markdown
@@ -141,13 +141,13 @@ sensor:
 
 Now, you only need to configure Energy dashboard's params using the UI (first to values use the newly created entities, other ones are just the same as in the _"feed in path"_ config):
 
-- For `Grid consumption` use the meters `energy_bough` entity.
-- For `Return to grid` use the meters `energy_sold` entity.using `energy_bough`.
-- For `Solar production`: 
-  - If no battery is connected to an inverter: Add each inverters `Energy total` entity.
-  - If a battery is connected to an inverter: Use [Riemann sum](/integrations/integration/) over `Power photovoltaics` entity (from power_flow endpoint found in your `SolarNet` device)
-- `Battery systems` aren't supported directly. Use [Template](/integrations/template) to split and invert negative values of `Power battery` entity (from power_flow) for charging power (W) and positive values for discharging power (W). Then use [Riemann sum](/integrations/integration/) to integrate each into energy values (kWh).
-- For `Devices` use the Ohmpilots `Energy consumed` entity.
+- For **Grid consumption** use the meters `energy_bough` entity.
+- For **Return to grid** use the meters `energy_sold` entity.using `energy_bough`.
+- For **Solar production**: 
+  - If no battery is connected to an inverter: Add each inverters **Energy total** entity.
+  - If a battery is connected to an inverter: Use [Riemann sum](/integrations/integration/) over **Power photovoltaics** entity (from power_flow endpoint found in your **SolarNet** device)
+- **Battery systems** aren't supported directly. Use [Template](/integrations/template) to split and invert negative values of **Power battery** entity (from power_flow) for charging power (W) and positive values for discharging power (W). Then use [Riemann sum](/integrations/integration/) to integrate each into energy values (kWh).
+- For **Devices** use the Ohmpilots **Energy consumed** entity.
 
 
 ## Note
--- a/source/_integrations/fronius.markdown
+++ b/source/_integrations/fronius.markdown
@@ -78,77 +78,28 @@ When a device is not responding correctly the update interval will increase to 1
 
 ## Energy dashboard
 
-Home Assistant's Energy Dashboard expects some sources of data to be configured for it to work properly. They mainly are the energy you get from the grid, the energy you send to the grid, the energy produced by your solar panels and others relative to batteries and devices you can see below.
-
-The meter integrated with de Fronius can be installed (and configured) in two different positions: _"feed in path"_ (`meter_location` = 0) or _"in consumption path"_ (`meter_location` = 1). In the first case, the meter provides all data you need but, in the second case, it doesn't so you need to calculate some of them by using helpers.
-
-How to know which is your case? Just go to your entities list and search `meter_location` to see the value it has.
-
-Below you can find instructions for configuring both cases.
-
-### Feed in path
-
-Recommended energy dashboard configuration for meter location in feed in path (`Meter location: 0`):
-
-- For `Grid consumption` use the meters `Energy real consumed` entity.
-- For `Return to grid` use the meters `Energy real produced` entity.
 - For `Solar production`: 
   - If no battery is connected to an inverter: Add each inverters `Energy total` entity.
   - If a battery is connected to an inverter: Use [Riemann sum](/integrations/integration/) over `Power photovoltaics` entity (from power_flow endpoint found in your `SolarNet` device)
 - `Battery systems` aren't supported directly. Use [Template](/integrations/template) to split and invert negative values of `Power battery` entity (from power_flow) for charging power (W) and positive values for discharging power (W). Then use [Riemann sum](/integrations/integration/) to integrate each into energy values (kWh).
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @luixal, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#22320).

Thanks!

source/_integrations/fronius.markdown
```diff
@@ -78,6 +78,16 @@ When a device is not responding correctly the update interval will increase to 1
 
 ## Energy dashboard
 
+Home Assistant's Energy Dashboard expects some sources of data to be configured for it to work properly. They mainly are the energy you get from the grid, the energy you send to the grid, the energy produced by your solar panels and others relative to batteries and devices you can see below.
+
+The meter integrated with de Fronius can be installed (and configured) in two different positions: _"feed in path"_ (`meter_location` = 0) or _"in consumption path"_ (`meter_location` = 1). In the first case, the meter provides all data you need but, in the second case, it doesn't so you need to calculate some of them by creating some entities in your configuration file.
```
farmio:
```suggestion
The meter integrated with de Fronius can be installed (and configured) in two different positions: _"feed in path"_ (`meter_location` = 0) or _"in consumption path"_ (`meter_location` = 1). In the first case, the meter provides all data you need but, in the second case, it doesn't so you need to calculate some of them by using helpers.
```

farmio (commented):
Helper sensors (like integral) can now not only be configured in yaml so we should not be too specific about that in the Fronius documentation.

You can relatively link to different integrations like
`[Integration - Riemann sum integral](/integrations/integration/)` or
`[State-based template sensors](/integrations/template/#state-based-template-binary-sensors-buttons-numbers-selects-and-sensors)` (although I guess this heading may change one day so it might be better to just link to /template/ )

luixal:
Not sure about which changes would you like me to do. It's my first time commiting to HA docs. I just placed the steps I followed to get this working :)

farmio:
I just think we should not use exact yaml configuration of other integrations in the Fronius documentation page. When other integrations change some details once, we would have to notice and update this page too - just like it happend in 2022.4: Riemann integral can now be configured from UI instead of yaml. Maybe yaml config for Riemann gets deprecated at some point in the future - then the Fronius documentation is outdated as well. 

So maybe try to provide an overview of how this should be done instead of a copy&paste-yaml.

You did acutally add such a high-level description: 
> If a battery is connected to an inverter: Use [Riemann sum](/integrations/integration/) over `Power photovoltaics` entity (from power_flow endpoint found in your `SolarNet` device)

I'd describe the power/energy sold/bought things just like this.

luixal:
Sorry, I just copy&paste the description you're pointing too from the same page (from the _feed in path_ original content), I didn't write that.

Thing is, the kind of explanation you want is out of my knowledge, I just wanted to "make it work" and put things together from the forum thread and the [github issue](https://github.com/home-assistant/home-assistant.io/issues/21804) you pointed me to and another user made it work.

It just feel better to publish this config, which is what I want "as a user" instead of keep it for myself. If you need a different documentation I think I won't have the time to dig deeper and understand it enough to write it properly.

source/_integrations/fronius.markdown
```diff
@@ -88,6 +98,58 @@ Recommended energy dashboard configuration for meter location in feed in path (`
...
+
+```yaml
+template:
+  - sensor:
+    # Positive values are from grid
+    - name: "power_bought"
+      unit_of_measurement: W
+      state: >-
+        {{ max(states('sensor.power_grid_fronius_power_flow_0_XXX_XXX_XXX_XXX') | float, 0) }}
+      device_class: power
```
frenck: This is a bug IMHO and not a documentable solution.

It is wrapping an existing sensor to solve an issue (as it seems).

source/_integrations/fronius.markdown
```diff
@@ -88,6 +98,58 @@ Recommended energy dashboard configuration for meter location in feed in path (`
...
+  - platform: integration
+    source: sensor.power_sold
+    name: energy_sold
+    round: 2
+    
+  - platform: integration
+    source: sensor.power_bought
+    name: energy_bought
+    round: 2
+```
```
frenck: This is not related to this integration and does not belong in this documentation. Instead you could write a note and link to the integration.

source/_integrations/fronius.markdown
```diff
@@ -88,6 +98,58 @@ Recommended energy dashboard configuration for meter location in feed in path (`
...
+
+Now, you only need to configure Energy dashboard's params using the UI (first to values use the newly created entities, other ones are just the same as in the _"feed in path"_ config):
+
+- For `Grid consumption` use the meters `energy_bough` entity.
```
frenck: "Grid consumption" isn't code, and should thus not be code fenced (same applies to a couple of other values below).

frenck (changes requested):

source/_integrations/fronius.markdown
farmio: The "Power Grid" entity as provided from Fronius API is positive on import and negative on export. Hence the `max(state, 0)` and `max(0- state, 0)` to split it up into two entities - import and export to be used in energy (after integration).

source/_integrations/fronius.markdown
luixal: These are needed as they are the ones you can select in the Energy Dashboard configuration. No idea why, but it is working that way.

source/_integrations/fronius.markdown
luixal: Ok, was just to give attention to the field names in the UI.

EDIT: in fact, I was just keeping consistence with previous content in this documentation. Not sure if I should edit it.

source/_integrations/fronius.markdown
frenck: Sure, but you can do a lot more with entities. It is not the goal of this page to teach about other integration (like the integration integration).

Thus, this does not belong on this page. Please remove it.

source/_integrations/fronius.markdown
frenck: > Not sure if I should edit it.

🤷 I was reviewing your changes; not the full document.

source/_integrations/fronius.markdown
luixal: I think our focus is different. I'm writting the docs that my future me, as a user, would like to find to configure it's solar installation in HA. This is a 100% needed part for it to work. If I remove this, I'll leave a documentation that doesn't fully allow the user to get it working.

What do you think? I remove this and then... add some text saying that this should be donde in order for it to work? or do I leave the average user (not interested in the technical parts) alone?

farmio: I think you are missing that this document here should *only* be documenting the Fronius integration. 
This does *not* include the `energy` integration which has its own [documentation documents](https://www.home-assistant.io/docs/energy).
And also not Riemann sum and state based template sensors - which also have their own sites. 

It would just add maintenance dept to document all this multiple times.

luixal: Then I got it wrong as the original document is based on Fronius integration but also related to configure the data you get from this integration in the energy integration. Using it with the energy integration is 1/3 of the document and is what I found missing here. In fact, the matter of the consumption path VS feed in grid is not touch until the energy integration section.

If this page souldn't include information/examples about the energy integration, the Riemann sum or template sensor, this PR for this problem in detail is totally senseless and it's best to close it.

farmio:
I've updated this PR with a brief version of a consupmtion path meter instruction, without unrelated integration documentation, so it is not stale-closed.

frenck (approved):
Thanks, @luixal 👍

PR #22486, opened 2022-04-23
Update conditions markdown

tomlut:
## Proposed change

Remove incorrect sun example.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/scripts/conditions.markdown
+++ b/source/_docs/scripts/conditions.markdown
@@ -325,14 +325,7 @@ condition:
 
 This is an example of 1 hour offset after sunset.
 
-```yaml
-condition:
-  - condition: sun
-    after: sunset
-    before: sunrise
-```
 
-This is 'when dark' - equivalent to a state condition on `sun.sun` of `below_horizon`.
 
 ```yaml
 condition:
@@ -343,8 +336,6 @@ condition:
 
 This is 'when light' - equivalent to a state condition on `sun.sun` of `above_horizon`.
 
-We cannot use both keys in this case as it will always be `false`.
-
 ```yaml
 condition:
   condition: or
@@ -354,6 +345,8 @@ condition:
     - condition: sun
       before: sunrise
 ```
+
+This is 'when dark' - equivalent to a state condition on `sun.sun` of `below_horizon`. We cannot use both keys without the or condition in this case as it will always be `false`.
 
 A visual timeline is provided below showing an example of when these conditions are true. In this chart, sunrise is at 6:00, and sunset is at 18:00 (6:00 PM). The green areas of the chart indicate when the specified conditions are true.
 


source/_docs/scripts/conditions.markdown
```diff
@@ -325,14 +325,7 @@ condition:
...
 This is an example of 1 hour offset after sunset.
 
-```yaml
-condition:
-  - condition: sun
-    after: sunset
-    before: sunrise
-```
 
-This is 'when dark' - equivalent to a state condition on `sun.sun` of `below_horizon`.
```
frenck, line 335: This is a bug?

source/_docs/scripts/conditions.markdown
```diff
@@ -355,6 +346,8 @@ condition:
       before: sunrise
 ```
 
+This is 'when dark' - equivalent to a state condition on `sun.sun` of `below_horizon`. We cannot use both keys without the or condition in this case as it will always be `false`.
```
frenck, line 349: This is a bug?

frenck (changes requested):

source/_docs/scripts/conditions.markdown
tomlut: No it's a feature :)

AFIK It's how the sun integration has always worked. It confuses many, many new users. So if it is possible to make supplying both options `after: sunset` and `before: sunrise` work across the midnight boundary (currently it uses sunrise and sunset from the same day, which gives an impossible time period) it would make Home Assistant more user friendly.

frenck: > No it's a feature :)

I don't agree, there is specific logic in the integration to handle this case.
This is either a bug that needs fixing or the integration needs to be cleaned up from code that doesn't belong there.

tomlut: If there is specific logic in the integration to use tomorrow's sunrise in this particular case then it isn't working as "why is my sun condition that checks if it is night always false" is a constant question on the forum. The answer has always been:

You cant have before sunrise AND after sunset on the same day. Use an or condition containing the two separate sun conditions. Or use a sun below horizon condition instead.

frenck: > You cant have before sunrise AND after sunset on the same day.

The integration should take that into account (and actually has logic for that, doing a day+1 in such cases).

As said, according to the code base, this should work: thus a bug.

tomlut: Ok, how about I test it with something like this:
```
trigger
  platform: time_pattern
  minutes: "/1"
condition:
  condition: sun
  before: sunrise
  after: sunset
action:
  service: input_number.increment
  target:
    entity_id: input_number.test
```

And open an issue if the input number fails to increment all night?

frenck: I'm not sure about your last comment. 

This PR tries to remove something that is IMHO a bug which should be reported 🤷

tomlut: Now you've lost me. 

This is a PR for a documentation update for what I thought was incorrect documentation. 

I have based this entirely on reports from other people (I use sun below horizon conditions). 

You say the documentation is correct and the integration is (or may be) at fault. 

So, I will test it for myself with the above automation, and if I confirm it is a bug and it does not operate as per the documentation then I will open an issue for the sun integration, and either way close this documentation PR.

tomlut: Actually, the documentation contradicts itself, so one of these has to go:

Screenshot 2022-04-26 at 21-16-13 Conditions

tomlut:
From my proposed test in the conversation above, it pretty obvious that this works:

```
condition:
  condition: or
  conditions:
    - condition: sun
      after: sunset
    - condition: sun
      before: sunrise
```

But this does not:

```
condition:
  - condition: sun
    after: sunset
    before: sunrise
```

  (it's after sunset here)

I'll close this and open the issue for the sun integration.

PR #22517, opened 2022-04-25
[tplink] Update docs: Local access possible again

RubenKelevra:
- Add link to vendor FAQ how to setup the devices for local access only
- Rewrite summary based on current status
- Comment out devices explicitly mentioned by TP-Link to not support the new local control

## Proposed change
Update the documentation on the availability of a new method to locally access the devices, after [removing the alert](https://github.com/home-assistant/alerts.home-assistant.io/pull/518).

## Type of change

## Additional information
Community members report that the process offered by TP-Link works for them:

https://community.home-assistant.io/t/tp-link-offers-way-to-add-local-api-back/248333/99

## Checklist

### DIFF
--- a/source/_integrations/tplink.markdown
+++ b/source/_integrations/tplink.markdown
@@ -25,9 +25,10 @@ ha_integration_type: integration
 
 The `tplink` integration allows you to control your [TP-Link Smart Home Devices](https://www.tp-link.com/kasa-smart/) such as plugs, power strips, wall switches and bulbs.
 
-TP-Link offers an [FAQ](https://www.tp-link.com/us/support/faq/2707/) on how to provision your newly purchased devices. The devices won't connect to the TP-Link cloud and become locally controllable afterward. If you already own devices that are connected to the cloud, you can convert them from TP-Link Cloud operated to local operation as well.
+You need to provision your newly purchased device to connect to your network before it can be added via the integration. This can be done by using the using the [kasa command-line tool](https://python-kasa.readthedocs.io/en/latest/cli.html#provisioning).
 
-In the past, more TP-Link used to support local access, but they removed this support. If your old device is on older firmware, it can still be controlled locally by Home Assistant. In this case, you need to provision your newly purchased device to connect to your network before it can be added via the integration. This can be done by using the [kasa command-line tool](https://python-kasa.readthedocs.io/en/latest/cli.html#provisioning).
+TP-Link has issued some firmware updates to their devices which disabled the local access which is necessary to control the device via Home Assistant. They revisited their decision in early 2022 and started offering a special 'local access only' provisioning which re-enables local access for Home Assistant on newer firmware versions. This provisioning is not compatible with all devices supported by this integration. Their [FAQ](https://www.tp-link.com/us/support/faq/2707/) outlines how to proceed.
+
 
 There is currently support for the following device types within Home Assistant:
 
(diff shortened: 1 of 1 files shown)


RubenKelevra:
@balloob done :)

RubenKelevra:
@probot-home-assistant nope. Branch is correct.

source/_integrations/tplink.markdown
```diff
@@ -25,7 +25,9 @@ ha_integration_type: integration
...
 
-You need to provision your newly purchased device to connect to your network before it can be added via the integration. This can be done either by using [kasa command-line tool](https://python-kasa.readthedocs.io/en/latest/cli.html#provisioning) or by adding it to the official Kasa app before trying to add them to Home Assistant. If you use the app, do not upgrade the firmware if it presents the option to avoid blocking the local access by potential firmware updates.
+TP-Link offers an [FAQ](https://www.tp-link.com/us/support/faq/2707/) how to provision your newly purchased devices. The devices won't connect to the TP-Link cloud and become locally controllable afterwards. If you already own devices which are connected to the cloud, you can convert them from TP-Link Cloud operated to local operation as well.
+
+Note: Previously TP-Link had cut the local access via a firmware update to their devices. This is no longer the case. SOme devices not listed here as supported might still work with local control, if the old which supported local control is not updated.
```
balloob:
```suggestion
In the past more TP-Link used to support local access but they removed this support. If your old device is on older firmware it can still be controlled locally by Home Assistant.
```

source/_integrations/tplink.markdown
```diff
@@ -25,7 +25,9 @@ ha_integration_type: integration
 
 The `tplink` integration allows you to control your [TP-Link Smart Home Devices](https://www.tp-link.com/kasa-smart/) such as plugs, power strips, wall switches and bulbs.
 
-You need to provision your newly purchased device to connect to your network before it can be added via the integration. This can be done either by using [kasa command-line tool](https://python-kasa.readthedocs.io/en/latest/cli.html#provisioning) or by adding it to the official Kasa app before trying to add them to Home Assistant. If you use the app, do not upgrade the firmware if it presents the option to avoid blocking the local access by potential firmware updates.
```
rytilahti, line 28: Please do not remove this as some people may want to provision their devices without the official app. I think adding a note about the FAQ entry & that newer models can also be converted to local operation would be fine.

source/_integrations/tplink.markdown
rytilahti: So far I have heard that only a very specific selection of devices (UK versions of some HS* models) are affected.

Has there been a change in that?

rytilahti: What probably should be done is to update the alert content to clarify this. I think I had seen an issue about its scary wording even when most of the users are not affected by locked down firmwares (at least that I know of).

source/_integrations/tplink.markdown
RubenKelevra: Okay, I'll incorporate this at the end when I mention the older firmware - as I'm not sure that the tool works with the latest firmware.

source/_integrations/tplink.markdown
RubenKelevra: > What probably should be done is to update the alert content to clarify this.

Well, I first updates the alert content, which was rejected by @balloob who encouraged to update this in the docs.

Would you mind discussing this with @balloob directly – if the docs or the alert should be updated? :)

RubenKelevra: If you find more recent info about the distribution of the “bad” firmware, feel free to add this info. I cannot confirm or deny this.

source/_integrations/tplink.markdown
```diff
@@ -25,7 +25,9 @@ ha_integration_type: integration
...
+TP-Link offers an [FAQ](https://www.tp-link.com/us/support/faq/2707/) how to provision your newly purchased devices. The devices won't connect to the TP-Link cloud and become locally controllable afterwards. If you already own devices which are connected to the cloud, you can convert them from TP-Link Cloud operated to local operation as well.
```
frenck:
```suggestion
TP-Link offers an [FAQ](https://www.tp-link.com/us/support/faq/2707/) on how to provision your newly purchased devices. The devices won't connect to the TP-Link cloud and become locally controllable afterward. If you already own devices that are connected to the cloud, you can convert them from TP-Link Cloud operated to local operation as well.
```

source/_integrations/tplink.markdown
```diff
@@ -25,7 +25,9 @@ ha_integration_type: integration
...
+
+In the past more TP-Link used to support local access but they removed this support. If your old device is on older firmware it can still be controlled locally by Home Assistant. In this case you need to provision your newly purchased device to connect to your network before it can be added via the integration. This can be done by using the [kasa command-line tool](https://python-kasa.readthedocs.io/en/latest/cli.html#provisioning).
```
frenck:
```suggestion
In the past, more TP-Link used to support local access but they removed this support. If your old device is on older firmware it can still be controlled locally by Home Assistant. In this case, you need to provision your newly purchased device to connect to your network before it can be added via the integration. This can be done by using the [kasa command-line tool](https://python-kasa.readthedocs.io/en/latest/cli.html#provisioning).
```

rytilahti: I don't agree with the wording, afaik, most of the available devices still support local controls no matter how they are provisioned.

@bdraco, if I'm not mistaken you have some newer devices that work still just fine?

source/_integrations/tplink.markdown
```diff
@@ -42,24 +44,24 @@ There is currently support for the following device types within Home Assistant:
...
+- KP105 (is not mentioned in the TP-Link FAQ as working/not working)
 - KP115 (supports consumption sensors)
 
 ### Strip (Multi-Plug)
 
-- HS107 (indoor 2-outlet)
-- HS300 (powerstrip 6-outlet) (supports consumption sensors)
-- KP303 (powerstrip 3-outlet)
-- KP400 (outdoor 2-outlet)
+<!-- - HS107 (indoor 2-outlet) # Does not support local control according to TP-Link FAQ (as of date 2022-04-25) -->
```
rytilahti: Not worth changing as these are based on what users have reported to be functioning even when tplink's docs / official stance disagree.

source/_integrations/tplink.markdown
bdraco: I have yet to find a device sold in the US that doesn't work just fine out of box without having to go through [this](https://www.tp-link.com/us/support/faq/2707/)

RubenKelevra: @bdraco thanks a lot for the feedback, I'll update the PR to reflect that!

source/_integrations/tplink.markdown
RubenKelevra: Alright, I made a table to reflect both infos to give the users a chance to get an overview :)

Thanks!

source/_integrations/tplink.markdown
```diff
@@ -35,50 +38,87 @@ There is currently support for the following device types within Home Assistant:
...
+### Plugs
 
-- HS107 (indoor 2-outlet)
-- HS300 (powerstrip 6-outlet) (supports consumption sensors)
-- KP303 (powerstrip 3-outlet)
-- KP400 (outdoor 2-outlet)
-- KP200 (indoor 2-outlet)
-- KP40 (outdoor 2-outlet)
-- EP40 (outdoor 2-outlet)
+| Model Number     | Support by Home Assistant | Support by TP-Link's 'local only mode'     | Note                                     |
```
frenck, line 45: Please avoid tables, these are nearly impossible to view nicely on a mobile device.

RubenKelevra: What would you recommend as an alternative formatting?

frenck: Anything but tables :)

RubenKelevra: I mean, the file already contains tables plus they really don't look *that* bad. I'm not sure there's a cleaner way to display this amount of information.

You're sure that's not just a personal preference? 🤔 

IMG_20220502_133803

frenck: We generally ask to avoid those, I think it is weird to make an exception in this case.

Thanks 👍

RubenKelevra: Ah okay - didn't know this. Is there a style guideline or something like that for the docs? :)

RubenKelevra: I was looking through the other integrations to see how this is done for other pages, and there are quite a few using tables as well:

- yeelight
- xiaomi_miio
- philips_js
- zha

Feels more like the norm than an exception :)

frenck: Please remove the use of tables as much as possible.

frenck (changes requested):
Marking this PR as changes requested until the above added tables are removed.

RubenKelevra:
Yeah, not sure how to proceed here. I feel like tables makes the most sense in being a clean way to display the information. Anybody got an idea?

source/_integrations/tplink.markdown
bdraco: I think the simplest way to remove the tables is to not enumerate which devices support local only mode and instead link to Tplink's article.  We really don't want to get issue reports for incorrect documentation because we have copied tplink's incorrect documentation (I know at least the KL420 requires a kasa account even though their docs don't mention that).  In short we only need to know what Home Assistant supports for Home Assistant's documentation.

AFAICT all the devices that removed the local api when linking to cloud have the option to restore it by following the procedure

source/_integrations/tplink.markdown
```diff
@@ -35,50 +38,87 @@ There is currently support for the following device types within Home Assistant:
...
-- KL110
-- KL120
-- KL125
-- KL130
-- KB130
+| Model Number     | Support by Home Assistant | Support by TP-Link's 'local only mode'     | Note                                     |
+|------------------|---------------------------|--------------------------------------------|------------------------------------------|
+| KB130            | ✅                        | ❓ <!-- not mentioned as of 2022-05-01 --> |                                          |
+| KL50             | ❓                        | ✅                                         |                                          |
+| KL60             | ❓                        | ✅                                         |                                          |
```
bdraco, line 83: These work. There are test fixtures for them here

https://github.com/python-kasa/python-kasa/tree/master/kasa/tests/fixtures

source/_integrations/tplink.markdown
```diff
@@ -35,50 +38,87 @@ There is currently support for the following device types within Home Assistant:
...
+
+| Model Number     | Support by Home Assistant | Support by TP-Link's 'local only mode'     | Note                                     |
+|------------------|---------------------------|--------------------------------------------|------------------------------------------|
+| All Kasa cameras | ❓                        | ❌ <!-- as of 2022-05-01               --> |                                          |
+
+### Smart Routers
+
+| Model Number     | Support by Home Assistant | Support by TP-Link's 'local only mode'     | Note                                     |
+|------------------|---------------------------|--------------------------------------------|------------------------------------------|
+| SR20             | ❓                        | ❌                                         |                                          |
```
bdraco, line 121: None of these are supported.  We should say all Range Extenders, Cameras, and Smart Routers are not supported instead of listing them out.

source/_integrations/tplink.markdown
```diff
@@ -35,50 +38,87 @@ There is currently support for the following device types within Home Assistant:
...
 
-- KL400
-- KL420
-- KL430
+| Model Number     | Support by Home Assistant | Support by TP-Link's 'local only mode'     | Note                                     |
+|------------------|---------------------------|--------------------------------------------|------------------------------------------|
+| KL400            | ✅                        | ❓ <!-- not mentioned as of 2022-05-01 --> |                                          |
+| KL420            | ✅                        | ❓ <!-- not mentioned as of 2022-05-01 --> |                                          |
+| KL430            | ✅                        | ❌                                         |                                          |
+| KL430E           | ❓                        | ❌                                         |                                          |
```
bdraco, line 101: Let's not list models we don't know if they work.

source/_integrations/tplink.markdown
```diff
@@ -35,50 +38,87 @@ There is currently support for the following device types within Home Assistant:
...
 
-- HS200
-- HS210
-- HS220 (acts as a light)
+| Model Number     | Support by Home Assistant | Support by TP-Link's 'local only mode'     | Note                                     |
+|------------------|---------------------------|--------------------------------------------|------------------------------------------|
+| ES10             | ❓                        | ✅                                         |                                          |
+| HS200            | ✅                        | ✅                                         |                                          |
+| HS210            | ✅                        | ❌                                         |                                          |
+| HS210KIT         | ❓                        | ❌                                         |                                          |
```
bdraco, line 74: This is the same as the HS210, we can drop this line

source/_integrations/tplink.markdown
```diff
@@ -35,50 +38,87 @@ There is currently support for the following device types within Home Assistant:
...
-- KP200 (indoor 2-outlet)
-- KP40 (outdoor 2-outlet)
-- EP40 (outdoor 2-outlet)
+| Model Number     | Support by Home Assistant | Support by TP-Link's 'local only mode'     | Note                                     |
+|------------------|---------------------------|--------------------------------------------|------------------------------------------|
+| HS100            | ✅                        | ✅                                         |                                          |
+| HS103            | ✅                        | ✅                                         |                                          |
+| HS105            | ✅                        | ✅                                         |                                          |
+| HS110            | ✅                        | ✅                                         | supports consumption sensors             |
+| KP100            | ❓                        | ✅                                         |                                          |
```
bdraco, line 51: These work.  I have a few of these.  I'll add fixtures for them to `python-kasa` when I get a chance.

bdraco: https://github.com/python-kasa/python-kasa/pull/343

rytilahti:
I'd say it'd would make sense to add a note (and link to the instructions) about the ability to "re-enable" the local mode for those that are affected, but I'm also against adding tables and lists that list any device details.

Instead of listing the devices, I'd propose adding a note that all kasa-branded switches, dimmers, lights, led strips, and power strips besides some corner cases should be supported. A link to the supported devices page of python-kasa (titled "Devices known to work" or something like that) could be added to allow checking out if some specific device is known to work with the integration.

frenck:
Because there hasn't been any activity on this PR for quite some time now, I've decided to close it for being stale.

Feel free to re-open this PR when you are ready to pick up work on it again 👍

../Frenck

PR #22590, opened 2022-05-02
Device class for covers can't be set on the group

parautenbach:
## Proposed change
Missing information.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/group.markdown
+++ b/source/_integrations/group.markdown
@@ -74,7 +74,7 @@ binary_sensor:
       - binary_sensor.door_right_contact
 ```
 
-Example YAML configuration of a cover group:
+Example YAML configuration of a cover group (the [`device_class`](https://www.home-assistant.io/integrations/cover/#device-class) must be set via the [customize section](https://www.home-assistant.io/docs/configuration/customizing-devices/) or UI):
 
 ```yaml
 # Example configuration.yaml entry


source/_integrations/group.markdown
```diff
@@ -74,7 +74,7 @@ binary_sensor:
       - binary_sensor.door_right_contact
 ```
 
-Example YAML configuration of a cover group:
+Example YAML configuration of a cover group (the [`device_class`](https://www.home-assistant.io/integrations/cover/#device-class) must be set via the [customize section](https://www.home-assistant.io/docs/configuration/customizing-devices/) or UI):
```
frenck, line 77: What does that have to do with that example?

parautenbach: In the example above the one for covers in the docs, it shows that `device_class` is allowed for a binary sensor group, but the same isn't true for cover groups. This one:

```
binary_sensor:
  - platform: group
    name: "Patio Doors"
    device_class: opening  # <==
    entities:
      - binary_sensor.door_left_contact
      - binary_sensor.door_right_contact
 ```

Last night, I was trying to set `device_class` on a new-style cover group like this, but it failed, complaining about the use of `device_class`:

```
- platform: group
  name: "Garage Doors"
  device_class: garage # <==
  entities:
    - cover.lhs_garage_door
    - cover.rhs_garage_door
 ```

Because I looked at the binary sensor example I thought it should be possible for the others too. But maybe I'm wrong, because `device_class` isn't listed as an option under the configuration variables (even though the binary sensor example uses it). Whatever the case, it seems inconsistent that it would work for some groups and not others, and I just wanted to clear that up. 

My personal problem is solved, since I did what I wanted to by using my `customize.yaml`, but that's not the point here.

frenck: > Whatever the case, it seems inconsistent 

Documenting inconsistencies will not solve them. Lets fix those instead.

PR #22594, opened 2022-05-02
Add hint to using MQTT Discovery as alternative to ebusd integration

john30:
## Proposed change
Extend the documentation of ebusd integration with a hint to the more generic MQTT discovery approach with [ebusd](https://github.com/john30/ebusd).

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/ebusd.markdown
+++ b/source/_integrations/ebusd.markdown
@@ -13,6 +13,8 @@ ha_integration_type: integration
 
 Integration between [ebusd](https://github.com/john30/ebusd/) daemon for communication with eBUS heating systems, and Home Assistant using sensor component.
 
+Note: [ebusd](https://github.com/john30/ebusd/) also supports a more generic approach using the [MQTT discovery feature](https://www.home-assistant.io/docs/mqtt/discovery/), which allows to make almost everything available as entities in home assistant automatically, see [documentation and setup instructions here](https://github.com/john30/ebusd/wiki/MQTT-integration).
+
 ## Configuration
 
 Enable the sensor by adding the following to your `configuration.yaml` file:


source/_integrations/ebusd.markdown
```diff
@@ -13,6 +13,8 @@ ha_integration_type: integration
 
 Integration between [ebusd](https://github.com/john30/ebusd/) daemon for communication with eBUS heating systems, and Home Assistant using sensor component.
 
+Note: [ebusd](https://github.com/john30/ebusd/) also supports a more generic approach using the [MQTT discovery feature](https://www.home-assistant.io/docs/mqtt/discovery/), which allows to make almost everything available as entities in home assistant automatically, see [documentation and setup instructions here](https://github.com/john30/ebusd/wiki/MQTT-integration).
```
frenck, line 16: This page is about the ebusd integration, and is not meant to promote custom add-ons that are not supported by the Home Assistant project.

As they are not supported, we don't want to add these things, as it might sounds like we do.

../Frenck

john30:
@frenck I don't get your point. the extra hint is for using another way of integrating ebusd to HA, which is even a more generic way of doing the integration in the first place.

In general I find it hard for newbies to find out which approach is better: a dedicated integration or MQTT. Even knowing that there might be alternative ways of integration with pros and cons for each is not obvious when looking at the HA documentation from my point of view.

frenck:
>  the extra hint is for using another way of integrating ebusd to HA

Sure, but not related to the integration and thus does not belong there.

Anyways, the suggested addition is about something we don't support; thus won't be adding it.

john30:
> Anyways, the suggested addition is about something we don't support; thus won't be adding it.

wrong, you just approved it in https://github.com/home-assistant/home-assistant.io/pull/22584#event-6532586151

ok, so assuming there are two ways of doing an integration, e.g. ESPHome via [MQTT](https://www.home-assistant.io/docs/mqtt/discovery/#support-by-third-party-tools) or via dedicated [integration](https://www.home-assistant.io/integrations/esphome/).
where should alternatives be documented?
or how else is the user supposed to know of alternative approaches?

frenck:
> wrong, you just approved it in https://github.com/home-assistant/home-assistant.io/pull/22584#event-6532586151

That is a specific section that advertises external resources that support MQTT discovery. It is related and on par for that integration (MQTT).

> where should alternatives be documented?

Not in our documentation, as it's not related to core functionality.

john30:
would it be acceptable to change the hint text to point to the HAs own documentation of MQTT discovery instead?

frenck:
No. This is about as much as you can get in for a third-party add-on.

PR #22607, opened 2022-05-02
add unavailable status with an example

jugla:
## Proposed change

Let's have a sensor that gives a numerical value from an integration. If there is an issue, the sensor can be either none/unknown or unavailable . `unavailable` state is not described in documentation

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/configuration/templating.markdown
+++ b/source/_docs/configuration/templating.markdown
@@ -72,7 +72,7 @@ Not supported in [limited templates](#limited-templates).
 - Iterating `states` will yield each state sorted alphabetically by entity ID.
 - Iterating `states.domain` will yield each state of that domain sorted alphabetically by entity ID.
 - `states.sensor.temperature` returns the state object for `sensor.temperature` (avoid when possible, see note below).
-- `states('device_tracker.paulus')` will return the state string (not the object) of the given entity, `unknown` if it doesn't exist, `unavailable` if the object exists but not yet available (e.g. due to an integration internal processing) 
+- `states('device_tracker.paulus')` will return the state string (not the object) of the given entity, `unknown` if it doesn't exist, `unavailable` if the object exists but is not yet available. 
 - `is_state('device_tracker.paulus', 'home')` will test if the given entity is the specified state.
 - `state_attr('device_tracker.paulus', 'battery')` will return the value of the attribute or None if it doesn't exist.
 - `is_state_attr('device_tracker.paulus', 'battery', 40)` will test if the given entity attribute is the specified state (in this case, a numeric value). Note that the attribute can be `None` and you want to check if it is `None`, you need to use `state_attr('sensor.my_sensor', 'attr') == None`. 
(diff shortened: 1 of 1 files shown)


source/_docs/configuration/templating.markdown
```diff
@@ -72,7 +72,7 @@ Not supported in [limited templates](#limited-templates).
 - Iterating `states` will yield each state sorted alphabetically by entity ID.
 - Iterating `states.domain` will yield each state of that domain sorted alphabetically by entity ID.
 - `states.sensor.temperature` returns the state object for `sensor.temperature` (avoid when possible, see note below).
-- `states('device_tracker.paulus')` will return the state string (not the object) of the given entity or `unknown` if it doesn't exist.
+- `states('device_tracker.paulus')` will return the state string (not the object) of the given entity, `unknown` if it doesn't exist, `unavailable` if the object exists but not yet available (e.g. due to an integration internal processing)
```
frenck: What does "due to an integration internal processing" mean?

source/_docs/configuration/templating.markdown
```diff
@@ -119,6 +119,14 @@ Other state examples:
   Paulus is at {{ states('device_tracker.paulus') }}.
 {% endif %}
 
+#check state and compute remaining time before a train departure time (in minutes)
+{% if (states('sensor.train_departure_time') != "unknown" and states('sensor.train_departure_time') != "unavailable") %}
+{{ ( (as_timestamp(states('sensor.train_departure_time')) - as_timestamp(now()))/60.0 ) | round (2) }}
+{% else %}
+unknown
+{% endif %}
```
frenck: This looks like an extremely overcomplicated example that isn't really helpful in this location IMHO.

What is it you are trying to explain with this example?

frenck (changes requested):

source/_docs/configuration/templating.markdown
jugla: Hello,
The root cause of this PR is because, I've got some error in templating with error such as 'None Type'.
When I look deeper in HomeAssistant source code, I've seen that for templating, I got "3 states" : 
- `STATE_UNKNOWN` (const.py)
- `STATE_UNAVAILABLE` (const.py)
- the native state value

Unknown was quite clear (e.g. 'none')
Unavailable can be set by the integration via the method available (or _attr_available attribute) (cf. `helpers/entity.py`)
That's why I put  

> integration internal processing

If you have another proposal, I take

source/_docs/configuration/templating.markdown
jugla: You're right ...
I just wanted to show an example with a test with unknown/unavailable state. Is the following example better ?
```
#check state and make a processing
{% if (states('sensor.train_departure_time') != "unknown" and states('sensor.train_departure_time') != "unavailable") %}
{{ ... }}
{% else %}
unknown
{% endif %}
```

NB: due to unavailable state , the jinja statement 'is defined' does not work (indeed the variable exists)

source/_docs/configuration/templating.markdown
frenck: My proposal would be:

```suggestion
- `states('device_tracker.paulus')` will return the state string (not the object) of the given entity, `unknown` if it doesn't exist, `unavailable` if the object exists but is not yet available.
```

Reasoning: I have no idea what internal processing means. `unavailable` has nothing to do with Home Assistant internals.

source/_docs/configuration/templating.markdown
frenck: Why are you transforming `unavailable` into `unknown` in this template? I'm not following that.

Anyways, it can also be written as `{% if states('sensor.train_departure_time') in ("unavailable", "unknown") %}`

That said, `unknown` should not be needed to return at all (as that is a so called `None` value), so the whole `else` part can be dropped too.

source/_docs/configuration/templating.markdown
jugla: I upate

source/_docs/configuration/templating.markdown
jugla: I upate

frenck (changes requested):
Thanks, @jugla 👍 


EDIT: Wrong button, retry done below 🤦

frenck (approved):
Thanks, @jugla 👍

PR #23084, opened 2022-06-11
Add documentation for landisgyr_heat_meter

vpathuis:
## Proposed change

Add documentation for the new Landis+Gyr Heat Meter integration

## Type of change

- [x] Added documentation for a new integration I'm adding to Home Assistant (`next` branch).
  - [x] I've opened up a PR to add logo's and icons in [Brands repository](https://github.com/home-assistant/brands).

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/73363
- Link to parent pull request in the Brands repository: https://github.com/home-assistant/brands/pull/3457

## Checklist


homeassistant:
Hi @pvianag, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#23084).

Thanks!

frenck:
Something seems to went wrong, could you please git rebase this PR? Thanks! 👍

vpathuis:
Mm, weird. I only did a sync fork on my fork so I could do some minor changes. I'll look into it.

klaasnicolaas:
Your branch is descended from the `current` branch, to solve this problem you must rebase your branch to the `next` branch. If this doesn't work, you can always start over from a new branch 😉

vpathuis:
Hi @frenck, I think I fixed the problem. I also added a small commit. Do you think this is now ready to merge so it can be released together with the parent PR?

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,88 @@
+---
+title: Landis+Gyr Heat Meter
+description: Instructions on how to integrate your Landis+Gyr Heat Meter device into Home Assistant.
+ha_category:
+  - Energy
+  - Sensor
+ha_iot_class: Local Polling
+ha_config_flow: true
+ha_release: 2022.7
```
klaasnicolaas:
```suggestion
ha_release: 2022.9
```

klaasnicolaas (dismissed):

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,88 @@
...
+ha_platforms:
+  - sensor
+ha_integration_type: integration
+---
+
+The Landis+Gyr Heat Meter integration for Home Assistant allows you to read the meter values from Ultraheat devices by Landis & Gyr. These devices are commonly used for district heating. The integration has been tested with the Landis & Gyr Ultraheat 50 (UH50) type LUGCUH50. Other models are likely to work as well.
+
+The device is read through the optical interface. An (USB) IR reader is needed and connected to Home Assistant.
+
+<img src='/images/integrations/landisgyr_heat_meter/usb_ir_reader.png' />
```
frenck: Please use Markdown syntax to embed images.

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+<div class='note warning'>
```
frenck: Is this really a warning? This could also just be a normal text at the end of the sensors paragraph.

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,88 @@
...
+</div>
+
+{% include integrations/config_flow.md %}
+
+## Sensors
+
+The integration will create the following sensors:
+
+- heat usage (MWh)
+- volume usage (m3)
```
frenck:
```suggestion
- Heat usage (MWh)
- Volume usage (m3)
```

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+To be compatible with the Home Assistant energy units of measurement, heat usage is converted to MWh, from usage in GJ which is supplied by the device, using conversion factor: 1 GJ = 0.277778 MWh.
```
frenck:
```suggestion
To be compatible with the Home Assistant energy units of measurement, heat usage is converted to MWh, from usage in GJ, which is supplied by the device, using a conversion factor: 1 GJ = 0.277778 MWh.
```

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,88 @@
...
+## Sensors
+
+The integration will create the following sensors:
+
+- heat usage (MWh)
+- volume usage (m3)
+
+To be compatible with the Home Assistant energy units of measurement, heat usage is converted to MWh, from usage in GJ which is supplied by the device, using conversion factor: 1 GJ = 0.277778 MWh.  
+
+Further data that is read from the device is added as diagnostic entities:
```
frenck:
```suggestion
Further data that is read from the device is added as diagnostic entities:
```

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,88 @@
...
+- return temperature max previous year c
+- operating hours
+- fault hours
+- fault hours previous year
+- yearly set day
+- monthly set day
+- meter date time
+- measuring range m3ph
+- settings and firmware
+- flow hours
```
frenck:
```suggestion
- Heat usage measured in GJ, as is read from the device before conversion
- Ownership number
- Volume previous year (m3)
- Heat previous year (MWh)
- Error number
- Device number
- Measurement period minutes
- Power max kW
- Power max previous year kW
- Flow rate max m3ph
- Flow rate max previous year m3ph
- Flow temperature max c
- Return temperature max c
- Flow temperature max previous year c
- Return temperature max previous year c
- Operating hours
- Fault hours
- Fault hours previous year
- Yearly set day
- Monthly set day
- Meter date time
- Measuring range m3ph
- Settings and firmware
- Flow hours
```

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+## Energy Dashboard
+
+Either heat usage or volume usage can be used as "Gas" on the energy dashboard. If you want to supply a price per MWh, make sure to apply the conversion factor first.
+
+## Polling the device
+
+Polling is not automatic (see battery warning). This means that after adding the integration all values will initially be unknown.
+To read the values from the device, create an automation that will update one of the entities. The other entities will be updated as well.
```
frenck: That sounds off? We sound not make a broken sensor out of the box, not should we ever allow sensors that require manually created automations to function.

I would suggest restoring the state of this sensor (so we don't need to poll on startup) and poll, e.g., once a week or something like that.

frenck (changes requested):

source/_integrations/landisgyr_heat_meter.markdown
vpathuis: Agreed. This text was a leftover from an earlier version. State is now indeed restored on startup. It's also polled after adding the integration so the user should not see unknown values.

vpathuis: Is that acceptable? I'd rather leave the polling to the user, but if it's not compliant with HA standards, of course I will change this.

frenck: Well it doesn't violate any standards I think, but no everybody knows or uses YAML, confronting them with a YAML blob to get a working sensor, IMHO isn't the user experience we should or want to provide.

> State is now indeed restored on startup. It's also polled after adding the integration so the user should not see unknown values.

That means it updates once and never again? That is also weird. So it worked, but doesn't anymore (from a user perspective).

Alternatively, poll daily or weekly, and keep the entity disabled by default (and don't poll when it is). This allows the user to enable it if they want.

If they want another polling schedule, they can still opt to do that (e.g., do it weekly, monthly, or every minute).

vpathuis: Thanks for the feedback. Disabling entities won't be necessary I think, because that would just disable all functionality and in that case the user could just remove the integration. 

Creating the automation would not require using YAML. The example is just shownn for convenience. I now understand it has the oposite effect :-o.

The idea behind the chosen solution is:
- the user probably only wants to poll daily as a maximum, due to the risk of battery drain (this could only be replaced by a specialist, so expensive). This is ultimately for the user to decide. 
- the user wants as much control on getting the most relevant detail as possible, within this limitation, so probably polling late in the day, or week, or perhaps before and after the weekend. Allowing the user to do it manually will give max control. 

I also looked at options to set a time in config flow, but I felt that was not the right place. 

What I could do is add a setting in the config flow to enable or disable polling? If enabled, it will poll once weekly (without time option). If disabled, the user can do it manually using automations. How does that sound? 

This will require changes to the PR that is already merged. How does that work, should I just add another PR?

frenck: > Creating the automation would not require using YAML. The example is just shownn for convenience. I now understand it has the oposite effect :-o.

Things remain the same, if it is a polling integration, we need to provide a sane default and usable polling frequency. If one decides to change, that is up to them.

> What I could do is add a setting in the config flow to enable or disable polling?

That already exists. For every integration.

vpathuis: > Things remain the same, if it is a polling integration, we need to provide a sane default and usable polling frequency. If one decides to change, that is up to them.

Alright, understood.
 
> That already exists. For every integration.

Oh yes I see now. That's very convenient! :)
I'll apply the change and then also change the documentation.

vpathuis: I've submitted the change here: https://github.com/home-assistant/core/pull/77078
The integration will poll daily by default. It's what most people seem to do, so it's fine.
I've also changed the documentation accordingly. 
I look forward to your further suggestions or feedback!

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,90 @@
...
+
+Polling is by default done only once per day (and once right after adding the integration). Every time the Heat Meter values are read, battery time of the device will (supposedly) go down by about 30 minutes.
+
+For detailed control on polling and time of polling, consider disabling polling in the integration panel and poll manually. For instructions, see below.
+
+### Polling manually (optional)
+
+For detailed control on when the device is polled. disable default polling for this integration and create an automation that will update one of the entities. The other entities will be updated as well.
+
+If you're comfortable with yaml, this yaml could be used:
```
frenck:
```suggestion
For detailed control on when the device is polled, disable the default polling for this integration and create an automation that will update one of the entities (the other entities will be updated as well)

If you're comfortable with YAML, this game could be used:

```

source/_integrations/landisgyr_heat_meter.markdown
```diff
@@ -0,0 +1,90 @@
...
+trigger:
+  - platform: time
+    at: '23:30:00'
+condition: []
+action:
+  - service: homeassistant.update_entity
+    data: {}
+    target:
+      entity_id: sensor.heat_meter_heat_usage_gj
+mode: single
```
frenck:
```suggestion
alias: "Heat Meter manual update"
trigger:
  - platform: time
    at: "23:30:00"
action:
  - service: homeassistant.update_entity
    target:
      entity_id: sensor.heat_meter_heat_usage_gj
```

frenck (approved):
Thanks, @vpathuis 👍

PR #23168, opened 2022-06-22
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -7,8 +7,8 @@ on:
 jobs:
   lock:
     permissions:
-      issues: write  # for dessant/lock-threads to lock issues
-      pull-requests: write  # for dessant/lock-threads to lock PRs
+      issues: write
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - uses: dessant/lock-threads@v3
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -4,8 +4,14 @@ on:
   schedule:
     - cron: "0 * * * *"
 
+permissions:
+  contents: read
+
 jobs:
   lock:
+    permissions:
+      issues: write  # for dessant/lock-threads to lock issues
+      pull-requests: write  # for dessant/lock-threads to lock PRs
     runs-on: ubuntu-latest
     steps:
       - uses: dessant/lock-threads@v3
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -4,9 +4,6 @@ on:
   schedule:
     - cron: "0 * * * *"
 
-permissions:
-  contents: read
-
 jobs:
   lock:
     permissions:
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -4,8 +4,14 @@ on:
   schedule:
     - cron: "0 * * * *"
 
+permissions:
+  contents: read
+
 jobs:
   stale:
+    permissions:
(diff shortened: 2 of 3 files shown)


frenck:
I'm kinda amazed to see the same stuff in here that was declined/commented on in this PR: 
https://github.com/home-assistant/frontend/pull/12989. Maybe resolve that one first?

.github/workflows/lock.yml
```diff
@@ -4,8 +4,14 @@ on:
   schedule:
     - cron: "0 * * * *"
 
+permissions:
```
frenck: This should be set in repo settings, not here.

.github/workflows/stale.yml
```diff
@@ -4,8 +4,14 @@ on:
   schedule:
     - cron: "0 * * * *"
 
+permissions:
```
frenck: This should be set in repo settings, not here.

.github/workflows/test.yml
```diff
@@ -2,6 +2,9 @@ name: home-assistant.io Test
 
 on: [push, pull_request]
 
+permissions:
```
frenck: This should be set in repo settings, not here.

frenck (changes requested):

.github/workflows/lock.yml
naveensrinivasan: Took care of it. Thanks

.github/workflows/stale.yml
naveensrinivasan: Took care of it. Thanks

.github/workflows/test.yml
naveensrinivasan: Took care of it. Thanks

.github/workflows/lock.yml
```diff
@@ -6,6 +6,9 @@ on:
 
 jobs:
   lock:
+    permissions:
+      issues: write  # for dessant/lock-threads to lock issues
+      pull-requests: write  # for dessant/lock-threads to lock PRs
```
frenck: It the only thing in this workflow, so a bit redundant.

```suggestion
      issues: write
      pull-requests: write
```

.github/workflows/stale.yml
```diff
@@ -6,6 +6,9 @@ on:
 
 jobs:
   stale:
+    permissions:
+      issues: write  # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
```
frenck:
```suggestion
      issues: write
      pull-requests: write
```

.github/workflows/test.yml
```diff
@@ -2,6 +2,7 @@ name: home-assistant.io Test
 
 on: [push, pull_request]
 
+
```
frenck: Non-related change
```suggestion
```

frenck (approved):

PR #23261, opened 2022-07-03
Update LIFX for config entry per device reliability refactor

bdraco:
## Proposed change

Update LIFX for config entry per device reliability refactor

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/74316

## Checklist


frenck (approved):

Djelibeybi:
I need to double check what happens when you upgrade to the new discovery method but do not have an interface on the same subnet as the LIFX devices but either way, we should probably document that specific scenario so that we have something explicit to point people to in the docs when they all start opening issues and/or threads on the forum. :)

bdraco:
> I need to double check what happens when you upgrade to the new discovery method but do not have an interface on the same subnet as the LIFX devices but either way, we should probably document that specific scenario so that we have something explicit to point people to in the docs when they all start opening issues and/or threads on the forum. :)

Good idea.  Please inline suggestions once you have results.

Thanks!

Djelibeybi:
> Good idea. Please inline suggestions once you have results.

Will do. What's the expected/intended release version for this?

bdraco:
> > Good idea. Please inline suggestions once you have results.
> 
> 
> 
> Will do. What's the expected/intended release version for this? 

2022.8.0. If all goes well we can likely switch out the underlying library in 2022.9.0

Djelibeybi:
Things started to break with 2022.4.0 so I'm not going to be able to inline. I'm busy writing up the scenarios so I can simplify it down and we can discuss how to document it properly without confusing folks.

bdraco:
> Things started to break with 2022.4.0 so I'm not going to be able to inline. I'm busy writing up the scenarios so I can simplify it down and we can discuss how to document it properly without confusing folks.

Another option would be to send a PR to this branch

Djelibeybi:
Righto. Let me finish testing first. I had to go back a fair way to find the last version of HASS that would discover and instantiate bulbs with broken networking. :)

bdraco:
> > Things started to break with 2022.4.0 so I'm not going to be able to inline. I'm busy writing up the scenarios so I can simplify it down and we can discuss how to document it properly without confusing folks.
> 
> Another option would be to send a PR to this branch

Or maybe just another PR after this one to add it to keep it simple

Djelibeybi:
> Or maybe just another PR after this one to add it to keep it simple

Yeah, I think it's going to all be under the heading of "Upgrading from 2022.4 or older" because that's when discovery of actual devices changed.

Djelibeybi:
The problem of previously accessible LIFX devices being inaccessible due to a lack of locally connected interface only occurs from v2022.5.0 until the parent PR is merged. The changes in that PR restore the pre-2022.5 experience because devices discovered via Zeroconf/HK are connected to directly using whatever inteface is enabled. 

Given that routing and firewalls would have to be properly configured to permit this pre-2022.5, it'll work again now. So migrating hard-coded configurations from `configuration.yaml` to config entries works fine too.

Not sure if we need to call this out if it's not going to be a problem much longer. Easier just to tell people to upgrade because it's fixed in 2022.8.

bdraco:
> Not sure if we need to call this out if it's not going to be a problem much longer. Easier just to tell people to upgrade because it's fixed in 2022.8.

This is the way

PR #23807, opened 2022-08-18
Update sun condition

DubhAd:
## Proposed change

At some point the sun condition docs had their order reworked so the problem with the "after dark" condition is now confusing. I've restructured it and make it clearer that the example isn't valid.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/scripts/conditions.markdown
+++ b/source/_docs/scripts/conditions.markdown
@@ -396,15 +396,6 @@ condition:
   after_offset: "-01:00:00"
 ```
 
-This is 'when dark' - equivalent to a state condition on `sun.sun` of `below_horizon`:
-
-```yaml
-condition:
-  - condition: sun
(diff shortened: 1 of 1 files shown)


source/_docs/scripts/conditions.markdown
```diff
@@ -396,25 +396,25 @@ condition:
...
 condition:
   - condition: sun
-    after: sunrise
-    before: sunset
+    after: sunset
+    before: sunrise
 ```
 
-We cannot use both keys in this case as it will always be `false`.
+However, we cannot use both keys in this case as it will always be `false`, so instead you have to use:
```
frenck, line 417: This is actually a bug, this part of the documentation should be removed, as it is not the intended functionality.

frenck (changes requested):

PR #23819, opened 2022-08-20
Docs: add detail to if-actions

Pixdigit:
## Proposed change

As this [forum post](https://community.home-assistant.io/t/automation-with-for-each-and-if/422732/8) makes clear. One can not use a template in the `entity_id` in the condition of an if when checking for a state. This change adds a note in the documentation about this behaviour and includes a work-around.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -602,12 +602,14 @@ script:
 When testing for a state, you cannot use a template for the `entity_id` as it has to be a constant.
   
 If you need to check a dynamically changing entity use a template check instead:
+{% raw %}
 ```yaml
 if:
   condition: template
   value_template: "{{ is_state(repeat.item, 'off') }}"
 
 ```
+{% endraw %}
 </div>
 
 This action supports nesting, however, if you find yourself using nested if-then
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -598,6 +598,18 @@ script:
           message: "Skipped cleaning, someone is home!"
 ```
 
+<div class='note'>
+When testing for a state, you cannot use a template for the `entity_id` and it has to be a constant.
+  
+If you need to check a dynamically changing entity use a template check instead:
+```yaml
+if:
+  condition: template
+  value_template: "{{ is_state(repeat.item, 'off') }}"
+
+```
+</div>
+
 This action supports nesting, however, if you find yourself using nested if-then
 actions in the `else` part, you may want to consider using
 [choose](#choose-a-group-of-actions) instead.
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -608,7 +608,7 @@ If you need to check a dynamically changing entity use a template check instead:
 ```yaml
 if:
   condition: template
-  value_template: "{{ is_state('zone_home', 0) }}"
+  value_template: "{{ is_state('zone.home', 0) }}"
 ```
 {% endraw %}
 
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -599,7 +599,7 @@ script:
 ```
 
 <div class='note'>
-When testing for a state, you cannot use a template for the `entity_id` and it has to be a constant.
+When testing for a state, you cannot use a template for the `entity_id` as it has to be a constant.
   
 If you need to check a dynamically changing entity use a template check instead:
 ```yaml
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -599,17 +599,19 @@ script:
 ```
 
 <div class='note'>
+
 When testing for a state, you cannot use a template for the `entity_id` as it has to be a constant.
   
 If you need to check a dynamically changing entity use a template check instead:
+  
 {% raw %}
 ```yaml
 if:
   condition: template
-  value_template: "{{ is_state(repeat.item, 'off') }}"
-
+  value_template: "{{ is_state('zone_home', 0) }}"
 ```
 {% endraw %}
+
 </div>
 
 This action supports nesting, however, if you find yourself using nested if-then


homeassistant:
Hi @Pixdigit, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#23819).

Thanks!

source/_docs/scripts.markdown
```diff
@@ -598,6 +598,22 @@ script:
           message: "Skipped cleaning, someone is home!"
 ```
 
+<div class='note'>
+
+When testing for a state, you cannot use a template for the `entity_id` as it has to be a constant.
```
frenck, line 603: I'm not following, the forum post is about a condition, not if/else. 

Rhe condition is documented else where? What has this to do with the if/else statement?

frenck (changes requested):

Pixdigit:
@frenck The title of the post is "Automation with for_each and if" so it is clear to me the post is about using `if` in a specific case. The relevance becomes especially apparent when reading the last two responses to that post. I had a similar situation where I had an `if` within a `for` and wanted to check the state of the item in question. However as the last answer makes clear: it is not possible to use a jinja template for the `entity_id` in an `if` entirely.

Though this applies generally to conditions and triggers which is not explicitly marked in the documentation for `if`. It would appear to me that this is, what is referred to [limited templates](https://www.home-assistant.io/docs/configuration/templating/#limited-templates) elsewhere. However there are multiple issues with the documentation of that specific behaviour.

source/_docs/scripts.markdown
Pixdigit: The title of the post is "Automation with for_each and if" so it is clear to me the post is about using if in a specific case. The relevance becomes especially apparent when reading the last two responses to that post. I had a similar situation where I had an if within a for and wanted to check the state of the item in question. However as the last answer makes clear: it is not possible to use a jinja template for the entity_id in an if entirely.

Though this applies generally to conditions and triggers which is not explicitly marked in the documentation for if. It would appear to me that this is, what is referred to [limited templates](https://www.home-assistant.io/docs/configuration/templating/#limited-templates) elsewhere. However there are multiple issues with the documentation of that specific behaviour.

frenck: > The title of the post is "Automation with for_each and if"

It is not? This is a separate/new chapter.

Pixdigit: screenshot of the forum post

Correct me if I am wrong, but isn't that the title of the post?

frenck: Sure, but I am still not sure what this has to do with this chapter of the documentation in any way. I've read the forum post, which is about using a template for the `entity_id` in a state condition. That has nothing to do with if-actions. Sure, the if action accepts conditions, but your comment and question are specifically about the state condition, not the if action.

Pixdigit: So the condition of an action and of an if are exactly the same? Sorry, if this got a little heated but spending days on fixing such basic issues in my config has been really draining.

What follows is a fairly long insight into my way of thinking and interacting with Home Assistant. It is not meant as a criticism. Feel free to skip to the Tl;Dr.

For me the documentation in general is very confusing and I think this issue is just a manifestation of that. As a (admittedly junior) developer I interact often with documentation. Even though Home Assistant is somewhat larger compared to the biggest  frameworks I worked with (like Angular or Laravel) I struggle to get even basic stuff done. My guess is that this is due to the structure of the documentation. It is fairly fragmented, doesn't have a navigable over view and is not very searchable.
For example my case: I want my motion sensors to control some rooms and adjust the brightness the lights are turned on with depending of time of day.
Because I want something to be triggered automatically I need an Automation. Though because I want this for multiple rooms I and I don't want to repeat code I need a blueprint. Automations are written in YAML with Jinja2 templating and the schema for actions are called Scripting. 
While I have some experience with YAML and Jinja is not very hard to understand the basics, thought for some details I have the documentation open. Though as Home Assistant supplements the Jinja in-built functions I also need the Home Assistant Template documentation open. As I am writing a blueprint for an automation I also need the Automation and Blueprint documentation open. Then for the actions I need the scripting documentation. And then some documentation for auxiliary things like the Selectors in Blueprints.

With now ~7 tabs open and constantly switching between them it is often unclear what aspect belongs to what documentation, especially as they all interweave. To come back to my original problem: it came to me as a total surprise that the `entity_id` of an `if` isn't part of the `if`. It isn't even part of scripting but instead belongs to the Template documentation.

Now with all that said: this isn't supposed to be a rant or a critique of the countless hours of work you and other volunteers have spent not only making but also documenting and supporting the brilliant ecosystem that is Home Assistant. Though with the almost weeks as a somewhat experienced developer writing what is supposed to be a simple automation I feel like I am doing something wrong or that my use case is plainly too niche. I would be extremely thankful if you have the time and motivation to point out issues with my approach though I can very much understand if you don't have the time or plainly the motivation to deal with my personal problems.

Tl;Dr: I have trouble with the documentation and this specific behaviour has stumped me and others. While I think I understand what you mean and in that regard agree with you, I would still recommend _some_ change to the documentation.

As for the PR: I would still favour at least expanding the documentation in Templating to more explicitly document where and where not (most) templating can be used. I.e. not in conditions and triggers. Or even more explicitly that they can not be used as `entity_id` in if-actions, conditions for actions etc. Also a notice in the "Important Template Rules" warning, that not all of templating can be used everywhere would be beneficial as I regard it as a fairly heavy limitation of templating.

Feel free to resolve the conversation and/or close the PR. I don't think I would do a PR for the proposed changes unless there is prior agreement that the change is warranted.

frenck:
> So the condition of an action and of an if are exactly the same? 

An if-action accepts one or more conditions, if that pass, the sequence of actions is run. So, yes is the answer.

How to use conditions, can be found here: <https://www.home-assistant.io/docs/scripts/conditions>

That page contains a list of all conditions we have, including lots of examples.

Pixdigit:
I don't think this is going anywhere constructive so I'm just gonna close it and leave it at that.

frenck:
Ok, @Pixdigit 👍 

I think you have slight misunderstanding between actions and conditions (and what it means within the context of the if-action).

Nevertheless, thanks for being willing to contribute ❤️ 
Always appreciated!

../Frenck

Pixdigit:
@frenck 

>Tl;Dr: I have trouble with the documentation and this specific behaviour has stumped me and others. While I think I understand what you mean and in that regard agree with you, I would still recommend some change to the documentation.

Yes, that is what I was referring here.

>As for the PR: I would still favour at least expanding the documentation in "Templating" to more explicitly document where and where not (most) templating can be used. I.e. not in conditions and triggers. Or even more explicitly that they can not be used as entity_id in if-actions, conditions for actions etc. Also a notice in the "Important Template Rules" warning, that not all of templating can be used everywhere would be beneficial as I regard it as a fairly heavy limitation of templating.

Though as you didn't respond to my proposed changes to prevent such misunderstandings in the future I thought you either didn't read them or didn't agree. But having to repeat myself without sounding passive aggressive is difficult.

frenck:
There is no templating at all in this if-action. The if-action takes a list of conditions and a list of actions. Nothing more, nothing less.

Sure, you can use templates in some places, in some conditions. That is documented on the conditions page linked above (and several times on this actions page as well).

I'm not following any of the suggestions regarding templating you are trying to make in this section, as none of it applies here.

Pixdigit:
I mean more generally. A condition's `entity_id` can not be a template (or is it possible to use a  [Limited Templates](https://www.home-assistant.io/docs/configuration/templating/#limited-templates)?). It seems to me that this is currently not documented anywhere.

frenck:
> A condition's entity_id can not be a template

That is correct. The documentation doesn't list that, or does it somewhere?
Limited templates are available where mentioned, in this case, it isn't stated they can be used.

> It seems to me that this is currently not documented anywhere.

That is correct, we document when templates can be used. We document possibilities, not what isn't possible.

Pixdigit:
Oh, that is it then. My assumption was one could use templates (almost) anywhere. Would it then make sense to add a sentence to [Important Template Rules](https://www.home-assistant.io/docs/configuration/templating/#important-template-rules) of the likes: "Make you can use a template for the value you are trying use the template on"?

frenck:
That sounds odd IMHO, that applies to any value? You can't input a number, text string, array, mapping, float, boolean (or any other type) in all places either? Templates are not different from that.

Pixdigit:
Yes, but a template can return a number, string, float etc. And it seems reasonable to use templates on almost any value. Or the other way around: it's not always clear how things are intended to be done when they can not be solved via templating.

While I agree that in hindsight it should be obvious, the forum post indicates that I am not the only one having difficulties judging when and where what features can be used.

frenck:
> Yes, but a template can return a number, string, float etc.

No, it can't. Templates always return a string 😉 
We try to do smart stuff, but in the end, a template results in a string.

Anyways, you miss my point. IMHO it makes no sense to add such a waring about template, but not such a warning about e.g., booleans. The point is: Any input has documented values it can accept. If templates are accepted, that is documented. If not, it isn't documented (just like we don't document a value wanting a number, can't take a boolean).

> And it seems reasonable to use templates on almost any value.

That is not true, it can be rarely used. If I guessed, it would be less than 1% of all available options values in Home Assistant.

> While I agree that in hindsight it should be obvious, the forum post indicates that I am not the only one having difficulties judging when and where what features can be used.

Sure, it isn't an easy subject. I'm happy to see improvements or corrections. However, this PR is not correct and is placed incorrectly as well; as again, the if-action has no relation to this all.

Pixdigit:
I can not think of anything else to change except maybe to the documentation of the if-action. 

> This action allow you to conditionally (if) run a sequence of actions (then) and optionally supports running other sequence when the condition didn’t pass (else).

Would it make sense to rephrase to something along the lines of:

This action allows you to run a sequence of actions if a [condition](https://www.home-assistant.io/docs/scripts/conditions#state-condition) is met and optionally […]

I think this would make at least clear that the condition in an `if` is not inherent to an `if` but the same as conditions elsewhere.

>However, this PR is not correct and is placed incorrectly as well; as again, the if-action has no relation to this all.

Yes, but before spending time and effort on another change I want to make sure it's actually wanted.

frenck:
Again, nothing you have been describing is related to the if-action in any way.

Pixdigit:
Does the if-action not contain a condition?

frenck:
Sure, please refer to the conditions documentation.

Pixdigit:
I just think the phrasing "This action allow you to conditionally (if) run a sequence of actions" doesn't make it obvious that the condition is not inherent to the if-action but a concept that each action supports.

frenck:
? That comment makes no sense to me. sorry.

Pixdigit:
>This action allow you to conditionally (if) run a sequence of actions (then) and optionally supports running other sequence when the condition didn’t pass (else).

In the documentation in Scripting in the Section about If-Then that sentence is used. To me this indicates that this allows one to conditionally run one or more actions. Though the key focus there is the _sequence_ of actions as each action can already be conditionally be executed via the condition attribute. So the purpose of the `if` is to group several actions and then use a single condition on the whole sequence (and also provide an `else`). I would rephrase that sentence and also include a link to the condition documentation to avoid confusion as to what the `if` is for.

frenck:
>  To me this indicates that this allows one to conditionally run one or more actions.

Correct.

Automations and scripts are about three things.

- [Triggers](https://www.home-assistant.io/docs/automation/trigger) (automations only)
- [Conditions](https://www.home-assistant.io/docs/scripts/conditions)
- [Actions](https://www.home-assistant.io/docs/scripts/)

More about the basics be found here: <https://www.home-assistant.io/docs/automation/basics/>

The if-action is described as: 

> This action allow you to conditionally (if) run a sequence of actions (then) and optionally supports running other sequence when the condition didn’t pass (else).

I really really don't understand in this context what is unclear about that, how it it related to any of the template edits you wanted to make to the if-condition, nor do I understand the latter messages.

> Though the key focus there is the sequence of actions as each action can already be conditionally be executed via the condition attribute. 

Sure, but maybe you want to run parts of automation conditionally? The automation condition decides if an automation is triggered. Please read the documentation on automation basics.

I'm ending this conversation here, as I won't be accepting this PR, nor do I agree with many of the statements made. Lastly, it is almost might night on my Sunday, so going to finish a beer and get some sleep 😉 

I would recommend on reading up on the automation basics and using the community forums/discord for support.

../Frenck

Pixdigit:
As a last remark from my side: I am not advocating for this specific PR to be accepted. Though I think there are some improvements to be made as at leat I and the person from the forum had similar misunderstandings. For my recent discussion this is maybe the wrong place but I was asking if it would make sense to make _another_ PR with different changes, if that wasn't clear enough.

Anyway have a lovely night and thank you for your patience.

PR #23900, opened 2022-08-26
Note that force_update affects "last updated"

deviantintegral:
## Proposed change

While the current docs explain why `force_update` is useful for graphs, they don't specifically call out that it's needed to keep "Last Updated" in the UI current. When searching the current text, there's no match for "updated" which is what I expect an end user would look for when trying to troubleshoot this.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/sensor.mqtt.markdown
+++ b/source/_integrations/sensor.mqtt.markdown
@@ -144,7 +144,7 @@ expire_after:
   type: integer
   default: 0
 force_update:
-  description: Sends update events even if the value hasn't changed. Useful if you want to have meaningful value graphs in history.
+  description: Sends update events even if the value hasn't changed. Useful if you want to have meaningful value graphs in history. Also required to see real "Last Updated" values.
   required: false
   type: boolean
   default: false


source/_integrations/sensor.mqtt.markdown
```diff
@@ -144,7 +144,7 @@ expire_after:
   type: integer
   default: 0
 force_update:
-  description: Sends update events even if the value hasn't changed. Useful if you want to have meaningful value graphs in history.
+  description: Sends update events even if the value hasn't changed. Useful if you want to have meaningful value graphs in history. Also required to see real "Last Updated" values.
```
frenck, line 147: That sentence is misleading and technically incorrect. The fact is that forcing state updates can generate fake "Last updates" even if the state engine didn't update at all.

frenck (changes requested):
Small comment above

source/_integrations/sensor.mqtt.markdown
deviantintegral: Interesting - I wouldn't have expected that given the current text. I'm guessing this is most likely when restarting HA?

I compared to a zigbee2mqtt configured device, and the only other difference I could see affecting the last updated value might be `availability_mode` and `availability`?

What I'd like to do is get the docs to is a state where a user knows how to configure a sensor so that graph updates correct, and last updated works. Perhaps we actually remove or rewrite the note on "meaningful value graphs..." as I imagine that would lead more users to turn that on not realizing there's unintended consequences.

frenck: > What I'd like to do is get the docs to is a state where a user knows how to configure a sensor so that graph updates correct, and last updated works

That is what is provided out of the box, force_update has nothing to do with that.

deviantintegral: Your comment prompted me to take a deep dive into this.

What I was missing is that "Last updated" as shown here is a timestamp showing when [entity attributes are changed](https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L1086). This is confusing for entities that don't have any attributes, like this one. It's also confusing when they do have attributes, because the updated time isn't shown in the same container as attributes.

image

This is [documented reasonably](https://www.home-assistant.io/docs/configuration/state_object/) and I think it's just the frontend that needs improvement.

I think the above "meaningful value graphs in history" is misleading. I turned off `force_update` on a rain sensor (no rain here 😢 ), and the history graphs in HA seem to work fine. https://github.com/home-assistant/core/issues/11158 is the issue adding that text, but the original change only mentions _influxdb_ graphs and not HA's own graphs: https://github.com/home-assistant/core/pull/6492#issuecomment-285278748

I thought about adding a "last seen" entity attribute, but https://github.com/Koenkk/zigbee2mqtt/pull/7683 indicates entity attributes in general should be avoided.

My thoughts for next steps:

1. Replace this PR with one that updates "graphs" in the mqtt, rest, and any other integrations with more specific text referencing the influxdb integration, the state object docs, and a performance warning.
2. File a frontend issue to clarify Last Updated in the UI.
3. File an influxdb integration issue to consider adding something that walks through sensors periodically and reports them. Given the implications of `force_update`, it seems to me the "missing data" issue should be solved in that integration instead.
4. In the rtl_433 autodiscovery script, we'll change the "Time" entities being proposed at https://github.com/merbanan/rtl_433/pull/2102 to be disabled by default like how zigbee2mqtt does it.

Aside: It was _really neat_ to be able to spin up a local of HA in a dev container and point it to my mqtt broker for step debugging real messages, without affecting my actual home install. Fun!

deviantintegral:
> Replace this PR with one that updates "graphs" in the mqtt, rest, and any other integrations with more specific text referencing the influxdb integration, the state object docs, and a performance warning.

Done at https://github.com/home-assistant/home-assistant.io/pull/24092

> File a frontend issue to clarify Last Updated in the UI.

Discussion filed at https://github.com/home-assistant/frontend/discussions/13687

> File an influxdb integration issue to consider adding something that walks through sensors periodically and reports them. Given the implications of force_update, it seems to me the "missing data" issue should be solved in that integration instead.

This has already been discussed and proposed and declined at https://github.com/home-assistant/core/pull/42007 with continuing discussion at https://community.home-assistant.io/t/send-metrics-to-influxdb-at-regular-intervals/9096.

> In the rtl_433 autodiscovery script, we'll change the "Time" entities being proposed at https://github.com/merbanan/rtl_433/pull/2102 to be disabled by default like how zigbee2mqtt does it.

I commented there so we'll do that before that PR is merged upstream.

source/_integrations/sensor.mqtt.markdown
frenck: > This is confusing for entities that don't have any attributes, like this one.

That statement is incorrect. All entities have these properties, and are documented as well.

<https://www.home-assistant.io/docs/configuration/state_object/>

Please note: Home Assistant is not a statistics collector, but a state engine. Changing a state from value X to the same value X in an update is not a considered a state change. 

There are many reasons for that, and is working as intended.

For performance, SD life-time, storage sizes, and quite a bunch of other reasons: `force_update` is a really bad thing, that better is avoided and should maybe even removed from HA completely.

PR #24092, opened 2022-09-11
Clarify force_update is not needed for History graphs

deviantintegral:
## Proposed change

I believe the documentation is misleading by indicating that `force_update` is needed for "graphs in general", when notably it's InfluxDB that is hard to generate graphs with when there are missing datapoints. As well, given how `force_update` causes state updates, it means that a user who enables it for many devices may unwittingly be causing much more significant CPU or IO overhead than they realize.

## Type of change

## Additional information

- Follows up from https://github.com/home-assistant/home-assistant.io/pull/23900

## Checklist


source/_integrations/sensor.mqtt.markdown
```diff
@@ -148,7 +148,7 @@ expire_after:
   type: integer
   default: 0
 force_update:
-  description: Sends update events even if the value hasn't changed. Useful if you want to have meaningful value graphs in history.
+  description: Sends update events even if the value hasn't changed. Useful if you want to have meaningful value graphs in integrations like [InfluxDB](/integrations/influxdb). Use this option sparingly as it has performance and disk IO implications. This is not needed for Home Assistant's built-in graphs.
```
frenck, line 151: I am not following this one? Why would it be needed for Influx but not HA graphs? And what makes it needed for Influx? 

Would this be better written on the influxdb page instead?

I kinda feel saying: "if you want to have meaningful graphs" (which was also in the old description) is not good, as is kinda implies our current graphing is meaningless (which is not true).

frenck (changes requested):

source/_integrations/sensor.mqtt.markdown
deviantintegral: Influx and Grafana will not draw a graph if all of the datapoints exist outside the selected time range. I turned off `force_update`, and then did a query for the last 10 minutes. Home Assistant generates a line with the current results, while Grafana says "no data". `fill(previous)` won't draw a graph either unless there's a datapoint in the time frame.

image

10 minutes is not a big deal, but if a sensor doesn't change over days or weeks it's hard to generate decent graphs.

> Would this be better written on the influxdb page instead?

Sure, but then what would we put here?

> as is kinda implies our current graphing is meaningless (which is not true).

Yes, it totally does, and until I looked into it deeply I thought exactly that.

I think the PR at https://github.com/home-assistant/core/pull/42007 was closest to a good long term solution, as it would match what other systems (collectd, telegraf, etc) do, which is write a datapoint to influx for every observation. It was also good having it optional to make loading an SD card with writes opt-in. It would also allow us to remove `force_update` support from this and the rest integration. In the short-term, just getting the docs improved would be a good step forward.

deviantintegral:
I reviewed a few open docs PRs. Any next steps or anything I can do to help move this along?

frenck:
I'm going to friendly decline this PR. Let me explain why:

I'm ok with adding documentation to the InfluxDB page on how there is a mismatch in things like these between HA & InfluxDB.

However, Home Assistant and InfluxDB serve different purposes and goals. Home Assistant is foremost a state/automation engine and not a time series database.

Adding comments like this on non-InfluxDB-related pages can cause confusion for users that are not running or depend on this specific behavior and may thus lead to making incorrect configuration decisions.

As this PR mainly edits two places in non-InfluxDB pages, I'll go ahead and close it. I think opening a new PR towards the InfluxDB integration documentation would be discussable and nice to see.

../Frenck

PR #24108, opened 2022-09-12
Integrations: fix category getting reset

mislav:
When browsing integrations in a specific category, for example https://www.home-assistant.io/integrations/#health, the category would sometimes get reset to `#all` without the user's intent to do so. This is because the search bar is configured to automatically receive focus on page load and also handles `keyup` events, which will activate on _any_ keyboard interaction with the browser (such as switching browser tabs or navigating back/forward using the keyboard).

Automatically giving focus to an element on the page is considered inaccessible[^1] and should be only used when the search bar is absolutely the main element on the page, such as the search bar on Google Search page, for example.

The solution applied here is twofold:

1. Remove autofocus behavior from search field; and
2. Replace `keyup` handler with `input` handler which only triggers when the field's value has been changed[^2], and not on virtually any keystroke that happened while the field was focused.

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#autofocus
[^2]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event


homeassistant:
Hi @mislav, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#24108).

Thanks!

frenck (changes requested):
> Automatically giving focus to an element on the page is considered inaccessible[1](https://github.com/home-assistant/home-assistant.io/pull/24108#user-content-fn-1-c9a4dca73785b28e6df2486f9f61af48) and should be only used when the search bar is absolutely the main element on the page, such as the search bar on Google Search page, for example.

I would argue that that is the case on this page. The page itself doesn't provide much (except a couple of highlighted featured integrations).

Its main goal is to start searching. So let's restore the autofocus.

mislav:
Thank you for the review, and sorry for the delayed response. I have restored the autofocus attribute, but I still maintain my position that using autofocus like this leads to inaccessible sites. In addition to users of screen readers becoming disoriented, people who browse websites without a mouse will find themselves unable to scroll the page up/down using the keyboard unless they manually remove the focus from the search box.

Note that I have not restored the JS-based Internet Explorer workaround for autofocus. IE has [supported autofocus since 2012](https://caniuse.com/autofocus).

frenck:
As you stated in your PR description:

> should be only used when the search bar is absolutely the main element on the page, such as the search bar on Google Search page, for example.

When entering this page, it is absolutely the main element on the page.

Yet, you now disagree with that by saying:

> but I still maintain my position that using autofocus like this leads to inaccessible sites. In addition to users of screen readers becoming disoriented, people who browse websites without a mouse will find themselves unable to scroll the page up/down using the keyboard unless they manually remove the focus from the search box.

So your original reasoning in the PR description was false?

../Frenck

mislav:
My original reasoning wasn't well-explained. However, I did link to an authoritative resource on this, [which is the MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#autofocus):

> **Warning:** Automatically focusing a form control can confuse visually-impaired people using screen-reading technology and people with cognitive impairments. When autofocus is assigned, screen-readers "teleport" their user to the form control without warning them beforehand.
>
> Use careful consideration for accessibility when applying the autofocus attribute. Automatically focusing on a control can cause the page to scroll on load. The focus can also cause dynamic keyboards to display on some touch devices. While a screen reader will announce the label of the form control receiving focus, the screen reader will not announce anything before the label, and the sighted user on a small device will equally miss the context created by the preceding content.

For me, for example, it's really interesting to browse HA integrations by category instead of searching by name. I am a sighted user, however, so categories like “Health” and “Energy” are always easily visible for me in the sidebar. However, the sidebar appears earlier in the DOM than the search bar itself. For screen reader users, the autofocus property will cause their screen reading position to immediately jump to the search bar, skipping the categories list and making them essentially “invisible”.

Some more links:
- https://www.a11yproject.com/checklist/#avoid-using-the-autofocus-attribute
- https://brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/

PR #24187, opened 2022-09-20
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/add_prs_to_project.yml
+++ b/.github/workflows/add_prs_to_project.yml
@@ -5,8 +5,11 @@ on:
     types:
       - opened
 
+permissions: {}
 jobs:
   add-to-project:
+    permissions:
+      repository-projects: write # to add to projects 
     name: Add new PR to project
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,9 @@ name: home-assistant.io Test
 
 on: [push, pull_request]
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   markdown-lint:
     name: Lint Markdown


homeassistant:
Hi @sashashura, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant.io#24187).

Thanks!

.github/workflows/add_prs_to_project.yml
```diff
@@ -5,8 +5,11 @@ on:
     types:
       - opened
 
+permissions: {}
```
frenck, line 8: This is our default on the repo level  (well, the default is read, but since this is open source, everybody can read it already. This doesn't add any value).

```suggestion
```

.github/workflows/add_prs_to_project.yml
```diff
@@ -5,8 +5,11 @@ on:
...
 jobs:
   add-to-project:
+    permissions:
+      repository-projects: write # to add to projects
```
frenck, line 12: This is invalid, this doesn't use this permission at all.

You are actually lowering security by handing out unneeded write permissions.

.github/workflows/test.yml
```diff
@@ -2,6 +2,9 @@ name: home-assistant.io Test
 
 on: [push, pull_request]
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
```
frenck, line 7: This is already our repo default and thus redundant.

```suggestion
```

frenck (changes requested):

frenck:
Closing this PR, because the added permissions are either redundant / not needed or actually invalid.

This is not the first phantom PR by the ossf at this point @sashashura. It is kinda getting annoying.

.github/workflows/test.yml
sashashura: Apologies, the setting is not visible externally.

.github/workflows/add_prs_to_project.yml
sashashura: That was my impression from reading https://github.com/actions/add-to-project/blob/338ac1805ece459f9c25a3e7a2b749fec994576d/src/add-to-project.ts#L143 that is referenced in `uses: actions/add-to-project@v0.3.0`. I truly do not have intention to lower your security. If this is the case I'll won't suggest the permission for the action.

frenck: > That was my impression from reading

Yup, but it gets fed a different (more scoped) application token in this very same action.

sashashura:
I just want to say that my intentions are good. The read-only setting is invisible to outsiders (personally I think it is better to declare the read-only permission in every workflow explicitly instead of having some secret switch) and I thought the projects write is indeed needed based on the https://github.com/actions/add-to-project/blob/338ac1805ece459f9c25a3e7a2b749fec994576d/src/add-to-project.ts#L143.
No need to call me phantom if in a hundred PRs 3 or 4 mistakes happen. I always fixing them if I see something is broken.

.github/workflows/add_prs_to_project.yml
sashashura: I see, I missed that. Really didn't want to weaken any security.

frenck:
> personally I think it is better to declare the read-only permission in every workflow explicitly instead of having some secret switch

I disagree hardly on that. It is now locked down for the whole repository, regardless of the code. There are more people with write access compared to admin access on repos (in general). There is also additional GitHub security (2FA, and having to re-auth) for these kinds of changes and audit logs.

There is no way on earth you can sell putting it in the workflow is better in terms of security.

> No need to call me phantom

I'm not calling you phantom, I'm calling the PRs caused by ossf in this effort. I know you guys have a good goal and take things seriously, however, we have had multiple PRs on this theme. Most have been faulty/incorrect, redundant, and not all of them have been followed up correctly in reviews.

Please note, that we already have high workloads in reviews and getting everything reviewed. Having this discussion over and over again doesn't help with that at all.  

For me, personally, it is becoming annoying at this point.

.github/workflows/test.yml
frenck: Let's not commit assumptions in that case. Thanks 👍

sashashura:
Have a good day

frenck:
You too! ❤️

PR #24372, opened 2022-10-04
Describe behavior of template entities in some edge cases more explicitly

akloeckner:
## Proposed change

This describes the behavior of template entities in some edge cases more explicitly, as especially self-referencing seems to raise questions. See here, for example: https://github.com/home-assistant/core/issues/78358 .

## Type of change

## Additional information

## Checklist


akloeckner:
...

source/_integrations/template.markdown
```diff
@@ -339,6 +339,20 @@ If you are using the state of a platform that might not be available during star
...
 
+### Template errors
+
+If any of the templates generate an error when rendered, the sensor will become `unavailable`.
+
+### Caveats using the `this` variable
+
+The `this.state` data will be initialized with `unknown` during startup. 
+
+The `this.attributes` will be initialized with the empty mapping `{}` during startup. For that reason, using any attribute in your template, such as `this.attributes.my_attribute` will result in an `UndefinedError` during startup. To avoid this, use `state_attr(this.entity_id, 'my_attribute')` instead.
```
frenck, line 352: This is odd advise, as the provided solution is different compared to what the original code did.

source/_integrations/template.markdown
```diff
@@ -339,6 +339,20 @@ If you are using the state of a platform that might not be available during star
...
+
+The `this` variable will hold the state at the time of triggering an update. It will _not_ change during the update. Thus, if you want to use self-referencing in a trigger-based entity, you will likely have to add an explicit state trigger on the entity itself. Use `to: ~`, if you want to ignore attribute changes in the trigger. State-based entities will infer this self-referencing trigger autoamtically, if needed.
```
frenck: Why not just not-reference instead? 🤷

frenck (changes requested):

source/_integrations/template.markdown
akloeckner: You mean `this.attributes.my_attribute` is different from `state_attr(this.entity_id, 'my_attribute')` in that the former is fixed to the trigger-time, while the latter _might_ change? While that is true, of course, it will likely not make a difference in template entities, because they don't have any delays. 

The obvious alternative is to use `if`, but it will become clumsy.

Would you be fine to point out both alternatives and describe their difference?

(The other difference between the suggestion and the original code is that the suggestion guards against undefined attributes, but that is intended.)

source/_integrations/template.markdown
akloeckner: Because you would have to repeat some code. Suppose you have a (very complicated) `state` template and want to use `this.state` in the `icon`. You will need a trigger on the very entity itself in this case.

See the linked issue. https://github.com/home-assistant/core/issues/78358

source/_integrations/template.markdown
frenck: > You mean this.attributes.my_attribute is different from state_attr(this.entity_id, 'my_attribute') in that the former is fixed to the trigger-time, while the latter might change? While that is true, of course, it will likely not make a difference in template entities, because they don't have any delays.

That is not entirely true. Anyways, they are not the same thing :)

source/_integrations/template.markdown
frenck: Lots of that issue don't make sense to me. There is no need to self-reference. If you can calculate a state, you can calculate the icon in the same go, without triggering multiple times?

source/_integrations/template.markdown
akloeckner: I agree. But we'll have to change something, if we want to avoid the error. ;-)

What would you suggest?

frenck: > But we'll have to change something, if we want to avoid the error.

What error? If you handle it right, there is no error? I'm lost on why that difference is needed. When accessing an attribute from a dictionary, one needs to ensure the key exists, otherwise, an KeyError will be thrown.

The described solution, is not a solution, but a workaround. The solution, is to ensure a key exists before accessing it on a dict.

source/_integrations/template.markdown
akloeckner: Of course you can. But the code will become complicated, because you'll have to repeat things. I'll paste an example below. It's without a trigger, but the idea should come through. Internally, the sensor is triggered by itself, anyways. The example calculates if an aircraft is landing in one direction of Cologne airport's runway or the other. The direction (`14` or `32`) is used to derive the icon. I re-use the sensor for a second runway. This is the reason for the yaml anchors.

```yaml
  - name: Runway 14/32
    << : &runway_sensor
      icon: >
        {%- set runway = states(this.entity_id) | int(-9999) -%}
        mdi:{{ 'arrow-top-thin-circle-outline'          if runway > 33.75 else
               'arrow-top-left-thin-circle-outline'     if runway > 29.25 else
               'arrow-left-thin-circle-outline'         if runway > 24.75 else
               'arrow-bottom-left-thin-circle-outline'  if runway > 20.25 else
               'arrow-bottom-thin-circle-outline'       if runway > 15.75 else
               'arrow-bottom-right-thin-circle-outline' if runway > 11.25 else
               'arrow-right-thin-circle-outline'        if runway >  6.75 else
               'arrow-top-right-thin-circle-outline'    if runway >  2.25 else
               'arrow-top-thin-circle-outline'          if runway > -999  else
               'crosshairs-question' }}
      state: >
        {%- set data = state_attr(this.entity_id, 'data') %}
        {{- data.runway if data else None }}
    attributes:
      coordinates: >
        {#- Runway coordinates. Click and copy from Google maps! #}
        {#-               Latitude            Longitude          #}
        {%- set latlon = [50.880621184585635, 7.128884445352688,
                          50.85527028338135,  7.165619978822633] %}
        {{- latlon }}
      << : &runway_attributes
        direction: >
          {#- Calculate bearing from first point to second point #}
          {#- https://www.movable-type.co.uk/scripts/latlong.html #}
          {%- set latlon = state_attr(this.entity_id, 'coordinates') | default([0,0,1,1], true) %}
          {%- set lat1 = latlon[0]/180*pi %}
          {%- set lon1 = latlon[1]/180*pi %}
          {%- set lat2 = latlon[2]/180*pi %}
          {%- set lon2 = latlon[3]/180*pi %}
          {%- set dlon = lon2 - lon1 %}
          {%- set y = sin(dlon) * cos(lat2) %}
          {%- set x = cos(lat1) * sin(lat2)
                    - sin(lat1) * cos(lat2) * cos(dlon) %}
          {%- set bearing = atan2(y, x) % (2*pi) %}
          {{- bearing /pi*180}}
        offset_tolerance: >
          {#- Tolerance for cross-track distance as fraction of runway length -#}
          0.05
        direction_tolerance: >
          {#- Tolerance for aircraft heading in degrees -#}
          10
        opensky_sensor: >
          {#- The entity ID of the OpenSky to use -#}
          sensor.opensky
        data: >
          {#- Collect settings -#}
          {%- set latlon              = state_attr(this.entity_id, 'coordinates')         | default([0,0,1,1], true) %}
          {%- set direction           = state_attr(this.entity_id, 'direction')           | int(0) %}
          {%- set offset_tolerance    = state_attr(this.entity_id, 'offset_tolerance')    | float(0.05) %}
          {%- set direction_tolerance = state_attr(this.entity_id, 'direction_tolerance') | int(10) %}
          {%- set opensky_sensor      = state_attr(this.entity_id, 'opensky_sensor')      | default('sensor.opensky', true) %}

          {#- Collect aircraft data -#}
          {%- set value_json = {
               "time":   state_attr(opensky_sensor, 'time')   | default(None, true),
               "states": state_attr(opensky_sensor, 'states') | default([],   true),
          } %}

          {#- Loop over aicraft -#}
          {%- set name = (direction/10) | round %}
          {%- set ns = namespace({'runway': None}) %}
          {%- for aircraft in value_json.states %}

            {#- Abbreviate data #}
            {%- set callsign   = aircraft[1]  | trim %}
            {%- set latitude   = aircraft[6]  | float(0) %}
            {%- set longitude  = aircraft[5]  | float(0) %}
            {%- set altitude   = aircraft[7]  | float(0) %}
            {%- set velocity   = aircraft[9]  | float(0) %}
            {%- set vertical   = aircraft[11] | float(0) %}
            {%- set heading    = aircraft[10] %}

            {#- Calculate aircraft position in runway coordinates #}
            {%- set runway_vector     = [latlon[2]-latlon[0], latlon[3]-latlon[1] ] %}
            {%- set aircraft_vector   = [latitude -latlon[0], longitude-latlon[1] ] %}
            {%- set runway_length     = sqrt(runway_vector  [0]**2 + runway_vector  [1]**2) %}
            {%- set aircraft_length   = sqrt(aircraft_vector[0]**2 + aircraft_vector[1]**2) %}
            {%- set aircraft_position = (runway_vector[0]*aircraft_vector[0] + runway_vector[1]*aircraft_vector[1]) / runway_length %}
            {%- set aircraft_offset   = sqrt(aircraft_length**2 - aircraft_position**2) %}

            {#- Check, if aircraft is on runway, airborne and heading along runway #}
            {%- if true and
                   aircraft_position > -1.0*runway_length and
                   aircraft_position < +2.0*runway_length and
                   aircraft_offset | abs < offset_tolerance * runway_length and
                   velocity > 30 and
                   altitude > 0 and
                   altitude < 300 and
                   heading != None and
                   ((heading-direction+90) % 180 - 90) | abs < direction_tolerance%}

              {#- Populate namespace object #}
              {%- set ns.callsign  = callsign %}
              {%- set ns.latitude  = latitude %}
              {%- set ns.longitude = longitude %}
              {%- set ns.altitude  = altitude %}
              {%- set ns.velocity  = velocity %}
              {%- set ns.vertical  = vertical %}
              {%- set ns.heading   = heading %}
              {%- set ns.runway    = '%02d' % (name if (heading-direction) | abs < 90 else ((name+18)%36) ) %}
            {%- endif %}
          {%- endfor %}

          {#- Output data #}
          {{- {
            'timestamp':  value_json.time,
            'datetime':   value_json.time | as_datetime | as_local | string,
            'callsign':   ns.callsign,
            'latitude':   ns.latitude,
            'longitude':  ns.longitude,
            'altitude':   ns.altitude,
            'velocity':   ns.velocity,
            'vertical':   ns.vertical,
            'heading':    ns.heading,
            'runway':     ns.runway,
          } if ns.runway else state_attr(this.entity_id, 'data')}}
```

Edit: Yes, I might be using templates in an exaggerated way. But there will be less involved, yet still complicated, template sensors out there.

source/_integrations/template.markdown
akloeckner: During startup (i.e. before the first evaluation of the template sensor) there will be no attributes in `this`. So, there's no way for the user to make sure the key exists. And that results in `UndefinedError`s, when accessing the attributes. Basically, this applies to *all* usages of `this.attributes.anything`, I believe.

(Unless we change the `core` code to initialize all defined attributes with `None`. Then, we wouldn't need the workaround anymore. Which might be a good idea in the medium run. The user would still have to make sure, their templates handle `None` results properly, though.)

It is really only an issue during startup. See the second halve of this comment, e.g.: https://github.com/home-assistant/core/issues/78358#issuecomment-1267127022 . Here, it doesn't harm, but it bothers. Maybe there are more involved examples, which will break the sensor permanently. (I tried to come up with an example, but I cannot test it currently, so, I'll rather not guess into the wild.)

akloeckner:
.

akloeckner:
I'm lost what is next with this issue. Do I need to do anything? If there's no interest in clarifying this bit of information, I'll just let the bot close this PR next time.

PR #24410, opened 2022-10-06
Update ibeacon.markdown

WhimsySpoon:
## Proposed change
The `UNAVAILABLE_TIMEOUT` is currently set to 180 seconds. I couldn't ascertain why, after disabling an iBeacon it was still being shown as "Home" for another three minutes. It was only after reviewing `const.py` did I understand why.

This updates the documentation to make it clear.

## Type of change

## Additional information

## Checklist

- [ x] This PR uses the correct branch, based on one of the following:
- [x ] The documentation follows the Home Assistant documentation [standards].

### DIFF
--- a/source/_integrations/ibeacon.markdown
+++ b/source/_integrations/ibeacon.markdown
@@ -41,6 +41,10 @@ Consider setting up your iBeacons with a schema similar to the following:
 
 iBeacon devices that do not have stable Major and Minor values are not supported. The system automatically removes iBeacon devices with unstable Major and Minor values once ten (10) or more Major and Minor values have been seen with the same UUID from an iBeacon device with a fixed MAC address.
 
+## Considering an iBeacon Away
+
+iBeacons will be marked as "Away" after 180 seconds of not being seen.
+
 ## Fixed MAC addresses
 
 iBeacons with a fixed MAC address will get their own set of entities for each UUID, major, minor, and MAC address combination, enabling distance and presence detection per physical device basis. In this type of setup, it is permissible to have multiple iBeacons broadcasting the same UUID, Major, and Minor combination as long as you do not exceed five devices.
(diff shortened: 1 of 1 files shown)


bdraco:
That timeout value is only used for groups of devices with random mac addresses

WhimsySpoon:
It seems to apply to single iBeacon devices too. When I turn off the transmitter on the android companion app, it takes 3 minutes before it's marked as away. Is there another setting I'm missing?

bdraco:
The other timeout is provided by the bluetooth stack itself which is configured by bluez.  The default is 180s but may be changed

bdraco:
I don't think its a good idea to document a specific value as it may be different per system and the integration may change the value in the future.

WhimsySpoon:
Ah ok, didn't realise it was per-system. I think it would still be a good idea to document _something_ as there may be an expectation that it will be instant/a few seconds.

bdraco:
I think the best you could say is "a few minutes".

It's even more complicated because the ibeacon can broadcast at different intervals based on the firmware and settings

WhimsySpoon:
Good idea. I've updated the file and PR.

source/_integrations/ibeacon.markdown
```diff
@@ -41,6 +41,10 @@ Consider setting up your iBeacons with a schema similar to the following:
 
 iBeacon devices that do not have stable Major and Minor values are not supported. The system automatically removes iBeacon devices with unstable Major and Minor values once ten (10) or more Major and Minor values have been seen with the same UUID from an iBeacon device with a fixed MAC address.
 
+## Considering an iBeacon Away
+
+Due to various factors such as individual system settings and iBeacon firmware, iBeacons may not be marked as "Away" immediately. This could take several minutes.
```
bdraco: I think we can definitely say that will not be

WhimsySpoon: Updated accordingly.

frenck (approved):
Thanks, @WhimsySpoon 👍

PR #24478, opened 2022-10-09
Added a Device Registry example

ketilmo:
## Proposed change

The following example shows how you can connect MQTT sensors to the device registry.

## Type of change

## Additional information

## Checklist


source/_integrations/sensor.mqtt.markdown
```diff
@@ -350,6 +350,52 @@ mqtt:
 
 {% endraw %}
 
+### Using the Device Registry
+
+If your device has several entities, e.g. a weather station with temperature, humidity, and pressure sensors, it can be practical to add it to the [device registry](https://developers.home-assistant.io/docs/en/device_registry_index.html). The device registry groups the different entities under one device, and allows for modifying the device entities including their name and entity-IDs, through the user interface. To add a device to the device registry, ensure that each of the entities you want to add to the device has a `unique_id` specified. The ID can be a value you create, for instance, a GUID. Note that using the same ID several places will throw an error message. Next, assign all the entities to the same `device`, which needs to have a unique `identifier` of its own. The example below shows how this works in practice with a simple weather device.
```
frenck, line 355: I do think adding instructions around the device registry makes sense in a way for MQTT.
(otherwise, it would be a little bit of a blackhole).

Yet, I do think it is odd to add it to the sensor platform documentation only. What do you think?

../Frenck

ketilmo: Thanks for your feedback, @frenck. I added this passage after figuring out how to utilize the device registry for MQTT-driven sensors after finding my way through the instructions on this page. My perspective is that of a user, and I don't have a complete overview of how the functionality is structured across different modules and concepts. So if this example would be a better fit elsewhere, I am all in. But I would need a little pointer as to where this would belong. :)

frenck: > My perspective is that of a user, and I don't have a complete overview of how the functionality is structured across different modules and concepts. 

I agree, but at the same time, there s more than just the sensor platform. We have lights, climate, binary sensor and so many more. It feels wrong to add it here to sensors, but duplicating it all seems wrong too.

ketilmo: I fully understand your thoughts here, @frenck. I guess adding the example to this page would not be correct, either: https://developers.home-assistant.io/docs/device_registry_index/#what-is-a-device

An argument for keeping this example on the sensor page would be that the example is specific to sensors. Instead of duplicating it to the other types (lights, switches, etc), a better approach might be to create new examples relevant for each type.

Issue #16996, opened 2021-03-14
Add a filters in search bar


AalianKhan:
### Feedback

Hello,
I really like how there is a quality scale to see how good this integration is with home assistant. I often see people on forums asking what is the best fans, air conditioners, etc that work well with home assistant. A possible solution to this might be to make a filter in the search bar to see all the components with platinums, golds, etc to see which products/brands work well with home assistant which can influence mine and other people's buying decisions. I think some people will like a filter for IoT class too.

I think it would be best if it works how filters work for platforms. Such as when you click alarm you see all the alarm platforms
### URL

https://www.home-assistant.io/docs

### Version

2021.3.4

![Screen Shot 2021-03-14 at 1 00 20 AM](https://user-images.githubusercontent.com/39784853/111059073-c3b3cf00-8460-11eb-81a3-13ed4ad721ce.png)

AalianKhan:
Maybe a better implementation might be to have a sort button where a person clicks on the platform on the sidebar he/she is looking for and sorts it by quality

frenck:
This has been discussed a couple of time before, hacking it into a "category" list is a bit off. It would be better if we added an actual filter option.

MichaelAldrich:
Please do this. I'm someone excited about home assistant who is about to purchase their first ever smart home products. For me being able to choose the most well integrated products is a big priority. I woud like filters for IOT class as well, so I can quickly find the local push options.

AalianKhan:
This will be still highly useful today

ludeeus:
I made a concept to add filters yesterday <https://deploy-preview-18512--home-assistant-docs.netlify.app/integrations/>.
If anyone want to continue on that approach [here are the source](https://github.com/home-assistant/home-assistant.io/pull/18512/files)

It was fast thrown together so it's far from ready but should give a starting point, it would need at least:
- Better styling at least for desktop (mobile looks "fine" as IMO better than the current view)
- IOT Class and quality scale should be multi-select
- It should be possible to combine filters
- Maybe an easy way to reset filters
- The filter functions needs adjustments (this concept har misusing the search and that will not work with combining and multi-select)
- The filter needs to be backward compatible since there are links out there directly to searches/categories.

PR #24642, opened 2022-10-20
Overhaul of filters on integrations page

### LINKED ISSUE #9944: Mobile experience on Integrations page
Maybe it's just me, but I think the mobile experience for the [Integrations page](https://www.home-assistant.io/components/) can be improved.

I'm primarily referring to the filter type stuff. This is displayed as a nice left sidebar on desktop, but that's obviously not possible with limited width of mobile screens. So instead it's displayed as like 50+ buttons at the top of the page. Meaning a lot of swiping involved. [Screencapture](https://user-images.githubusercontent.com/9065557/61609671-da226c80-ac24-11e9-9b02-e19d7f7d3ff0.png).

But besides that, the other more important issue imo is that when the user _does_ tap on one of those buttons to filter the list, it doesn't appear like anything happened for two reasons: 
1. User doesn't get a loading bar in their browser because it's not loading a new page, simply filtering the current one.
2. The page doesn't visibly change either. If you tap on Calendar button you'd expect to see the eight calendar integrations, but instead you're still looking at the same list of buttons as before. Page actually does change but at the very bottom, and the user wouldn't necessarily know that. This can be kinda confusing and cause people to leave the page. Tap on one of the filters to see what I mean. 

I think that maybe this stuff should get collapsed to a single "Filter" button or input select type thing. There is already a select on that page for the "Added in." 

I don't think it's super crucial or anything that needs to be done right away, just noting it here as a potential area for future improvement. (There's been a lot of site improvements lately btw - nice job @frenck 😃).

kknopper:
## Proposed change
A rework of #24500 / #18512 to add proper filtration to the integrations page. Also addresses #9944

[Netlify preview / demo](https://deploy-preview-24642--home-assistant-docs.netlify.app/integrations/)

used @ludeeus 's notes from here as a guideline: [#16996 (comment)](https://github.com/home-assistant/home-assistant.io/issues/16996#issuecomment-880515679)

- [x] Better styling for filters. Thanks @ludeeus for the starting point
- [x] added filtration of IoT Classes, and Quality Scale 
- [x] all filtration items have been moved into `<select>` dropdowns. (Category, Version, IoT, and Quality)
- [x] It's now possible to combine filters (Search, IoT, Version, Category, and Quality)
- [x] an easy way to reset filters (buttons get added to the dom that can be clicked to remove filters)
- [x] complete rework of the `applyFilter()` function that allows for combining filters
- [x] update is backwards compatible with old `#` urls
- [x] on page load, if no filters are present in the url, the category is set to `Featured` like the production site 
- [x] integrations can now be searched for by their iot class and quality scale. You can also search for Featured integrations

Completely reworked the javascript for the integrations page. 
I opted to use the [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) browser API, over the hash method, as this api makes it easy to add and remove querys/filters to the url.

Created a `buildQueryFromURL()` function to build a query object of active filters any time the url gets updated. This object is then compared against the integrations array for filtration. Added more helper functions as well.

Old `#` Urls still work as expected. Also allowed for new hash urls for IoT and Quality which look like so:
`.../integrations/#iot_class/local-polling` and `.../integrations/#quality_scale/internal`

If a `#` url is present, any change to a filter will convert the URL to use the URLSearchParams API format instead.

On page load, or any change to the url, the current active filters (select dropdowns / search bar) will get updated to match the current active filters

Filter removal buttons are dynmaically added to the dom as well, and also update to match current active filters

Filters can be removed by clicking these buttons, or setting a filter dropdown to its initial state

### Caveats:
- URLSearchParams is not supported by IE
- this update contains a lot of ES6 javascript which is not supported by IE
- if IE needs to be supported, there are [polyfills](https://github.com/jerrybendy/url-search-params-polyfill) available for the URLSearchParams api

This is a pretty large change, because of this, I tried to make my code as readable as possible with plenty of comments.
Also left console log's present for reviewing purposes.  

## Type of change

## Additional information

- This PR fixes or closes issue: fixes #9944

## Checklist


source/integrations/index.html
```diff
@@ -23,43 +23,87 @@
...
+  <div class="grid__item six-sixths lap-one-whole palm-one-whole">
+    <div class="component-search">
+      <form onsubmit="event.preventDefault(); return false">
+        <input type="text" name="search" id="search" data-id="search" class="search" placeholder="Search integrations..." autofocus />
+      </form>
+    </div>
+  </div>
+  <div class="grid__item three-sixths lap-one-whole palm-one-whole">
+      <div class="integration_filter">
+        <select name="category" data-id="cat">
```
ludeeus: Why not use `ha_category` here (and everywhere else)?

source/integrations/index.html
```diff
@@ -23,43 +23,87 @@
...
-            {%- endfor -%}
-          </optgroup>
+  <div class="grid__item six-sixths lap-one-whole palm-one-whole">
+    <div class="component-search">
+      <form onsubmit="event.preventDefault(); return false">
+        <input type="text" name="search" id="search" data-id="search" class="search" placeholder="Search integrations..." autofocus />
+      </form>
+    </div>
+  </div>
+  <div class="grid__item three-sixths lap-one-whole palm-one-whole">
```
ludeeus: I think `one-quarter` looks better:
image

`three-sixths`:
image

source/integrations/index.html
```diff
@@ -23,43 +23,87 @@
...
+    </div>
+  </div>
 
-      <a href='#other' class="btn" onclick="document.querySelector('.page-content').scrollTop = 0">Other ({{ components | where: 'ha_category', 'Other' | size }})</a>
+  <div class="grid__item three-sixths lap-one-whole palm-one-whole">
+    <div class="integration_filter">
+      <select name="iot_classes" data-id="iot_class">
+        <option value="#">IoT Class</option>
+        {%- for iot_class in iot_classes -%}
+          {%- assign iot_class_count = components | where: 'ha_iot_class', iot_class | size -%}
```
ludeeus: Counts do not consider active filters.
I wonder if we should show it at all.

source/integrations/index.html
```diff
@@ -171,99 +318,241 @@
...
+          }
+          return true;
+        });
+        buttonsToRemove.forEach(button => button.remove());
+      } else {
+        //create a filter button if one doesn't already exists
+        for (key in queryFilter) {
+          if (document.querySelector(`.component-search .active-filter#${key}`) === null) {
+            let button = document.createElement('button');
+            button.innerHTML = `<span>${getButtonText(key)}<i class="icon-plus-sign"></i></span>`;
```
ludeeus: this should include the filtered value as well, like `version: 2022.11.0` and `Category: featured`.

source/integrations/index.html
```diff
@@ -103,60 +147,163 @@
...
+      }
+
+      if (hash.indexOf('#featured') === 0 || queryParams.get('cat') === 'featured') {
+        // only show those with featured = true
+        queryFilter.cat = 'featured';
+      }
+
+      if (hash.indexOf('#search/') === 0 || queryParams.has('search')) {
+        let query = getQueryValue('search');
+        queryFilter.search = query;
```
ludeeus, line 258:
```suggestion
        queryFilter.search = query.toLowerCase();
```

source/integrations/index.html
```diff
@@ -171,99 +318,241 @@
...
+          text = 'Quality';
+          break;
+        default:
+          text = id;
+      }
+      return text;
+    }
+
+    // update all filter buttons to match current obj filter
+    function updateFilterButtons(queryFilter) {
```
ludeeus, line 522: Lets reserve the space these take up so the UI does not jump when adding/removing

sass/custom/_component_page.scss
```diff
@@ -12,6 +12,35 @@
...
     }
   }
+
+  .integration_filter select {
+    width: 100%;
+    margin-bottom: 10px;
+    text-align: center;
+  
+    &[data-id="iot_class"], &[data-id="quality_scale"] {
+    margin-bottom: 15px;
```
ludeeus: With a single row this can be removed

sass/custom/_component_page.scss
```diff
@@ -12,6 +12,35 @@
       border-color: #7c7c7c #c3c3c3 #ddd;
     }
   }
+
+  .integration_filter select {
+    width: 100%;
+    margin-bottom: 10px;
```
ludeeus:
```suggestion
    margin-bottom: 15px;
```

With a single row this can be 15 instead, 16 maybe?

sass/custom/_component_page.scss
```diff
@@ -12,6 +12,35 @@
...
+    text-align: center;
+  
+    &[data-id="iot_class"], &[data-id="quality_scale"] {
+    margin-bottom: 15px;
+    }
+  }
+  
+  .active-filter {
+    margin: 10px 5px 0px 0;
```
ludeeus:
```suggestion
    margin: 10px 5px 0px 0px;
```

source/integrations/index.html
```diff
@@ -103,60 +147,163 @@
...
+
+      if (hash.indexOf('#search/') === 0 || queryParams.has('search')) {
+        let query = getQueryValue('search');
+        queryFilter.search = query;
+      } else {
+        // reset search box when not searching
+        $searchFilter.val(null);
+      }
+      
+      if (hash.indexOf('#version/') === 0 || queryParams.has('version')) {
```
ludeeus:
```suggestion
      if (hash.startsWith('#version/') || queryParams.has('version')) {
```
The same goes for the rest of these

source/integrations/index.html
```diff
@@ -103,60 +147,163 @@
...
     init();
 
-    /**
-     * filter all components, based on the location's hash and render them into the component box
-     */
-    function applyFilter() {
+    // get query value pair that matches provided id from url. // for buildQueryFromURL()
+    function getQueryValue(id) {
+      let index = id.length + 2; // id + '#' and '/' => '#search/'
+      if (id === 'cat') index = 1;
```
ludeeus:
```suggestion
      if (id === 'cat') {
        index = 1;
      }
```

source/integrations/index.html
```diff
@@ -171,99 +318,241 @@
...
+      let text;
+      switch(id) {
+        case 'cat':
+          text = 'Category';
+          break;
+        case 'iot_class':
+          text = 'IoT';
+          break;
+        case 'quality_scale':
+          text = 'Quality';
```
ludeeus:
```suggestion
          text = 'Quality scale';
```

source/integrations/index.html
```diff
@@ -171,99 +318,241 @@
...
+
+    // filter button name cleanup. used inupdateFilterButtons()
+    function getButtonText(id) {
+      let text;
+      switch(id) {
+        case 'cat':
+          text = 'Category';
+          break;
+        case 'iot_class':
+          text = 'IoT';
```
ludeeus:
```suggestion
          text = 'IoT class';
```

source/integrations/index.html
kknopper: updated the `updateFilterButtons` function. Now there is another check to test if the filter object values have changed, and it updates the button text accordingly. Also switched to use the data attribute over id

source/integrations/index.html
kknopper: updated to use `one-quarter` class for select dropdown width

source/integrations/index.html
kknopper: added a new `ha_category` class for the 4 select drop downs

source/integrations/index.html
kknopper: added height to the container holding the filter buttons, so the ui doesnt jump anymore

sass/custom/_component_page.scss
kknopper: removed

sass/custom/_component_page.scss
kknopper: updated to 15px

sass/custom/_component_page.scss
kknopper: cleaned this up

source/integrations/index.html
kknopper: updated all instances of `indexOf` to use `startsWith` method

source/integrations/index.html
kknopper: cleaned this up

source/integrations/index.html
kknopper: updated filtered button name

source/integrations/index.html
kknopper: updated filter button names

kknopper:
@ludeeus addressed PR feedback. Currently I'm leaving in the counts, as this has been useful for testing. 

When you think this PR is review ready, I can add an additional commit to remove all filter counts, along with all the current js console logs.

source/integrations/index.html
kknopper: @ludeeus not sure if there's a better way we wanna handle this for mobile. The filter buttons can stack, and create a lot more space. Adjusted my initial fix to use `min-height`, so it grows for mobile.

Screen Shot 2022-11-30 at 9 13 43 AM

ludeeus:
> When you think this PR is review ready, I can add an additional commit to remove all filter counts, along with all the current js console logs.

Please do that.

kknopper:
@ludeeus I think this PR is in a pretty good place. 

The only thing i would like to change would be to properly order the dropdown of the Quality Scale select element. Was trying to do this in the liquid syntax, but I couldn't find an elegant solution. I think the simplest way would be to manually declare an array variable, instead of mapping it from the integrations. This doesn't make it dynamic though. Let me know your thoughts on this, along with any other feedback you might have. Would love to get this merged soon!

frenck:
I think my main grind with this approach is:

CleanShot 2023-01-31 at 09 39 03

The "Featured" is set by default, which is OK, but is also hides the "categories". When getting to this page now the first time, it is not clear where all the categories are...

../Frenck

kknopper:
@frenck We could add custom dropdown titles above the select element to make this more clear, similar to the version select thats on the integrations page now:

Screen Shot 2023-02-03 at 11 07 36 AM

kknopper:
@frenck @ludeeus 

- Added a title for each of the dropdown selectors
- moved selectors above search bar
- Cleaned up styling for dropdown selectors - made sure styles were compatible cross browser
  - added a wrapper div around dropdowns to make sure widths matched across major browsers (Chrome / FF / Safari)
- Now the default for every selector will be `All` for when no filter is applied 
- adjusted how the `All` filter works for the Categories dropdown
  - old `/integrations/#all` url still works as expected
- Added a check for the dropdowns - now if a user changes the url to a filter that is not present, the dropdown will only change if a matching option is available

kknopper:
@frenck @ludeeus If you guys have time to take another look at this PR, that would be awesome!

PR #24652, opened 2022-10-22
Small clarification in documentation

jacobstim:
I tried adding my own HA URL instead of literally "my.home-assistant.io" as the redirect URI -- this URI leans to being read as "oh, I need to fill out my own URI here", which is not the case here :). This small documentation addition might avoid some facepalms amongst users like me.


source/_integrations/neato.markdown
```diff
@@ -38,7 +38,7 @@ Visit [the Neato Developer Network](https://developers.neatorobotics.com/applica
 
 You will have to enter a name, a description and your redirect URL.
 
-Use `https://my.home-assistant.io/redirect/oauth` as the redirect URL.
+Use `https://my.home-assistant.io/redirect/oauth` (as-is, without customization for your own Home Assistant URLs) as the redirect URL.
```
frenck: This is a standard text across all our documentation, I think it is odd just to change it here. What do you think?

Also, this uses the My Home Assistant services, should we link that instead?

frenck: @jacobstim A small and friendly notification.

Are you planning on moving forward on this PR? 
I didn't see any activity recently, while there are still things to do. I hope you consider picking it up soon! 👍

../Frenck

jacobstim: Not sure what is still needed to move this PR forward? It's a single line addition to the documentation :). 

I agree it is strange to only change it here, but it's a good idea to start somewhere, right? (Tracing it back through the entire documentation seems a daunting effort). The alternative is not to change it all, but I would at least recommend changing the phrasing at least *somewhere* to clarify that this is a literal URL to copy/paste, without thinking, without changes. For example, the prior sentence "You will have to enter a name, a description and **_your_** redirect URL." seems to suggest customization of the URL is needed, where-as it is not the case.

frenck: > Not sure what is still needed to move this PR forward? 

See above questions, which have been left without a response.

> I agree it is strange to only change it here, but it's a good idea to start somewhere, right? (Tracing it back through the entire documentation seems a daunting effort).

Search and replace will help a lot. And yes, I think consistency is important. I don't agree with merging a single instance.

jacobstim: I did a [search through the documentation ](https://github.com/home-assistant/home-assistant.io/search?q=redirect%2Foauth) and indeed, the original Neato documentation is out of line compared to how most other integration specify the creation of an app in an external platform. For other integrations, it's much clearer that one just needs to copy-paste the URI, without modifications.

I've updated the PR to be consistent with the other integrations... then we have both the clarification and consistency in one go... woohoo! 😉

source/_integrations/neato.markdown
```diff
@@ -35,13 +35,14 @@ There is support for the following platform types within Home Assistant:
 Visit [the Neato Developer Network](https://developers.neatorobotics.com/applications) and create a new app.
 
 <div class='note'>
-
-You will have to enter a name, a description and your redirect URL.
-
-Use `https://my.home-assistant.io/redirect/oauth` as the redirect URL.
+You will have to enter a name, a description and the redirect URI:
```
frenck:
```suggestion

You will have to enter a name, a description, and the redirect URI:

```

source/_integrations/neato.markdown
```diff
@@ -36,12 +36,15 @@ Visit [the Neato Developer Network](https://developers.neatorobotics.com/applica
 
 <div class='note'>
 
-You will have to enter a name, a description and your redirect URL.
+You will have to enter a name, a description, and the redirect URI:
 
-Use `https://my.home-assistant.io/redirect/oauth` as the redirect URL.
+- **Name**: can be anything you like, for example, "HomeAssistant".
```
frenck:
```suggestion
- **Name**: can be anything you like, for example, "Home Assistant".
```

frenck (approved):

PR #24671, opened 2022-10-23
Expand dsmr_reader documentation

Glodenox:
## Proposed change
Add more information on how the sensors in DSMR Reader are best used within the Energy Dashboard. Quite a few people were having issues with this (mostly due to the changes in the state class of some entities).

Additionally, I've better clarified the difference with the DSMR integration. I'm not sure why I previously stated that it provides an instance of DSMR Reader directly, as it has nothing to do with DSMR Reader at all.

## Type of change

## Additional information
No other information

## Checklist


source/_integrations/dsmr_reader.markdown
```diff
@@ -33,6 +33,16 @@ To use this DSMR Reader sensor integration, you need to have a DSMR Reader insta
 
 {% include integrations/config_flow.md %}
 
+## Configuring the Energy Dashboard
+
+It is most advisable to use the `sensor.dsmr_reading_*` sensors in the energy dashboard over the "merged" and "day" sources. The "reading" sensors provide the most stable source of data for Home Assistant to use. These MQTT values are part of the "Telegram: Split topic" MQTT values within DSMR Reader, so make sure to enable it.
```
frenck: Why does that matter? (reading, merged, day) Every source that meets the requirements is viable for use, the result should make any difference?

source/_integrations/dsmr_reader.markdown
```diff
@@ -33,6 +33,16 @@ To use this DSMR Reader sensor integration, you need to have a DSMR Reader insta
...
+
+| Section          | Sensor entities to configure                                                   |
+| ---------------- | ------------------------------------------------------------------------------ |
+| Grid consumption | `dsmr_reading_electricity_delivered_1`, `dsmr_reading_electricity_delivered_2` |
+| Return to grid   | `dsmr_reading_electricity_returned_1`, `dsmr_reading_electricity_returned_2`   |
+| Gas consumption  | `dsmr_consumption_gas_delivered`                                               |
```
frenck: Don't use entity IDs (object IDs in this case), those are not guaranteed to be assigned.

frenck (changes requested):

source/_integrations/dsmr_reader.markdown
Glodenox: The reading values contain a running total, while the day values get reset each day at midnight. In some cases, there might be some data loss at that time. This is especially true when the Home Assistant instance was down for a day or more.

Slightly tangential: after your change to the state classes, the other ones actually aren't viable sources any more. I still need to revert that change, as they recently fixed the bug in DSMR Reader.

source/_integrations/dsmr_reader.markdown
Glodenox: Do you mean to say I should use `sensor.dsmr_reading_electricity_delivered_1` and so on instead? The code blocks overflowed the fixed table layout in that case, as it wasn't possible to break at any word. I'm all ears for a solution to that, as I wasted quite a bit of time on finding some working solution for that. The script will always assign these sensor entities with these IDs and those texts are also part of the unique IDs.

If you prefer to use the names of the entities instead: they aren't unique within the integration. There's 4 of each of them. We wouldn't be helping people by using those names.

frenck: > Do you mean to say I should use sensor.dsmr_reading_electricity_delivered_1 and so on instead?

No, I'm saying not to use entity IDs or object IDs at all. Just describe them instead.

> There's 4 of each of them. We wouldn't be helping people by using those names.

I suggest we improve their names in that case (if that is really problematic, is should be addressed to being with). Object and entity IDs are not guaranteed to be assigned.

Glodenox: In that case I'll just leave out the table for now, as I can't describe something without using the one thing that actually describes them. I'm looking at maybe translating those default names, so I'll combine those two changes in the integration.

Glodenox: A renaming of the sensors is planned, so the names will be unique. As such, I added the table again with the regular sensor names.

frenck (approved):
Thanks, @Glodenox 👍 

../Frenck

PR #24778, opened 2022-11-01
Added troubleshooting for 0 devices in home assistant

mitchcapper:
Some concise steps to check, most importantly about the account not being linked under Users in their portal.

## Type of change

Technically it is just adding additional troubleshooting tips not correcting something wrong.

## Checklist

- [x ] The documentation follows the Home Assistant documentation [standards].

### DIFF
--- a/source/_integrations/tuya.markdown
+++ b/source/_integrations/tuya.markdown
@@ -114,9 +114,9 @@ Click the created project to enter the `Project Overview` page and get the `Auth
 
 {% configuration_basic %}
 
-If no devices show up in home assistant:
+If no devices show up in Home Assistant:
  - First make sure the devices show up in Tuya's cloud portal under the devices tab. 
- - In the Tuya IOT configuration cloud portal you must NOT link your non-developer account under the "Users" tab.  Doing so will work, and you can even still add the devices under the devices tab, but the api will send 0 devices down to home assistant.  You must only link the account under the Devices->"Link Tuya App Account".  If it shows up on the users  tab be sure to delete it.
+ - In the Tuya IoT configuration cloud portal you must NOT link your non-developer account under the "Users" tab.  Doing so will work, and you can even still add the devices under the devices tab, but the api will send 0 devices down to Home Assistant.  You must only link the account under the Devices->"Link Tuya App Account".  If it shows up on the users  tab be sure to delete it.
  - Your region may not be correctly set
  - Make sure your cloud plan does not need to be renewed (see error #28841002 on this page)
 
--- a/source/_integrations/tuya.markdown
+++ b/source/_integrations/tuya.markdown
@@ -114,6 +114,13 @@ Click the created project to enter the `Project Overview` page and get the `Auth
 
 {% configuration_basic %}
 
+If no devices show up in home assistant:
+ - First make sure the devices show up in Tuya's cloud portal under the devices tab. 
+ - In the Tuya IOT configuration cloud portal you must NOT link your non-developer account under the "Users" tab.  Doing so will work, and you can even still add the devices under the devices tab, but the api will send 0 devices down to home assistant.  You must only link the account under the Devices->"Link Tuya App Account".  If it shows up on the users  tab be sure to delete it.
+ - Your region may not be correctly set
+ - Make sure your cloud plan does not need to be renewed (see error #28841002 on this page)
+
+
 "1004: sign invalid":
   description: Incorrect Access ID or Access Secret. Please refer to the **Configuration** part above.
 
--- a/source/_integrations/tuya.markdown
+++ b/source/_integrations/tuya.markdown
@@ -115,6 +115,7 @@ Click the created project to enter the `Project Overview` page and get the `Auth
 {% configuration_basic %}
 
 If no devices show up in Home Assistant:
+  description: >
     - First make sure the devices show up in Tuya's cloud portal under the devices tab. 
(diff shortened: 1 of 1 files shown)


source/_integrations/tuya.markdown
```diff
@@ -113,6 +113,13 @@ Click the created project to enter the `Project Overview` page and get the `Auth
 
 {% configuration_basic %}
 
+If no devices show up in Home Assistant:
```
frenck, line 117: - This is a configuration basic block, and requires YAML (not Markdown) see examples below.
- Why not link to the section about linking above? Instead of making new/duplicate instructions?

source/_integrations/tuya.markdown
```diff
@@ -113,6 +113,13 @@ Click the created project to enter the `Project Overview` page and get the `Auth
...
+ - First make sure the devices show up in Tuya's cloud portal under the devices tab. 
+ - In the Tuya IoT configuration cloud portal you must NOT link your non-developer account under the "Users" tab.  Doing so will work, and you can even still add the devices under the devices tab, but the api will send 0 devices down to Home Assistant.  You must only link the account under the Devices->"Link Tuya App Account".  If it shows up on the users  tab be sure to delete it.
```
frenck: Double spaces, please remove those.

source/_integrations/tuya.markdown
```diff
@@ -113,6 +113,13 @@ Click the created project to enter the `Project Overview` page and get the `Auth
...
+ - Your region may not be correctly set
+ - Make sure your cloud plan does not need to be renewed (see error #28841002 on this page)
```
frenck: What is `#28841002` referring too?

source/_integrations/tuya.markdown
```diff
@@ -113,6 +113,13 @@ Click the created project to enter the `Project Overview` page and get the `Auth
 
 {% configuration_basic %}
 
+If no devices show up in Home Assistant:
+ - First make sure the devices show up in Tuya's cloud portal under the devices tab. 
+ - In the Tuya IoT configuration cloud portal you must NOT link your non-developer account under the "Users" tab.  Doing so will work, and you can even still add the devices under the devices tab, but the api will send 0 devices down to Home Assistant.  You must only link the account under the Devices->"Link Tuya App Account".  If it shows up on the users  tab be sure to delete it.
+ - Your region may not be correctly set
```
frenck: Wouldn't this cause a different error? (as already listed below)

frenck (changes requested):

source/_integrations/tuya.markdown
mitchcapper: Ha .markdown did get me;) I think the primary difference was for the spacing on the list which should be corrected now.

The instructions are correct above, however if the user does not follow them exactly or does additional steps, things fail in a non obvious way.   Coming from making the mistake myself and spending hours trying to figure out why the API was returning nothing and no error.
If a user was trying to troubleshoot why home assistant has 0 entities due to an expired service, wrong region, project to old, or other error (that shows no error message) and notices the user account does not show up under users they may add the account there.   No error will be created but all future efforts will be thwarted until it is removed.

If the suggestion is to just put the warning in the linking instructions that would likely also suffice, but I figured if someone gets past that step they may not think to go back to review it, but would hopefully go to the troubleshooting section to troubleshoot.

source/_integrations/tuya.markdown
mitchcapper: done

source/_integrations/tuya.markdown
mitchcapper: If it expired and you are trying to add new I believe you will get "28841002: No permissions. Your subscription to cloud development plan has expired"   however if it is already in home assistant I do not believe an error is generated.

source/_integrations/tuya.markdown
mitchcapper: If a user has multiple devices spanning regions (/ have multiple regions on their project as possible) I believe you can get nothing to show up but no error.  I did not experience this but took it from reading some of the other tickets.

source/_integrations/tuya.markdown
```diff
@@ -114,6 +114,13 @@ Click the created project to enter the `Project Overview` page and get the `Auth
 
 {% configuration_basic %}
 
+If no devices show up in Home Assistant:
```
frenck, line 117:
```suggestion
If no devices show up in Home Assistant:
  description: >
```

source/_integrations/tuya.markdown
```diff
@@ -114,6 +114,17 @@ Click the created project to enter the `Project Overview` page and get the `Auth
...
+  description: >
+    - First make sure the devices show up in Tuya's cloud portal under the devices tab.
+
+    - In the Tuya IoT configuration cloud portal you must NOT link your non-developer account under the "Users" tab. Doing so will work, and you can even still add the devices under the devices tab, but the api will send 0 devices down to Home Assistant. You must only link the account under the Devices->"Link Tuya App Account". If it shows up on the users  tab be sure to delete it.
+
+    - Your region may not be correctly set.
+
+    - Make sure your cloud plan does not need to be renewed (see error #28841002 on this page).
```
frenck:
```suggestion
    - First, make sure the devices show up in Tuya's cloud portal under the devices tab.

    - In the Tuya IoT configuration cloud portal, you must NOT link your non-developer account under the "Users" tab. Doing so will work, and you can even still add the devices under the devices tab, but the API will send 0 devices down to Home Assistant. You must only link the account under the Devices->"Link Tuya App Account". If it shows up on the users tab, be sure to delete it.

    - Your region may not be correctly set.

    - Make sure your cloud plan does not need to be renewed (see error #28841002 on this page).
```

frenck (approved):
Thanks, @mitchcapper 👍 

../Frenck

Issue #24801, opened 2022-11-03
Watermeter on energy dashboard: entity specs ?


jjvandenberg:
### Feedback

Hi, can you add to the docs, the necessary entity definition for the watermeter, like 
- state_class
- unit_of_measurement
-device_class

Thanks !

### URL

https://www.home-assistant.io/docs/energy/water/

### Version

2022.11

fvschie:
I have issues where my water meter sensors (of type 'total', both in liters and m3) aren't showing up. It's possible it needs a total_increasing, which would be annoying, as that's harder to fabricate.

So I second a list of requirements for the sensor.

helmuto-maker:
I have these settings - but doesn't work and does not show up as possible entity to choose in water - energy dashboard. My gas-meter does show up as possible entity in the water-section:
state_class: total_increasing
unit_of_measurement: m³
icon: mdi:chart-line
friendly_name: Total Water Use Daily
last_reset: '1970-01-01T00:00:00+00:00'
device_class: water

helmuto-maker:
I figured this out. My sensor had two different units in the long term statistics. After fixing this (in developer tools -> statistics), the sensor is available to choose.

neuhausf:
I had that too, but removing the old statistics didn‘t help. My sensor is a restful, total_increasing with unit m3. Additionally I have setup utility_meters for different periods (hourly, daily, weekly, monthly and yearly). Statistics are generating fine. What am I missing?
![884605E0-05B4-4F5E-88ED-2D7AE25559A1](https://user-images.githubusercontent.com/7963239/199894249-bb258a38-6ac3-4e77-a88d-f7b0fbf76214.jpeg)

frenck:
That sounds backwards? You don't adjust entities to make them work with the energy dashboard... instead configure entities correct according to the data they provide. If they provide the right data, they become available for the energy dashboard.

jjvandenberg:
So if the question was "how do i need to configure my entities so the energy dashboard can use them for the water meter", it would have been answered? 
why so unfriendly?

for all that had the same qustion: 
unit_of_measurement: m³ (use an uppercase 3, not a regular 3, that's where I went wrong)
device_class: water

seems more people have this question , like #24813

jjvandenberg:
solved it
unit_of_measurement: m³ 
device_class: water

mrandt:
I had to fiddle with this, too. My watermeter integrates via MQTT.

[...]

PR #24815, opened 2022-11-04
Added manual configuration instructions for adding a water sensor to Energy dashboard

mcsquared88:
## Type of change

## Additional information

## Checklist


anatoly-savchenkov:
HA seems to be picky about the way how measurement units are defined for cubic meters, see discussion here https://github.com/home-assistant/home-assistant.io/issues/24801. Only the format below is accepted (not m3, for instance):
````
      unit_of_measurement: "m³"
      device_class: water
      state_class: total_increasing
````
Could you please update your MR with information for metric world :)

**Comment by mcsquared88** [comment]
> HA seems to be picky about the way how measurement units are defined for cubic meters, see discussion here #24801. Only the format below is accepted (not m3, for instance):
> 
> ```
>       unit_of_measurement: "m³"
>       device_class: water
>       state_class: total_increasing
> ```
> 
> Could you please update your MR with information for metric world :)

Great suggestion. I've updated the MR to mention both "gal" and "m³"

File: source/_docs/energy/water.markdown [hunk 1]
```diff
@@ -27,3 +27,14 @@ Alternatively, the following shops sell ESPHome-based devices, that use a proxim
...
 For any of the above-listed options, make sure it actually works with the type of water meter you have before getting one.
+
+## Manual Configuration
+
+Go to Settings > Dashboards > Energy and then click Add Water Source.
+
+If your water meter sensor isn't listed, be sure your sensor has the following parameters:
+- unit_of_measurement: "gal" or "m³" [the 3 must be superscript]
+- device_class: water
+- state_class: total_increasing
```
frenck, line 38: This is the wrong way around. One does not adjust a sensor to work with the energy dashboard. Instead, if a sensor provides the right data, it will work with the energy dashboard.

This instruction will lead to setting non-matching information set on sensors that are actually not suitable for use.

frenck (changes requested):

source/_docs/energy/water.markdown
mcsquared88: I don't know how common of a situation this is, but I've got smart meters that I'm able to read via a RTL-SDR dongle. I'm sending the readings to HA via MQTT. When I saw that HA now includes water meters on the Energy dashboard, I took a look at my Energy dashboard but didn't see my water meter. These are the steps I figured out (via posts by others) to get it to appear. It wasn't that the sensor wasn't providing the "right" data, it's that I hadn't formatted the sensor definition correctly. I was using "g" as the unit_of_measurement instead of "gal". I also had to go into the Deveoper Tools > Statics screen and find and fix the error about the unit changing. I learned of that trick from https://github.com/home-assistant/home-assistant.io/issues/24801

I thought these instructions should be included in the official documentation to make it easier for others to figure this out, so I submitted this change. How about I change line 35 to "Your water meter won't be listed unless it has the following parameters. If your water meter uses a different unit of measure, try creating a template sensor to recalculate the reading into gallons or cubic meters."

frenck: That is very MQTT specific. When looking at the instruction provided:

> Go to Settings > Dashboards > Energy and then click Add Water Source.

> If your water meter sensor isn't listed, be sure your sensor has the following parameters:

For almost all integrations, there is no such thing as providing these parameters. In the end, when an integration (or in case you when defining templates/MQTT/protocol-based sensors), are responsible for providing as much and correct information as possible. It is not just this, but also for things like: icons 🤷 

The problem with adding this is that it isn't an end-user concern in almost all cases and, above all, isn't use-fixable. Adding statements like this will thus cause more issues then they resolve IMHO.

frenck:
Based on the above reasoning, I'm going to friendly decline this PR for now.

Nevertheless, thank you for willing to contribute 👍 

../Frenck

mcsquared88:
I guess it all depends on your sense of who the end users are and what their level of expertise is. I assume most HA users are very experienced tinkerers. I'm fine with deciding this is an edge case and shouldn't be in the "main" documentation. But I don't like the thought that others will have to search for multiple message board posts like I did. I feel like there should be official "developer" documentation somewhere that describes what an integration needs to do for this to work, and maybe use MQTT as a generic example. And there can then be a link (with necessary caveats) in the "main" documentation that points to the developer documentation.

PR #25179, opened 2022-12-07
One char change to fix a doc typo

danielhoherd:
The brand name for the USB-BT4LE is "Plugable" with a single G. <https://plugable.com/products/usb-bt4le/>

This typo led me to buy another bluetooth device before realizing my current one was in the list, but was spelled wrong, and thus didn't show up in my cmd-f result.


frenck (approved):
Thanks, @danielhoherd 👍

PR #25253, opened 2022-12-12
CoIoT is using UDP as the protocol

BerendWouters:
## Proposed change
The documentation doesn't describe that the port should be exposed via udp protocol instead of tcp.

## Type of change

## Additional information

## Checklist


source/_integrations/shelly.markdown
```diff
@@ -271,6 +271,8 @@ shelly:
   coap_port: 12345
 ```
 
+Additionally, make sure the firewall or network boundary allows 5683 (or the customized port) on the udp protocol.
```
frenck: This is not a normal use case, but an advanced networking case. An average user reading will be confused or has no idea what to do with this, they won't be able to apply it on their Home Assistant OS-based system (which is the bigger majority of our users).

IMHO, we should not add this for that reason.

User managing and creating firewall, should know how they handle their traffic. The ports are extensively documented in the documentation already, so that is not an open end that is being solved.

thecode: I was waiting for another review because I had the same feeling it is to confusing for most users, I think if we change:
```
In some cases, it may be needed to customize the CoAP port
```
to:
```
In some cases, it may be needed to customize the CoAP UDP port
```
It may be enough not to confuse average users, but still allow advanced users to understand they need to map UDP port.

frenck: I think that makes sense @thecode

BerendWouters: Good point regarding the advanced level of networking. In the end, the issue was also quickly resolved by just googling the protocol of CoAP.
The suggestion made by @thecode would've probably saved me 5 minutes and a quick search.

frenck: Agree, could you adjust the PR accordingly, @BerendWouters?

BerendWouters: I've changed the PR.

thecode (approved):
Thanks!

PR #25258, opened 2022-12-12
Indicate that MPD can connect with Hostnames

theTaikun:
## Proposed change

Currently, the documentation states that connection to an MPD server must be done via an IP address, however this is untrue.

This integration uses the python-mpd2 package to establish connection to an MPD server. That package has the ability to connect via TCP using a hostname or IP addresses. [It does this by using socket.getaddrinfo() to do an IP lookup of the hostname](https://github.com/Mic92/python-mpd2/blob/d632d7b48fd8e5ed4ce39651755ef40b61d5276d/mpd/base.py#L742). I've confirmed this works within home assistant with two separate MPD servers.

Python-mpd2 documentation specifies that it can also connect to local sockets using absolute paths starting with `/` or via the MPD convention using `@`. Can be seen in [this if statement and the following one](https://github.com/Mic92/python-mpd2/blob/d632d7b48fd8e5ed4ce39651755ef40b61d5276d/mpd/base.py#L791) I haven't tested with sockets, so I haven't added it to the proposed changes.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/mpd.markdown
+++ b/source/_integrations/mpd.markdown
@@ -26,7 +26,7 @@ media_player:
 
 {% configuration %}
 host:
-  description: IP address of the Host where Music Player Daemon is running.
+  description: Hostname or IP address of the Host where Music Player Daemon is running.
   required: true
   type: string
 port:


frenck (approved):
Thanks, @theTaikun 👍

PR #25313, opened 2022-12-16
Add hint why some entities are not selectable after choosing the first one

Bascht74:
## Proposed change
Add a hint because I didn't get any in the ui, why the entities list ist filtered. 
With the history graph card this works so I think it is worth it

## Type of change

## Additional information
none

## Checklist

### DIFF
--- a/source/_dashboards/statistics-graph.markdown
+++ b/source/_dashboards/statistics-graph.markdown
@@ -30,7 +30,7 @@ type:
   type: string
 entities:
   required: true
-  description: "A list of entity IDs or `entity` objects (see below), or an external statistic id"
+  description: "A list of entity IDs or `entity` objects (see below), or an external statistic id. Note: After specifying the first one, all others can only be selected with the same unit of measurement."
   type: list
 days_to_show:
   required: false


source/_dashboards/statistics-graph.markdown
```diff
@@ -30,7 +30,7 @@ type:
   type: string
 entities:
   required: true
-  description: "A list of entity IDs or `entity` objects (see below), or an external statistic id"
+  description: "A list of entity IDs or `entity` objects (see below), or an external statistic id. Note: After specifying the first one, all others can only be selected with the same unit of measurement."
```
frenck, line 33: This describes the YAML configuration parameters, there is no selection or UI in such cases? I'm confused about what that means in YAML?

Bascht74: Hi, I don't use YAML, trying to do stuff in the UI. So I don't know what is happening if you select entities with different unit of measurements via YAML. I can only speak for the UI. There the entities are filtered right after you select the first entities. After that, only entities of the same unit of measurements are shown. The card doesn't behave like the history graph card. There is no such restriction and a new graph is added with the new unit.

I tried in yaml, and the behavior is more than strange: 
I can add another entity with another uni there, but it is shown in the same graph with the wrong unit (= unit of the first entity). 

image

Power entity only: 
image

History graph card: 
image

cannot select power entity via UI:
image

can select power entity via UI with history graph card:
image

What do you think?

frenck: > Hi, I don't use YAML, trying to do stuff in the UI.

You suggested change is made on the YAML configuration default in the documentation, hence my comment on this PR. 
The rest of your UI-based comments are thus not relevant 🤷

Bascht74: Then we leave it as it is. 
There is no section for UI in the documentation right now. 

> The rest of your UI-based comments are thus not relevant 🤷

I wrote that, because the behavior via yaml is inconsistent as well, but in a different way.
Should I open an issue for that or does it work as designed from your point of view? (still learning)

frenck:
Closing the PR as per above.

source/_dashboards/statistics-graph.markdown
Bascht74: btw. in the documentation it doesn't say YAML, just `CONFIGURATION VARIABLES` after describing how to add the card via UI. Later, below `Alternatively, the card can be configured using YAML`. 

So I read it the way, that the standard is configuration via "UI" and alternatively you can do it via YAML. So I thought the variables are at least for both ways (maybe with additional infos like `list` ) for people doing it via YAML.

PR #25445, opened 2022-12-29
Installation in VirtualBox -> Deactivate paravirtualization

lenaschimmel:
## Proposed change

The [issue #71278](https://github.com/home-assistant/core/issues/71278) documents severe problems with timed automations, which occur since at least 2019 when using a VirtualBox VM. I can confirm that this still occurs with recent version (Home Assistant 2022.12.8, Supervisor 2022.12.1, Operating System 9.4), and that it affects more time-related features, which in sum render home assistant useless for many use cases.

The issue contains a fix, which has to be performed by the user by changing a setting of the VM. This setting can also be applied at installation time, so that the problem is prevented from the start. I can confirm that this setting indeed fixes the automation timing issue and several other timing issues that I experienced.

In the issue, there is no sign of any (current or future) change in the code base to fix the bug inside the home-assistant code. So this PR adds the needed instruction to the documentation.

## Type of change

## Additional information

- This PR fixes or closes issue: fixes [core / #71278](https://github.com/home-assistant/core/issues/71278)
- If this PR is rejected, I request to re-open the mentioned issue, so that other solutions can be discussed

## Checklist

### DIFF
--- a/source/_includes/installation/operating_system.md
+++ b/source/_includes/installation/operating_system.md
@@ -248,6 +248,7 @@ _All these can be extended if your usage calls for more resources._
     Using Wi-Fi on your VirtualBox host is unsupported.
     </div>
     6. Then go to "Audio" and choose "Intel HD Audio" as Audio Controller.
+    7. Go to "System" > "Acceleration" and set the Paravirtualization Interface to "None".
     <div class="note info">
 
     By default VirtualBox does not free up unused disk space. To automatically shrink the vdi disk image


lenaschimmel:
PS: As a user, I would prefer if the instructions would explain *why* the setting is needed. I omitted that explanation because I was not sure how to include this in the documentation without breaking the style.

frenck:
Disabling paravirtualization seems like an incorrect change, that is merely done to work around an issue. We should not adjust documentation to document workarounds, I do not accept such changes.

Things like these should be tracked in issues. Please report this issue with the HA operating system repository instead.

../Frenck

lenaschimmel:
I tried to write an issue, but since there were already many related issues (most of them closed due to inactivity), I was not able to come up with a new issue that adds valuable, new input.

Instead, I collected the existing information and tried to summarize it in a [new discussion](https://github.com/home-assistant/operating-system/discussions/2292). I hope that's ok.

PR #25465, opened 2022-12-30
Systemmonitor to integration key

gjohansson-ST:
## Proposed change

Move yaml for Systemmonitor to integration key

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/84859

## Checklist

### DIFF
--- a/source/_integrations/systemmonitor.markdown
+++ b/source/_integrations/systemmonitor.markdown
@@ -19,17 +19,15 @@ add the following to your `configuration.yaml` file:
 
 ```yaml
 # Example configuration.yaml entry
-sensor:
-  - platform: systemmonitor
-    resources:
-      - type: disk_use_percent
-        arg: /config
-      - type: memory_free
+systemmonitor:
+  - type: disk_use_percent
+    arg: /config
+  - type: memory_free
 ```
 
 {% configuration %}
-resources:
-  description: Contains all entries to display.
+systemmonitor:
+  description: integration domain
   required: true
   type: list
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
Doesn't this integration fall under ADR10 devices / services? Ie we want it to move to config flow only, eventually.

https://github.com/home-assistant/architecture/blob/master/adr/0010-integration-configuration.md

gjohansson-ST:
> Doesn't this integration fall under ADR10 devices / services? Ie we want it to move to config flow only, eventually.
> 

Not sure. My guess was that this one should be both so started with moving yaml to integration key.
I have been wrong before 😄

MartinHjelmare:
I don't think the integration falls under any of the other bullets. The system monitor retrieves state from psutil which is a service running on the host computer.

gjohansson-ST:
Given responses I had before (for example with `scrape`) I think perhaps it can be defined within "Integrations that integrate transports. These integrations allow users to define their own protocol. Examples include MQTT, serial, GPIO." as the user defines what the sensor should display among the options available/possible.

MartinHjelmare:
No, we don't define a transport for system monitor. MQTT is a transport, serial is a transport. There's no special transport in the system monitor integration. We're just using a service.

gjohansson-ST:
Alright. So we'll close this one then and move to config flow instead. No point in a debate.

MartinHjelmare:
👍

PR #25551, opened 2023-01-04
Add E1 Zoom to tested models

ghost:
## Proposed change

Adding the "E1 Zoom" model of Reolink camera to the list of supported models.

This model provides the main, sub and snapshots entities.  The Images entity noted in the docs is not provided and this is noted in the proposed change.

The E1 Zoom was confirmed working and provides a stream using 2023.1.0 using  E1 Zoom Firmware: v3.0.0.136_20121108
and Hardware: IPC_515SD6.  There is a ~5-10 second delay in the stream.  This delay was present in other methods used to steam the camera with HA also.

### DIFF
--- a/source/_integrations/reolink.markdown
+++ b/source/_integrations/reolink.markdown
@@ -37,7 +37,7 @@ The following models have been tested and confirmed to work:
 - RLN8-410 NVR
 - RLC-811A
 - RLC-420-5MP
-- E1 Zoom (Images Stream Not Supported)
+- E1 Zoom (Images stream not supported)
 - Reolink Video Doorbell PoE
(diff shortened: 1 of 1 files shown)


source/_integrations/reolink.markdown
```diff
@@ -37,6 +37,7 @@ The following models have been tested and confirmed to work:
 - RLN8-410 NVR
 - RLC-811A
 - RLC-420-5MP
+- E1 Zoom (Images Stream Not Supported)
```
frenck:
```suggestion
- E1 Zoom (Images stream not supported)
```

frenck (approved):
Thanks, @SilvrrGIT 👍

PR #25599, opened 2023-01-06
Recommend 32-bit on RPI3

### LINKED ISSUE #25598: Recommend 32-bit OS for Pi3
### Feedback

The Raspberry Pi 3 has 1GB of memory which is not enough to reliably run the 64-bit HAOS.  It becomes unstable and prone to silently crashing.  We should recommend 32-bit OS on this device.

### URL

https://www.home-assistant.io/installation/raspberrypi

### Version

_No response_

### Additional information

_No response_

NeilRickards:
Fixes https://github.com/home-assistant/home-assistant.io/issues/25598

### DIFF
--- a/source/_includes/installation/operating_system.md
+++ b/source/_includes/installation/operating_system.md
@@ -130,8 +130,10 @@ sudo apt install libfuse2
 
     {% if variant.key == "odroid-n2" %}
     [Guide: Flashing Odroid-N2 using OTG-USB](/hassio/flashing_n2_otg/)
-    {% elsif variant.key == "rpi4" or variant.key == "rpi3" %}
+    {% elsif variant.key == "rpi4" %}
       _(64-bit is recommended)_
+    {% elsif variant.key == "rpi3-64" %}
+      _(32-bit is recommended)_
     {% endif %}
 
 {% endfor %}


frenck:
Sorry, we actually don't recommend using 32-bits at all.

Support for 32 bits things are dropping hard in the Linux world, lots of things will not work or stop working in the near future. Hence we explicitly recommend 64-bits.

../Frenck

NeilRickards:
Should we then drop support for the RPI 3, as it's proving very unstable?  It doesn't degrade gracefully or restart under memory pressure, it just stops.  Or I can file another ticket requesting that HAOS handles memory pressure better?

frenck:
> Should we then drop support for the RPI 3

There is no need to?

>  as it's proving very unstable?

We are not aware of such issues at this point. At the time of writing this response, around 11.67% of all HA instances out there, use an RPi3 on 64-bits OS (versus just 1.86% RPi3 on 32-bits).

> Or I can file another ticket requesting that HAOS handles memory pressure better?

If you think there is a bug, please by all means raise an issue in the appropriate issue tracker.

../Frenck

NeilRickards:
Thanks very much for the response.
Possibly I have a higher than average number of integrations or add-ins.  Or maybe I've misdiagnosed the issue.
I've filed a separate ticket* (https://github.com/home-assistant/core/issues/85308) and linked to a discussion where others seem to have a similar problem (https://github.com/home-assistant/operating-system/issues/1999)

(* please let me know if this isn't in the right place)

PR #25721, opened 2023-01-12
Clarify use of template shorthand with if-then

TomBrien:
## Proposed change

When using if-then(-else) block, you can't use a template shorthand condition as the value to the `condition` key. Instead it has to be used directly with the `if`. Currently the template shorthand documentation can lead to confusion as we imply (out right say) you can use shorthand wherever there is a `condition(s)` key, This adds an example to the if-then docs to clarify the correct usage.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -604,7 +604,7 @@ This action supports nesting, however, if you find yourself using nested if-then
 actions in the `else` part, you may want to consider using
 [choose](#choose-a-group-of-actions) instead.
 
-[Template shorthand notation](https://www.home-assistant.io/docs/scripts/conditions/#template-condition-shorthand-notation) can be used with if-then. The code below is equivalent to the example above.
+[Template shorthand notation][shorthand-template] can be used with if-then. The code below is equivalent to the example above.
 
 ```yaml
 script:
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -604,6 +604,22 @@ This action supports nesting, however, if you find yourself using nested if-then
 actions in the `else` part, you may want to consider using
 [choose](#choose-a-group-of-actions) instead.
 
+[Template shorthand notation](https://www.home-assistant.io/docs/scripts/conditions/#template-condition-shorthand-notation) can be used with if-then. The code below is equivalent to the example above.
+
+```yaml
+script:
+  - if: "{{ is_state('zone.home', 0) }}"
+    then:
+      - alias: "Then start cleaning already!"
+        service: vacuum.start
+        target:
+          area_id: living_room
+    else:
+      - service: notify.notify
+        data:
+          message: "Skipped cleaning, someone is home!"
+```
+
 ## Choose a Group of Actions
 
 This action allows you to select a sequence of other actions from a list of sequences.
--- a/source/_docs/scripts.markdown
+++ b/source/_docs/scripts.markdown
@@ -606,6 +606,8 @@ actions in the `else` part, you may want to consider using
 
 [Template shorthand notation][shorthand-template] can be used with if-then. The code below is equivalent to the example above.
 
+{% raw %}
+
 ```yaml
 script:
   - if: "{{ is_state('zone.home', 0) }}"
@@ -619,6 +621,8 @@ script:
         data:
           message: "Skipped cleaning, someone is home!"
 ```
+
+{% endraw %}
 
 ## Choose a Group of Actions
 


source/_docs/scripts.markdown
```diff
@@ -604,6 +604,22 @@ This action supports nesting, however, if you find yourself using nested if-then
 actions in the `else` part, you may want to consider using
 [choose](#choose-a-group-of-actions) instead.
 
+[Template shorthand notation](https://www.home-assistant.io/docs/scripts/conditions/#template-condition-shorthand-notation) can be used with if-then. The code below is equivalent to the example above.
```
frenck: Don't use full links

source/_docs/scripts.markdown
```diff
@@ -604,6 +604,22 @@ This action supports nesting, however, if you find yourself using nested if-then
...
+
+```yaml
+script:
+  - if: "{{ is_state('zone.home', 0) }}"
```
frenck, line 613: This feels misplaced? As in, like an oddity to mention in the middle out of the blue?
This page is not aimed at showing how conditions work.

The conditions page has IMHO a perfectly fine explanation for this already?

frenck (changes requested):

frenck:
> When using if-then(-else) block, you can't use a template shorthand condition as the value to the condition key

I think you are confusing a "condition" versus a list of c `conditions` (the array).

That sounds weird and not correct either? 


> Currently the template shorthand documentation can lead to confusion as we imply (out right say) you can use shorthand wherever there is a condition(s) key,

That is not this page, so not sure why it is "fixed" on this page either.

Additionally, I don't see the confusion?

../Frenck

TomBrien:
Currently we say "This shorthand notation can be used everywhere in Home Assistant where conditions are accepted" and have examples on the conditions docs showing short hand in place of either the whole list `conditions: "{{ ... }}"` and also in place of a single condition.

```yaml

conditions:
  - "{{ (state_attr('device_tracker.iphone', 'battery_level')|int) > 50 }}"
  - condition: state
    entity_id: alarm_control_panel.home
    state: armed_away
  - "{{ is_state('device_tracker.iphone', 'away') }}"
```

There is also a specific example that talks about condition actions (I agree not the same thing as here). That shows `- condition: {{ ... }}`.

My reasoning for clarifying here was not to dilute the conditions page with a specific example for this syntax since, to me at least, it's not clear that `if` is equivalent to `conditions`. I assumed people struggling with and if-then would come to this page. I'm happy to make changes to that page or just to close this. Maybe it's just my confusion but I'm working on the concept that if I got confused others probably have too.

frenck:
> This shorthand notation can be used everywhere in Home Assistant where conditions are accepted

That is correct 🤷 

> There is also a specific example that talks about condition actions (I agree not the same thing as here). That shows - condition: {{ ... }}.

There is no such thing as a condition action.

>  it's not clear that if is equivalent to conditions. 

I don't agree with that. The paragraph explains that in the first line? "This action allows you to conditionally (`if`)..."

> I assumed people struggling with and if-then would come to this page.

I've not came across such a case at this point before, to be honest.

../Frenck

source/_docs/scripts.markdown
TomBrien:
```suggestion
[Template shorthand notation][shorthand-template] can be used with if-then. The code below is equivalent to the example above.
```

source/_docs/scripts.markdown
TomBrien: We have something pretty much equivalent for [Repeat-until](https://www.home-assistant.io/docs/scripts/#repeat-until) on this page, which is why I thought here was sensible

source/_docs/scripts.markdown
```diff
@@ -604,6 +604,22 @@ This action supports nesting, however, if you find yourself using nested if-then
...
+    then:
+      - alias: "Then start cleaning already!"
+        service: vacuum.start
+        target:
+          area_id: living_room
+    else:
+      - service: notify.notify
+        data:
+          message: "Skipped cleaning, someone is home!"
+```
```
TomBrien, line 621:
```suggestion
{% raw %}

```yaml
script:
  - if: "{{ is_state('zone.home', 0) }}"
    then:
      - alias: "Then start cleaning already!"
        service: vacuum.start
        target:
          area_id: living_room
    else:
      - service: notify.notify
        data:
          message: "Skipped cleaning, someone is home!"
```

{% endraw %}
```

TomBrien:
Fair enough

PR #25777, opened 2023-01-16
Fix example Time Pattern Value

garritye:
Original sample code was rejected by Home Assistant.  When I change code to remove the surrounding quotes, Home Assistant saw this as valid.  My changed code works properly.  I didn't check the other examples.

## Type of change

## Additional information

## Checklist

- [  x ] This PR uses the correct branch, based on one of the following:

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -763,7 +763,7 @@ automation 3:
   trigger:
     - platform: time_pattern
       # You can also match on interval. This will match every 5 minutes
-      minutes: "/5"
+      minutes: /5
 ```
 
 <div class='note warning'>


source/_docs/automation/trigger.markdown
```diff
@@ -763,7 +763,7 @@ automation 3:
   trigger:
     - platform: time_pattern
       # You can also match on interval. This will match every 5 minutes
-      minutes: "/5"
+      minutes: /5
```
frenck, line 766: This change is incorrect. Time pattern is a string and, therefore, should be enquoted.

frenck (changes requested):

garritye:
Hi Frank,

First of all, I thank you for authoring this triggering function.  It's
exactly what I needed.  However, I still think that quoting creates a
problem, as shown below.

It does not allow me to save when I quote this, either in single or double
quotes.  These are the messages I receive:
[image: image.png]
and...
[image: image.png]

But, when I remove the quotes, it works fine.  This is the same screen
after changing and saving.

[image: image.png]

Is there something else I can use to help document this?

Thank you. I appreciate your prompt attention.

Ed Garrity.

frenck:
> I thank you for authoring this triggering function.

I didn't create it, sorry.

> However, I still think that quoting creates a problem, as shown below.

Nothing is visible, as you used email to respond. Please don't do that. It doesn't work (well just partly).

../Frenck

garritye:
Frenck, Thanks for the reply.

The issue is with the visual editor it seems.
When I enter `"/5"` as the time pattern in the visual editor, then look at the YAML it generates, it's attempting to escape the quote.
It generates the following YAML code:

```
  - platform: time_pattern
    hours: "\"/5\""
```

This produces the following error that's shown on screen: `Message malformed: invalid time_pattern value for dictionary value @ data['hours']`.

When I don't include the quotes in the visual editor, it produces the following YAML:

```
  - platform: time_pattern
    hours: /5
```

And this does work.

The documentation looks like it's for YAML, so it's not wrong.
I've changed my fork to restore the quotes, and added a note that in the visual editor, quotes should not be added.

Is that the best solution, or do you think I should drop the PR and open an issue regarding the visual editor generating bad code if quotes are used?

Thanks and have a great day !

frenck:
> When I enter "/5" as the time pattern in the visual editor

Yes? That is expected. It will make a string for you.

Sounds like you want to report an issue with the Home Assistant frontend. If so, please raise an issue in the frontend issue tracker. The change you are suggesting in this PR is, however, incorrect.

../Frenck

garritye:
I agree.  Thank you.

Have a great day!

Ed

PR #25856, opened 2023-01-20
Change value_json.XXX to value_json['XXX'] in docs

jflecool2:
## Proposed change
Although I understand the aesthetics and simplicity for beginners of using periods for accessing items of a dict and calling methods (`a.key.method()`), it should not be done in value_json specifically. value_json implies values from the outside world: a radio REST API, a robot vacuum, or a cheap MQTT device, none of them made specifically for home assistant. If the vendor stumble upon naming something that's a python dict's method... well, debugging will (was) unpleasant. 
As an example:
```
{% set my_test_json = {
  "update": {"time":"123"},
  "u": {"time":"456"}
} %}
bracket update: {{ my_test_json['update']['time'] }}
bracket u: {{ my_test_json['u']['time'] }}
period update: {{ my_test_json.update.time }} <- Will break because update is a python method for dict
period u: {{ my_test_json.u.time }}
```
[Here's a list of word that will break templating when using period](https://www.w3schools.com/python/python_ref_dictionary.asp)
I fully understand this PR implies decision-making, weighting user-friendlyness and consistency across the hassio documentation... 

## Type of change
Change value_json.XXX to value_json['XXX'] in docs

## Additional information
None

## Checklist

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -165,7 +165,7 @@ automation:
     - platform: mqtt
       topic: "living_room/switch/ac"
       payload: "on"
-      value_template: "{{ value_json.state }}"
+      value_template: "{{ value_json['state'] }}"
 ```
 
 {% endraw %}
--- a/source/_docs/configuration/templating.markdown
+++ b/source/_docs/configuration/templating.markdown
@@ -588,7 +588,7 @@ Example using `is_defined` to parse a JSON payload:
 {% raw %}
 
 ```text
-{{ value_json.val | is_defined }}
+{{ value_json['val'] | is_defined }}
 ```
 
 {% endraw %}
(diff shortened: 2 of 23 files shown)


jbouwh (changes requested):
I do not agree with the changes you want to make in general. Though it would be good to update the general template docs and not that the rules.
The reason I do not agree is that it will make templates longer then needed, and IMO brackets should only be used is special cases.

jflecool2:
I know I followed the REST sensor religiously: copied pasted and change the names and read the troubleshooting steps of that sensor. I'd never guess to look into templating docs, as the template provided in the docs are sure to work already. This is why the scope of my changes limits to json_value, as its the only one where the probability of those words are higher, IMO out-weighing the length argument. But I do understand the "short" argument.

PR #26080, opened 2023-02-02
Correct API Naming

ghost:
Correcting incorrect references to Drive API and changing "Google Mail API" to "Gmail API" to match the actual naming used by Google.


frenck (approved):
Thanks, @SilvrrGIT 👍 

This BTW depends a bit on where in the world you are. In some countries, `GMail` can't be used by Google.

../Frenck

PR #26104, opened 2023-02-04
Amend code snippet for Calendar schedule example to render actions in the UI

marcjay:
## Proposed change
~Original 'if' action appeared to have invalid syntax. Correcting with a working version.~
Original 'if' action cannot be rendered in UI. Alternative syntax does render, and may be easier for beginners

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/calendar.markdown
+++ b/source/_integrations/calendar.markdown
@@ -130,7 +130,8 @@ automation:
         value_template: "{{ 'Front Lights' in trigger.calendar_event.summary }}"
     action:
       - if:
-          - "{{ trigger.event == 'start' }}"
+        - condition: template
+          value_template: "{{ trigger.event == 'start' }}"
         then:
           - service: light.turn_on
             entity_id: light.front


source/_integrations/calendar.markdown
```diff
@@ -130,7 +130,8 @@ automation:
         value_template: "{{ 'Front Lights' in trigger.calendar_event.summary }}"
     action:
       - if:
-          - "{{ trigger.event == 'start' }}"
```
balloob, line 133: This is valid short-hand notation for a template condition

marcjay: Oh, hmm, well then I'm not sure why the Automations editor didn't treat it as valid syntax until I made this change - action wouldn't show up in the UI and wouldn't save. On the latest HA version. Will have another look.

balloob: Not all valid YAML can be correctly represented in the UI. Try running it and see if it works 👍

marcjay: Ok, I had a chance to retry this - it does indeed save and work. Not sure why it wasn't saving before.

Whilst it would be best if the UI could represent this YAML, would it not be a better user experience if the sample code did indeed render in the UI? 
I'll rename the PR as it's not a 'fix', but as a user playing with this, I'd personally find YAML that renders in the UI as:
image
easier to use than YAML that renders as:
image

frenck: The current example is valid, if it results in a non-working UI, it should be reported in the issue tracker of the frontend repository.

Not all features available in YAML are available in the UI, however, in those cases a YAML block/editor should be shown in the UI.

frenck:
Closing the PR on this end, as the documented example is currently correct.

../Frenck

PR #26106, opened 2023-02-04
MQTT Button - Add a hint for other button services

MarZab:
I spent several hours trying to get MQTT button working for my automation while what I really needed was `device_trigger`. This hint would have saved my day.

## Proposed change
Added a strategic hint to MQTT Button to guide users to alternative options.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/button.mqtt.markdown
+++ b/source/_integrations/button.mqtt.markdown
@@ -212,3 +212,8 @@ mqtt:
       entity_category: "config"
       device_class: "restart"
 ```
+
+### Automation
+
+MQTT Button offers limited automation capabilites - to make a remotely controlled button/press event see [`device_trigger.mqtt`](/integrations/device_trigger.mqtt/) or [`binary_sensor.mqtt`](/integrations/binary_sensor.mqtt/)
+


source/_integrations/button.mqtt.markdown
```diff
@@ -212,3 +212,8 @@ mqtt:
       entity_category: "config"
       device_class: "restart"
 ```
+
+### Automation
+
+MQTT Button offers limited automation capabilites - to make a remotely controlled button/press event see [`device_trigger.mqtt`](/integrations/device_trigger.mqtt/) or [`binary_sensor.mqtt`](/integrations/binary_sensor.mqtt/)
```
frenck, line 218: I'm not following what you are trying to say here? Could you clarify?
These are three completely different and unrelated concepts, so I don't see the connection (well the latter two have overlap a bit, as binary sensors tied to a device will provide device triggers automatically), but I don't see what a button has to do with that?

MarZab: A [MQTT Switch](https://www.home-assistant.io/integrations/switch.mqtt/) can be used to toggle on the interface and to control it via MQTT - finding the MQTT Button would logically follow that you can do the same - have it on the UI while also offering the option to trigger it via MQTT - this however is not possible (or completely unrelated) and I wanted to point the reader in the correct direction.

Thank you for your patience - I might still be wrong here :)

frenck: > A [MQTT Switch](https://www.home-assistant.io/integrations/switch.mqtt/) can be used to toggle on the interface and to control it via MQTT

Correct.

> finding the MQTT Button would logically follow that you can do the same

Yes, it can trigger things on a device, just like a switch in that regard (or any other entity that provides controls).

>  have it on the UI while also offering the option to trigger it via MQTT

That is not what a button does. You can have it on the UI to trigger a change on the  device. But the option to trigger it from MQTT? This is not a sensor, it is control.

A button entity is an entity that can fire an event / trigger an action towards a device or service but remains stateless from the Home Assistant perspective.

If it wasn't stateless (thus two-way), one wouldn't be using a button to begin with, it would have been a switch or binary sensor.

The other side of the spectrum is events. Events are basically reverted buttons. Not control from Home Assistant, but controls from the device sending in events to Home Assistant (e.g. doorbell button presses). But events are mainly momentary, like a button press, otherwise it would be a sensor if it was an uncontrollable remote state.

Events can be used in e.g., trigger templates, automation triggers, device triggers and many more places. Sensors can be that too, but hang on to their state (and can thus be displayed again).

So, to recap, from the above, the line added to this document, doesn't make much sense IMHO.

MarZab: I agree with all points above and thank you for the extensive explanation.

To explain what I wanted to accomplish with MQTT Button that I ultimately did with a `device_trigger.mqtt` - once I deducted the points above - I have a hardware click/toggle button that does not keep a state and it does not have a controllable remote state. I use this as an event to trigger my automations.

I hope you can see where I got confused within the difference between MQTT Button and MQTT Switch and how one could use/misuse them. In my mind, there is no difference between pressing a button on the UI and pressing the same button remotely (ie. keeping the last pressed event and MQTT as the source of the event) as the stateful equivalent can be acomplished with MQTT Switch.

So here is a final jab at this and I will close this PR if you think this comment does not add value to the documentation.

> MQTT Button does not have a state itself (beyond the last trigger event) and it is only used as an event source from the Home Assistant frontend. To send an event to Home Assistant, you can use [`device_trigger.mqtt`](/integrations/device_trigger.mqtt/) instead.

frenck: That sentence suggested at the end is not correct.

- Buttons are not limited to  the frontend
- Events and device triggers are not the same things. Device triggers can leverage events, but not related either.

I'm going to close the PR on this end, as I  think most  of it is caused by misunderstanding the Core building blocks (and not MQTT entities itself). Some of my text above was copy pasted from the button documentation.

If those pages need more extensive documentation, I'd welcome suggestion on that. I donn't think the MQTT page should explain our building blocks (as MQTT is merely a consumer of those, like any other integration).

../Frenck

PR #26221, opened 2023-02-12
Add section describing broken devices

StevenLooman:
## Proposed change

Add a section to describe problems with some devices.

## Type of change

## Additional information

- This PR fixes or closes issue: fixes https://github.com/home-assistant/core/issues/82032, https://github.com/home-assistant/core/issues/82599

## Checklist

### DIFF
--- a/source/_integrations/upnp.markdown
+++ b/source/_integrations/upnp.markdown
@@ -43,3 +43,15 @@ logger:
 ```
 
 When creating an issue, please include the (relevant) logging with the issue. Any sensitive information such as IPs can be obfuscated.
+
+## Warnings statistics sensors
+
+Some devices do not follow the UPnP/IGD standard or are simply broken and provide invalid values when the router's traffic counters are read. According to the standard, the traffic counters should provide an increasing value between 0 bytes and 4.294.967.294 bytes. After the maximum value is reached, the counter should roll over to 0 and restart counting.
+
+There are several devices which do not adhere to this standard causing the traffic counters to be unusable. As a result, the traffic sensors might not be working properly, or Home Assistant might even given a warning such `Entity sensor.fritz_box_upload_summe from integration upnp has state class total_increasing, but its state is not strictly increasing. Triggered by state .... Please create a bug report at ...`. This is something the integration can nothing do about.
+
+Devices/brands with **known** defects (incomplete list!):
+* Fritz!
+* Compal Broadband Networks, Inc CH7465LG
+
+When this message is seen, the sensors regarding traffic should be disabled.


source/_integrations/upnp.markdown
```diff
@@ -43,3 +43,15 @@ logger:
 ```
 
 When creating an issue, please include the (relevant) logging with the issue. Any sensitive information such as IPs can be obfuscated.
+
+## Warnings statistics sensors
+
+Some devices do not follow the UPnP/IGD standard or are simply broken and provide invalid values when the router's traffic counters are read. According to the standard, the traffic counters should provide an increasing value between 0 bytes and 4.294.967.294 bytes. After the maximum value is reached, the counter should roll over to 0 and restart counting.
+
+There are several devices which do not adhere to this standard causing the traffic counters to be unusable. As a result, the traffic sensors might not be working properly, or Home Assistant might even given a warning such `Entity sensor.fritz_box_upload_summe from integration upnp has state class total_increasing, but its state is not strictly increasing. Triggered by state .... Please create a bug report at ...`. This is something the integration can nothing do about.
```
frenck, line 51: If HA can detect it, sure as hell the integration can do that too and handle it properly?

It sounds odd and weird to document errors and just consider it a normal thing?

../Frenck

frenck (changes requested):

source/_integrations/upnp.markdown
```diff
@@ -43,3 +43,15 @@ logger:
...
+
+Devices/brands with **known** defects (incomplete list!):
+* Fritz!
+* Compal Broadband Networks, Inc CH7465LG
+
+When this message is seen, the sensors regarding traffic should be disabled.
```
frenck, line 57: Same as above, if this is know, why does the integration creates them?

source/_integrations/upnp.markdown
StevenLooman: Should the  integration disable the sensors? This seems like automatic behavior which can cause lots of confusion.

frenck: Documenting an error is normal not an acceptable option either...

StevenLooman: What do you propose then?

source/_integrations/upnp.markdown
StevenLooman: It is not always known. Perhaps there are Fritz!boxes that do provide proper data (is it the model, the firmware version, ...)?

source/_integrations/upnp.markdown
frenck: Making sure this error doesn't end up in the logs / happens.

StevenLooman: How would you propose to prevent this error from happening? Add some machine learning to the component and/or integrate ChatGPT4 to let it figure out when it is an error or regular behavior?

Adding the /s, just in case...

PR #26287, opened 2023-02-17
Update influxdb.markdown

Vortexmind:
Unless a port is specified in the yaml configuration, integration with 2.xx failed with an HTTP 405 error

## Type of change

## Checklist

- [ x ] The documentation follows the Home Assistant documentation [standards].

### DIFF
--- a/source/_integrations/influxdb.markdown
+++ b/source/_integrations/influxdb.markdown
@@ -56,8 +56,8 @@ host:
   default: localhost
 port:
   type: integer
-  description: Port to use. 2.xx - No default port for 2.xx, otherwise 8086.
-  required: false
+  description: Port to use. 2.xx - Must specify port for 2.xx, otherwise 8086.
+  required: true
   default: 8086
 path:
   type: string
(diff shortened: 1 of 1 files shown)


source/_integrations/influxdb.markdown
```diff
@@ -56,8 +56,8 @@ host:
   default: localhost
 port:
   type: integer
-  description: Port to use. 2.xx - No default port for 2.xx, otherwise 8086.
-  required: false
+  description: Port to use. 2.xx - Must specify port for 2.xx, otherwise 8086.
+  required: true
```
frenck: This change is not correct, it isn't required. See the schema defined in our code here:

<https://github.com/home-assistant/core/blob/289bab6f875698c0bc6e419df979cba88ff70100/homeassistant/components/influxdb/const.py#L141>

Unless, I've missed something? Is there a PR or changing pending that matches this documentation change?

../Frenck

frenck (changes requested):

source/_integrations/influxdb.markdown
Vortexmind: You are right. The `port` is not required strictly speaking, but with Influx 2.xx it must be included in the config (even with the default `8086` value) otherwise the integration will not work, see [the error here](https://community.home-assistant.io/t/external-influxdb-bucket-is-not-accessible-due-to-405/476018/3) which I also ran into. 

I have changed my commit to only update the `description` field.

frenck (approved):
Thanks, @Vortexmind 👍 

../Frenck

PR #26450, opened 2023-03-03
Add RLC-410 Support

ghost:
Adding RLC-410 to list of supported devices. 

Camera streaming and motion sensor are both confirmed to be working.


klaasnicolaas (approved):
✅ Approved

Please keep our pull request template as it is next time 😉

PR #26998, opened 2023-04-13
use the correct type

Mariusthvdb:
for the format: option on datetime sensors. 
this is a longstanding confusing guideline. the Frontend doesn't expect a timestamp as output here, following definitions in https://www.home-assistant.io/docs/configuration/templating/#time.
It does require a `device_class: timestamp`.

## Type of change

## Additional information

## Checklist


source/_dashboards/entities.markdown
```diff
@@ -85,7 +85,7 @@ secondary_info:
   type: string
 format:
   required: false
-  description: "How the state should be formatted. Currently only used for timestamp sensors. Valid values are: `relative`, `total`, `date`, `time` and `datetime`."
+  description: "How the state should be formatted. Currently only used for datetime sensors with `device_class: timestamp`. Valid values are: `relative`, `total`, `date`, `time` and `datetime`."
```
frenck, line 88: This seems incorrect. The existing wording is right.
We don't have datetime sensors (not yet, that is actually in process). The device class already defines what type of entity (in this case sensor), thus  the existing "timestamp sensors" term is correct.

../Frenck

Mariusthvdb: would you agree on this:
```
Currently only used for sensors with `device_class: timestamp`
```

because that is the improvement I am seeking to make (shouldn't have mentioned the datetime in the first place).

adding that would make it more succinct, and prevent users from trying the entity option in the card on a sensor that actually outputs a timestamp (1681456993.59573) but does not carry the `device_class: timestamp`.

frenck: `device_class: timestamp` <- is a technical detail on the classification looks like in code. This is normal text.

> that actually outputs a timestamp (1681456993.59573)

That isn't a timestamp sensor. But a sensor with a (possible) number or textual value.

Mariusthvdb: Thanks for pulling me through this (....), because that is the exact confusion I am trying to prevent..:

above number was the output of `{{as_timestamp(now())}}`, and which one might believe to be a timestamp sensor because of the `as_timestamp()` in the template.

However, one can only get a true timestamp sensor in the sense of the documentation, by setting a `device_class: timestamp` .

```
# this is expected in the card (an `as_datetime`) sensor with `device_class: timestamp` set to it 

        state: >
          {% set base = states('sensor.next_alarm_base')|int(default=0) %}
          {{as_datetime(base)|as_local}}
        device_class: timestamp

# the as_timestamp does not work although it is outputting a timestamp

        state: >
          {% set base = states('sensor.next_alarm_base')|int(default=0) %}
          {{as_timestamp(as_datetime(base)|as_local)}}
```

PR #27107, opened 2023-04-23
Update templating docs

veleek:
## Proposed change
Clarified the initial description of the `trigger` variable.  Added links from the sections about the data available from each platform to the specific platform in the Trigger page.

## Type of change

## Additional information

I don't know if there are guidelines around whether or not to include links in headers, but for this list it seemed like the best option, but I'm happy to move the links into sentences right underneath the headers (e.g. "This is the data available in the [Calendar Trigger](/docs/automation/trigger/#calendar-trigger)").

I was also considering back linking from the trigger section to the data section since that's a very common use case for me, so I'm assuming others do the same type of lookups.

## Checklist


source/_docs/automation/templating.markdown
```diff
@@ -13,18 +13,20 @@ The variable `this` is the [state object](/docs/configuration/state_object) of t
...
+Triggers from all platform will include the following data.
 
 | Template variable | Data |
 | ---- | ---- |
-| `trigger.id` | Optional trigger `id`, or index of the trigger.
+| `trigger.id` | The [`id` of the trigger](/docs/automation/trigger/#trigger-id).  If `id` is omitted, this will be set to the trigger index.
 | `trigger.idx` | Index of the trigger. (The first trigger idx is `0`.)
 
-### Calendar
+### [Calendar](/docs/automation/trigger/#calendar-trigger)
```
frenck: I think adding links to different pages on headings is actually not a good idea.

Instead, we should link to the paragraph itself.

This comment applies to every occurrence below.

source/_docs/automation/templating.markdown
```diff
@@ -13,18 +13,20 @@ The variable `this` is the [state object](/docs/configuration/state_object) of t
...
 
 ### All
 
-The following describes trigger data associated with all platforms.
+Triggers from all platform will include the following data.
 
 | Template variable | Data |
 | ---- | ---- |
-| `trigger.id` | Optional trigger `id`, or index of the trigger.
+| `trigger.id` | The [`id` of the trigger](/docs/automation/trigger/#trigger-id).  If `id` is omitted, this will be set to the trigger index.
```
frenck: This text is confusing, as this is an event result: How would one omit this here?
Let's avoid the term `omitted` in this case to avoid confusion.

source/_docs/automation/templating.markdown
```diff
@@ -13,18 +13,20 @@ The variable `this` is the [state object](/docs/configuration/state_object) of t
 
 ## Available Trigger Data
 
-The following tables show the available trigger data per platform.
+The variable `trigger` is an object that contains details about which trigger activated the automation.  The data can be used by templates to modify the actions that are performed by the automation or displayed in a message.  For example, you could create an automation that that can be triggered by multiple sensors and then use the location of the sensor to specify a light to activate; or you could send a notification that contains the friendly name of the sensor that triggered it.
```
frenck:
```suggestion
The variable `trigger` is an object that contains details about which trigger triggered the automation.

Templates can use the data to modify the actions performed by the automation or displayed in a message. For example, you could create an automation that multiple sensors can trigger and then use the sensor's location to specify a light to activate; or you could send a notification containing the friendly name of the sensor that triggered it.
```

source/_docs/automation/templating.markdown
```diff
@@ -13,18 +13,20 @@ The variable `this` is the [state object](/docs/configuration/state_object) of t
...
+
+Each [trigger platform](/docs/automation/trigger/#event-trigger) can include additional data specific to that platform.
 
 ### All
 
-The following describes trigger data associated with all platforms.
+Triggers from all platform will include the following data.
```
frenck:
```suggestion
Triggers from all platforms will include the following data.
```

frenck (changes requested):

source/_docs/automation/templating.markdown
veleek: @frenck - I agree in this context it's a bit confusing.  It is copied from [the existing Trigger ID documentation](https://www.home-assistant.io/docs/automation/trigger/#trigger-id) almost verbatim.  I didn't like the fact that the old docs which don't really make it clear how it works.

Do you have a suggestion for a better way?  What If I just drop the whole second sentence and keep the link to the Trigger ID page?

source/_docs/automation/templating.markdown
veleek: @frenck - Can you clarify what you mean by "link to the paragraph"?  Do you mean something similar to what I suggested in the PR description?

> I don't know if there are guidelines around whether or not to include links in headers, but for this list it seemed like the best option, but I'm happy to move the links into sentences right underneath the headers (e.g. "This is the data available in the [Calendar Trigger](/docs/automation/trigger/#calendar-trigger)").

If so, I can add a line of text at the start of each table that includes a link like that.

source/_docs/automation/templating.markdown
frenck: > I agree in this context it's a bit confusing. It is copied from [the existing Trigger ID documentation](https://www.home-assistant.io/docs/automation/trigger/#trigger-id) almost verbatim.

Sure, but that part is about user input (defining the trigger). This documentation is output, thus not "omittable".

> Do you have a suggestion for a better way?

I think the current was pretty clear. Or maybe:

```
The `id` of the trigger. This is either a custom [`id` of the trigger](/docs/automation/trigger/#trigger-id) or this will be set to the trigger index if that wasn't set on the trigger configuration.

Comment (veleek) [reply source/_docs/automation/templating.markdown:28]: Yup, like I said, I agree that it's a bit confusing.  Do you just want me to revert this change, or do you have a suggestion for a better alternative?

Comment (veleek) [reply source/_docs/automation/templating.markdown]: My beef with the original one was the fact that the first word was `optional` even though it would always have a value.  You have to go to the original definition of Trigger ID to understand what it means anyway. :P

How about we just go dead simple and just include the first sentence of you suggestion?

```
The [`id` of the trigger](/docs/automation/trigger/#trigger-id).
```

The linked page already has a definition that clarifies exactly what the value will be, so let's just keep the definition in one place and then we don't even have to worry about the input vs. output semantics.

source/_docs/automation/templating.markdown
veleek: Most of these sections ONLY contain a table, so there was no obvious place to add a link without adding extra text.

frenck: > If so, I can add a line of text at the start of each table that includes a link like that.

Yeah or a simple text/link to each paragraph with "Read more about configuring this trigger here".

The problem is that more often, headers in documentation (in general) are anchor links, in this case, it links to a different piece of documentation; which I think is unexpected/not good behavior.

veleek: Ah yes, I've been using the header anchors a BUNCH for this PR!  That'd definitely be weird to have both together.  

I'll add the links inside the sections this afternoon.

frenck (approved):
Nice! Thanks @veleek 👍 

../Frenck

PR #27176, opened 2023-04-27
Update Packages folder - simplify

ArveVM:
Replacement/simplification of point I was trying to make in #26320. Not sure what happend to that one, I might have cocked up something so it got stuck;)

have tried to simplify;
- removed the option for !include_dir_merge_named from origninal doc, so now this is all about !include_dir_named
- added info on this way of including keep indentation "as if written in configuration.yaml"

## Proposed change

The purpose of this PR is to be able to refer to documentation when urging people to use "packages-folders with the indentation of configuration.yaml" 

Currently I've seen several examples where they have read the documentation and set up with !include_dir_merge_named and having to keep up with different indentation than most examples that can be found in communities etc.

## Type of change

## Additional information

## Checklist


source/_docs/configuration/packages.markdown
```diff
@@ -74,33 +74,16 @@ Components inside packages can only specify platform entries using configuration
 
 ## Create a packages folder
 
-One way to organize packages is to create a folder named "packages" in your Home Assistant configuration directory. In the packages directory you can store any number of packages in a YAML file. This entry in your `configuration.yaml` will load all packages:
+One way to organize packages is to create a folder named "packages" in your Home Assistant configuration directory. In the packages directory you can store any number of packages in a YAML files. This entry in your `configuration.yaml` will load all YAML-files in Packages-folder and subfolders:
```
c0ffeeca7:
```suggestion
One way to organize packages is to create a folder named "packages" in your Home Assistant configuration directory. In the packages directory, you can store any number of packages in a YAML file. This entry in your `configuration.yaml` will load all YAML-files in this _packages_ folder and its subfolders:
```

source/_docs/configuration/packages.markdown
```diff
@@ -74,33 +74,16 @@ Components inside packages can only specify platform entries using configuration
...
 
 ```yaml
 homeassistant:
   packages: !include_dir_named packages
 ```
 
-This uses the concept splitting the configuration and will include all files in a directory with the keys representing the filenames.
-See the documentation about [splitting the configuration](/docs/configuration/splitting_configuration/) for more information about `!include_dir_named` and other include statements that might be helpful. The benefit of this approach is to pull all configurations required to integrate a system, into one file, rather than across several.
+The benefit of this approach is to pull all configurations required to integrate a system, into one file, rather than across several.
```
c0ffeeca7:
```suggestion
The benefit of this approach is to pull all configurations required to integrate a system into one file&mdash;rather than keeping them spread across several files.
```

c0ffeeca7 (changes requested):

ArveVM (commented):
Great improvements - thanks :)

source/_docs/configuration/packages.markdown
```diff
@@ -74,33 +74,16 @@ Components inside packages can only specify platform entries using configuration
...
+One way to organize packages is to create a folder named "packages" in your Home Assistant configuration directory. In the packages directory, you can store any number of packages in a YAML file. This entry in your `configuration.yaml` will load all YAML-files in this _packages_ folder and its subfolders:
 
 ```yaml
 homeassistant:
   packages: !include_dir_named packages
 ```
 
-This uses the concept splitting the configuration and will include all files in a directory with the keys representing the filenames.
-See the documentation about [splitting the configuration](/docs/configuration/splitting_configuration/) for more information about `!include_dir_named` and other include statements that might be helpful. The benefit of this approach is to pull all configurations required to integrate a system, into one file, rather than across several.
+The benefit of this approach is to pull all configurations required to integrate a system into one file&mdash;rather than keeping them spread across several files.
```
c0ffeeca7, line 84: I just noticed that the reference to the documentation on configuration splitting somehow got dropped here. Was this intentional?

source/_docs/configuration/packages.markdown
```diff
@@ -74,33 +74,16 @@ Components inside packages can only specify platform entries using configuration
...
+You can use other !include-methods for packages folder, but with this one you get yaml-indentation 'as if the content was written in configuration.yaml'.
 
-The following example allows to have subfolders in the `packages` folder, which could make managing multiple packages easier by grouping:
```
c0ffeeca7, line 87: Remind me, why would this documentation be better off without this example section? Is it incorrect?

source/_docs/configuration/packages.markdown
ArveVM: the intention was to simplify.
There is still a reference to splitting configuration in first paragraph of the page - so it applies to everytning Packages - so why repeat in 'create packages folder'-subsection?

source/_docs/configuration/packages.markdown
ArveVM: not tecnically incorrect, but for noobs - it is not clear that the docs show two separate ways of doing this - and that both include files in subfolders.
Tried to separate them more clearly in #26320, but that bounced back on "why the two examples" :)

So just trying to simplify,,, I guess the super advanced users see the splitting config-principles and can do it this way,,, or they'll include by functionality - or another way :)

For myself it is much simpler to explain to peers to use "same indentation as in configuration yaml" ,,, which is not what they currently read out of the current documentatjon (have assisted two who have gone for the !include_dir_merge_named packages option because that supposedly is the only one including subfolders :(  -> therefore the request for simplification :)

source/_docs/configuration/packages.markdown
ArveVM: should it perhaps be included again,, to compensate for the removal of the !include_dir_merge_named packages - example?

source/_docs/configuration/packages.markdown
c0ffeeca7: Yes, ok, I think I'm starting to see your point. Let me rephrase this in my own words to be sure I really understand. 

1. it is not clearly indicated that this example is using the `include_dir_merge_named` method (while the above example was using the `include_dir_named` method). It is easily overlooked by readers that this is a different method.
2. Because this example is listed under the _Create a packages folder_ title, people work with this example because they think this is the way to go if you want to work with subfolders.
3. The thing then doesn't work if they copy paste content from the config - because `include_dir_merge_named` requires a different indentation.
4. All in all, for new users it is easier to start with just one method, the  `include_dir_named` method, because it allows to copy paste from the config file.

ArveVM: Brilliant ! - why was I not able to write it like that in the first place  :)    

Thanks for spelling it out a bit more structured,
(I'm struggeling a bit, not native english nor techy)

So should we proceed with the simplification on just one sample,, or revert to #26320 and explain better the two separate options?
IMHO it is better to keep documentation and example small, and I don't see that the second example have a great audience/usability - so this is the better way? - - but I'm not the expert ;)

c0ffeeca7: could we continue this conversation in a week? I'm on the road and writing on the phone is not my favorite. I'm hesitant to merge because I'm not sure if deleting the example is the best way forward. Maybe we can do some editing.

ArveVM: sure :)
thanks for taking the time to review - and to clarify "my issue"  :)

ArveVM: > could we continue this conversation in a week? I

Got some time for this one now ??

source/_docs/configuration/packages.markdown
```diff
@@ -74,33 +74,16 @@ Integrations inside packages can only specify platform entries using configurati
...
 
 ```yaml
 homeassistant:
   packages: !include_dir_named packages
 ```
 
-This uses the concept splitting the configuration and will include all files in a directory with the keys representing the filenames.
-See the documentation about [splitting the configuration](/docs/configuration/splitting_configuration/) for more information about `!include_dir_named` and other include statements that might be helpful. The benefit of this approach is to pull all configurations required to integrate a system, into one file, rather than across several.
+The benefit of this approach is to pull all configurations required to integrate a system into one file&mdash;rather than keeping them spread across several files.
+You can use other !include-methods for packages folder, but with this one you get yaml-indentation 'as if the content was written in configuration.yaml'.
```
c0ffeeca7:
```suggestion
You can use other `!include` methods for packages; for example `!include_dir_merge_named`. However, unlike `!include_dir_merge_named`, the `!include_dir_named` method uses the same indentation as the configuration.yaml'. This means that you can copy and paste elements from the config file. With `!include_dir_named`, the file name is used as the package name. File names must be unique.
```

source/_docs/configuration/packages.markdown
```diff
@@ -74,33 +74,16 @@ Integrations inside packages can only specify platform entries using configurati
...
 
-The following example allows to have subfolders in the `packages` folder, which could make managing multiple packages easier by grouping:
```
c0ffeeca7, line 87:
```suggestion
Another approach is to organize packages using the `!include_dir_merge_named` method.
This method allows you to have subfolders in the `packages` folder, which could make managing multiple packages easier by grouping:
```
How about leaving the example in, but adding a bit more info before the example?

c0ffeeca7:
```suggestion
With the `!include_dir_merge_named` method, the package name has to be included in the file. The configuration below then needs to be indented accordingly - which means you cannot directly copy and paste from the configuration file.
```
And then, add this information after the example?

source/_docs/configuration/packages.markdown
```diff
@@ -74,33 +74,16 @@ Integrations inside packages can only specify platform entries using configurati
...
-and in `packages/subsystem1/functionality1.yaml`:
-
-```yaml
-subsystem1_functionality1:
-  input_boolean:
-  ...
-  binary_sensor:
-  ...
-  automation:
-```
```
c0ffeeca7, line 86:
```suggestion
With the `!include_dir_merge_named` method, the package name has to be included in the file. The configuration below then needs to be indented accordingly. This means you cannot directly copy and paste from the configuration file.
```
And then, after the example, add the information that you need to be careful when copy-pasting?

c0ffeeca7 (changes requested):
Yes. Thank you for sticking around. 
I've reworked the text a bit. I'm hoping to preserve the second example while pointing out the differences between the two methods. 
What do you think? Would this be a way to move forward?

c0ffeeca7 (approved):
Thank you, @ArveVM :+1:

PR #27180, opened 2023-04-27
Roborock: Remove false "cloud support" section

RubenKelevra:
## Proposed change

Both, Miio and Roborock, needs the cloud to configure the devices.

Remove section which claims otherwise.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/roborock.markdown
+++ b/source/_integrations/roborock.markdown
@@ -18,7 +18,7 @@ ha_integration_type: integration
 
 The Roborock integration allows you to control your [Roborock](https://us.roborock.com/pages/robot-vacuum-cleaner) vacuum while continuing to use the Roborock app.
 
-In contrast to [Xiaomi Miio](/integrations/xiaomi_miio/) integration, this integration provides more complete support but requires cloud access to set up and device configuration using the Roborock app.
+In contrast to [Xiaomi Miio](/integrations/xiaomi_miio/) integration, this integration provides more complete support using the Roborock app.
 
 Once you log in with your Roborock account, the integration will automatically discover your Roborock devices and get the needed information to communicate locally with them. Please ensure your Home Assistant instance can communicate with the local IP of your device. We recommend setting a static IP for your Roborock Vacuum to help prevent future issues.
 


RubenKelevra:
> It seems that this PR is targeted against an incorrect branch. Documentation updates which apply to our current stable release should target the `current` branch. Please change the target branch of this PR to `current` and rebase if needed. If this is documentation for a new feature, please add a link to that PR in your description.

That's wrong. `next` is the correct target branch.

rytilahti:
To my understanding, the differences are:
- xioami-miio requires initial setup using the mi home app and the device token extraction, e.g., using the cloud. The device communications are done locally.
- roborock requires initial setup using the roborock app. The device communications are routed through their cloud.

Is this correct?

RubenKelevra:
Nope.

Both require a that you pair the device with an app with a registration in the cloud.

Aftwards you need cloud access to register the device with HA. For xiaomi-miio that's done with a token, for Roborock it's via a pin per Mail.

Then both won't require the cloud to work, as they send commands directly to the device.

rytilahti:
Ah, my mistake on not understanding how it works then... Maybe reword the doc to be more clear that this provides a more complete support but that the setup needs to be done using the roborock app _instead_ of the mi one?

It would be great if there is  no confusion about the differences, especially if (or rather when?) registering a device using the roborock app breaks it for xiaomi-miio and the mi home app.

source/_integrations/roborock.markdown
```diff
@@ -18,7 +18,7 @@ ha_integration_type: integration
 
 The Roborock integration allows you to control your [Roborock](https://us.roborock.com/pages/robot-vacuum-cleaner) vacuum while continuing to use the Roborock app.
 
-In contrast to [Xiaomi Miio](/integrations/xiaomi_miio/) integration, this integration provides more complete support but requires cloud access to set up and device configuration using the Roborock app.
+In contrast to [Xiaomi Miio](/integrations/xiaomi_miio/) integration, this integration provides more complete support using the Roborock app.
```
rytilahti, line 21:
```suggestion
In contrast to [Xiaomi Miio](/integrations/xiaomi_miio/) integration, this integration provides a more complete support but requires that the device is configured using the Roborock app.
```
Maybe something like this? I'm not sure about the proper wording, see my other comment. If the device can still be controlled using the mi app even after it has been configured using the roborock one, this suggestion is irrelevant.

starkillerOG:
Actually the xiaomi Miio integration does not nessesarly require cloud acces.
Using the cloud credentials is by far the easiest way to set it up, but if you insist on not using the cloud, a user can supply the token directly during config flow withouth using the cloud.
The token can be obtained in other complex/advanced ways.

RubenKelevra:
@starkillerOG 

Well, you have to connect your robot via an app to your wifi first. The app and the robot needs Internet connection for this.

Connecting a robot to a wifi without either of these both apps is not possible.

In addition: In both cases the robot will keep a connection to the cloud at all times.

source/_integrations/roborock.markdown
RubenKelevra: But this Integration does not feature a more complete support. The contrary is the case: This integration is lacking some key features.

The point of the original text is, that using this integration allows to use the **Roborock app** and it is what allows more features than using the Xiaomi Home app.

frenck: Let's remove the whole line.

```suggestion
```

frenck (changes requested):
Let's remove the line. It is not worth discussing and above all, not true.

Please change the target of this PR to the `current` branch and git rebase if needed.

../Frenck

PR #27401, opened 2023-05-14
Add formatting note for Reusable Templates Formatting

SirGoodenough:
## Proposed change

This was pointed out to me on Discord, and I offered to extend the documentation to include the information.

QUOTE:
TheFes — 05/12/2023 1:05 PM
Important to know:

    The result of an imported macro is always a string
    If you don't add a hyphen to all your opening and closing tags (so use {%-, -%}, {{- and -}}) you're result will include unintended whitespace.

## Type of change

## Additional information

https://discord.com/channels/330944238910963714/672223497736421388/1106643448943095938

## Checklist

### DIFF
--- a/source/getting-started/index.markdown
+++ b/source/getting-started/index.markdown
@@ -7,7 +7,7 @@ show_title: true
 
 {% include getting-started/next_step.html step="Installation" link="/installation/" %}
 {% include getting-started/next_step.html step="Onboarding" link="/getting-started/onboarding/" %}
-{% include getting-started/next_step.html step="Onboarding" link="/getting-started/onboarding_dashboard/" %}
+{% include getting-started/next_step.html step="Edit the dashboard" link="/getting-started/onboarding_dashboard/" %}
 {% include getting-started/next_step.html step="Concepts and terminology" link="/getting-started/concepts-terminology/" %}
 {% include getting-started/next_step.html step="Integration" link="/getting-started/integration/" %}
 {% include getting-started/next_step.html step="Automation" link="/getting-started/automation/" %}


source/_docs/configuration/templating.markdown
```diff
@@ -98,6 +98,11 @@ In your automations, you could then reuse this macro by importing it:
 
 {% endraw %}
 
+NOTE: Two important formatting concerns to remember when reusing templates:
+
+1. The value returned is always a string.  If you want a number or a boolean you need to use ```|int()```, ```|float(),``` or ```|bool``` filters.
```
c0ffeeca7:
```suggestion
1. The value returned is always a string.  If you want a number or a boolean, you need to use ```|int()```, ```|float(),``` or, ```|bool``` filters.
```

source/_docs/configuration/templating.markdown
```diff
@@ -98,6 +98,11 @@ In your automations, you could then reuse this macro by importing it:
...
+1. If you don't add a hyphen to all your opening and closing tags (so use `{%-`, `-%}`, `{{-`, and `-}}`) you're result will include unintended whitespace.
```
c0ffeeca7:
```suggestion
1. If you don't add a hyphen to all your opening and closing tags (so use `{%-`, `-%}`, `{{-`, and `-}}`), your result will include unintended whitespace.
```

c0ffeeca7 (changes requested):

frenck:
Hoe if this different from any other template? 🤷‍♂️

source/_docs/configuration/templating.markdown
SirGoodenough: agree

source/_docs/configuration/templating.markdown
SirGoodenough: agree

frenck: This is not specific to re-using templates, but general information on the Jinja2 syntax that is put here. But putting it here, it seems like this is something specific to the marco's, which it is not.

../Frenck

SirGoodenough: It's specific in this use case because the casual operator may be expecting say a boolean returned or a number or a list, and that's not what they will be getting.  Only returns a string.  I used these words as that was what someone on the HA Discord helped me with.  So if not these exact words, some reminder that it returns a string would be very helpful to the prople that need the docs to use the feature.

PR #28116, opened 2023-07-07
Renames CO2Signal integration to Electricity Maps

madsnedergaard:
## Proposed change

The team at Electricity Maps is replacing the CO2Signal API that was built years ago, with an integration directly to our main API. With this change we can provide more scalable infrastructure, higher data quality, and add new features and improvements faster.

The migration will happen in several steps to ensure a smooth transition and **the existing URLs and API keys will continue working for current users.**

The only user-impacting change is that new users of the integration will have to go to a new website ([api-portal.electricitymaps.com](https://api-portal.electricitymaps.com/catalog/prd_8e3ood60rcv8f51h)) to get their api key.

We are keeping the domain as `co2signal` to avoid breaking changes.

## Type of change

_Since we also have changes in other repositories, I assume targeting `next` is correct?_

  - [x] I've opened up a PR to add logo's and icons in [Brands repository](https://github.com/home-assistant/brands).

## Additional information

Not sure if we need to keep the original `co2signal.markdown` file name here, or if we can/should replace it? 🤔 

The [PR in brands](https://github.com/home-assistant/brands/pull/4521) that adds the new logo should be merged first.

- Pull request in `core`: https://github.com/home-assistant/core/pull/96252
- Pull request in `brands`: https://github.com/home-assistant/brands/pull/4521

## Checklist


madsnedergaard:
It seems like the bot is saying that both `current` and `next` are wrong, so not sure what to do here 😅

c0ffeeca7:
> It seems like the bot is saying that both `current` and `next` are wrong, so not sure what to do here sweat_smile

`next` is correct, since it looks like you'll be adding changes to core as well.

janiversen (changes requested):
Why do you change all these integrations, that is not correct.

madsnedergaard:
Sorry for all the tagging, I changed the target branch back to `next` before doing a rebase (which should have been done first due to code owners! 😓)

**Edit:** Doesn't seem like I have permission to remove reviewers again either - so sorry people :/

source/_integrations/electricity_maps.markdown
```diff
@@ -0,0 +1,28 @@
+---
```
madsnedergaard: Should I rename this file back to `co2signal.markdown` and just change content, or is this okay with the redirect?

jbouwh: You can do both, so old links do not break. You can search for references to that can be replaced.

frenck: That is incorrect. The file need to match the domain of the integration.

madsnedergaard: Okay thanks, is there no way we can show a `/electricity_maps` URL instead of `/co2signal` without changing the domain? :)

frenck: The domain key will be around forever, unless you have a way to migrate it in core so it won't break any system.

For example, <https://www.home-assistant.io/integration/hassio>, which is the documentation for our own Supervisor application. `hassio` hasn't been used as a term since years anymore 🤷

madsnedergaard: Thanks for the clarification, I have renamed the file and removed the redirect 👍

balloob (approved):

PR #28122, opened 2023-07-07
Revert "Add Alarm Control Panel category for Alarm Control Panel, Alarm Control Panel Template and Manual Alarm"

frenck:
Reverts home-assistant/home-assistant.io#28112

I'm suggesting to revert this PR.

This PR adds more integrations to the "Alarm Control Panel" category, which makes no sense to me, as we already have an "Alarm" category. It adds no value and only adds more confusion.

If anything, we should try to get rid of the "Alarm Control Panel" category.

../Frenck

/CC @c0ffeeca7 @dougiteixeira

### DIFF
--- a/source/_integrations/alarm_control_panel.markdown
+++ b/source/_integrations/alarm_control_panel.markdown
@@ -3,7 +3,6 @@ title: Alarm Control Panel
 description: Instructions on how to integrate Alarm Control Panels into Home Assistant.
 ha_category:
   - Alarm
-  - Alarm Control Panel
 ha_release: 0.7.3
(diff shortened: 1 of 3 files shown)


c0ffeeca7 (approved):
Thank you, @frenck :pray:

dougiteixeira:
> If anything, we should try to get rid of the "Alarm Control Panel" category.

I agree with that… but if the “Alarme Control Panel” category exists, it's strange that they don't appear when filtered.

Can we remove the “Alarm Control Panel” category altogether then?

PR #28162, opened 2023-07-10
Update history-graph.markdown

ildar170975:
Currently `hours_to_show` may be like "0.0833", "1.5".

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_dashboards/history-graph.markdown
+++ b/source/_dashboards/history-graph.markdown
@@ -36,7 +36,7 @@ entities:
   type: list
 hours_to_show:
   required: false
-  description: Hours to show in graph. Minimum is 1 hour. Big values can result in delayed rendering, especially if the selected entities have a lot of state changes.
+  description: Hours to show in graph. Can have a fractional part. Big values can result in delayed rendering, especially if the selected entities have a lot of state changes.
   type: integer
   default: 24
 title:


ildar170975:
Issue for UI editor:
https://github.com/home-assistant/frontend/issues/15933

source/_dashboards/history-graph.markdown
```diff
@@ -36,7 +36,7 @@ entities:
   type: list
 hours_to_show:
   required: false
-  description: Hours to show in graph. Minimum is 1 hour. Big values can result in delayed rendering, especially if the selected entities have a lot of state changes.
+  description: Hours to show in graph. Can have a fractional part. Big values can result in delayed rendering, especially if the selected entities have a lot of state changes.
```
frenck, line 39: - I think this is a bug and not an actual feature (the default step size of a number select is `1`)
- This PR removes the mention of a minimal of 1 hour, which is defined as the minimum in the source code:

Ref: 
- <https://github.com/home-assistant/frontend/blob/bffdfcf61c4833293e9782aef4bdcc7c7921152a/src/panels/lovelace/editor/config-elements/hui-history-graph-card-editor.ts#L46>
- <https://github.com/home-assistant/frontend/blob/bffdfcf61c4833293e9782aef4bdcc7c7921152a/src/components/ha-selector/ha-selector-number.ts#L45>

frenck (changes requested):

source/_dashboards/history-graph.markdown
ildar170975: > which is defined as the minimum in the source code:

Exactly, and I think that UI should be adapted (https://github.com/home-assistant/frontend/issues/15933).

ildar170975: Currently we may use `hours_to_show` with fractional part, and this is great 'cause we can define small intervals <1hr.
Also, this fractional `hours_to_show` allows to create graphs "from 00:00 till NOW" with a help of template cards like `config-template-card`.
So, this fractional values are GREAT.

frenck: > Exactly, and I think that UI should be adapted.

That is an opinion and not what the intention of the code currently is, making the change here in the documentation incorrect.

> Currently we may use hours_to_show with fractional part, and this is great 'cause we can define small intervals

Same, that seems to be an unintended side-effect. It is cool that it works, but that doesn't mean we should document unintended workings.

See also the references pointed out above in our code base.

ildar170975: > That is an opinion

Correct, it becomes an "opinion" if you are sure that `hours_to_show` MUST BE integer and the current behaviour "value may be float" is a bug. I guess many users will be disappointed if `hours_to_show` again becomes integer.

ildar170975: > See also the references pointed out above in our code base.

I do not see a reason why UI (selector is integer and starts from 1) is more important than the current implementation of the graph card.

frenck: > I do not see a reason why UI (selector is integer and starts from 1) is more important than the current implementation of the graph card.

Because you are trying to document something different to compared what the intention was. For that reason, you could be documenting unintended functionality or even a bug; which we should not (especially the latter) document but fix first.

../Frenck

ildar170975: > but fix first.

So, for clarity: are you going to disable a possibility to use a float `hours_to_show` (a feature which was added several month ago - or call it a bug) and make it again integer?

frenck: > are you going to disable a possibility to use a float hours_to_show (a feature which was added several month ago - or call it a bug) and make it again integer?

I didn't say that, I said there is unclarity compared to what the intention is looking at the code base. This means the codebase currently doesn't reflect the actual function/idea you are trying to document. This can either mean unintentional behavior or a bug.

> a feature which was added several month ago

Could you point me to it?

../Frenck

ildar170975: > Could you point me to it?

I cannot ((((.
Have no idea after which particular PR it became possible to use float `hours_to_show`.
It happened a few months a ago; I compare `custom:mini-graph-card` and `history-graph` cards rather often, and was happy when realized that `history-graph` allows using float values.
As a result, I believed that float values are the CURRENT implementation - then proposed corr. changes in UI & Docs.

frenck: Looking at the file history, I cannot find evidence of it specifically being made available:

<https://github.com/home-assistant/frontend/commits/dev/src/panels/lovelace/editor/config-elements/hui-history-graph-card-editor.ts>

The thing I did find, though, is this PR from 2021, which specifically (and intentionally) set it to a minimum of 1: <https://github.com/home-assistant/frontend/pull/9464>

ildar170975: Sorry, cannot help you... Here I am mainly a user/tester, not a developer.
But I am sure that some time ago (at least at the end of 2020 when I started using HA) period was minimum 1hour.

ildar170975: My test setup 2022.7.4 - value "0.08333" treated as "1":

image

ildar170975: Same for 2022.12.8:
image

I noticed the change on March 2023.

ildar170975: > Looking at the file history, I cannot find evidence of it specifically being made available:

Could it be a part of "streamline history" changes?

frenck: I've checked this with the frontend team, and this is an unintended change. Meaning it should be reported as a bug or fixed and not documented as a feature.

../Frenck

frenck:
Closing the PR as it documents an unintended functionality (which is a bug that should be fixed).

../Frenck

source/_dashboards/history-graph.markdown
ildar170975: Will users be able to define fractional `hours_to_show` ?

frenck: > Will users be able to define fractional hours_to_show ?

They should not. That was the point and the issue here.

ildar170975: Is there any harm for HA when a user defines `hours_to_show: 0.5`?
There are plenty of usecases when similar fractional values may be useful.
What is a meaning of reducing a functionality?

frenck: > Is there any harm for HA when a user defines hours_to_show: 0.5?

Yes, as not the the full fraction is supported. As said, this was not intended functionality and thus a bug.

> There are plenty of usecases when similar fractional values may be useful.

🤷 Not sure what that has to do with documenting.

> What is a meaning of reducing a functionality?

I think you missed the part where this was unintended and thus a bug at this point.

../Frenck

ildar170975: > I think you missed the part where this was unintended

Sometimes unintended actions create unexpected & positive outcome.
Too sad that you want to delete this functionality instead of rectifying possible negative consequences.

> Not sure what that has to do with documenting.

Now this is not about documenting, this is about a deleting a useful functionality.

frenck: > Sometimes unintended actions create unexpected & positive outcome.

They can also cause issues.

> Too sad that you want to delete this functionality instead of rectifying possible negative consequences.

That comment makes no sense whatsoever. A user is now able to configure it in an unusable and broken way. In no world, that is a positive outcome.

../Frenck

ildar170975: If a useful feature (even unexpected) can cause issues - then this feature should be improved, re-designed etc - to provide a similar functionality for users. Not deleted.

frenck: Don't get me wrong, we are open for contributions that improve the situation. Feel free to join in on the development in the frontend repository.

../Frenck

ildar170975: Well, I am trying as I can.
I just wonder what negative consequences can have fractional `hours_to_show`. Did you notice any registered issue like "smth works wrong due to fractional hours_to_show"?
OK, no need to answer, I quit.

frenck: I'm just relaying a message on this matter, that it isn't correctly to document it like this and it will cause issues. Did they tell me they rather see a refactor of the `hours_to_show` in the future (remove it all together and add a generic `duration` as we use in multiple places).

ildar170975: Then for me this is a different story then. More hopeful. Thanks for explaining, really hopeful.
OK, there is no need to document an unintended change - especially if this whole `hours_to_show` attribute will be replaced by a duration (which hopefully will include minutes).

PR #28340, opened 2023-07-26
Update default_config.markdown

ertechdesign:
## Proposed change
    Example of configuration allowing a user to load desired integrations only.

## Type of change

## Additional information
  Users might require help how to replace default_config with custom configuration

## Checklist


source/_integrations/default_config.markdown
```diff
@@ -64,3 +64,51 @@ To integrate this into Home Assistant, add the following section to your `config
 # Example configuration.yaml entry
 default_config:
 ```
+
+If you wish to load specific integrations only, you can list them instead.
```
frenck, line 68: This is documented in each integration, I don't think we need to maintain this here.

frenck (changes requested):

source/_integrations/default_config.markdown
ertechdesign: This PR was submitted as a result of the struggle of multiple users in [Community - Let Home assistant show “home” or “away” not precise location](https://community.home-assistant.io/t/let-home-assistant-show-home-or-away-not-precise-location/229867/7)

Petro31: I think the point here is that people want something quickly to copy/past instead of copying each integration and adding a colon.  I would suggest we change the list to yaml, or provide an example of them all listed out.  It's cumbersome copying/pasting each integration and adding a colon.

frenck: I think one should simply not remove default config.
Additionally, we should not maintain the list multiple times / multiple places.

If one want to take the advanced route and remove things like these, they have to know what they are doing.

I really don't understand what this has to do with the linked community topic.

ertechdesign: @frenck I understand your concerns but I'm just a user myself. I **must** remove the map as it tracks my lodgers in the same household and home/away is required to operate heating. Showing their location is a serious issue.
@Petro31 informed multiple users trying to do the same that the best route is to remove the map. I wasn't sure if I can just list the desired integrations with colons or also need to add some minimal configuration thereof.

I don't think users should be left guessing a valid configuration. Isn't this exactly why a documentation is published?

frenck: > Showing their location is a serious issue.

Wait, if that is a serious issue, you still have a serious issue; as disabling maps will not solve that. The location is still tracked and can be queried/viewed by anyone at any time that has access to HA (regardless if you disable the map or not).

If that motivates this change for you, I'm sorry but that is a false sense of security you've gotten there.

Petro31: @frenck this topic comes up multiples times per month on the forums and discord.  It has been in both months of WTH.  The least we can do is make it a simple copy/paste.  You don't have to answer these questions constantly but many other people do.  The least we can do is make it somewhat easier to maintain.  Purposely making it hard does not deter people from removing it, it just makes it take longer.  Wasting peoples time because of an 'ideal' is not really a nice thing to do.  Especially when it really doesn't matter in the grand scheme of things.  It's a documents page, not extra code.

Maintaining the list twice is a good reason to have it autogenerated using some jinja on this page.

frenck: > It has been in both months of WTH.

No, what has been in WTH was the ability to exclude some of those. You are mixing things up now.

> You don't have to answer these questions constantly but many other people do

I do, and tell them to use the default config.

> urposely making it hard does not deter people from removing it, it just makes it take longer.

They should not remove it.

> Maintaining the list twice is a good reason to have it autogenerated using some jinja on this page.

Sorry, that sentence make no sense to me whatsoever, this page is not powered by Jinja, nr can it access this information.


That said, this PR is driven by something that actually create a wrong perception of what this does and will do. Which is a great example in the response above by the author of the PR.

Not just that, there are still dependencies from other integration and subdependencies that will pull in. Regardless if you use this or not. Stimulating the removal will make that issue bigger.

ertechdesign: That is true, however, I didn't find a working solution to stop showing the location of users while keeping home/away status.

I tried methods described in [Location](https://companion.home-assistant.io/docs/core/location/) but it didn't quite worked.

Removing the map from the frontend would be at least a partial solution. In the past I didn't need my lodgers' location as I tracked their presence by the Tomato integration but this one is broken for many users and nobody cares to look at it:
[issue 8345](https://github.com/home-assistant/core/issues/84315)

I still need to operate my heating though. I invested a lot of time and money into Home Assistant and don't want to switch to different solution (or buy a different router) only because of my issue.

ertechdesign: > They should not remove it.

Users were told in another forum there won't be any feature allowing Home Assistant to track users location and completely hide it from the front-end. The backend tracking the location and only showing Home/Away on Front End would be acceptable for most users concerned about their privacy. Personally I'd love a feature like this.

If devs don't want to introduce features users need they shouldn't be surprised some users will go long lengths to make the app working for them.

frenck: @ertechdesign Every user that can access your HA instance... Including your lodgers, will have full access to HA (like full admin access).

Regardless of your other motivation in this PR or anything else you do with default config...

Exactly this you are now convinced about, is not what you thought it would do. How is that a good motivator for this change?

> If devs don't want to introduce features users need they shouldn't be surprised some users will go long lengths to make the app working for them.

Our users are the devs...

Petro31: > No, what has been in WTH was the ability to exclude some of those. You are mixing things up now.

Yes, but you aren't understanding the pain point:  People don't want default configuration because they want to remove things.  E.g. Auto detecting things via DHCP Discovery, Bluetooth, or zero config.

> Sorry, that sentence make no sense to me whatsoever, this page is not powered by Jinja, nr can it access this information.

I was under the impression that we can use jinja to generate repeated information.  What's the point of using {% raw %} in all documentation if that's not the case?

> Not just that, there are still dependencies from other integration and subdependencies that will pull in. Regardless if you use this or not. Stimulating the removal will make that issue bigger.

Yes I understand that, but it still doesn't mean we can't supply a simple segment of YAML to easily copy/paste.  Sometimes it's nice to help out the little guy even if it goes against your coding morals.

If you'd like, I can CC you in every time this question arises so that you see the true impact of these poor decisions.  I 100% agree that we should not have an exclude in the yaml, but I also think we should make it easier for people to copy/paste.

Petro31: I can tell you that this question comes up a lot.  I can also tell you I've never heard anyone ask "I've removed XYZ from default config but it's still loading".  Not to mention, the answer would be "It's used by something else you're using so it's required".  No one will be upset with that response.

frenck: > If you'd like, I can CC you in every time this question arises so that you see the true impact of these poor decisions. 

Not sure why that comment is needed, feels more like a threat to be honest which I think stops this conversion right here.

There is a serious concern in the motivation of this PR which is a broader problem with this.

frenck: > "I've removed XYZ from default config but it's still loading"

It happens quite a lot and causes issues to be filed as well. E.g., esphome that loads the voice assistants come up quite a lot now.

frenck: Let's put it different, if we add this, we need to be clear on the consequence and explain that; additionally explain some things may still load and that this is not a security feature.

Putting that in an advanced detail box.

Additionally, this list in this PR is not complete/cannot be used copied/pasted, as some are loaded dynamically, which is environment dependent. So that needs to be handled as well.

ertechdesign: > Every user that can access your HA instance... Including your lodgers, will have full access to HA (like full admin access).

Perhaps I'm wrong but users that are not administrators don't seem to have access to the Location of user devices other than the Map integration.

I honestly think there should be a reasonable way for an admin to prevent Home Assistant leaking exact location of each user (without having to ask each user to have their phone handed over and try changing settings that are different on iOS and Android).

And if such feature is not accessible please don't blame users for removing the Map.

frenck: > Perhaps I'm wrong but users that are not administrators don't seem to have access to the Location of user devices other than the Map integration.

You are wrong, they could even add the map back. The map integration just display existing data.

Also note our user account documentation warns you every user has full access.

> I honestly think there should be a reasonable way for an admin to prevent Home Assistant leaking exact location of each use

That is a feature request, not something that is solved in this PR. If you think that was the case, then I have to disappoint you.

> And if such feature is not accessible please don't blame users for removing the Map.

☝️ This is my concern. That is simply not solving your issue. Your lodgers can still get that information without the map. Heck, then can just add the map back if can access your system. Your motivation for this PR is concerning as it is not solving your case.. yet you think it does.

Petro31: > Not sure why that comment is needed, feels more like a threat to be honest which I think stops this conversion right here.

I'm only saying it because it doesn't impact you.  But it impacts everyone else who helps.  So your decision here is basically saying "we don't respect your time because of a belief... but we still want you to support these people with the same questions every month/week".  It shouldn't stop the conversation, we should be able to reach a compromise, which it seems you're unwilling to do.  At this point, only you are the barrier, which is why I said I can CC you.  Sorry if that offends you but this is a real problem and you aren't taking it seriously and just dismissing it.  IMO ending the conversation is just a cop-out to avoid a solution.

ertechdesign: > That is a feature request, not something that is solved in this PR. If you think that was the case, then I have to disappoint you.

I believe this feature request was repeatedly raised and rejected. Plus I wouldn't have been wasting my time if somebody finally did something with [84315](https://github.com/home-assistant/core/issues/84315)

I can only say it was first time and last time I tried to suggest anything at all regarding Home Assistant. 🤮

frenck: >  Sorry if that offends you but this is a real problem and you aren't taking it seriously and just dismissing it.

I'm not just dismissing it, I'm even providing a [counter suggestion above](https://github.com/home-assistant/home-assistant.io/pull/28340#discussion_r1275246895). That said, I do ask for motivation in PRs; and the motivation in this case is plain dead scary wrong :)

As the person in this case got a false sense of security. That is really problematic, concerning and super serious IMHO.

ertechdesign: @frenck please stop patronising me.

You state in your profile people should think about the environment. That's exactly what I'm doing by geofencing my heating (to conserve resources).

You overcomplicate things. I need to remove the location from the frontend as a temporary solution because everything else failed me. Nobody in my household will reverse engineer the app to spy on roommates (people have better things to do). I just don't want to see location of my lodgers on the front end and they want the same.

This is only a temporary solution until [84315](https://github.com/home-assistant/core/issues/84315) will get fixed or until somebody starts thinking about users privacy in Home Assistant so don't blame this on me OK?

What's problematic, concerning and super serious IMHO is people trying to help on their time are being talked down.

frenck: >  I just don't want to see location of my lodgers on the front end and they want the same.

You still will see them. Just click on the person or device tracker entity 🤷 That is not a hack, it is on every dashboard, including search.

Anyways, regardless, I'm fine with adding changes in this PR, under the conditions put out above to set expectations right for future users.

ertechdesign: > You still will see them. Just click on the person or device tracker entity 🤷 That is not a hack, it is on every dashboard, including search.

I mentioned at least 5 times this is a temporary solution until [84315](https://github.com/home-assistant/core/issues/84315) is fixed. The problem here is I don't want to have front-end with map showing people's location in the UI so I want to disable so at least we don't have it as a feature. Yeah in theory every user could search for other people entity, but nobody cares. We just don't want to see each other location directly, it's creepy.

> Anyways, regardless, I'm find with adding thing like this under the conditions put out above to set expectations right for future users.

Honestly, at this point I couldn't care less. Make edits to my PR, accept it, reject it. I'm done with ever trying to collaborate on anything regarding Home Assistant with anybody.

frenck: > Make edits to my PR, accept it, reject it. I'm done with ever trying to collaborate on anything regarding Home Assistant with anybody.

Sorry to hear that, if you are not planning on continuing working on this PR, then we only can close it I guess.

I'll make a note for myself to look at a possible PR later, right now I'm working on the beta so that won't be today.

../Frenck

Petro31: > Let's put it different, if we add this, we need to be clear on the consequence and explain that; additionally explain some things may still load and that this is not a security feature.
> 
> Putting that in an advanced detail box.
> 
> Additionally, this list in this PR is not complete/cannot be used copied/pasted, as some are loaded dynamically, which is environment dependent. So that needs to be handled as well.

I missed this post.  All valid points.  Is this information available somewhere or is it just a "known thing to happen"?

frenck: No problem. Will look into improving it tomorrow or Friday. Is it OK if I mention you for a review?

Petro31: > No problem. Will look into improving it tomorrow or Friday. Is it OK if I mention you for a review?

Yeah, sure. I can help. I can make the PR too if needed.  I'm not a strong writer though.

PR #28351, opened 2023-07-27
Adding note warning about ding not working

apekatten:
Adding a warning that the integration has been broken since february
https://github.com/home-assistant/core/issues/88705

### DIFF
--- a/source/_integrations/ring.markdown
+++ b/source/_integrations/ring.markdown
@@ -24,7 +24,7 @@ ha_integration_type: integration
 ---
 
 <p class='note'>
-This integration have a bug that makes it impossible to make automations based on visitors presing the doorbell. [Issue #88705](https://github.com/home-assistant/core/issues/88705)
+This integration currently has a bug that makes it impossible to create automations based on visitors pressing the doorbell. [Issue #88705](https://github.com/home-assistant/core/issues/88705)
 </p>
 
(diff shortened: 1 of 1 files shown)


source/_integrations/ring.markdown
```diff
@@ -23,6 +23,10 @@ ha_platforms:
 ha_integration_type: integration
 ---
 
+<p class='note'>
+This integration have a bug that makes it impossible to make automations based on visitors presing the doorbell. [Issue #88705](https://github.com/home-assistant/core/issues/88705)
```
c0ffeeca7:
```suggestion
This integration currently has a bug that makes it impossible to create automations based on visitors pressing the doorbell. [Issue #88705](https://github.com/home-assistant/core/issues/88705)
```

c0ffeeca7 (commented):
tiny tweak

c0ffeeca7 (approved):
Thank you, @apekatten :+1:

source/_integrations/ring.markdown
```diff
@@ -23,6 +23,10 @@ ha_platforms:
 ha_integration_type: integration
 ---
 
+<p class='note'>
+This integration currently has a bug that makes it impossible to create automations based on visitors pressing the doorbell. [Issue #88705](https://github.com/home-assistant/core/issues/88705)
+</p>
```
frenck, line 28: We don't document issues, please use and track the issue tracker for that.

Reason is that we have 2K open issue right now, if we would document them all (and have to remove them each time one is solved), it would become unmaintainable.

../Frenck

PR #28580, opened 2023-08-14
Update Apprise example to include name

mjforan:
## Proposed change
Minor update to documentation, clarifies use of Apprise notification service
- add name, as defined in https://github.com/home-assistant/core/blob/6be47b1fbde48d810cc1d5cfd9824f41513805fd/homeassistant/components/notify/__init__.py#L37

## Type of change

## Additional information
The Apprise integration is meant to be called as a service. To register this service, the configuration needs to have the "name" parameter configured. Copying the examples as-is will not currently work because they are calling the `notify.apprise` service, which was not registered because the "name" parameter was not used. I added this parameter so the example will work.

## Checklist

### DIFF
--- a/source/_integrations/apprise.markdown
+++ b/source/_integrations/apprise.markdown
@@ -22,7 +22,8 @@ To use Apprise supported notifications, add the following to your `configuration
 ```yaml
 # Example configuration.yaml entry using URLs
 notify:
-  - platform: apprise
+  - name: apprise
+    platform: apprise
     url: YOUR_APPRISE_URLS
 ```
 
@@ -32,7 +33,8 @@ You can also pre-define your own configuration files while storing them either r
 # Example configuration.yaml entry using externally located Apprise
 # Configuration Files/Sites:
 notify:
-  - platform: apprise
+  - name: apprise
+    platform: apprise
     config: YOUR_APPRISE_CONFIG_URLS
 ```
 
@@ -41,7 +43,8 @@ There is no restriction on the number of URLs or Apprise Configuration locations
 ```yaml
 # Example configuration.yaml entry using all options
 notify:
-  - platform: apprise
+  - name: apprise
+    platform: apprise
     config: YOUR_APPRISE_CONFIG_URLS
     url: YOUR_APPRISE_URLS
 ```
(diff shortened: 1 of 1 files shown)


source/_integrations/apprise.markdown
```diff
@@ -22,7 +22,8 @@ To use Apprise supported notifications, add the following to your `configuration
 ```yaml
 # Example configuration.yaml entry using URLs
 notify:
-  - platform: apprise
+  - name: apprise
```
c0ffeeca7:
```suggestion
  - name: NOTIFIER_NAME
```
I think this would make it clearer that the name can be user-defined.

source/_integrations/apprise.markdown
```diff
@@ -32,7 +33,8 @@ You can also pre-define your own configuration files while storing them either r
 # Example configuration.yaml entry using externally located Apprise
 # Configuration Files/Sites:
 notify:
-  - platform: apprise
+  - name: apprise
```
c0ffeeca7:
```suggestion
  - name: NOTIFIER_NAME
```

source/_integrations/apprise.markdown
```diff
@@ -41,7 +43,8 @@ There is no restriction on the number of URLs or Apprise Configuration locations
 ```yaml
 # Example configuration.yaml entry using all options
 notify:
-  - platform: apprise
+  - name: apprise
```
c0ffeeca7:
```suggestion
  - name: NOTIFIER_NAME
```

c0ffeeca7 (changes requested):
Tiny tweak. I think this way of writing it would make it clear that the name can be defined by the user.

mjforan (commented):
Good idea. I made a commit with these changes and also modified the name in the service call.

mjforan:
One thing I wasn't sure of is if the name field should be listed as "required". Technically HA will accept a config without the name, but then the service won't be accessible by the user.

mjforan (commented):
Addressed in separate commit

c0ffeeca7 (approved):
Thank you, @mjforan :+1:

PR #28882, opened 2023-09-12
Add initial property to imap email_content event

jbouwh:
## Proposed change

Add initial property to imap email_content event

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/100171

## Checklist

### DIFF
--- a/source/_integrations/imap.markdown
+++ b/source/_integrations/imap.markdown
@@ -86,7 +86,7 @@ Email providers may limit the number of reported emails. The number may be less 
 
 ### Using events
 
-When a new message arrives that meets the search criteria the `imap` integration will send a custom [event](/docs/automation/trigger/#event-trigger) that can be used to trigger an automation.
+When a new message arrives or a message is removed within the defined search command scope the `imap` integration will send a custom [event](/docs/automation/trigger/#event-trigger) that can be used to trigger an automation.
 It is also possible to use to create a template [`binary_sensor` or `sensor`](/integrations/template/#trigger-based-template-binary-sensors-buttons-numbers-selects-and-sensors) based the [event data](/docs/automation/templating/#event).
 
 The table below shows what attributes come with `trigger.event.data`. The data is a dictionary that has the keys that are shown below.
--- a/source/_integrations/imap.markdown
+++ b/source/_integrations/imap.markdown
@@ -86,7 +86,7 @@ Email providers may limit the number of reported emails. The number may be less 
 
 ### Using events
 
-When a new message arrives or a message is removed within the defined search command scope the `imap` integration will send a custom [event](/docs/automation/trigger/#event-trigger) that can be used to trigger an automation.
+When a new message arrives or a message is removed within the defined search command scope, the `imap` integration will send a custom [event](/docs/automation/trigger/#event-trigger) that can be used to trigger an automation.
 It is also possible to use to create a template [`binary_sensor` or `sensor`](/integrations/template/#trigger-based-template-binary-sensors-buttons-numbers-selects-and-sensors) based the [event data](/docs/automation/templating/#event).
 
 The table below shows what attributes come with `trigger.event.data`. The data is a dictionary that has the keys that are shown below.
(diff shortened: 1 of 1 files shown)


source/_integrations/imap.markdown
```diff
@@ -86,7 +86,7 @@ Email providers may limit the number of reported emails. The number may be less
 
 ### Using events
 
-When a new message arrives that meets the search criteria the `imap` integration will send a custom [event](/docs/automation/trigger/#event-trigger) that can be used to trigger an automation.
+When a new message arrives or a message is removed within the defined search command scope the `imap` integration will send a custom [event](/docs/automation/trigger/#event-trigger) that can be used to trigger an automation.
```
c0ffeeca7:
```suggestion
When a new message arrives or a message is removed within the defined search command scope, the `imap` integration will send a custom [event](/docs/automation/trigger/#event-trigger) that can be used to trigger an automation.
```

source/_integrations/imap.markdown
```diff
@@ -120,6 +120,8 @@ headers:
   description: The `headers` of the message in the for of a dictionary. The values are iterable as headers can occur more than once.
 custom:
   description: Holds the result of the custom event data [template](/docs/configuration/templating). All attributes are available as a variable in the template.
+initial:
+  description: Returns `True` if this is the initial event for the last message received. When a message within the search scope is removed and the last message has not been changed then an `imap_content` event is generated and `initial` property is set to `False`. Note that if no `Message-ID` header was set on the triggering email the `initial` property will always be set to `True`.
```
c0ffeeca7:
```suggestion
  description: Returns `True` if this is the initial event for the last message received. When a message within the search scope is removed and the last message received has not been changed, then an `imap_content` event is generated and the `initial` property is set to `False`. Note that if no `Message-ID` header was set on the triggering email, the `initial` property will always be set to `True`.
```
For my own understanding. What do you mean by "_last message_"? Are we not talking about one specific message?
I read the explanation in 100171, but I do not understand it. Particularly this bit. "_If a message within the search context is removed the message UID of the last message received might change_."
The UID of the message that was removed is being changed? Or the UID of whatever message has the most recent timestamp? What is the point of an UID if it is being changed? :confused:

jbouwh: The last message is the last message received within the search scope. A UID of a message is a unique identifier of the message. But now it seems it can change if e.g. messages are removed from a mailbox. Lets say the search scope is all unread mails in the INBOX. When an unread email is removed (not the last received) the last received message gets an new UID. When this happens, an event is triggered though. The content in the event is about the last email, if it was not changed `initial` will be set to `False`.

c0ffeeca7: Thank you for explaining this. So in other words, if 'initial' is set to 'true', provided we have an ID, this means:
1. Out of all the messages within the search context, this is the message with the latest time stamp.
2. Only one message in the search scope can have 'initial' set to 'true', because the moment something changes within that scope, it is set to false.
If I understand this correctly, 'initial' means _the latest message within this specific set_?

c0ffeeca7 (approved):
Thank you, @jbouwh :+1:

source/_integrations/imap.markdown
jbouwh: That is correct. Another example: If the last message changes because it is marked as `read`. Then the content of the last message changes because it is no longer in scope of the search. The new message is marked as initial. When the last received message is marked unread again, and it comes back in the search scope, it will again be marked as initial because it was nit the previous last message.

c0ffeeca7: complex...so, if in the same client I have multiple folders representing different filter settings, this one message can have a different state of the "initial", all at the same time?
I realize this is outside the scope of this PR. The reason I'm asking these questions is because I'm wondering whether something like _latest message within this filter set_ would be a more informative name for this attribute than *initial*.

jbouwh: No, one config entry has one search on one folder. The search results depend on the search command. If it filters out `read `messages` then it can be items are removed from the scope if they are read, and they return in the list if they are marked unread. The last message though is the last received message that matches the search scope. If the last message it self is not changing, then `initial` is set to `False`.

jbouwh: You can monitor multiple folders, or use multiple searches, but you need additional config entries for that.

c0ffeeca7: Thank you!

jbouwh: yw

PR #29092, opened 2023-10-01
Adjust documentation for knx

elupus:
## Proposed change

Adjust documentation on KNX covers.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/knx.markdown
+++ b/source/_integrations/knx.markdown
@@ -823,6 +823,8 @@ The KNX cover platform is used as an interface to KNX covers.
 
 Unlike most KNX devices, Home Assistant defines 0% as closed and 100% as fully open in regards to covers. The corresponding value inversion is done internally by the KNX integration.
 
+Home Assistant will, by default, `close` a cover by moving it in the `DOWN` direction in the KNX nomenclature, and `open` a cover by moving it in the `UP` direction.
+
 </div>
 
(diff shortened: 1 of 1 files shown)


source/_integrations/knx.markdown
```diff
@@ -852,7 +854,7 @@ move_long_address:
   required: false
   type: [string, list]
 move_short_address:
-  description: KNX group address for moving the cover stepwise up or down. Used by some covers also as the means to stop the cover, if no dedicated `stop_address` exists on the actuator. *DPT 1*
+  description: KNX group address for moving the cover stepwise up or down. Used by some covers also as the means to stop the cover. Note. Stepwise moves are only mapped to tilt angle functions in Home Assistant, as no stepwise move of cover position is generally supported by the architecture. If tilt angle is not supported, prefer the use of a `stop_address`. *DPT 1*
```
farmio: We don't even use it to set tilt if there is a `angle_address` configured. So in most installations it is actually only used to as "stop". 

I think there should be a `:` instead of a `.` in "Note. "

elupus: Yup, but there:in lies the grub. Ie by adding the addresses, we enable tilt features which may not exist in the hardware. So I like to push people to the stop address. Maybe we should have some other option, or make use of the device class.

farmio (dismissed):
Thank you! 👍

farmio (approved):

PR #29472, opened 2023-10-22
Update arris_tg2492lg.markdown

ArbenP:
Add ireland as virgin media is also in ireland with this router


frenck (approved):
Thanks, @ArbenP 👍 

../Frenck

PR #29497, opened 2023-10-24
Update matter.markdown

rtvanhook:
Add Leviton

## Checklist

### DIFF
--- a/source/_integrations/matter.markdown
+++ b/source/_integrations/matter.markdown
@@ -196,6 +196,17 @@ Did you test a device that is not listed below? It would be greatly appreciated 
 [Eve Door & Window on Amazon](https://amzn.to/3RIU6ml)
 [Eve Motion on Amazon](https://amzn.to/3jDujiP)
 
+### Leviton Decora Smart Switches (D215S)
+
+- You will need a My Leviton account and app, and must first connect the switch to the app through the cloud.
+- [Upgrade to Matter-supported firmware](https://decorasmartsupport.leviton.com/hc/en-us/articles/11638969410331)
+- If you want to break Leviton's connection to the switch (and the hole in your firewall), factory reset the swtich before pairing with Matter.
+- Ensure IPv6 is enabled on your HA host, and your phone and HA host are on the same network/VLAN as the switch.
+- The Matter pairing code for the switch is found in the My Leviton app.  There's also a QR code, but it's worthless with a single phone.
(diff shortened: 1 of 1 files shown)


source/_integrations/matter.markdown
```diff
@@ -196,6 +196,17 @@ Did you test a device that is not listed below? It would be greatly appreciated
...
+### Leviton Decora Smart Switches (D215S)
+
+- You will need a My Leviton account and app, and must first connect the switch to the app through the cloud.
+- [Upgrade to Matter-supported firmware](https://decorasmartsupport.leviton.com/hc/en-us/articles/11638969410331)
+- If you want to break Leviton's connection to the switch (and the hole in your firewall), factory reset the swtich before pairing with Matter.
+- Ensure IPv6 is enabled on your HA host, and your phone and HA host are on the same network/VLAN as the switch.
+- The Matter pairing code for the switch is found in the My Leviton app.  There's also a QR code, but it's worthless with a single phone.
+- Place the switch in pairing mode by holding the top for about 7 seconds until the LED turns amber.
+- Add the device through "Add Matter Device" with the pairing code.
+- Doesn't always work on the first try.
```
frenck, line 208:
```suggestion
### Leviton Decora Smart Switches (D215S)

- You will need a My Leviton account and app and must first connect the switch to the app through the cloud.
- [Upgrade to Matter-supported firmware](https://decorasmartsupport.leviton.com/hc/en-us/articles/11638969410331)
- If you want to break Leviton's connection to the switch (and the hole in your firewall), factory reset the switch before pairing it using Matter.
- Ensure IPv6 is enabled on your Home Assistant host and your phone and Home Assistant host are on the same network/VLAN as the switch.
- The Matter pairing code for the switch is found in the My Leviton app. There's also a QR code, which isn't very helpful with a single phone.
- Place the switch in pairing mode by holding the top for about 7 seconds until the LED turns amber.
- Add the device through "Add Matter Device" with the pairing code. It doesn't always work on the first try.
```

frenck (changes requested):

marcelveldt:
Sorry, but we're about to remove all device specific instructions from the Matter integration docs as there is no more reason to specifically list it:

- Matter devices have become (somewhat) widely available
- The list is no longer complete, nor will we get it complete
- It is no longer needed to join beta programs and such.

So, this PR is obsolete, sorry.

PR #29504, opened 2023-10-24
Add additional information about Shelly.click events

madAndroid:
## Proposed change

Several users have been caught out by `shelly.click` events not working as they expect for relay devices like Shelly 1/1pm/2.5 etc - the integration does not fire an event when a switch attached to the relay in `detached` mode is turned from `on` to `off`, as explained in this forum post [here](https://community.home-assistant.io/t/shelly1-configured-as-detached-edge-switch/322213)

## Type of change

Adds additional context for Shelly relay devices, with a link to the forum article, to help other users that might assume that using the `shelly.click` event would work for them as they expect.

## Additional information

- Isssue in core: https://github.com/home-assistant/core/issues/102583

## Checklist

### DIFF
--- a/source/_integrations/shelly.markdown
+++ b/source/_integrations/shelly.markdown
@@ -125,6 +125,7 @@ For generation 2 hardware it's possible to select if a device's input is connect
 ## Event entities (generation 1)
 
 If the **BUTTON TYPE** of the switch connected to the device is set to `momentary` or `detached switch`, the integration creates an event entity for this switch. You can use this entity in your automations.
+(Important to note: for relay devices, this is not the same as the binary sensor - it doesn't track the state of a relay, it only fires events when an detached switch is turned from `off` to `on`, and doesn't fire when a detached switch is turned from `on` to `off`)
 
 ## Event entities (generation 2)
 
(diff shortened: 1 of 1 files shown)


source/_integrations/shelly.markdown
```diff
@@ -125,6 +125,7 @@ For generation 2 hardware it's possible to select if a device's input is connect
 ## Event entities (generation 1)
 
 If the **BUTTON TYPE** of the switch connected to the device is set to `momentary` or `detached switch`, the integration creates an event entity for this switch. You can use this entity in your automations.
+(Important to note: for relay devices, this is not the same as the binary sensor - it doesn't track the state of a relay, it only fires events when a detached switch is turned from `off` to `on`, and doesn't fire when a detached switch is turned from `on` to `off`)
```
bieniu, line 128: IMO this statement is not true. Both the event entities and `shelly.click` events inform about actions performed on the button (momentary button, not switch) and are not related to the state of the relay. If you use a bistable switch in detached mode, you should track its status using the input binary sensor.

madAndroid: Okay, please can you suggest an alternative - you've asked me to update the docs, and I've done so based on my experience with Shelly relays (1/1pm and 2.5) - the statement I've made here is _exactly_ the behaviour I've observed with these relays when the switch is set to detached mode.

bieniu: First of all, I didn't ask you to update the documentation.

Currently, there is an information in the documentation about the binary input sensors you should use in your setup.

obraz

Adding information for events/event entities that they do not correspond to changes in relay states is, in my opinion, unnecessary because they are, by design/definition, not related to the relay states, those are "input events". I know I may not be very representative when it comes to understanding the documentation and how Shelly devices work, but the change you proposed would be confusing for users in my opinion.
Perhaps it would be better to add a note that if you are using a toggle (bistable) switch you should use the input binary sensor rather than events/event elements in the automation.

madAndroid: Wow, how do people work with you??? lol .. whatever man, I'll close this then

madAndroid: Tell me, how is that _*several*_ people made the same mistake I did? we all didn't read the docs, right??
I'm sick of maintainers like you that treat people that want to contribute, like dirt. I will not be attempting to contribute again, if this is the response.

madAndroid: I'm abandoning this now, I was making changes that made sense to me, to hopefully help others that had the same issue. I don't have the appetite or endurance to respond here, closing the PR, maybe someone else will hit up against this same issue and submit a more sensible doc update, but that's clearly not me 👍

PR #29516, opened 2023-10-24
Revert "Update README.md"

frenck:
Reverts home-assistant/home-assistant.io#29512

This PR modifies a badge we carry as set out / created by Netlify. In exchange, they sponsor our project. We should not modify that at will.

https://www.netlify.com/press/

CC @c0ffeeca7 @HarshRanaOC


c0ffeeca7 (approved):
Thank you for the catch, @frenck :+1:

PR #29610, opened 2023-10-30
Improve IMAP integration SSL settings copy

strugee:
## Proposed change

Just some minor grammar/typo fixes.

## Type of change

## Additional information

## Checklist

- [ ] The documentation follows the Home Assistant documentation [standards]. (Leaving this unchecked since I used bold for emphasis instead of italics, and I need to amend locally)

### DIFF
--- a/source/_integrations/imap.markdown
+++ b/source/_integrations/imap.markdown
@@ -71,7 +71,7 @@ Yahoo also requires the character set `US-ASCII`.
 If the default IMAP server settings do not work, you might try to set an alternate SSL cipher list.
 The SSL cipher list option allows you to select the list of SSL ciphers to be accepted from this endpoint: `default` (_system default_), `modern` or `intermediate` (_inspired by [Mozilla Security/Server Side TLS](https://wiki.mozilla.org/Security/Server_Side_TLS)_).
 
-If you are using self signed certificates, you can turn off SSL verification. Be careful - normally this is **dangerous** and not necessary.
+If you are using self signed certificates, you can turn off SSL verification.
 
 <div class='note info'>
 
--- a/source/_integrations/imap.markdown
+++ b/source/_integrations/imap.markdown
@@ -66,12 +66,12 @@ Yahoo also requires the character set `US-ASCII`.
 
 </div>
 
-### Selecting an alternate SSL cipher list or disable SSL verification (advanced mode)
+### Selecting an alternate SSL cipher list or disabling SSL verification (advanced mode)
 
-If the default IMAP server settings do not work, you might try to set an alternate SLL cipher list.
-The SSL cipher list option allows to select the list of SSL ciphers to be accepted from this endpoint. `default` (_system default_), `modern` or `intermediate` (_inspired by [Mozilla Security/Server Side TLS](https://wiki.mozilla.org/Security/Server_Side_TLS)_)
+If the default IMAP server settings do not work, you might try to set an alternate SSL cipher list.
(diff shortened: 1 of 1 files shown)


source/_integrations/imap.markdown
```diff
@@ -66,12 +66,12 @@ Yahoo also requires the character set `US-ASCII`.
...
-### Selecting an alternate SSL cipher list or disable SSL verification (advanced mode)
+### Selecting an alternate SSL cipher list or disabling SSL verification (advanced mode)
 
-If the default IMAP server settings do not work, you might try to set an alternate SLL cipher list.
-The SSL cipher list option allows to select the list of SSL ciphers to be accepted from this endpoint. `default` (_system default_), `modern` or `intermediate` (_inspired by [Mozilla Security/Server Side TLS](https://wiki.mozilla.org/Security/Server_Side_TLS)_)
+If the default IMAP server settings do not work, you might try to set an alternate SSL cipher list.
+The SSL cipher list option allows you to select the list of SSL ciphers to be accepted from this endpoint: `default` (_system default_), `modern` or `intermediate` (_inspired by [Mozilla Security/Server Side TLS](https://wiki.mozilla.org/Security/Server_Side_TLS)_).
 
-If you are using self signed certificates can can turn of SSL verification.
+If you are using self signed certificates, you can turn off SSL verification. Be careful - normally this is **dangerous** and not necessary.
```
jbouwh:
```suggestion
If you are using self signed certificates, you can turn off SSL verification.
```

jbouwh: It is possible to have self signed certificates and still use validation.
In that case your custom CA record need to be in the trusted store. But I am not sure this is possible for HA core.

jbouwh: I think that when custom certificates are used on the email server, you might have no other option.

strugee: Hmm, true. What about something like:

> You can also turn off SSL verification. This is dangerous, a security risk, and not normally necessary - be very sure you know what you're doing. If you're using self-signed certificates, consider adding them to Home Assistant's CA trust store as an alternative.

jbouwh: Is there documentation on how to do this? In that case we could link to that. Otherwise I suggest leaving the additional text.
Setting a self-signed certificate on an imap server is not very common. And users that are capable of doing that, do not need the extra warning IMO.

jbouwh: https://community.home-assistant.io/t/add-private-cas-to-certificate-store/267452

It is still a feature request.

strugee: > Setting a self-signed certificate on an imap server is not very common. And users that are capable of doing that, do not need the extra warning IMO.

Yeah it's unclear to me who would really want to do this, especially since Let's Encrypt exists now. My primary concern is warning people who don't seriously know what they're doing away from disabling cert verification. I added the last sentence as a harm reduction thing but I'm totally fine dropping it too and just leaving it at "you probably shouldn't be doing this".

jbouwh (approved):
I reverted the suggested phrase that it would be dangerous to disable SSL validation for now, as it is clear to the user running an own mailserver what the consequences are.

Thanx for the improments @strugee !

strugee:
@jbouwh you're welcome, but is it? What if it's an inexperienced user running something like [Mailu](https://mailu.io/2.0/) or [Mail-in-a-Box](https://mailinabox.email/)?

jbouwh:
> @jbouwh you're welcome, but is it? What if it's an inexperienced user running something like [Mailu](https://mailu.io/2.0/) or [Mail-in-a-Box](https://mailinabox.email/)?

Turning of validation, does not necessarily turn off encryption. Especially for cases where there is no trusted certificate validation can be turned off. Many custom IMAP email servers do not have a trusted certificate.

PR #30132, opened 2023-12-04
[Google Assistant SDK] Elaborate on IPv6 issues

Daniel15:
## Proposed change
Updates the Google Assistant SDK documentation to elaborate on the IPv6 issues, and potential workarounds.

## Type of change

## Additional information

## Checklist


source/_integrations/google_assistant_sdk.markdown
```diff
@@ -69,13 +69,19 @@ The integration setup will next give you instructions to enter the [Application
 
 {% enddetails %}
 
+## IPv6
+
+Due to limitations in Google's SDK, broadcasts do not currently support IPv6. In order for broadcasts to work, IPv6 needs to be disabled for both your Home Assistant server as well as all speakers. To disable IPv6 on your Home Assistant server, set `net.ipv6.conf.all.disable_ipv6=1` in `/etc/sysctl.conf` or your Docker configuration.
```
agners: There is no such thing as broacasts in IPv6.

As we don't have sysctl configuration support in Home Assistant OS, disabling IPv6 that way is not possible.

What is the problem here exactly?

Daniel15: @agners I used that sysctl for my Docker container and it works well there.  I haven't used HAOS though, but does it not support sysctls at all? I thought it was Linux based?

The problem is that the docs currently say to disable IPv6 across the entire network, when really you *only* need to disable IPv6 on the Home Assistant server and the Google speakers. Everything else on the network can continue to use it. In my case, I have several IPv6-only servers "in the cloud" so I can't totally disable IPv6, but I still managed to get broadcasts working. 

Disabling IPv6 for the entire network is a bad idea in 2023, especially as many ISPs outside the USA are using CGNAT for IPv4. In the case of CGNAT, IPv6 is the only way to make systems externally accessible (for the cases where you can't use a VPN).

agners: HAOS is Linux based and uses sysctl. However, it is largely statelesss and much of the rootfs is read-only, including `/etc/sysctl.conf`. So that won't work on HAOS as-is.

> Disabling IPv6 for the entire network is a bad idea in 2023, especially as many ISPs outside the USA are using CGNAT for IPv4. In the case of CGNAT, IPv6 is the only way to make systems externally accessible (for the cases where you can't use a VPN).

100% agree, in fact with Thread and Matter based on IPv6, I'd say one should not disable IPv6 on Home Asssistant even :smile: 

I'd like to understand what the issue here is at all. In a quick try in my home, where all devices have global and ULA and link-local IPv6 addresses, asking the speaker t broadcast things worked just fine. Is it only a HA issue? :thinking: Are there Core issues I could read up about this?

c0ffeeca7: Marking this as *Draft* as there is an ongoing discussion. Please undraft it once you feel the discussion has been resolved.

Daniel15: >  Are there Core issues I could read up about this?

@agners Sorry for the delay in replying.  It's an issue with the Google Assistant SDK rather than anything specific to Home Assistant.

Please see prior discussions:
 - https://greghesp.github.io/assistant-relay/docs/other/faq/#i-tried-a-broadcast-but-i-dont-hear-anything (from a different software product)
 - https://stackoverflow.com/a/62772434 (the answer here lacks substance though)
 - https://community.home-assistant.io/t/google-assistant-sdk-notify-doesnt-work/514036/2
 - https://github.com/home-assistant/core/issues/84726

Google have deprecated the Python Google Asssistant SDK and I doubt they'll ever fix this issue. It seems like they don't care about development of Google Assistant any more.

agners: Ok, thanks for these links!

I've also reproduced this on my end, and I can confirm the same observation.

However, it doesn't seem necessary to disable IPv6 entirely using sysctl. Disabling the SLAAC/DHCPv6 to assign ULA/global addresses seems sufficient. Or in other words: Google Assistant broadcast works even with link-local IPv6 addressing enabled. That is exactly what the "Disabled" option of the Home Assistant OS network configuration does:

image

Can you adjust the text to suggest this method on Home Assistant OS installations?

Also, I think it makes sense to link the Core issue here to make it clear that this is a recognized problem we can't fix.

agners: Maybe make the title a bit more verbose, e.g. "IPv6 limitations when using Broadcast feature" or "Broadcast feature and IPv6". :thinking:

agners: Let's get this done, @Daniel15 does this sound good to you?

```suggestion
## Broadcast message feature and IPv6

Due to limitations in Google's SDK, broadcasting messages to Google Assistant speakers and displays isn't supported in networks using IPv6. In order for broadcasts to work, IPv6 needs to be disabled for both your Home Assistant server as well as all speakers and displays.

To disable IPv6 on Home Assistant Operating System or Supervised installations, go to Settings > System > Network and set IPv6 to "Disabled". Note that leaving link-local addressing on seems to be sufficient, e.g. if your Server uses NetworkManager the `ipv6.method` `link-local` is sufficient.
```

agners:
Closing as this seems to be fixed. See also #31338.

Daniel15:
This is great! I'll try it out soon. Hopefully this means I can finally remove my "No IPv6" VLAN that only exists for the purpose of working around this issue.

PR #30696, opened 2024-01-07
add information to help configure the energy dashboard with Solax Power

brew-your-own:
Solax Power integration

## Proposed change
This change adds a section to the Solax power integration documentation to help users configure the energy dashboard.

## Type of change

## Checklist


source/_integrations/solax.markdown
```diff
@@ -39,6 +39,22 @@ template:
...
 
+### Configuring the Energy Dashboard
+
+There are generally 3 sensors from your inverter that you need to configure in the energy dashboard:
+
+- the total consumption sensor (in kWh) for the "Grid Consumption"
+- the total feed-in sensor (in kWh) for the "Return to grid", i.e. the solar panel energy that you do not consume and return to the grid instead
+- the total on-grid yield sensor (in kWh) for the "Solar production"
+
+These are "running total" sensors, do not use the "daily" sensors.
```
frenck: Why not? That is totally valid.

If you can select them on the energy dashboard it means they are compatible.

source/_integrations/solax.markdown
```diff
@@ -39,6 +39,22 @@ template:
...
+
+Depending on the type of inverter, the sensors you need may be named differently. Here is an example with the `X1HybridGen4` inverter:
+
+- `sensor.solax_XXXXXXXX_total_consumption` for the "Grid Consumption"
+- `sensor.solax_XXXXXXXX_total_feed_in_energy` for the "Return to grid"
+- `sensor.solax_XXXXXXXX_on_grid_total_yield` for the "Solar production"
```
frenck: We should not reference them by entity IDs.

They are not guaranteed and above all translated.

Let's remove that.

source/_integrations/solax.markdown
```diff
@@ -39,6 +39,22 @@ template:
 
 {% endraw %}
 
+### Configuring the Energy Dashboard
+
+There are generally 3 sensors from your inverter that you need to configure in the energy dashboard:
+
+- the total consumption sensor (in kWh) for the "Grid Consumption"
+- the total feed-in sensor (in kWh) for the "Return to grid", i.e. the solar panel energy that you do not consume and return to the grid instead
+- the total on-grid yield sensor (in kWh) for the "Solar production"
```
frenck:
```suggestion
- The total consumption sensor (in kWh) for the **Grid Consumption**.
- The total feed-in sensor (in kWh) for the **Return to grid**, like the solar panel energy that you do not consume and return to the grid instead.
- The total on-grid yield sensor (in kWh) for the **Solar production**.
```

source/_integrations/solax.markdown
```diff
@@ -39,6 +39,22 @@ template:
 
 {% endraw %}
 
+### Configuring the Energy Dashboard
+
+There are generally 3 sensors from your inverter that you need to configure in the energy dashboard:
+
+- the total consumption sensor (in kWh) for the "Grid Consumption"
+- the total feed-in sensor (in kWh) for the "Return to grid", i.e. the solar panel energy that you do not consume and return to the grid instead
```
frenck: Isn't this country depended? I know this ain't going to work in the Netherlands for example.

frenck (changes requested):

source/_integrations/solax.markdown
brew-your-own: ok the problem I am trying to solve is that  it is not obvious which entity tracks what - the naming conventions are not clear for non-electricians ;)
In my solax sensors, I have 4 different sensors that are compatible with the "Solar Production" section in the energy dashboard: which one should I pick?
Using a practical example might help make this easier for other users. WDYT?

source/_integrations/solax.markdown
brew-your-own: ok I was not aware of this nuance.
how do we make it easier to users which entities to put where in the energy dashboard?
I could qualify this section as an illustration with the X1HybridGen4 inverter in Spain, to make it clearer this is just an example. What would it look like in the NL ?

source/_integrations/solax.markdown
brew-your-own: Do you mean that it does not matter which one I use? I have both "daily" and "total" sensors - do they both work in the energy dashboard?
If yes, then we should change this to clarify, indeed.

klaasnicolaas: Yes you can use them both

source/_integrations/solax.markdown
klaasnicolaas: What we normally do is that we do not literally name the entity_id, but describe what kind of entity you have.

brew-your-own: ok I dropped this section in the latest update.

source/_integrations/solax.markdown
brew-your-own: I removed the line stating (incorrectly) that the running total must be used
I removed "total" in each of the items of the list.

source/_integrations/solax.markdown
frenck: Depends on the setup. There are many ways this can be hooked up. For example, inverters don't have to be connected directly to the grid (this is not specific to NL)

source/_integrations/solax.markdown
```diff
@@ -39,6 +39,14 @@ template:
 
 {% endraw %}
 
+### Configuring the Energy Dashboard
+
+There are generally at least 3 sensors from your inverter that you need to configure in the energy dashboard:
+
+- The consumption sensor (in kWh) for the **Grid Consumption**.
+- The feed-in sensor (in kWh) for the **Return to grid**, i.e. the solar panel energy that you do not consume and return to the grid instead.
+- The on-grid yield sensor (in kWh) for the **Solar production**.
```
frenck:
```suggestion
- The consumption sensor (in kWh) for the **Grid Consumption**.
- The feed-in sensor (in kWh) for the **Return to grid**, for example, the solar panel energy you do not consume and return to the grid instead.
- The on-grid yield sensor (in kWh) for the **Solar production**.
```

frenck (approved):
Thanks, @brew-your-own 👍 

../Frenck

PR #30907, opened 2024-01-18
Remove double quotes from the username and password in the example co…

benseven:
…nfiguration

I just spent an hour trying to work out what I was getting wrong with adding Evohome to HA. If there's no quotes needed around the username and password strings, there shouldn't be quotes around them in the example code.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/evohome.markdown
+++ b/source/_integrations/evohome.markdown
@@ -34,9 +34,11 @@ To set up this integration, add the following to your `configuration.yaml` file:
 ```yaml
 # Example configuration.yaml entry
 evohome:
-  username: "YOUR_USERNAME"
-  password: "YOUR_PASSWORD"
+  username: YOUR_USERNAME
+  password: YOUR_PASSWORD
 ```
+
+After you have added these details, restart Home Assistant. 
 
 {% configuration %}
 username:


source/_integrations/evohome.markdown
```diff
@@ -34,10 +34,12 @@ To set up this integration, add the following to your `configuration.yaml` file:
 ```yaml
 evohome:
-  username: "YOUR_USERNAME"
-  password: "YOUR_PASSWORD"
+  username: YOUR_USERNAME
+  password: YOUR_PASSWORD
```
frenck, line 38: These changes are not correct. String should be enquotes, as per our YAML documentation standards.

../Frenck

frenck:
Thanks for your PR. The current documentation, is however, correct.

../Frenck

benseven:
With quotes, the configuration does not work.

Without quotes, the configuration works.

If you haven't tested that - as I have this morning, with a now working Evhohome integration based on these changes and these changes alone - then you are entirely wrong to refuse this PR.

source/_integrations/evohome.markdown
benseven: Which is preferable - documentation that meets your standards, or documentation that causes the software to actually work?

frenck:
> With quotes, the configuration does not work.
> Without quotes, the configuration works.

There is no difference in YAML's perspective, meaning you might have a different issue at hand.
If you suspect a bug, please raise an issue in our Core issue tracker.

source/_integrations/evohome.markdown
frenck: If it doesn't work, then it sounds like a bug. As there is no difference in those syntaxes for the outcome. Both are considered a string by YAML. As a matter of fact, enquoting is better, as it is more explicit (hence it is the standard for documenting string values).

../Frenck

benseven: I've retested and added the quotes back in, and it still works so this PR can be ditched. For whatever reason HA didn't pick up or start working on the first restart, but on the second restart (when I had removed the quotes) it did  - my apologies.

PR #31023, opened 2024-01-24
Update statistics-graph.markdown

spacecabbie:
Add definitions of min, max, mean, and sum

## Proposed change
Just to clearify the meaning of the option mainly mean might confuse people.

## Type of change
- [x ] Spelling, grammar or other readability improvements (`current` branch).

## Checklist

- [ x] The documentation follows the Home Assistant documentation [standards].

### DIFF
--- a/source/_dashboards/statistics-graph.markdown
+++ b/source/_dashboards/statistics-graph.markdown
@@ -17,7 +17,7 @@ Screenshot of the statistics graph card with none metered entities and `chart_ty
 Screenshot of the statistics graph card with a metered entity and `chart_type` `bar`.
 </p>
 
-Statistics are gathered every 5 minutes for sensors that support it. It will either keep the `min`, `max`, and `mean` of a sensor's value for a specific hour or the `sum` for a metered entity.
+Statistics are gathered every 5 minutes for sensors that support it. It will either keep the `min`[^1], `max`[^1], and `mean`[^1] of a sensor's value for a specific hour or the `sum`[^1] for a metered entity.
 
 If your sensor doesn't work with statistics, check [this](/more-info/statistics/).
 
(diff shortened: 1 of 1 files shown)


source/_dashboards/statistics-graph.markdown
```diff
@@ -17,7 +17,7 @@ Screenshot of the statistics graph card with none metered entities and `chart_ty
 Screenshot of the statistics graph card with a metered entity and `chart_type` `bar`.
 </p>
 
-Statistics are gathered every 5 minutes for sensors that support it. It will either keep the `min`, `max`, and `mean` of a sensor's value for a specific hour or the `sum` for a metered entity.
+Statistics are gathered every 5 minutes for sensors that support it. It will either keep the `min`[^1], `max`[^1], and `mean`[^1] of a sensor's value for a specific hour or the `sum`[^1] for a metered entity.
```
c0ffeeca7:
```suggestion
Statistics are gathered every 5 minutes for sensors that support it. It will either keep the <abbr title="minimum">`min`</abbr>, <abbr title="maximum">`max`</abbr>, and <abbr title="average">`mean`</abbr> of a sensor's values for a specific hour or the `sum` for a metered entity.
```
@spacecabbie. Counterproposal: how about just adding the term average in brackets? In this way, the information is right where it is needed.

spacecabbie: I get wat you are saying in my case only mean was confusing, but IMHO we should always be as complete as possible when it comes to documenting. But I leave it up to you how to merge it.

c0ffeeca7: I agree with you in that it is good practice to spell out abbreviations. Amended the suggestion to use the `abbr` tag. This will show a pop up.
image

c0ffeeca7 (approved):
Thank you, @spacecabbie :+1:

spacecabbie:
> Thank you, @spacecabbie 👍

thank you aswel I don't do this ofthen it was helpfull to see how its done so i can do more of these document addons/suggestions.

Ps. made 1 more pull request because accidently removed the `` around min

PR #31029, opened 2024-01-24
unprivileged doc: Warn about running installation script again

odinho:
I did the very unfortunate thing to run the dpkg installation again as suggested to 'quick fix' the issue Home Assistant was seeing. It was not a quick fix, it's quite invasive since it can block you out from the entire home assistant computer. (it changed the /etc/interfaces etc, even if those were in fact made by home assistant itself)

This install did show up as 'supported', so even if it had some extra packages on the outside (ssh, and gnome) everything else was as expected.

So now I'm locked out of a computer that's in an empty house 7 hours one way away. :'(

I did think it would take over the computer, which is okay, it's only used for home assistant, but it making itself inaccessible from internet should probably get an extra mention here I think :)

## Proposed change

Mention that the suggested fix can change internet settings.

## Type of change

## Checklist

### DIFF
--- a/source/more-info/unhealthy/privileged.markdown
+++ b/source/more-info/unhealthy/privileged.markdown
@@ -16,3 +16,7 @@ If you are running an older version of our Home Assistant OS, update it in the
 If this is not our Home Assistant OS, your operating system might be out of date. Try checking for and
 installing updates, then restarting your system. If this doesn't work, you may need to re-run our
 [convenience installation script](https://github.com/home-assistant/supervised-installer).
+
+Note this may lock you out of your instance unless you have physical access, since it will reset all
+network settings and take over the machine again. If you are able, a more targeted fix is probably
+warranted.
--- a/source/more-info/unhealthy/privileged.markdown
+++ b/source/more-info/unhealthy/privileged.markdown
@@ -14,9 +14,9 @@ If you are running an older version of our Home Assistant OS, update it in the
 {% my configuration title="Configuration" %} panel.
 
 If this is not our Home Assistant OS, your operating system might be out of date. Try checking for and
-installing updates, then restarting your system. If this doesn't work, you may need to re-run our
-[convenience installation script](https://github.com/home-assistant/supervised-installer).
+installing updates, then _restarting your system_.
 
-Note this may lock you out of your instance unless you have physical access, since it will reset all
-network settings and take over the machine again. If you are able, a more targeted fix is probably
-warranted.
+If this doesn't work, you may need to re-run our
+[convenience installation script](https://github.com/home-assistant/supervised-installer). Note this 
+may lock you out of your instance unless you have physical access, since it will reset all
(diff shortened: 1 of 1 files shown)


odinho:
<del>Reading more up on this. I think the docs should actually mention that this is a known bug and you should simply try doing a restart of the supervisor container or the host system first.

Now I did a real stupid thing (should have checked what the `.deb` file actually was going to do), for something that would have had a really simple workaround that _actually_ works.</del>

https://github.com/home-assistant/supervisor/issues/4381

Unsure whether to close this and make a new PR, or just push the overwritten update here. Ah, I'll push a new commit, that way they're both in.

---

Edit: Well that makes me both look and feel super stupid. It's actually there in the text. I must have somehow glanced over it or didn't consider it relevant. Had I known the issue was supervisor got an update I could understand how a reboot might fix it, I guess I ignored it.

Really time to go to bed. I made the restart bit stick out a bit and kept the warning about the installer deb ('convenience script') in.

agners:
Ugh, I see, yeah that is unfortunate :cry: 

Adding the emphasis on restart make sense.

However, the note about resetting the network settings are probably better placed in the [supervised-installer](https://github.com/home-assistant/supervised-installer) REDME.md. Ideally probably with a warning emphasis (https://github.com/orgs/community/discussions/16925).

source/more-info/unhealthy/privileged.markdown
```diff
@@ -14,5 +14,9 @@ If you are running an older version of our Home Assistant OS, update it in the
...
 
 If this is not our Home Assistant OS, your operating system might be out of date. Try checking for and
-installing updates, then restarting your system. If this doesn't work, you may need to re-run our
-[convenience installation script](https://github.com/home-assistant/supervised-installer).
+installing updates, then _restarting your system_.
+
+If this doesn't work, you may need to re-run our
+[convenience installation script](https://github.com/home-assistant/supervised-installer). Note this 
+may lock you out of your instance unless you have physical access, since it will reset all
+network settings.
```
agners, line 22: This is also no longer a install script actually, so we probably should suggest this:

```suggestion
If this doesn't work, you may need to re-install the [Home Assistant Supervised Debian Package](https://github.com/home-assistant/supervised-installer).
```

@ikifar2012 more generally, do we have instructions how to update/maintain a Supervised installation? What is the general expectation?

agners (changes requested):

source/more-info/unhealthy/privileged.markdown
ikifar2012: The expectation is that the Debian package and OS Agent are supposed to always be on the latest version... it seems the [ADR](https://github.com/home-assistant/architecture/blob/master/adr/0014-home-assistant-supervised.md) mentions the agent always being up to date but not the installer package itself, maybe it should explicitly mention that in the ADR?

PR #31161, opened 2024-01-31
Update statistics-graph.markdown

spacecabbie:
accidently removed the `` around min restored this.


c0ffeeca7 (approved):
Thank you for the catch, @spacecabbie :+1:

PR #31273, opened 2024-02-06
service-calls: added info that target values should be lower-cased

MacDada:
Added info about values case–(in)sensitivity.

I was surprised when my `area_id` did not work, only later to notice, that the value is case–sensitive. Apparently, it is different from `entity_id`, which is case–insensitive.

## Type of change

### DIFF
--- a/source/_docs/scripts/service-calls.markdown
+++ b/source/_docs/scripts/service-calls.markdown
@@ -28,7 +28,7 @@ Instead of targeting an entity, you can also target an {% term area %} or {% ter
 This is done with the `target` key.
 
 A `target` is a map that contains at least one of the following: `area_id`, `device_id`, `entity_id`.
-Each of these can be a list.
+Each of these can be a list. The values should be lower-cased.
 
 The following example uses a single service call to turn on the lights in the
 living room area, 2 additional light devices and 2 additional light entities:


source/_docs/scripts/service-calls.markdown
```diff
@@ -28,7 +28,7 @@ Instead of targeting an entity, you can also target an {% term area %} or {% ter
 This is done with the `target` key.
 
 A `target` is a map that contains at least one of the following: `area_id`, `device_id`, `entity_id`.
-Each of these can be a list.
+Each of these can be a list. The values of `entity_id` are case–insensitive, while `area_id` and `device_id` are case–sensitive.
```
frenck: Everything should be considered case-sentitive. Let's adjust the text to reflect that.

frenck (changes requested):

source/_docs/scripts/service-calls.markdown
MacDada: Actually, when I'm trying to create areas `foo` and `FOO` at http://homeassistant.local:8123/config/areas/dashboard -> I'm getting `The name FOO (foo) is already in use` error.

That would mean, that the areas should also be case–**in**sensitive when being a target. Or the other way,  if they _should_ be case–sensitive, that would mean that there is a bug in the dashboard…

Anyway: I'd love it, if the docs reflected the real behaviour, not a wishful thinking.

frenck: > That would mean, that the zones should also be case–insensitive when being a target. 

You are mixing up scopes.

> Anyway: I'd love it if the docs reflected the real behaviour, not the wishful thinking.

Yup, well, there is a backward compatibility layer active, but entities IDs have naming limits and are case sensitive (capital letters cannot be used in entity names, try it yourself, rename an entity ID).

MacDada: > You are mixing up scopes.

Kinda, probably, yeah. I'm looking for consistency -> some way to lower [WTF–factor](https://www.reddit.com/r/ProgrammerHumor/comments/cd1xio/wtfs_per_minute/) ;-)

> capital letters cannot be used in entity names, try it yourself, rename an entity ID

When I edit entity_id with changed caps, the error is that the entity_id already exists -> proving in yet another place, that entity_ids are case–insensitive, not case–sensitive.

The error message does not state that capital letters are invalid:

Screenshot 2024-02-06 at 23 16 24

---

EDIT1: when I change it entirely with upper caps, the error message is different -> just "invalid", without pointing to disallowed capital letters, but I trust that it's the reason:

Screenshot 2024-02-06 at 23 18 46

---

EDIT2: the experiment is telling me that maybe capital letters are not allowed on entity–management, but entity_ids are case **in**sensitive anyway.

frenck: Still, they are consider so. The point is: We have some backward compatible logic/history in place. That doesn't make it a good idea or feature.

Please adjust the text as suggested, thanks!

MacDada: > Everything should be considered case-sentitive.

I'm not sure how to phrase that… Maybe just say the values should be lower-cased (if this is an actual rule for IDs)?

I mean, if the letters cannot be (can't be set to) uppercase, then the point about case–sensitivity is a bit moot.

MacDada: Maybe this?

```
Each of these can be a list. As always, be mindful about values [case sensitivity](/docs/configuration/yaml/#upper-and-lower-case).
```

Is there a better way to make the link?

frenck: > I'm not sure how to phrase that… Maybe just say the values should be lower-cased

Perfect! 👌

MacDada: > > I'm not sure how to phrase that… Maybe just say the values should be lower-cased
> 
> Perfect! 👌

Rebased and pushed a new commit.

frenck (commented):
Thanks, @MacDada 👍 

../Frenck

frenck (approved):
Thanks, @MacDada 👍 

../Frenck

PR #31275, opened 2024-02-06
templating: added info that the Jinja2 engine is modified

MacDada:
Which should indicate that there might be differences to the original engine.

Source: https://community.home-assistant.io/t/documentation-for-functions-and-operations/398690/5

## Type of change

### DIFF
--- a/source/_docs/configuration/templating.markdown
+++ b/source/_docs/configuration/templating.markdown
@@ -16,7 +16,7 @@ Templating is a powerful feature that allows you to control information going in
 
 ## Building templates
 
-Templating in Home Assistant is powered by the [Jinja2](https://palletsprojects.com/p/jinja) templating engine. This means that we are using their syntax and make some custom Home Assistant variables available to templates during rendering. Jinja2 supports a wide variety of operations:
+Templating in Home Assistant is based on a modified [Jinja2](https://palletsprojects.com/p/jinja) templating engine. This means that we are using their syntax and make some custom Home Assistant variables available to templates during rendering. Jinja2 supports a wide variety of operations:
 
 - [Mathematical operation](https://jinja.palletsprojects.com/en/latest/templates/#math)
 - [Comparisons](https://jinja.palletsprojects.com/en/latest/templates/#comparisons)


source/_docs/configuration/templating.markdown
```diff
@@ -16,7 +16,7 @@ Templating is a powerful feature that allows you to control information going in
 
 ## Building templates
 
-Templating in Home Assistant is powered by the [Jinja2](https://palletsprojects.com/p/jinja) templating engine. This means that we are using their syntax and make some custom Home Assistant variables available to templates during rendering. Jinja2 supports a wide variety of operations:
+Templating in Home Assistant is based on a modified [Jinja2](https://palletsprojects.com/p/jinja) templating engine. This means that we are using their syntax and make some custom Home Assistant variables available to templates during rendering. Jinja2 supports a wide variety of operations:
```
frenck, line 19: That isn't a correct statement.

We use the default Jinja2 from upstream. We load additional extensions, which has been documented already on this same page.

../Frenck

frenck (changes requested):

source/_docs/configuration/templating.markdown
MacDada: [According to tom_l, a moderator](https://community.home-assistant.io/t/documentation-for-functions-and-operations/398690/5), it is _modified_, not just _extended_.

No matter the technical mechanism, if it is _copy -> paste -> change -> commit_ or _download with dependency manager -> override stuff with inheritance/config/whatever_ => if the result is that the API is different between the "original" engine and the HA's implementation ("different" as _incompatible_ changes, not just _adding_ stuff) => then I'd argue Tom is correct: it is a _modified_ engine.

**I've just tested the forum's example of `float` filter's _modified behaviour_ with http://jinja.quantprogramming.com/ – actually `'foo'|float` gives the default `0.0` for Jinja "original", while it throws in HA.**

My guess, after quick grepping, is that the default `float` filter [gets overriden here](https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/template.py#L2499), while – for example – `abs` filter isn't overridden, but I may be mistaken as for the place of _modification_ ;-)

MacDada: yep, the original Jinja's `float` [returns the default](https://github.com/pallets/jinja/blob/main/src/jinja2/filters.py#L988), while HA's _modification_ [throws](https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/template.py#L2009)

MacDada: > Templating in Home Assistant is powered by the [Jinja2](https://palletsprojects.com/p/jinja) templating engine **with a bit of monkey-patching**.

:D

frenck: > [According to tom_l, a moderator](https://community.home-assistant.io/t/documentation-for-functions-and-operations/398690/5), it is modified, not just extended.

Sorry, that statement is just incorrect.

We load extensions, sure. The engine is the same.

MacDada: > We load extensions, sure.

If the "loaded extensions" just added functionalities -> I'd agree.

But I've shown above, HA _changes_ (_modifies_) the original behaviour of Jinja (`float` filter is an example) -> therefore I call it _modification_. The result being: original Jinja docs do _not_ apply to HA in all cases. Therefore one shall not be surprised that, if they do something according to Jinja, it might not work in HA, or it might work differently. Therefore, my PR.

> The engine is the same.

I don't want to get into metaphysics, but… HA does [monkey patching](https://en.wikipedia.org/wiki/Monkey_patch) on Jinja – it's not _rewriting_ the original code, but _modifying_ the behaviour at runtime. Anyway: _modifying_.

MacDada: Mind that the next step is to list those changes somewhere. Or at least add some kind of a warning in places, where HA's implementation differs from Jinja's original.

If one already _knows Jinja_ and reads that HA uses Jinja internally, they shouldn't assume their Jinja knowledge is valid and shall make corrections accordingly.

frenck: Sorry, don't agree.

Even in Jinja itself these are extensions (we also don't load all Jinja shipped extensions).

The engine, is the same and unmodified.

MacDada: > Sorry, don't agree.

Yet, u don't disprove my arguments.

> The engine, is the same and unmodified.

Is monkey patching a modification or you don't agree with wikipedia definition as well?

MacDada: Let me ask another way: do you see that the docs need an improvement, to acknowledge the fact, that using Jinja docs to learn how to use templates in HA is mildly saying "not enough"? That it can lead to a wrong code in HA?

Then we might rephrase my original PR, sure. I don't think the "engine vs extension" distinction actually makes any difference to the reader though.

> `Templating in Home Assistant is powered by the [Jinja2](https://palletsprojects.com/p/jinja) templating engine. This means that we are using their syntax, make some custom Home Assistant variables available to templates during rendering; with [a few changes](the-list-shall-be-created-with-later-pr].`

Or sth like that. Anyway: if you link to Jinja docs with an intention to learn from them, i'd advise to point out what one should NOT learn from them.

frenck: > Yet, u don't disprove my arguments.

I did.

I don't agree.

frenck: Going to close this, as this as I don't agree and don't see any value in having a discussion over this.

The added word is wrong, the engine is the same.

Nevertheless, thanks for being willing to contribute.

../Frenck

PR #31348, opened 2024-02-10
Revert "Document scan_interval support for RESTful switch"

frenck:
Reverts home-assistant/home-assistant.io#31330

Reverts this change, as this is not a property of this integration at all.

Platform specific options are documented here:

<https://www.home-assistant.io/docs/configuration/platform_options/#scan-interval>

See also the source, which doesn't have it:

https://github.com/home-assistant/core/blob/845071f8f0b1297073d3314c5d465de7787faa17/homeassistant/components/rest/switch.py#L66-L86

../Frenck

/CC: @c0ffeeca7 @holysoles

### DIFF
--- a/source/_integrations/switch.rest.markdown
+++ b/source/_integrations/switch.rest.markdown
@@ -96,11 +96,6 @@ verify_ssl:
   required: false
   type: boolean
   default: true
-scan_interval:
-  description: Define the frequency for calling the REST endpoint to get the current state in seconds.
-  required: false
-  type: integer
-  default: 30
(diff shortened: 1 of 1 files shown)


holysoles:
@frenck apologies for this, I had a misunderstanding of what should be documented at the platform level. After having tested it respecting the configuration option, I thought this was good to go. I spent some time reviewing the code and the HA architecture documentation and now understand. Thanks for catching the error.

c0ffeeca7 (approved):
Thank you for the catch, @frenck :+1:

PR #31355, opened 2024-02-11
Positioning terminology tooltip to be always fully visible

### LINKED ISSUE #31341: Terminology tooltip does is not always fully visible
### Feedback

It sometimes does not "fit" in the visible part of a page:

image

I've started experimenting with some JS to fix the positioning of the tooltip, so that it'd be always visible to the reader:

```js
[...document.getElementsByClassName('terminology-tooltip')].forEach((t) => {
    const rect = t.getBoundingClientRect();
    const marginLeft = parseInt(window.getComputedStyle(t).marginLeft);
    const minimalMargin = 20;
    const siteHeaderHight = document.getElementsByClassName('site-header')[0].clientHeight;
    
    console.log(rect)
    
    if (rect.left < 0) {
        console.log(`calc(50% + ${Math.abs(rect.left) + minimalMargin}px)`);
        // t.style.marginLeft = (marginLeft - rect.left + minimalMargin) + 'px';
        t.style.left = `calc(50% + ${Math.abs(rect.left) + minimalMargin}px)`;
    } else if (rect.right + minimalMargin > window.innerWidth) {
        // t.style.marginLeft = (marginLeft - (rect.right - window.innerWidth) - minimalMargin) + 'px';
        t.style.left = `calc(50% - ${rect.right - window.innerWidth + minimalMargin}px)`;
    }

    if (rect.top < siteHeaderHight) {
        t.style.bottom = 'auto';
        t.style.top = '1lh';
    }
});
```

It finds if the tooltip box fits in the visible part of the window, and repositions it if it doesn't.

It's a POC, but I can finish/polish it and make a PR, if you think this is the right direction.

### URL

https://www.home-assistant.io/docs/

### Version

_No response_

### Additional information

_No response_

MacDada:
Currently it does not always fit within the page:

image

#resolves https://github.com/home-assistant/home-assistant.io/issues/31341

## Type of change

## Checklist


source/javascripts/terminology_tooltip.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+[...document.getElementsByClassName('terminology')].forEach($terminology => {
```
ludeeus: Lets not prefix with `$`

source/javascripts/terminology_tooltip.js
```diff
@@ -0,0 +1,44 @@
...
+    const horizontalMargin = 20;
+
+    const topMargin = document
+        .getElementsByClassName('site-header')[0]
+        .clientHeight;
+
+    const terminology = $terminology.innerText;
+
+    const debug = message => {
```
ludeeus: This do not need to be here.

source/javascripts/terminology_tooltip.js
ludeeus: there is a few other variables that also do this in this file

source/javascripts/terminology_tooltip.js
```diff
@@ -0,0 +1,44 @@
...
+        console.debug(`Terminology tooltip (${terminology}): ${message}`);
+    }
+
+    const $tooltip = $terminology.querySelector('.terminology-tooltip');
```
ludeeus: While unlikely, this can be undefined if a class of "terminology" is added to an element that is not using the term tag

source/javascripts/terminology_tooltip.js
MacDada: > Lets not prefix with `$`

sure.

just to explain myself: this is a convention kinda taken from jQuery -> vars containing DOM objects are prefixed with `$` -> so it is easy to differentiate them from "normal" objects/var.

like in this case: `$terminology` (the current DOM object) vs `terminology` (the current "term")

source/javascripts/terminology_tooltip.js
MacDada: I understand you want to me to get rid of `console.debug()`, not just `debug()` helper? ;)

ludeeus: Yeah, both of them 👍

source/javascripts/terminology_tooltip.js
MacDada: Yep. It will be a NPE. It will throw in the console. It will not reposition the tooltip. That's good enough for me.

Do you want me to check for `$tooltip` existence and explicitly throw, instead of the implicit NPE?

ludeeus: I think we can just return if the queryselector returns nothing for this

MacDada: > I think we can just return if the queryselector returns nothing for this

That would silence a most-likely error on the programmer part -> someone forgetting to grep when attaching the class name in a new context.

I know, in theory, it could be possible for the two contexts to coexist with the same class name  -> but I'd strongly advise against that, as it could lead to unexpected results (like "their" new code messing with these tooltips).

source/javascripts/terminology_tooltip.js
MacDada: https://github.com/home-assistant/home-assistant.io/pull/31355/commits/aa23edcb8f62957c2bfa1c6b416415dedd81d254

source/javascripts/terminology_tooltip.js
MacDada: https://github.com/home-assistant/home-assistant.io/pull/31355/commits/9fc502028c3aac90751ed5a8f5f6042d9c976e5c

source/javascripts/terminology_tooltip.js
ludeeus: In a web application with a lot of moving parts I would agree.
But in a statically hosted documentation site that is as likely as "terminology" being added to some random element.

I have no strong feelings either way for this 👍

ludeeus (approved):
Thanks 👍

source/javascripts/terminology_tooltip.js
MacDada: i've just stumbled upon name collisions MANY times, and i became pretty pedantic to keeping global names distinct and separated

> In a web application with a lot of moving parts I would agree.

haha, yeah - that's my background -> i'm probably overthinking this one ;)

PR #31518, opened 2024-02-20
Fix example template sensor for weather.get_forecasts

jaydeethree:
## Proposed change

There are two problems with this example:

1. Using a time-based trigger here will cause Home Assistant to log an error. See https://github.com/home-assistant/core/issues/99914 for details.

2. It makes more sense to update the template sensor when the weather entity or its attributes change, instead of updating it on a fixed schedule.

## Type of change

## Additional information

## Checklist


c0ffeeca7:
Does this work if my weather service only updates every 3 hours or so?

emontnemery:
The PR description states:

> Using a time-based trigger here will cause Home Assistant to log an error. See https://github.com/home-assistant/core/issues/99914 for details.

That seems to be a bug, which we should fix, not hide by changing this example.

> It makes more sense to update the template sensor when the weather entity or its attributes change, instead of updating it on a fixed schedule.

Why does this makes sense? What does the weather entity's state have to do with the forecast?

jaydeethree:
>That seems to be a bug, which we should fix, not hide by changing this example.

That's fair.

>Why does this makes sense? What does the weather entity's state have to do with the forecast?

Since the weather entity contains the forecast data, I would expect that both the forecast data and the weather entity's state/attributes would all be updated at the same time. Perhaps that's not actually true though? That's how I do it in [my weather integration](https://github.com/jaydeethree/Home-Assistant-weatherdotcom/), but maybe other integrations work differently (in which case I agree that this doesn't make sense).

emontnemery:
@jaydeethree the weather entity was recently updated to deprecate the forecast property from the state attributes as documented in this blog post: https://developers.home-assistant.io/blog/2023/08/07/weather_entity_forecast_types, that means there may not be a state change when the forecast is updated.

PR #31684, opened 2024-03-01
Remove "NOTSET" logging level

rpatterson:
## Proposed change

The `logging.NOTSET` level has special meaning in Python and when used results in confusing behavior. Looking at the previous listing of levels a admin not familiar with Python might reasonably assume setting `homeassistant.components.command_line` to `notset` would get them all possible log messages. In fact, it results in no more log messages and thus less log messages than setting `homeassistant.components.command_line` to `debug`.

Better not to list it as an available level. If it needs to be mentioned at all, it should be a special mention that describes the effect it has.

## Type of change

## Checklist

### DIFF
--- a/source/_integrations/logger.markdown
+++ b/source/_integrations/logger.markdown
@@ -115,7 +115,6 @@ Possible log severity levels, listed in order from most severe to least severe, 
 - warn
 - info
 - debug
-- notset
 
 ### Log filters
 


edenhaus:
From the python docs about [`logging.NONSET`](https://docs.python.org/3/library/logging.html#logging.NOTSET):
> When set on a logger, indicates that ancestor loggers are to be consulted to determine the effective level.

In other words, the parent logger will be consulted.

> In fact, it results in no more log messages and thus less log messages than setting  homeassistant.components.command_line to debug.

This depends on what you have set for the parent loggers and as HA sets the level to info for the root logger, it could happen, if you didn't change the default level or any parent logger.

edenhaus (changes requested):
Instead of removing a valid level, we should add a link to https://docs.python.org/3/library/logging.html#logging-levels

Something like Read more about ... or similar

rpatterson:
> This depends on what you have set for the parent loggers and as HA sets the level to info for the root logger, it could happen, if you didn't change the default level or any parent logger.

Precisely. It's behavior is different and much more complicated to understand than just including it in the same list as the normal levels whose behavior is quite simple and obvious. Also, while this describes the behavior technically it does not describe any common use cases as documentation like this should.

> Instead of removing a valid level, we should add a link to https://docs.python.org/3/library/logging.html#logging-levels

The audience for those docs is Python developers. The audience for these docs are HA admins and maybe HA users, and many, possibly most, of those are not Python developers.

I think removing this item make these docs better than they were before the change. I am unclear on the use case for using the `notset` level, but someone who is could write that and *add* it to these docs. I think that would make them even better, but I don't think waiting for someone to do that should block this easy and minor improvement. IOW, in order of best to worst:

1. Including `notset` and describing it's complicated behavior with use cases
2. Not including `notset`
3. Including `notset` and describing it's complicated behavior but without use cases
4. Including `notset` without describing it's complicated behavior or use cases

It would be a shame to leave the docs at #.4 waiting for #.1 when this simple change is available now.

edenhaus:
@rpatterson Sorry for the delay but somehow I missed the notification of this PR

We should keep the level but describe it as special, and users should read the documentation before using it. I would not describe the behavior instead we should add a link to the python docs (See the link in an earlier comment)

The reason to keep it is that the docs should document the code functionality, where it is possible to set it to `nonset`.

Could you update the PR to says the this level has a special behavior and should only be used after reading about it on the python docs (link above)?

rpatterson:
> Could you update the PR to says the this level has a special behavior and should only be used after reading about it on the python docs (link above)?

Documentation should be oriented towards use cases. From [my previous comment](https://github.com/home-assistant/home-assistant.io/pull/31684#issuecomment-1974128546):

> I am unclear on the use case for using the `notset` level, but someone who is could write that and add it to these docs.

The rest of that comment says all I can think to say about my argument and resulting opinion.

edenhaus:
> Documentation should be oriented towards use cases

The documentation should document the code, and `nonset` is a valid level to set. See https://github.com/home-assistant/core/blob/dev/homeassistant/components/logger/const.py#L19

We don't need to add use cases for each level as we document the code, not use cases.

Drafting the PR as I asked for changes in [my previous comment](https://github.com/home-assistant/home-assistant.io/pull/31684#issuecomment-2008997875)

rpatterson:
Why not just...

PR #31883, opened 2024-03-14
Revert "Update govee_light_local.markdown"

klaasnicolaas:
Reverts home-assistant/home-assistant.io#31877

This is not something we should allow in the docs


c0ffeeca7 (approved):
Thank you, @klaasnicolaas :+1:

PR #31893, opened 2024-03-15
Modbus, configuration without entities

janiversen:
## Proposed change

The modbus integration do not allow a configuration of a device without at least a single entity.

## Type of change

## Additional information

https://github.com/home-assistant/core/pull/113516

## Checklist

### DIFF
--- a/source/_integrations/modbus.markdown
+++ b/source/_integrations/modbus.markdown
@@ -339,7 +339,7 @@ modbus:
 
 modbus entities are grouped below each modbus communication entry.
 
-**REMARK** Each modbus device must have at least 1 entity defined.
+**REMARK** Each modbus device must have at least 1 entity defined, otherwise the integration will not be loaded.
 
 Experience has shown that a modbus configuration without entities leads to instability issues.
 
--- a/source/_integrations/modbus.markdown
+++ b/source/_integrations/modbus.markdown
@@ -341,10 +341,6 @@ modbus entities are grouped below each modbus communication entry.
 
 **REMARK** Each modbus device must have at least 1 entity defined, otherwise the integration will not be loaded.
 
-Experience has shown that a modbus configuration without entities leads to instability issues.
-
-The normal use case for the integration is to have entities, however there are some special cases and custom components that used the modbus integration without entities. The solution to these special cases is to add a dummy entity.
-
 Please refer to [Parameter usage](#parameters-usage-matrix) for conflicting parameters.
 
 All modbus entities have the following parameters:
--- a/source/_integrations/modbus.markdown
+++ b/source/_integrations/modbus.markdown
@@ -339,9 +339,15 @@ modbus:
 
 modbus entities are grouped below each modbus communication entry.
 
-All modbus entities have the following parameters:
+**REMARK** Each modbus device must have at least 1 entity defined.
+
+Experience has shown that a modbus configuration without entities leads to instability issues.
+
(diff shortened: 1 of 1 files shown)


source/_integrations/modbus.markdown
```diff
@@ -339,10 +339,16 @@ modbus:
 
 modbus entities are grouped below each modbus communication entry.
 
-All modbus entities have the following parameters:
+**REMARK** Each modbus device must have at least 1 entity defined.
+
+Experience have shown that a modbus configuration without entities leads to instability issues, which in worst case impact HA negatively.
```
davet2001:
```suggestion
Experience has shown that a modbus configuration without entities leads to instability issues.
```
Grammar correction and suggest removing last part of a sentence which seems redundant.

source/_integrations/modbus.markdown
```diff
@@ -339,10 +339,16 @@ modbus:
...
+
+The normal use case for the integration is to have entities, however there are some special cases and custom_components that used the modbus integration without entities. The solution to these special cases is to add a dummy entity.
```
davet2001:
```suggestion
The normal use case for the integration is to have entities, however there are some special cases and custom components that used the modbus integration without entities. The solution to these special cases is to add a dummy entity.
```

davet2001 (changes requested):

davet2001 (commented):
The text looks good to me now. 

Is there a way we can prevent the configuration from being created without entities, to stop this from occurring rather than relying on documentation which may or may not be read/understood?

janiversen:
The Pr in core, which is already merged, takes care of that.

janiversen:
It is referenced above, if you want to check.

janiversen:
Rebaased to newest Next (seems there was a couple of dependaboot changes, that potentially could affect the PR).

janiversen:
@davet2001 you forgot to approve, of course unless you have additional changes.

davet2001:
Ok understood, good that it's now already blocking the situation you want to avoid.

In that case the documentation could be a bit clearer.

I read it as though the user needed to take special care to avoid a problem.

Perhaps something like:

"The integration requires at least one valid entity for Home Assistant to operate reliably, and will be automatically blocked if none are configured.

For special cases such as use by custom components, please add a dummy entity to allow the integration to load, as prompted in the settings panel."

source/_integrations/modbus.markdown
```diff
@@ -339,10 +339,16 @@ modbus:
 
 modbus entities are grouped below each modbus communication entry.
 
-All modbus entities have the following parameters:
+**REMARK** Each modbus device must have at least 1 entity defined, otherwise the integration will not be loaded.
+
+Experience has shown that a modbus configuration without entities leads to instability issues.
+
+The normal use case for the integration is to have entities, however there are some special cases and custom components that used the modbus integration without entities. The solution to these special cases is to add a dummy entity.
```
davet2001:
```suggestion
The normal use case for the integration is to have entities, however there are some special cases and custom components that use the modbus integration without entities. The solution to these special cases is to add a dummy entity.
```

davet2001 (changes requested):
Looks good now, just spotted one minor typo. Thanks!

source/_integrations/modbus.markdown
janiversen: Removed all the explaining, which really do not bring anything. The remark states clearly it is not allowed, and will prohibit the integration from loading.

davet2001 (approved):

PR #31952, opened 2024-03-20
Update tuya integration document

yuqiufeng:
tell developer how to extend standard function


c0ffeeca7:
Hi there, @yuqiufeng 
Thank you for contributing. The content you're proposing here is developer documentation. However, this documentation page is targeted at end-users. This is not the right place for this kind of content. For this reason, I'm closing this PR.

PR #32087, opened 2024-03-29
Update esphome.markdown

Phalen:
Added note for thos who have advanced networks to quickly realise mdns is in use and that they will need to update their configuration accordingly.

## Type of change

## Checklist

- [ X] This PR uses the correct branch, based on one of the following:
- [ X] The documentation follows the Home Assistant documentation [standards].

### DIFF
--- a/source/_integrations/esphome.markdown
+++ b/source/_integrations/esphome.markdown
@@ -42,6 +42,9 @@ works_with:
 
 This integration allows [ESPHome](https://esphome.io) devices to connect directly to Home Assistant with the [native ESPHome API](https://esphome.io/components/api.html).
 
+
+`Note: This system uses MDNS for ease of use on Home networks. Multiple subnets and vlans will require the use of additional options for devices to show online. Please see the forums for needed additions to configurations to allow devices to be detected.`
+
 {% include integrations/config_flow.md %}
 
 ## Home Assistant service calls


source/_integrations/esphome.markdown
```diff
@@ -42,6 +42,9 @@ works_with:
 
 This integration allows [ESPHome](https://esphome.io) devices to connect directly to Home Assistant with the [native ESPHome API](https://esphome.io/components/api.html).
 
+
+`Note: This system uses MDNS for ease of use on Home networks. Multiple subnets and vlans will require the use of additional options for devices to show online. Please see the forums for needed additions to configurations to allow devices to be detected.`
```
c0ffeeca7, line 46: what do you mean by *show online*? Are you referring to the status shown in the addon here
image

If this is what you mean, I don't think this note is in the right place here. This is the documentation of the integration, not the add-on.

In that case, the information might go into a troubleshooting section here https://esphome.io/guides/getting_started_hassio

frenck, line 46: This is an advanced networks setup, which we generally do not support. If you manage/set up such networks, we will assume you are capable of working with those and knows what that entails.

Additionally, this note is not about Home Assistant, but about the ESPHome Dashboard itself; and thus, IMHO, this is not the right place to add this.

../Frenck

Phalen: yes that is the screen mentioned.  The problem is esphome treats that as gospel and does not pole or update the device unless you know exactly what to change in the config. if you have no idea that its due to a combination of configuration/development choices and the fact that you are not any joe off the street and have more than expcted by the developer it makes it hard for NEW users to Home Assistant to know what to look for to fix their issue.   To me esphome was some crappy slaped together project due to my first impression of  [tasmota installs and works with minimal configuration, esphome does not on the same device.]

Phalen: thats quite the unfrendly attitude to bring towards people just because we all cant just download 20 years of development knowledge of  the esphome project from day one like you have.  i know of DBA's who are quite increadible at building a massive enterprise sql database system and have troubles with turning on the pc they work on or following prompts of software installs.  

This is the most appropriate location as it pertains to Home Assistant users who are looking to add functionality and would most benifit from being aware of the addon's limitations/design choices as it pertains to them and their intrest of adding it to their setup.

frenck: > thats quite the unfrendly attitude

Sorry, it isn't meant like that, nor did I put it like that.

> This is the most appropriate location as it pertains to Home Assistant users who are looking to add functionality and would most benifit from being aware of the addon's limitations/design choices as it pertains to them and their intrest of adding it to their setup.

Sorry, I'm in disagreement, as you are documenting a behavior in an advanced network setup about the ESPHome Dashboard. Not Home Assistant.

> and would most benifit from being aware of the addon's limitations/design choices as it pertains to them and their intrest of adding it to their setup.

It is not an add-on specific limitation. It is a limitation of your network setup, which will occur in many other things, including non-add-on usage of ESPHome.

../Frenck

frenck:
Going to friendly decline this PR, as it aimed to document part of the ESPHome Dashboard, while this page is aimed at being the ESPHome Home Assistant integration documentation.

../Frenck

PR #32093, opened 2024-03-30
Update modbus.markdown

nicelight:
changed parity bit by default in examples

### DIFF
--- a/source/_integrations/modbus.markdown
+++ b/source/_integrations/modbus.markdown
@@ -274,7 +274,7 @@ modbus:
     baudrate: 9600
     bytesize: 8
     method: rtu
-    parity: E
+    parity: N
     stopbits: 1
 ```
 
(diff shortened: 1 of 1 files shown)


source/_integrations/modbus.markdown
```diff
@@ -274,7 +274,7 @@ modbus:
     baudrate: 9600
     bytesize: 8
     method: rtu
-    parity: E
```
janiversen, line 277: Why is 'N' better than 'E' ?

my line is setup with Even parity.

These are just examples.

janiversen (changes requested):
I fail to see why this is better.

source/_integrations/modbus.markdown
nicelight: Ok, its up to you. 
But even in your documentaiont it is written ( N - by default). 
Can you say what modbus devices do you use with E parity?

janiversen: Most devices can configure the parity.

PR #32361, opened 2024-04-17
YAML syntax: add headings and tiny rephrase

c0ffeeca7:
## Proposed change

YAML syntax: 
- add headings
- Move the section on private data to config.yaml page, as it is unrelated to YAML syntax
- tiny rephrase

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/configuration.markdown
+++ b/source/_docs/configuration.markdown
@@ -26,10 +26,6 @@ _If you use {% term "Home Assistant Operating System" %}, you can find `configur
 
 _If you use {% term "Home Assistant Core" %} , you can find `configuration.yaml` in the config folder passed to the `hass` command (default is `~/.homeassistant`)._
 
-### Do not store private data in your configuration.yaml file
-
-Avoid storing private data (passwords, API keys, etc.) directly in your `configuration.yaml` file. Private data can be stored in either a [separate file](/docs/configuration/secrets/) or in [environmental variables](/docs/configuration/yaml/#using-environment-variables), which circumvents this security problem.
-
 ## Reloading configuration changes
 
 Most integrations in Home Assistant that do not interact with {% term devices %} or {% term services %} can reload changes made to their configuration in `configuration.yaml`.
@@ -58,6 +54,7 @@ The method for running a configuration check depends on your [installation type]
 
 - [YAML syntax](/docs/configuration/yaml/)
 - [Configuration check on Operating System](/common-tasks/os/#configuration-check)
+- [Storing credentials in `secrets.yaml` file](/docs/configuration/secrets)
 
 ### Backups
 
--- a/source/_docs/configuration.markdown
+++ b/source/_docs/configuration.markdown
@@ -26,6 +26,10 @@ _If you use {% term "Home Assistant Operating System" %}, you can find `configur
 
 _If you use {% term "Home Assistant Core" %} , you can find `configuration.yaml` in the config folder passed to the `hass` command (default is `~/.homeassistant`)._
 
+### Do not store private data in your configuration.yaml file
+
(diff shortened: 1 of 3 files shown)


source/_docs/configuration/yaml.markdown
```diff
@@ -1,13 +1,17 @@
...
 ---
 
-Home Assistant uses the [YAML](https://yaml.org/) syntax for configuration. YAML might take a while to get used to but is powerful in allowing you to express complex configurations.
+Home Assistant uses the [YAML](https://yaml.org/) syntax for configuration. While most integrations can be configured through the UI, some integrations require you to edit your [`configuration.yaml`](/docs/configuration/) file to specify its settings.
 
-While more and more integrations are configured through the UI, for some, you will add code in your [`configuration.yaml`](/docs/configuration/) file to specify its settings.
+## YAML Style Guide
 
-The following example entry assumes that you would like to set up the [notify integration](/integrations/notify) with the [pushbullet platform](/integrations/pushbullet).
+This page gives a high-level introduction to the YAML syntax used in Home Assistant. For a more detailed description and more examples, refer to the [YAML Style Guide for Home Assistant developers](https://developers.home-assistant.io/docs/documenting/yaml-style-guide/).
```
klaasnicolaas, line 10: Do we want to send end users to our developer website? In the past we kept this strictly separate.

source/_docs/configuration/yaml.markdown
```diff
@@ -120,3 +124,26 @@ Not quoting the value may generate an error such as:
...
+
+With all these indents and rules, it is easy to make a mistake. The best way to check if your YAML syntax is correct (validate) depends on the editor you use. We can't list them all here.
+
+- If you edit the files directly in Home Assistant, refer to the section: [Validating the configuration](/docs/configuration/#validating-the-configuration)
+
+## Related topics
+
+### YAML
+
+- [YAML Style Guide for Home Assistant developers](https://developers.home-assistant.io/docs/documenting/yaml-style-guide/)
```
klaasnicolaas, line 138: Same here

source/_docs/configuration/yaml.markdown
c0ffeeca7: Excellent question. Yes, I agree with that in principle. And I was very hesitant putting that link there... But...

I think the Style Guide is a not actually a technical dev documentation, describing developpery things like implementing an async component. It is a Style Guide like the MS Style Guide, but for YAML. The topics discussed in the YAML Style Guide are topics you are going to be exposed to when you start messing with YAML in HA. The reason we don't want to send people to the dev docs is because the topics there are usually dev topics. But if we have users mess with YAML, this Style Guide gives more context and examples than what we currently have in the user docs. I think not providing that information is not a good option.

The alternative would be creating such a Guide in the User Docs. But I think the result would be a copy of the existing YAML Guide. This is probably a better solution from a user perspective. But it comes with the maintenance cost of duplicate (or similar but different) content.

source/_docs/configuration.markdown
```diff
@@ -26,6 +26,10 @@ _If you use {% term "Home Assistant Operating System" %}, you can find `configur
 
 _If you use {% term "Home Assistant Core" %} , you can find `configuration.yaml` in the config folder passed to the `hass` command (default is `~/.homeassistant`)._
 
+### Do not store private data in your configuration.yaml file
+
+Avoid storing private data (passwords, API keys, etc.) directly in your `configuration.yaml` file. Private data can be stored in either a [separate file](/docs/configuration/secrets/) or in [environmental variables](/docs/configuration/yaml/#using-environment-variables), which circumvents this security problem.
```
frenck: Which security problem?

c0ffeeca7: Oh, good question. I copied this over from the YAML syntax file, because it is not about syntax. probably this is referring to the fact that you share your information by accident when sharing configs. Which is not mentioned here. Let me rephrase...

klaasnicolaas (approved):

PR #32418, opened 2024-04-23
Update ecobee.markdown

CDejoEcobee:
Updated the documentation to indicate that ecobee is not accepting new developer subscriptions

### DIFF
--- a/source/_integrations/ecobee.markdown
+++ b/source/_integrations/ecobee.markdown
@@ -33,6 +33,8 @@ The **ecobee** {% term integration %} lets you control and view sensor data from
 
 You will need to obtain an API key from ecobee's [developer site](https://www.ecobee.com/developers/) to use this integration. To get the key, your thermostat must be registered on ecobee's website (which you likely would have already done while installing your thermostat). Once you have done that, perform the following steps.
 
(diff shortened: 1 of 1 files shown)


source/_integrations/ecobee.markdown
```diff
@@ -33,6 +33,8 @@ The **ecobee** {% term integration %} lets you control and view sensor data from
 
 You will need to obtain an API key from ecobee's [developer site](https://www.ecobee.com/developers/) to use this integration. To get the key, your thermostat must be registered on ecobee's website (which you likely would have already done while installing your thermostat). Once you have done that, perform the following steps.
 
+*As of March 28th, 2024, ecobee is no-longer accepting new developer subscriptions, and there is no ETA for when new developer subscriptions will be allowed, existing develoiper access will not be affected*
```
frenck:
```suggestion
<div class='note warning'>
As of March 28th, 2024, ecobee is no longer accepting new developer subscriptions, and there is no ETA for when they will be allowed again. Existing developers are not affected.
</div>
```

frenck (approved):
Thanks, @CDejoEcobee 👍 

../Frenck

PR #32445, opened 2024-04-24
General settings: replace section on reloading config by link

c0ffeeca7:
## Proposed change

General settings: replace section on reloading config 
- by a link to procedure on reloading the configuration to apply changes
- add a reference to Home Assistant Core integration
- Update the referenced reload procedure on the configuration.yaml page

## Type of change

## Additional information

## Checklist


source/_docs/configuration.markdown
```diff
@@ -58,3 +45,20 @@ The method for running a configuration check depends on your [installation type]
...
 - [Configuration check on Container](/common-tasks/container/#configuration-check)
 - [Configuration check on Core](/common-tasks/core/#configuration-check)
+
+## Reloading the configuration to apply changes
+
+For configuration changes to become effective, the configuration must be reloaded. Most integrations in Home Assistant (that do not interact with {% term devices %} or {% term services %}) can reload changes made to their configuration in `configuration.yaml` without needing to restart Home Assistant.
+
+1. There are two main ways to reload configuration changes.
+   - Under **Settings**, select the three dots menu (top right), select **Restart Home Assistant** > **Quick reload**.
+2. Another way to reload the configuration is by using the **Developer Tools**.
```
frenck: Should we even have this in our docs? As in, the dev tools is not really meant for day2day end-users?

source/_docs/configuration.markdown
```diff
@@ -35,19 +35,6 @@ _If you use {% term "Home Assistant Operating System" %}, you can find `configur
...
-
-    ![Reload configuration changes](/images/docs/configuration/reloading_config.png)
-
-2. If the integration is listed, select it to reload the settings.
-3. If integration is not listed, you need to restart Home Assistant for changes to take effect:
-   - [Validate the configuration](#validating-the-configuration). Then, select the **Restart** button.
-
 ## Validating the configuration
 
 After changing configuration or automation files, check if the configuration is valid.
```
frenck: This part is actually no longer valid and/or needed. Reload/restart of Home Assistant will already run validation.

source/_docs/configuration.markdown
c0ffeeca7: Yes, I was wondering... the reason I left it is because I thought someone put it there for a reason. I'm not sure how many people use this.

source/_docs/configuration.markdown
c0ffeeca7: I see. 
we refer to this procedure from this page https://www.home-assistant.io/docs/configuration/yaml/
image

this section ould probably be removed as well?

source/_docs/configuration.markdown
frenck: How many people still use this: I don't know either..

I think the gist is: They shouldn't need to use it anymore. There basically isn't really a reason to use it I guess

source/_docs/configuration.markdown
frenck: Yeah I guess that makes sense. Back in the day, you had to check before restarting (to ensure HA would start). This safeguard is nowadays just part of the restart/reload itself (and above all, Home Assistant will **always** start now).

source/_docs/configuration.markdown
klaasnicolaas: I always use option 2 😄

I'm not in favor of removing it from the docs, it's an existing option and the end-user can decide which route they use. Also the `Quick reload` button from option 1 is hidden under the `Restart Home Assistant` button. A user will initially not expect a reload option to appear after pressing the restart button

image

source/_docs/configuration.markdown
klaasnicolaas: But what if you just want to validate without immediately restarting? I would not delete this section, but rewrite it with text that validation always happens during a restart.

Another option is to create a YAML section here: https://www.home-assistant.io/docs/tools/dev-tools/ and move this validation to here?

c0ffeeca7: another question: maybe we have a bit of a Matryoshka doll situation here.
image
The Validation section points to the configuration check procedures for the different installation types.
image
for Container
image

Are they all no longer needed?

source/_docs/configuration.markdown
c0ffeeca7: I'm in favor of removing the **Developper tools** option from this specific procedure.
I think documentation should in general follow a minimalist approach. If there are multiple ways to get somewhere, and they have the same result, within that procedure, document one way only. (The reason I left both in was because I wasn't entirely sure if they are identical). 

This would give us one simple procedure we can link to from everywhere else where reload is needed.

Having said that I think the other option should be explained in the docs on the dev tools (as discussed with klaasnicolaas offline). If users stumble across that button and wonder what it does and how it differs from the **quick reload**, they can look it up there.

c0ffeeca7: Let me rework this a bit...

c0ffeeca7: Moving those steps to Dev tools page: https://github.com/home-assistant/home-assistant.io/pull/32449, in preparation for removing them from this section here.

source/_docs/configuration.markdown
ludeeus: It is not needed for before restarting, but its needed for a validation check when you are creating things.

frenck (approved):
Thanks, @c0ffeeca7 👍 

../Frenck

c0ffeeca7:
thanks!

PR #32461, opened 2024-04-25
Revert "Document group state changes when in cases with one specific state"

jbouwh:
Reverts home-assistant/home-assistant.io#32443

Parent PR was reverted


c0ffeeca7 (approved):
Thank you, @jbouwh :+1:

PR #32564, opened 2024-05-02
Revert "Remove references to UniFi WLAN password sensor"

frenck:
Reverts home-assistant/home-assistant.io#32556

As this is a generated file/git commit history.

We can't change history :)

../Frenck


c0ffeeca7 (approved):
Thank you, @frenck :+1:

PR #32652, opened 2024-05-08
Resolve terminology links causing new table rows

RosemaryOrchard:
Related to PR #32640, this will prevent the terminology links from causing new table rows which was done because they add  a `<br>`. By making them `<small>` instead, they'll stay inline but not be too in the way.

Before:
CleanShot 2024-05-08 at 17 51 29

After:
CleanShot 2024-05-08 at 17 51 01


c0ffeeca7:
This did not work for me. It is smaller, but the issue persists.
image

frenck (approved):
Thanks, @RosemaryOrchard 👍 

../Frenck

PR #33241, opened 2024-06-12
Improve Generic x86-64 Installation Documentation

RosemaryOrchard:
## Proposed change
Generic x86-64 method 2 has a warning to do method 1 first, but it wasn't highlighted as a warning.
It was also missing the extract step which made it fail.

## Type of change

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Added a warning note in Method 2 of HAOS installation, advising users to use this method only if Method 1 fails.
  - Included steps for extracting the archive before flashing the image.
  - Renumbered the subsequent steps for better clarity.

Here is the changed page on the netifly build: https://deploy-preview-33241--home-assistant-docs.netlify.app/installation/generic-x86-64#method-2-installing-haos-directly-from-a-boot-medium

### DIFF
--- a/source/_includes/installation/operating_system.md
+++ b/source/_includes/installation/operating_system.md
@@ -140,7 +140,9 @@ To write the HAOS image to the boot medium on your x86-64 hardware, there are 2 
 
 ### Method 2: Installing HAOS directly from a boot medium
 
-<div class='note warning'>Use this method only if Method 1 does not work for you.</div>
+<div class='note warning'>
+Use this method only if Method 1 does not work for you.
+</div>
 
 #### Required material
 
@@ -205,8 +207,8 @@ To write the HAOS image to the boot medium on your x86-64 hardware, there are 2 
 *Select and copy the URL or use the "copy" button that appear when you hover it.*
 
 5. Paste the URL into your browser to start the download.
-6. Extract the archive.
-7. Select **Flash from file** and select the image you just downloaded.
+6. Extract the file you just downloaded.
+7. Select **Flash from file** and select the image you just extracted.
    - Do not use **Flash from URL**. It does not work on some systems.
 
   ![Screenshot of the Etcher software showing flash from URL selected.](/images/installation/etcher1_file.png)
--- a/source/_includes/installation/operating_system.md
+++ b/source/_includes/installation/operating_system.md
@@ -140,7 +140,7 @@ To write the HAOS image to the boot medium on your x86-64 hardware, there are 2 
 
 ### Method 2: Installing HAOS directly from a boot medium
 
-Use this method only if Method 1 does not work for you.
+<div class='note warning'>Use this method only if Method 1 does not work for you.</div>
 
 #### Required material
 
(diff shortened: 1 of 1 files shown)


source/_includes/installation/operating_system.md
```diff
@@ -205,15 +205,16 @@ Use this method only if Method 1 does not work for you.
 *Select and copy the URL or use the "copy" button that appear when you hover it.*
 
 5. Paste the URL into your browser to start the download.
-6. Select **Flash from file** and select the image you just downloaded.
+6. Extract the archive.
```
frenck: This should not be needed.

ludeeus: True, but it is.

frenck: How is that? As in, we originally used URLs, and that one is broken in combination with `xz` files (and, unfortunately, still an open issue upstream). As a result of that issue, we have tested across multiple platforms on what the best solution would be.

The above-listed procedure is a result of that. So, wondering, if this addition is now really needed; what changed? Was this intentional? Or is it an issue? In the case of the latter, can we link it up with a known upstream issue, or do we need to create one?

frenck: /CC @agners I think we should reconsider the release on the OS-level and maybe publish in gzip format as well; as all of these issues, are likely (at least one confirmed one) are fed by `xz`.

agners: Yeah unfortunately it is currently required when using the latest version of balenaEtcher :cry: 

There is a bug in balenaEtcher, see https://github.com/balena-io/etcher/issues/4064. The project appears quite stale actually :cry: . At least at one point we were start considering alternatives.

I thought at one point we recommended to use an older version (@c0ffeeca7, do you remember?)

agners: Ah, we changed to install from file for that reason, see: #28624.

@ludeeus so you tried to install from file and that failed too for you now? What happened exactly?

agners: Oh I see:

image

Great, they broke that too. :see_no_evil: 

I'd rather prefer to move away from balenaEtcher. It will only be a question of time when gz also breaks, either from URL or file.

ludeeus: I did not try it personally, but based on reports by others and the last comment (from 2 weeks ago) on the previously opened issue, made it seem like this is now an issue.
<https://github.com/balena-io/etcher/issues/4064#issuecomment-2143419306>

agners: Actually, gz is currently broken too:

image

I'd say probably best to just merge this for now? :thinking:

agners (dismissed):

source/_includes/installation/operating_system.md
ludeeus: > I'd say probably best to just merge this for now?

I agree, better to have working documentation.

source/_includes/installation/operating_system.md
```diff
@@ -140,7 +140,7 @@ To write the HAOS image to the boot medium on your x86-64 hardware, there are 2
 
 ### Method 2: Installing HAOS directly from a boot medium
 
-Use this method only if Method 1 does not work for you.
+<div class='note warning'>Use this method only if Method 1 does not work for you.</div>
```
ludeeus:
```suggestion
<div class='note warning'>
Use this method only if Method 1 does not work for you.
</div>
```

source/_includes/installation/operating_system.md
```diff
@@ -205,15 +205,16 @@ Use this method only if Method 1 does not work for you.
 *Select and copy the URL or use the "copy" button that appear when you hover it.*
 
 5. Paste the URL into your browser to start the download.
-6. Select **Flash from file** and select the image you just downloaded.
+6. Extract the archive.
+7. Select **Flash from file** and select the image you just downloaded.
```
ludeeus:
```suggestion
7. Select **Flash from file** and select the image you just extracted.
```

source/_includes/installation/operating_system.md
ludeeus:
```suggestion
6. Extract the file you just downloaded.
```

ludeeus (commented):
Some small suggestions

source/_includes/installation/operating_system.md
frenck: Agree, thanks for the additional context here @agners 👍 

../Frenck

frenck (approved):
Thanks, @RosemaryOrchard 👍 

And thanks @ludeeus & @agners for chipping in here as well!

../Frenck

PR #33379, opened 2024-06-23
clarify prometheus supported domains

llamafilm:
## Proposed change
Clarify that Prometheus metrics only include certain domains, regardless of filter configuration.

## Type of change

## Additional information
- This PR fixes or closes issue: https://github.com/home-assistant/core/issues/119803

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Added a note in the Prometheus section indicating that metrics are exported only for specific domains.
  - Included a list of supported metrics for specific domains in the Prometheus integration documentation.

### DIFF
--- a/source/_integrations/prometheus.markdown
+++ b/source/_integrations/prometheus.markdown
@@ -122,6 +122,9 @@ prometheus:
 
 {% include common-tasks/filters.md %}
 
+> [!TIP]
+> Metrics are exported only for certain domains. To find out which ones are supported you can search for `def _handle_` in the [source](https://github.com/home-assistant/core/blob/dev/homeassistant/components/prometheus/__init__.py).
+
 ## Full Example
 
 Advanced configuration example:
--- a/source/_integrations/prometheus.markdown
+++ b/source/_integrations/prometheus.markdown
@@ -209,4 +209,4 @@ For example:
 
 Metrics are exported only for the following domains:
 
-`automation`, `binary_sensor`, `climate`, `cover`, `counter`, `device_tracker`, `fan`, `humidifier`, `input_boolean`, `input_number`, `light`, `lock`, `number`, `person`, `sensor`, `update`, `switch`, `zwave`
+`automation`, `binary_sensor`, `climate`, `cover`, `counter`, `device_tracker`, `fan`, `humidifier`, `input_boolean`, `input_number`, `light`, `lock`, `number`, `person`, `sensor`, `update`, `switch`
--- a/source/_integrations/prometheus.markdown
+++ b/source/_integrations/prometheus.markdown
@@ -122,8 +122,9 @@ prometheus:
 
 {% include common-tasks/filters.md %}
 
-> [!TIP]
-> Metrics are exported only for certain domains. To find out which ones are supported you can search for `def _handle_` in the [source](https://github.com/home-assistant/core/blob/dev/homeassistant/components/prometheus/__init__.py).
+<div class='note'>
+Metrics are exported only for certain domains. For details, see the `def _handle_` functions in the [Prometheus integration source code](https://github.com/home-assistant/core/blob/dev/homeassistant/components/prometheus/__init__.py).
+</div>
 
 ## Full Example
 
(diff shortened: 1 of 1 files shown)


source/_integrations/prometheus.markdown
```diff
@@ -122,6 +122,8 @@ prometheus:
 
 {% include common-tasks/filters.md %}
 
+> Note: Metrics are exported only for certain domains. For details, see the `def _handle_` functions in the [Prometheus integration source code](https://github.com/home-assistant/core/blob/dev/homeassistant/components/prometheus/__init__.py).
```
frenck: - These are GitHub syntax notes, which isn't supported by our website. Please make sure you test a PR before submitting it.
- We should send end-users into code; if there is information on the linked code that is relevant, we should add it here in normal text.

frenck (changes requested):
Hi there @llamafilm 👋 

Thanks for the PR! I've reviewed it, and left a comment. Could you take a look? Thanks! 👍 

../Frenck

source/_integrations/prometheus.markdown
llamafilm: ok, done

llamafilm: I've removed the code links and netlify preview looks good now

source/_integrations/prometheus.markdown
```diff
@@ -204,3 +204,9 @@ For example:
 - record: "known_temperature_c"
   expr: "temperature_c unless entity_available == 0"
 ```
+
+## Supported metrics
+
+Metrics are exported only for the following domains:
+
+`automation`, `binary_sensor`, `climate`, `cover`, `counter`, `device_tracker`, `fan`, `humidifier`, `input_boolean`, `input_number`, `light`, `lock`, `number`, `person`, `sensor`, `update`, `switch`, `zwave`
```
knyar: This certainly helps set expectations, but I worry it will become a maintenance burden. Now every time someone adds support for another domain, they will need to update the documentation page.

Also, support for a particular domain is not binary: for example, we already had metrics for the `climate` domain, but [your recent PR](https://github.com/home-assistant/core/pull/120267/) adds more (thank you!).

I wonder if there is a way to communicate the fact that `prometheus` integration does not provide automatic instrumentation of all entities and state attributes without hard-coding a list of supported domains here.

frenck: `zwave` is not an entity component or entity component platform? How can it be in this list?

frenck (changes requested):

source/_integrations/prometheus.markdown
llamafilm: I don't have any zwave devices so I'm not sure how they work.  I just made a list of all `_handle_` functions based on the code here:
https://github.com/home-assistant/core/blob/e2f9ba5455712c3f9ecc8c4b0a5356b06bf2ce01/homeassistant/components/prometheus/__init__.py#L205

There is a function for zwave:

https://github.com/home-assistant/core/blob/e2f9ba5455712c3f9ecc8c4b0a5356b06bf2ce01/homeassistant/components/prometheus/__init__.py#L687

llamafilm: I agree it's too much of a maintenance burden.  I would prefer the following note instead, but @frenck asked me to remove the link to code.

> Note: Metrics are exported only for certain domains. For details, see the `def _handle_` functions in the [Prometheus integration source code](https://github.com/home-assistant/core/blob/dev/homeassistant/components/prometheus/__init__.py).

frenck: To me, that sounds like a blast from the past that is a leftover from the old open Z-Wave that Home Assistant used to use many moons ago.

llamafilm: Ok, then I'll just remove it from this list.  I suppose it doesn't hurt to leave it in the code for backwards compatibility.

frenck: > Ok, then I'll just remove it from this list. 

👍 

>  I suppose it doesn't hurt to leave it in the code for backwards compatibility.

It doesn't hurt, but the code doesn't provide any backward compatibility, as all traces of the old Zwave integration have been removed from the codebase already. So, to me... it is like dead code 😄

frenck (approved):
Thanks, @llamafilm 👍 

../Frenck

PR #34168, opened 2024-08-07
Use presence detection in automations

c0ffeeca7:
## Proposed change

Add tutorial page to getting started
- using presence detection in automations
- add tutorial on presence detection as condition
- presence detection as trigger is topic of separate PR

## Type of change

## Additional information

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a new navigation link for "Presence detection in automations" under the "Getting started" section for easier access.
  - Introduced a comprehensive guide on implementing presence detection as a condition in automation setups, enhancing practical application.
  
- **Content Updates**
  - Updated an existing markdown link to direct users to relevant content about integrating presence detection into automations.


frenck (commented):
Isn't the more interesting use case here to also trigger when you get home and the lights turn on?

c0ffeeca7:
> Isn't the more interesting use case here to also trigger when you get home and the lights turn on?

- yes, topic for another PR
- point here is to work with the sunset automation, which we already created. and where it would make sense to run it only when someone is home.

frenck:
> yes, topic for another PR

👍 

> point here is to work with the sunset automation, which we already created. and where it would make sense to run it only when someone is home.

Don't agree with that statement. The automation is set up in an incorrect/inflexible way, making it hard to do now.

From an automation perspective, this automation is a wrong practice from the get-go (which happens a lot in our community, I don't blame them, as our triggers contain to much "hidden condition logic" in them).

In general, automation should trigger things, check conditions, and do actions.

Right now, the automation has a mix-up of the sunset trigger that is "conditionally", making it hard to add more triggers as it will mess up stuff.

A better approach, in general, is to trigger "sun" changes and have sunset as an actual condition already. Adding a presence condition and trigger at that point, will make the automation work in the case set here, but also when arriving home (as it will trigger an check the conditions).

c0ffeeca7:
> > yes, topic for another PR
> 
> 👍
> 
> > point here is to work with the sunset automation, which we already created. and where it would make sense to run it only when someone is home.
> 
> Don't agree with that statement. The automation is set up in an incorrect/inflexible way, making it hard to do now.
> 
> From an automation perspective, this automation is a wrong practice from the get-go (which happens a lot in our community, I don't blame them, as our triggers contain to much "hidden condition logic" in them).
> 
> In general, automation should trigger things, check conditions, and do actions.
> 
> Right now, the automation has a mix-up of the sunset trigger that is "conditionally", making it hard to add more triggers as it will mess up stuff.
> 
> A better approach, in general, is to trigger "sun" changes and have sunset as an actual condition already. Adding a presence condition and trigger at that point, will make the automation work in the case set here, but also when arriving home (as it will trigger an check the conditions).

Thank you for the detailed explanation. Let me rework it.

PR #34336, opened 2024-08-17
Count sensor documentation improvements

mekaneck:
## Proposed change

- Clarify the count sensor doesn't count rising edges
- Add a suggestion to change the `state_class` for count sensors which have non-sliding time windows
- Document how often the history_stats sensor is updated

## Type of change

## Additional information

## Checklist

## Summary by CodeRabbit

- **Documentation**
	- Enhanced clarity of the `history_stats` integration documentation.
	- Updated description of the `count` attribute to better distinguish its behavior, including initial state considerations.
	- Added guidance on using the `state_class` attribute for count sensors to track cumulative statistics.
	- Clarified update frequency of the history stats sensor, noting changes and timing for fixed durations.


source/_integrations/history_stats.markdown
```diff
@@ -121,7 +121,11 @@ Depending on the sensor type you choose, the `history_stats` integration can sho
 
 - **time**: The default value, which is the tracked time, in hours
 - **ratio**: The tracked time divided by the length of your period, as a percentage
-- **count**: How many times the tracked entity matched the configured state during the time period
+- **count**: How many times the tracked entity matched the configured state during the time period. Unlike a rising edge counter, this will also include a count if the entity is already in the matched state at the beginning of the time period.
```
c0ffeeca7: I'm not a microcontroller developer and I'm not sure rising edge counter is a commonly understood term.
The current description states that the counter only increases if the actual state is matched within the time period. It doesn't say anywhere that we're counting falling or rising edges. So I'm not sure why one would assume we're counting edges here. To me, this addition is confusing. Maybe you can add an example instead that illustrate how it counts?

source/_integrations/history_stats.markdown
```diff
@@ -121,7 +121,11 @@ Depending on the sensor type you choose, the `history_stats` integration can sho
...
+
+{% note %}
+For a count sensor with a time period that does not overlap (such as one that resets hourly, daily, or monthly), consider using [customization](/docs/configuration/customizing-devices/#customizing-an-entity-in-yaml) to change the `state_class` to `total_increasing` to generate useful statistics.
```
c0ffeeca7: `For a count sensor with a time period that does not overlap`
 this references a target entity (as defined by the entity_id of this sensor?  What would be an example of that? A step counter like FitBit?
`that does not overlap`
 overlap with what?

`to change the state_class to total_increasing to generate useful statistics`
is this referring to the state_class of this (history_stat) entity ?

c0ffeeca7 (changes requested):

source/_integrations/history_stats.markdown
mekaneck: > `For a count sensor with a time period that does not overlap`
this references a target entity (as defined by the entity_id of this sensor? What would be an example of that? A step counter like FitBit?

This note applies to any history_stats sensor that is set to `type: count`. Which would be anything you want to count, like how many times a light is `on` in a day. 

> `that does not overlap`
overlap with what?

I will work on how to clarify this. I meant it doesn't overlap with itself, meaning it isn't a sliding time window (e.g. a trailing 60 minute window). It would instead be a time window whose duration increases while the start time is fixed, and once the duration reaches a maximum value the start time will jump, and the cycle repeats. For example, a time window that starts at `today_at('0:00:00')` and ends at `now()`. Essentially, any type of window that would result in a sensor that acts like a utility_meter with a specified reset cycle. 

> `to change the state_class to total_increasing to generate useful statistics`
is this referring to the state_class of this (history_stat) entity ?

Yes. A shortcoming with this integration is that it cannot detect how the user set up the time window and therefore it cannot set the `state_class` of the new sensor accordingly. So I believe that the documentation should make the user aware of this so the user can take action themselves. The integration will always create the sensor with `state_class: measurement` but often the user would be better off with `state_class: total_increasing`

source/_integrations/history_stats.markdown
mekaneck: Yes I'll think about how best to put an example here. For wording, instead of `rising edge` I could use `event` but that word means something else in HA. Perhaps:
> Unlike counting transitions to the state, this will also include a count if the entity is already in the matched state at the beginning of the time period

I could then show a history graph of a light that was already 'on' at the start of the window, and show that the states counter would calculate e.g. 4 while the number of times the light got _turned on_ during the window was only e.g. 3. 

The point of all this is that, quite often, people want to know how many times something was turned on or turned off, and this `count` feature in history_stats doesn't actually count that. Instead, this feature counts how many times something was in that state. The fact that there is a difference between those two counting methods isn't obvious unless it's pointed out.

c0ffeeca7: Thank you for clarifying. I find your explanation very helpful.

mekaneck: I didn't include a history graph image but hopefully the new text is clear enough.

source/_integrations/history_stats.markdown
mekaneck: Reworded the description of the time window to be more clear.

c0ffeeca7 (approved):
Thank you, @mekaneck :+1:

Issue #33603, opened 2024-07-07
Install docs do not match what is seen when installing. 


kurtjcu:
### Feedback

Actual install requires country, tuya Iot Access ID, Tuya IOT Access Secret, account and password.

### URL

https://www.home-assistant.io/integrations/tuya/

### Version

2024.7.1

Issue #33565, opened 2024-07-04
The Tuya integration in home assistant does not ask me for the user code and instead asks for Tuya credentials which the docs don’t show how to find


Bilalm354:
### Feedback

The Tuya integration in home assistant does not ask me for the user code and instead asks for Tuya credentials which the docs don’t show how to find

![IMG_1131](https://github.com/home-assistant/home-assistant.io/assets/53113053/4f6d9ef2-66b9-4221-aeb0-d45d92a829cd)

### URL

https://www.home-assistant.io/integrations/tuya/

### Version

2024.7.0

kurtjcu:
Same login here as well

KEMBL:
Users who have an old version of Home assistant (from 2023 year) also will see that 'Tuya IoT Access ID', if so - try to update HA version

Issue #32243, opened 2024-04-10
Tuya Insstuction seems incomplete


EdFenner:
### Feedback

Step 4 syas: At the bottom, User Code will be shown; you need to when setting up this integration.
This is correct. But when you configure it, none of the fields are User Code.
Instead you need. Tuya Iot Access ID, Tuya Iot Access Secret, Account (maybe this is email?) There is no instruction on these 3 requirements. What they are, or where to find them.

### URL

https://www.home-assistant.io/integrations/tuya/

### Version

2024.4.2

GlennChubak:
This needs some attention as there is no clear way to use the integration by following the instructions.

dirkesquire:
Me too. Why does the Help say "enter the code shown on the User Code field. " when the HomeAssistant screen is actually asking for Access ID, Access Secret, Account password? It doesn't match. The text on the Tuya integration and the Tuya Integration help seem completely out of date

dirkesquire:
So has the out of date documentation / UI been fixed in #31663 and #34350 or not, I can't really tell. Or are we stuck with out of date UI?

Issue #32009, opened 2024-03-23
Tuya docs wrong 


TimMensch:
### Feedback

The Tuya docs page says all you need is a user code. The actual "add device" page for Tuya asks for an access ID and secret.

I managed to figure out how to do this at one point, but I can't for the life of me find it again. So now I can't add a new device and I have no idea how to add a device! 

### URL

https://www.home-assistant.io/integrations/tuya/

### Version

2024.3.3

### Additional information

I'm on 2024.1.2/Docker.

TimMensch:
Fun times. The Tuya site doesn't let you see the access ID/Client ID if you're on a mobile browser (!!!).

Found the codes, but the instructions are still wrong. And the Tuya login doesn't work either. Maybe it's the Smartlife login!?

TimMensch:
Yes, the Smartlife login. It really should be more clear.

frenck:
> Tuya asks for an access ID and secret.

This can mean either of two things:

- You are running an older version of Home Assistant. Please upgrade, as this changed a few months ago.
- You are not using the built-in version of the Tuya integration but installed it as a custom integration. In that case, this documentation doesn't apply. Please consult the custom integration documentation.

../Frenck

PR #34350, opened 2024-08-18
Updates tuya.markdown

rafpato:
Adds important section to help users who are using an older version of HA and see different configuration options when setting up Tuya integration

## Proposed change

An update to the Tuya integration has left some users puzzled since the documentation's configuration is only present in recent versions of Home Assistant + Tuya integration (unsure where the breaking change was introduced).
You can see this confusion in the [documentation feedback](https://github.com/home-assistant/home-assistant.io/issues?q=is:issue+is:open+sort:updated-desc+label:%22integration:+tuya%22). 

With this PR, we'll add a callout that helps users identify this easily and suggests an update

## Type of change

## Additional information

  * #32243
  * #33565
  * #33603  

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added an informational note in the Tuya integration documentation to warn users about outdated versions of Home Assistant and encourage updates for optimal functionality.


klaasnicolaas:
I think it is not always because people use an outdated version of Home Assistant, but mainly because they use a Tuya integration from HACS, that still uses the old method.

./Klaas

rafpato:
> I think it is not always because people use an outdated version of Home Assistant, but mainly because they use a Tuya integration from HACS, that still uses the old method.
> 
> ./Klaas

That can be true, but this PR applies to the Home Assistant integration and not the HACS one.
I don't think anyone will get to these docs if they're advanced users using HACS.

klaasnicolaas:
Yes, but users confuse these and if a custom integration also uses the same integration domain/brand, the custom integration is loaded over that of core. In my opinion that is mainly the case and not that someone is a few versions behind with HA.

So what I'm mainly trying to say is that your current PR change actually only provides the solution to part of the problem that people encounter with Tuya.

Reference: https://github.com/home-assistant/home-assistant.io/issues/32009#issuecomment-2016437127

What do you think about it @frenck?

rafpato:
> Yes, but users confuse these and if a custom integration also uses the same integration domain/brand, the custom integration is loaded over that of core. In my opinion that is mainly the case and not that someone is a few versions behind with HA.
> 
> So what I'm mainly trying to say is that your current PR change actually only provides the solution to part of the problem that people encounter with Tuya.
> 
> Reference: [#32009 (comment)](https://github.com/home-assistant/home-assistant.io/issues/32009#issuecomment-2016437127)
> 
> What do you think about it @frenck?

I don't think HACS integrations override oficial integrations from Home Assistant, here's a screenshot showing how they show up as their own integration/brand. 
image

I'm also not sure if we should mention HACS in the docs like that since it's an advanced user topic that can confuse users but maybe it's not a bad idea.

I'll let @frenck give his thoughts on this PR but I'm happy with any solution that can help users understand what's wrong and fast track their troubleshooting 🤗

source/_integrations/tuya.markdown
```diff
@@ -67,6 +67,8 @@ Tuya Smart app. You can find it by opening the app and:
 
 {% include integrations/config_flow.md %}
 
+{% important %} If the integration does not ask you for a User Code but instead asks you for Tuya IoT credentials, then you're most likely using an outdated version of Home Assistant. Please update your instance to use the latest Tuya integration. {% endimportant %}
```
frenck, line 70: Our documentation always reflects the current stable version of Home Assistant.

Essentially, you could write this warning on every line of every single piece of documentation of Home Assistant.

Additionally, one can run custom integration. More often, these are results of people overriding the internals of Home Assistant that way (which is already unsupported).

For the above reasons, I'm going to decline this PR.

../Frenck

rafpato: Fair enough, I just found it easier to put a warning since there was a massive change in the integration configuration. I found myself not understanding what was wrong until I got to the documentation feedback that has 3 issues related to the point I'm making. 

I don't run a custom integration, I just don't have a way to automatically update Home Assistant when every stable version comes out which might as well be most of the cases imho.

But all good, thanks for looking at it 🤗

PR #35162, opened 2024-10-10
Add signing instructions to Tesla Fleet

Bre77:
## Proposed change

Add signing instructions to Tesla Fleet

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/128407

## Checklist

## Summary by CodeRabbit

- **Documentation**
	- Updated instructions for the Tesla Fleet integration, clarifying command signing requirements for vehicles manufactured since late 2023.
	- Added detailed steps for generating a key pair and installing the public key via Bluetooth.
	- Included notes on error handling if the signing key is not added to the vehicle.

### DIFF
--- a/source/_integrations/tesla_fleet.markdown
+++ b/source/_integrations/tesla_fleet.markdown
@@ -64,7 +64,7 @@ Tesla restricts open-source integrations to the ["Discovery" plan](https://devel
 
 ## Command signing
 
-Certain vehicles, including all vehicles manufactured since late 2023, require vehicle commands to be signed with a private key. You will need to use Tesla's [command line tools](https://github.com/teslamotors/vehicle-command/blob/main/README.md#installation-and-configuration) to generate a key pair and install the public key on your vehicle.
+Certain vehicles, including all vehicles manufactured since late 2023, require vehicle commands to be signed with a private key. You will need to use Tesla's [command line tools](https://github.com/teslamotors/vehicle-command/blob/main/README.md#installation-and-configuration) to generate a key pair and install the public key on your vehicle using Bluetooth.
 
 ```
 tesla-keygen -key-file tesla_fleet.key create > tesla_fleet.pem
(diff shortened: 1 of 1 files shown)


source/_integrations/tesla_fleet.markdown
```diff
@@ -64,7 +64,14 @@ Tesla restricts open-source integrations to the ["Discovery" plan](https://devel
 
 ## Command signing
 
-Certain vehicles, including all vehicles manufactured since late 2023, require vehicle commands to be encrypted end-to-end and signed with a private key. The Tesla Fleet integration is unable to perform this encryption at this time, so certain features may be disabled or throw an exception when used.
+Certain vehicles, including all vehicles manufactured since late 2023, require vehicle commands to be signed with a private key. You will need to use Tesla's [command line tools](https://github.com/teslamotors/vehicle-command/blob/main/README.md#installation-and-configuration) to generate a key pair and install the public key on your vehicle using Bluetooth.
```
gjohansson-ST: Somehow here it should be mentioned what happens if the user does not perform these steps e.g. the vehicle requires signing but it's not performed.
Can it somehow be clarified what remains of the integrations in that case.

Bre77: I've added a line to try make this clearer.

I'm open to any suggestions on how to clearly communicate this:

- if the vehicle doesn't require signing, everything will work
- if the vehicle does require signing, and the key is installed, everything will work.
- if the vehicle does require signing, and the key isn't installed, all commands/actions/services will fail, but all the read-only/states/data will work.

gjohansson-ST: @c0ffeeca7 any help from your side. I think the last change was pretty good but maybe it can be clarified/highlighted better?
I think point 1 and 2 are given, just number 3 needs to be highlighted so the user knows the limitations of not providing this key.

c0ffeeca7: Just adding the third line looks good to me.
would it be worth mentioning that the corresponding private key is used to unlock these features, and that this private key is stored on the key key of each driver, or on their phone or wherever it is kept? Or do we assume that people who are able to do this know all about this public/private key pair encryption?

frenck (approved):
Thanks, @Bre77 👍 

../Frenck

PR #35573, opened 2024-11-05
Update nest.markdown to improve troubleshooting for *Something went wrong....* case

Zeugma4242:
## Proposed change

Added troubleshooting guidance for issues caused by having multiple Google accounts logged in during Google Nest integration setup. This update helps prevent errors due to unintended account switching, which often occurs if the configured account is not the default in the browser session.

## Type of change

## Additional information

## Checklist

## Summary by CodeRabbit

- **Documentation**
	- Expanded troubleshooting section for Google Nest integration, including guidance on OAuth authentication and account linking.
	- Added clarification on issues with multiple Google accounts in the same browser session.
	- Enhanced instructions for OAuth Client ID consistency between Home Assistant and Google Cloud Console.
	- Reiterated the importance of setting the OAuth Consent Screen to "Production" to avoid reauthentication issues.
	- Updated explanations for common error messages with clearer resolution steps.


allenporter (changes requested):
Thank you for reporting this. 

I'm assuming the problem you are talking about when the account in the device access console is different from the account used in the google cloud console. Is that right?

If so, I would rather tell users to select the same account using the account selection drop down in the corner. I don't want to tell them to logout.

Zeugma4242:
Actually, it does not, that was my case and I reproduced it twice. If your default Google account is say accounta@gmail.com and a second one accountb@gmail.com, and you’re adding the nest integration for accountb, two things are happening: 1. Google from one page to another will constantly pester their user to switch back to accounta while following the procedure. From one step to another, I had to « confirm it’s me » with the wrong account and switch back to b. Painful but no problem. However, on the very last step, I was not getting and confirmation screen but directly the error message from that PR. 
I search for quite some time to understand the issue, even found this troubleshooting section because of that. Out of options I decided to redo everything in a browser session with one account and that worked. I deleted everything and reproduced successfully the problem again with two accounts. (I haven’t tried with accountb marked as default in the browser session and I believe there would be no issue that way. )

allenporter:
I think we should get more specific here. I can't recognize the specific page you are referring to based on the short snippet. Which specific page sends you to the wrong place?  I'd like to get to the bottom of the issue either way and work with google to fix.

Zeugma4242:
I understand what you mean. Let me try to recap and please let me know if this is not enough info:

Context: the browser used to add the integration has at least 2 google accounts logged in. The one linked to Nest IS NOT the default one but a secondary one
Effect: throughout the entire setup process described by this documentation, Google constantly tries to go back to the default account to do the setup. Consequence: it is annoying as you have to switch to the secondary account and reauthenticate for each new page opened during the process to create the OAuth App Credientials, device access project etc. Again that part is just very annoying but can go through.

Issue: when you reach the setup process Section "Link Google Account" step 2 "Once all configuration information is entered in Home Assistant, a new tab opens, [...]", for that step Google doesn't pop up again their annoying page to "confirm it's me" by having automatically selected by the default account. For that step they force to open the page on the default account.
Consequence: as it is not the account used to setup Nest access, the error message in this current documentation I'm suggesting to amend with more info appears, and there is nothing you can do about it. No way to switch back to the secondary account, try again or anything. Just that cryptic message. 

If there is a way to work with Google to change that behavior, that is indeed the best way. I believe it's a general pattern on how they manage redirections to pages and it's not linked to that particular case with HA. So not sure if they would do something about or do it on a short timeline.

My suggested PR is to improve the documentation at short term and not get other users stuck like I was without understanding what's going on. Agreed logging out other accounts is not great but at least it solves the problem. Maybe I made a change by suggesting only incognito more?

allenporter (approved):

PR #35875, opened 2024-11-21
Document how to remove MQTT

jbouwh:
## Proposed change

Document how to remove MQTT

## Type of change

## Additional information

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a section on removing the MQTT integration, providing clear guidance for users.
	- Expanded documentation on MQTT discovery, detailing topic formats and payload requirements.

- **Documentation**
	- Updated mandatory fields for sharing device configuration to include `name`, `connection`, and `identifiers`.
	- Clarified naming conventions for unnamed entities based on their device class.

### DIFF
--- a/source/_integrations/mqtt.markdown
+++ b/source/_integrations/mqtt.markdown
@@ -51,6 +51,10 @@ MQTT (aka MQ Telemetry Transport) is a machine-to-machine or "Internet of Things
 
 {% include integrations/config_flow.md %}
 
+### Removing the MQTT integration
+
+The MQTT integration and its entities can be removed from by removing the MQTT config entry. Note that this does not remove the MQTT broker, or the MQTT broker data. Make sure you cleanup any MQTT related YAML configuration if you choose to remove MQTT.
+
 <a name="configuration-via-mqtt-discovery"></a>
 {% details "Configuration of MQTT components via MQTT discovery" %}
 
--- a/source/_integrations/mqtt.markdown
+++ b/source/_integrations/mqtt.markdown
@@ -59,7 +59,7 @@ The MQTT integration and its entities can be removed by following these steps:
 2. Find the MQTT integration and click on it
 3. Click the delete button to remove the MQTT config entry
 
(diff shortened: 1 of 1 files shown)


source/_integrations/mqtt.markdown
```diff
@@ -51,6 +51,10 @@ MQTT (aka MQ Telemetry Transport) is a machine-to-machine or "Internet of Things
 
 {% include integrations/config_flow.md %}
 
+### Removing the MQTT integration
+
+The MQTT integration and its entities can be removed from by removing the MQTT config entry. Note that this does not remove the MQTT broker, or the MQTT broker data. Make sure you cleanup any MQTT related YAML configuration if you choose to remove MQTT.
```
joostlek: So one example I like to think of when I think about this rule is Withings. Sometimes, people want to start anew to check if the issue they have is gone or they messed something up. Withings is an oauth2 integration, so that requires you to add application_credentials. but then if you remove the config entry, the application_credentials are still around and hard to find. So the next time someone sets up Withings, they don't have to enter the application_credentials.

I regularly get questions how to fix that. So I like to see this rule from the glasses of: What does a user need to do, to completely reset their integration?

One note specific to MQTT is that I have a friend who started to use MQTT on HAOS and I told him to look in mosquitto and he was looking at me like, I don't know what you are talking about. Only then I remembered that you now automatically setup mosquitto in the config flow since a few releases back, so he never really seen the mosquitto name or what it stands for.

So if I try to read this line through his glasses, I would have no clue what a MQTT broker is (I also haven't read the rest of the docs, maybe its explained there but I now like to finish my book). I would expect that something that sets up something automatically, also explains how to remove that again (and it would be a separate point if the MQTT docs should explain how addons work, but we could at least reference it (I am sure there are docs somewhere about this))

jbouwh: Right, added a link from `MQTT broker` to the section where is explained. It is very close near the place where we explain how to set up `mqtt`.
afbeelding

frenck (approved):
Thanks, @jbouwh 👍 

../Frenck

PR #36473, opened 2024-12-22
Update netdata.markdown

tinyhugh:
Repair examples

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration option, `scan_interval`, for the Netdata integration, allowing users to specify the polling interval for data retrieval.

- **Documentation**
	- Enhanced instructions for identifying `data_group` and `element` values for configuration.
	- Expanded guidance on using the Netdata API, including a JSON example for clarity.

### DIFF
--- a/source/_integrations/netdata.markdown
+++ b/source/_integrations/netdata.markdown
@@ -145,6 +145,7 @@ sensor:
     host: "192.168.1.2"
     port: "19999"
     name: SomeHostName
+    scan_interval: 00:00:10
     resources: 
       system_load:
         data_group: system.load 


source/_integrations/netdata.markdown
```diff
@@ -145,6 +145,7 @@ sensor:
     host: "192.168.1.2"
     port: "19999"
     name: SomeHostName
+    scan_interval: 00:00:10
```
frenck, line 148: This is not an option for this integration, but platform configuration.

Therefore, it doesn't belong here.

../Frenck

PR #36627, opened 2024-12-31
Replace broken "Copy Switch" example

Didgeridrew:
## Proposed change

This PR proposes to replace the "Copy Switch" example with an "Invert Switch" example.

The "Copy Switch" example was "fixed" in 2019; changing from a configuration that actually copied a switch to one that does not copy a switch... at least, not without additional automations which are not mentioned or described in the example. Recently, I have had a number of users on the forums and Discord who were attempting to learn about Template switches refer to this example with confusion and frustration because it does not work.

Even if the "Copy Switch" configuration worked, a straight copy of a switch without any alterations seems a pretty pointless example for a template entity. There are  already a couple configurations within the examples collection that are essentially copies of switches with small changes. As such, I have added an inverted switch as a replacement example that both works and illustrates additional uses for a Template Switch. 

## Type of change

## Additional information

- Link to pull request that broke the example: https://github.com/home-assistant/home-assistant.io/pull/11214

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Updated Template Switch integration documentation
  - Renamed section from "Copy Switch" to "Invert a Switch"
  - Clarified switch configuration with updated example
  - Added `availability_template` to improve switch functionality description


abmantis (commented):
Why doesn't the copy switch example works? What extra automatons would be required?

Didgeridrew:
> @abmantis Why doesn't the copy switch example works?


It doesn't work because the resulting switch, `switch.copy`, is neither a copy of `switch.source` nor a copy of `switch.target`.  

Based on the comments from the forum and discord users who have brought this up, the expected behavior of a "copy" is as follows:
  1. The states of both switches should be entrained. When switch A is "off", switch B should be "off".
  2. Performing a switch action on either should have the same end result.

Taking the current configuration example using `switch.source` as the original...

-   Expected Behavior 1 Fails: 
    - Changing the state of the original will result in the copy's state changing to match. However, changing the state of the copy *will have no such effect on the state of the original* because the `turn_on` and `turn_off` actions do not target the original.
-  Expected Behavior 2 Fails:
      - Switch actions targeted on the copy have no effect on the original, so any automations based on the state of the original will not be triggered by acting on the copy.

abmantis:
@Didgeridrew Ah, it is because it gets states from `switch.source` but performs actions on `switch.target`! I missed that initially. Thanks for the overview.
In that case, why not keep `copy` and change `switch.target` to `switch.source` in the actions?

Didgeridrew:
> In that case, why not keep `copy`

1. As a practical example, it's not very useful. Real-world use cases are going to copy a switch with some change such as a dynamic icon.  
2. Other [examples](https://www.home-assistant.io/integrations/switch.template/#change-the-icon) on the same page already show how to make a switch that copies another *with differences* like icon or entity picture changes.
3. Even when it consisted of a functional configuration, the "copy switch" provoke the response: "The ... example doesn't do anything."  as an explanation for the [previous PR](https://github.com/home-assistant/home-assistant.io/pull/11214) that changed the example from an actual copy to the current nonsense...

abmantis:
Thanks. I think it makes sense then.

abmantis (approved):

PR #36796, opened 2025-01-07
Update KNX connection and secure documentation

### LINKED ISSUE #36697: Wrong documentation for KNX Secure
### Feedback

https://deploy-preview-22203--home-assistant-docs.netlify.app/integrations/knx/#knx-secure states that the knykeys file should be exported from the project security details.
For ETS6, this is *wrong*. The knxkeys file must be created by the "Export Interface Information" command on the IP Tunnel that shall be used for HA.

### URL

https://www.home-assistant.io/integrations/knx/

### Version

2022.3.8

### Additional information

image

farmio:
## Proposed change

Also add new annotated screenshots using ETS 6 instead of ETS 5.

## Type of change

## Additional information

- This PR fixes or closes issue: fixes #36697

## Checklist

## Summary by CodeRabbit

- **Documentation**
	- Updated KNX integration documentation for improved clarity and consistency.
	- Enhanced descriptions of connection parameters and security features.
	- Expanded configuration examples for various KNX platforms with clearer instructions.
	- Added important notes on device configuration and security settings.
	- Improved overall structure and readability of the documentation.


source/_integrations/knx.markdown
```diff
@@ -122,32 +122,86 @@ knx:
...
+
+Assign all secured group addresses that Home Assistant will use to either the interface's tunnel endpoint or a dummy device in ETS before exporting the Keyring file.
+
+{% endimportant %}
+
+When updating secure groups, ensure all participating devices, routers, and couplers applications are updated as well. After making changes, load the updated Keyring file into Home Assistant.
+
+### Tunneling
+
+Tunneling uses a KNX IP Interface to connect to the KNX bus. Most KNX IP Routers support tunneling connections too. This is the recommended connection type and also used when selecting "Automatic" connection in the integration setup.
```
c0ffeeca7:
```suggestion
Tunneling uses a KNX IP Interface to connect to the KNX bus. Most KNX IP Routers support tunneling connections. This is the recommended connection type and is also used when selecting an "Automatic" connection in the integration setup.
```
"support tunneling connections, too". In addition to what?
could this "too" be removed?

source/_integrations/knx.markdown
```diff
@@ -122,32 +122,86 @@ knx:
...
+{% endnote %}
 
-![Obtain device authentication password in ETS](/images/integrations/knx/device_authentication_password.png)
+{% details "Manual IP Secure tunneling credentials" %}
 
-The user password can be obtained almost the same way as seen in the below screenshot.
+If you opt for manual configuration of IP Secure tunneling, you will need the following:
 
-![Obtain the user password in ETS](/images/integrations/knx/user_password.png)
+1. User-ID: Use a User-ID of 2 or higher. (IDs 0 and 1 are reserved.)
```
c0ffeeca7:
```suggestion
1. User-ID: Use a User-ID of 2 or higher. (IDs 0 and 1 are reserved).
```

source/_integrations/knx.markdown
```diff
@@ -122,32 +122,86 @@ knx:
...
-![Obtain device authentication password in ETS](/images/integrations/knx/device_authentication_password.png)
+{% details "Manual IP Secure tunneling credentials" %}
 
-The user password can be obtained almost the same way as seen in the below screenshot.
+If you opt for manual configuration of IP Secure tunneling, you will need the following:
 
-![Obtain the user password in ETS](/images/integrations/knx/user_password.png)
+1. User-ID: Use a User-ID of 2 or higher. (IDs 0 and 1 are reserved.)  
+The first tunnel endpoint in ETS will typically use User-ID `2`, the second `3`, and so on.
+2. User password
```
c0ffeeca7:
```suggestion
2. User password.
```

source/_integrations/knx.markdown
```diff
@@ -122,32 +122,86 @@ knx:
...
+3. Device authentication code (optional)
```
c0ffeeca7:
```suggestion
3. Device authentication code (optional).
```

source/_integrations/knx.markdown
```diff
@@ -122,32 +122,86 @@ knx:
...
+
+Assign all secured group addresses that Home Assistant will use to either the interface's tunnel endpoint or a dummy device in ETS before exporting the Keyring file.
+
+{% endimportant %}
+
+When updating secure groups, ensure all participating devices, routers, and couplers applications are updated as well. After making changes, load the updated Keyring file into Home Assistant.
+
+### Tunneling
+
+Tunneling uses a KNX IP Interface to connect to the KNX bus. Most KNX IP Routers also support tunneling connections. This is the recommended connection type and is also used when selecting an "Automatic" connection in the integration setup.
```
c0ffeeca7, line 152:
```suggestion
Tunneling uses a KNX IP Interface to connect to the KNX bus. Most KNX IP Routers support tunneling connections. This is the recommended connection type and is also used when selecting an "Automatic" connection in the integration setup.
```
about this "also support tunneling connections": i was wondering. in addition to what? could "also" just be omitted?

farmio: in addtion to "routing". "IP Routing" and "IP Tunneling" are the 2 connection methods we support. (Additionally "Automatic" which is also tunneling, but don't asks the user about which interface or tunnel endpoint - so it is a simplified way to configure).

Those methods are further divided to
- UDP Tunneling
- TCP Tunneling
- Secure Tunneling

or Routing
- Plain Routing
- Secure Routing

So Interfaces (the device type) support tunneling - mostly 4-9 concurrent connections. Those are used to configure other knx devices or connect software like HA.
Routers (the device type) are just like interfaces and additionally support routing too. With that they can communicate with other routers - or HA - via Multicast. So connectionless. Those cost ~100€ more for the different software stack 🤷

c0ffeeca7: ok. but routing would be the default? or are there routers that don't do routing? so the thing worth mentioning is tunneling. the "also" sent me looking for another, unmentioned option. Apologies if this seems overly nitpicky.. it's just something I tripped over when reading. hm, probably not worth discussing over..

c0ffeeca7 (approved):
Nice addition to the docs! Thank you, @farmio :+1:

source/_integrations/knx.markdown
farmio: No need to apologise - I get that this is not self explanatory, especially for people not familiar with knx. Those are things I don't think about anymore so really appreciate your comments.

Technically, if an installation only has one router, it doesn't do any routing. Until another router or some software like HA or Tasmota comes around.
We want users to prefer tunneling, as it is simpler to set up (thus less error prone). It doesn't really matter if there is some routing happening or not.

Using routing with HA is a good idea when there are multiple routers involved. Those connect sub-networks over IP and we can participate on that IP communication. If we used tunneling then, we would talk on such a sub-network, and the router would need to forward it up to IP again - creating unnecessary traffic on a TP line where traffic is expensive (twisted pair 9600 baud vs IP with 10 / 100 Mbit)

c0ffeeca7: oh, I see. there is a bit of background knowledge and thought that goes into the user's decision which method to use. Thank you for explaining!

PR #37218, opened 2025-01-30
Home Connect now adds entities automatically

Diegorro98:
## Proposed change

The devices entities are now added automatically when the appliance is connected, so the requirement to power on the appliances while loading the integration is no longer required.

## Type of change

## Additional information

- Link to parent pull request in the codebase: https://github.com/home-assistant/core/pull/136952

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Updated Home Connect integration documentation
  - Added note about appliance visibility when devices are powered off
  - Simplified configuration instructions for appliance setup by removing the requirement for appliances to be powered on during setup

### DIFF
--- a/source/_integrations/home_connect.markdown
+++ b/source/_integrations/home_connect.markdown
@@ -49,6 +49,10 @@ The integration will add one Home Assistant device for each connected home appli
 - Binary sensors that show binary states of the appliance.
 
 {% note %}
+Some appliances don't report data while they are turned off so they will not appear in the Home Connect integration after loading until they are turned on.
+{% endnote %}
+
+{% note %}
 Note that it depends on the appliance and on API permissions which of the features are supported.
 {% endnote %}
 
@@ -72,7 +76,6 @@ Some devices only have the state `on` and turn off is not supported by the appli
 
 {% important %}
 
-- **Power on** all your appliances during the integration configuration process; otherwise, appliance programs list will be empty.
 - To update the appliance programs list, you can reload the Home Connect integration when an appliance is turned on. If the re-initialization process is not triggered by reload, restart the Home Assistant when an appliance is turned on.
 - After performing the steps above, [log out](https://developer.home-connect.com/user/logout) of your Home Connect Developer account. If you don't do this, the configuration steps below will fail during OAuth authentication with the message `“error”: “unauthorized_client”`.
 - The provided Home Connect User Account email address **must** be all lowercase; otherwise, it will result in authentication failures.
--- a/source/_integrations/home_connect.markdown
+++ b/source/_integrations/home_connect.markdown
@@ -49,7 +49,7 @@ The integration will add one Home Assistant device for each connected home appli
 - Binary sensors that show binary states of the appliance.
 
 {% note %}
-Some appliances don't report data while they are turned off so they will not appear in the Home Connect integration after loading until they are turned on.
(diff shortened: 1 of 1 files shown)


source/_integrations/home_connect.markdown
```diff
@@ -48,6 +48,10 @@ The integration will add one Home Assistant device for each connected home appli
 - Multiple sensors that report the different states and events reported by the appliance.
 - Binary sensors that show binary states of the appliance.
 
+{% note %}
+Some appliances don't report data while they are turned off so they will not appear in the Home Connect integration after loading until they are turned on.
```
MartinHjelmare:
```suggestion
Some appliances don't report data while they are turned off so corresponding entities will not appear in the Home Connect integration after loading until the appliances are turned on.
```

source/_integrations/home_connect.markdown
```diff
@@ -72,7 +76,6 @@ Some devices only have the state `on` and turn off is not supported by the appli
 
 {% important %}
 
-- **Power on** all your appliances during the integration configuration process; otherwise, appliance programs list will be empty.
 - To update the appliance programs list, you can reload the Home Connect integration when an appliance is turned on. If the re-initialization process is not triggered by reload, restart the Home Assistant when an appliance is turned on.
```
MartinHjelmare, line 79: Is it still needed to reload the config entry to update programs?

MartinHjelmare (changes requested):

source/_integrations/home_connect.markdown
Diegorro98: Yes, it is, the user can also disconnect and connect the appliance to trigger the post-connection update.
Also, I don't know if there's any way to manually trigger the coordinator update.

MartinHjelmare: One can call the `homeassistant.update_entity` action targeting one of the integration entities.

MartinHjelmare: Should we make a request for the programs automatically when an appliance is turned on and we haven't requested the programs yet?

MartinHjelmare (approved):

source/_integrations/home_connect.markdown
Diegorro98: If the use case you are thinking of is the case where the appliance doesn't report any data because it disconnected while it is turned off because , we already fetch all the programs when it is turned on (because the CONNECTED event is received).
I can't think about any other use case right now

MartinHjelmare: Great! Thanks for clarifying.

MartinHjelmare: I suggest we rephrase the sentence to just mention that some devices may not report all programs until the device is turned on. The user shouldn't reload the integration.

Diegorro98: This is what we have right now:
https://github.com/home-assistant/home-assistant.io/blob/6b4953ef91a803f5fc9b2b93bdbe337baf82ac93/source/_integrations/home_connect.markdown?plain=1#L51-L53
It doesn't talk about the entities that use the programs specifically, but I think that the statement generalizes enough to include them.

MartinHjelmare: Right, that sentence is enough. Then we can just remove the sentence here on line 79.

PR #37266, opened 2025-02-03
Improving Webhook docs

carefulcomputer:
## Proposed change
1. added note about restarting haas after setting up webhook. 
2. added note on how to access webhook payload in automation.

## Type of change

## Additional information
none

## Checklist

## Summary by CodeRabbit

- **Documentation**
	- Enhanced guidance on setting up webhook triggers.
	- Added a note that a Home Assistant restart may be required during initial setup.
	- Provided an example for correctly passing JSON data to automation actions.
	- Included a caution about potential issues when the payload data type does not match expectations.

### DIFF
--- a/source/_docs/automation/trigger.markdown
+++ b/source/_docs/automation/trigger.markdown
@@ -868,7 +868,7 @@ See the [Persistent Notification](/integrations/persistent_notification/) integr
 
 ## Webhook trigger
 
-Webhook trigger fires when a web request is made to the webhook endpoint: `/api/webhook/<webhook_id>`. The webhook endpoint is created automatically when you set it as the `webhook_id` in an automation trigger.
+Webhook trigger fires when a web request is made to the webhook endpoint: `/api/webhook/<webhook_id>`. The webhook endpoint is created automatically when you set it as the `webhook_id` in an automation trigger. You may need to restart home assistant if you are setting up webhook for very first time.
 
 ```yaml
 automation:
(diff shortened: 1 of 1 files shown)


source/_docs/automation/trigger.markdown
```diff
@@ -868,8 +868,7 @@ See the [Persistent Notification](/integrations/persistent_notification/) integr
 
 ## Webhook trigger
 
-Webhook trigger fires when a web request is made to the webhook endpoint: `/api/webhook/<webhook_id>`. The webhook endpoint is created automatically when you set it as the `webhook_id` in an automation trigger.
-
+Webhook trigger fires when a web request is made to the webhook endpoint: `/api/webhook/<webhook_id>`. The webhook endpoint is created automatically when you set it as the `webhook_id` in an automation trigger. You may need to restart Home Assistant if you are setting up a webhook for the first time.
```
frenck, line 871: This not correct, as restant isn't needed.

Ik you run into a situation where in is, hou might have ran into a bug, which should be reported with reproduction steps instead.

source/_docs/automation/trigger.markdown
```diff
@@ -904,6 +903,12 @@ Note that to use JSON encoded payloads, the `Content-Type` header must be set to
 ```bash
 curl -X POST -H "Content-Type: application/json" -d '{ "key": "value" }' https://your-home-assistant:8123/api/webhook/some_hook_id
 ```
+If you want to pass webhook data to action, you will need to use something like this in your action
+```yaml
+data_template:
```
frenck, line 908: There is no need to use a data_template, it features has been added years ago, we now only use the data key.

source/_docs/automation/trigger.markdown
```diff
@@ -904,6 +903,12 @@ Note that to use JSON encoded payloads, the `Content-Type` header must be set to
...
+  payload: "{{trigger.json}}"
+```
+Note: if you pass json when template is expecting non-json data or vice versa, then automation will trigger but payload will be empty.
```
frenck, line 911: That sounds like a bug, please report it, as Home Assistant shouldn't just lose data or raise an error instead.

frenck (changes requested):

PR #38123, opened 2025-03-21
feat: enable View Transitions in supported browsers

coogie:
## Proposed change

Adds [View Transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API) to home-assistant.io, enabling a "single-page app" feel as users navigate between pages without the need for any userland JS or frameworks.  View Transitions, being a browser-native feature, can be implemented as a form of progressive enhancement -- browsers that don't support it yet (Firefox) simply behave as they currently do.

**Note:** I'm not quite sure about the inline `style` tag. Since its content is so small and intrinsically tied to the `meta` tag, I opted to make it inline instead of inside a CSS file. Happy to take guidance/feedback on it.

Browser Support Table

## Type of change

## Additional information

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced the user experience by introducing smoother view transitions, providing a more seamless single-page app feel in supported browsers.

### DIFF
--- a/source/_includes/site/head.html
+++ b/source/_includes/site/head.html
@@ -13,6 +13,10 @@
     <meta name="viewport" content="width=device-width">
     <link rel="canonical" href="{{ page.url | canonical_url }}">
 
+    <!-- Enable View Transition in supported browsers for single-page app feel -->
+    <meta name="view-transition" content="same-origin" />
+    <style>@view-transition { navigation: auto; }</style>
+
     <meta property="fb:app_id" content="{{ site.social.facebook.app_id }}">
     <meta property="og:title" content="{% if post.title %}{{ post.title }}{% elsif page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}">
     <meta property="og:site_name" content="{{ site.title }}">


mrdarrengriffin:
Hey @coogie

Cool to see ViewTransitions finally usable. I remember trying to implement this a few years ago and it just wasn't ready yet, so this is cool.

Whilst I love that this is just a progressive enhancement, I can't help but feel that the transition itself makes the site _feel_ slower. In my opinion, if a site is fast, having pages navigate with no transition feels much "snappier" than the fade in the PR preview. Not to say that there are ways we could improve certain areas to make pages load quicker.

I'll do some testing with it to see if we can speed up the transition and see if that feels better, but in it's current form, I think we should hold this for now.

Interested to hear your thoughts

coogie:
Yes, I don't think it will ever be measurably or perceptively faster than a
SPA setup because it will always require the round trip to resolve before
it can blend the views.

As for improving the perception, the first thing with exploring is a
different value from `auto` for the transition speed. I initially chose
this because I like letting the platform and any potential user preferences
take precedence -- fewer things to care about and work around 😅

After that, another thing that could be worth exploring is the additional
use of the [Speculation Rules API](
https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API).
Though I must admit I'm personally not a fan of this API due to how complex
the rule definitions/patterns can become. It also has no support in Gecko
or WebKit so won't actually improve the situation in those browsers. Which
isn't to say it's _not_ worth exploring since it is also a progressive
enhancement pattern.

frenck:
I agree we shouldn't do this. After playing with it, it seems not ready / the experience we should give.

Thanks for being willing to contribute, but going to close it from this end.

../Frenck

PR #40007, opened 2025-07-14
Z-Wave: switch from JS to JS UI add-on: add some steps

c0ffeeca7:
## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/zwave_js.markdown
+++ b/source/_integrations/zwave_js.markdown
@@ -987,7 +987,7 @@ Switching does not require renaming your devices.
    - Select the **Z-Wave** integration and select the three-dot {% icon "mdi:dots-vertical" %} menu.
    - From the menu, select **Reconfigure**, then **Reconfigure current adapter**.
    - Uncheck the **Use the Z-Wave JS Supervisor add-on**.
-   - Under **Websocket URL**, enter the correct address for the community add-on.
+   - Under **WebSocket URL**, enter the correct address for the community add-on.
      - To find the URL, in the Z-Wave JS UI add-on, open the **Documentation** tab. The URL is listed in the section "Setting up the Home Assistant Z-Wave JS integration".
 
 6. Uninstall the official Z-Wave JS add-on.
--- a/source/_integrations/zwave_js.markdown
+++ b/source/_integrations/zwave_js.markdown
@@ -977,18 +977,23 @@ Switching does not require renaming your devices.
 
 1. Note your network security keys from the official add-on.
 
-2. Install and start the Z-Wave JS UI add-on.
+2. Install and start the **Z-Wave JS UI** add-on.
+   - It may take a while for the add-on to start up.
 
-3. Configure the Z-Wave JS UI add-on with the added control panel, including setting the location of your Z-Wave device and the network security keys.
+3. Configure the **Z-Wave JS UI** add-on with the added control panel, including setting the location of your Z-Wave device and the network security keys.
+   - Open the **Z-Wave JS UI** web UI and go to **Settings** > **UI** > **Z-Wave**.
+   - Enter the security keys and region.
+   - Save your changes and in the top-right corner, select the {% icon "solar:restart-outline" %} restart button to reload the add-on configuration.
+   - Open the **Documentation** tab and copy the URL listed in the section **Setting up the Home Assistant Z-Wave JS integration**. You will need it in the next step.
 
 4. Reconfigure the adapter.
    - Select the **Z-Wave** integration and select the three-dot {% icon "mdi:dots-vertical" %} menu.
    - From the menu, select **Reconfigure**, then **Reconfigure current adapter**.
    - Uncheck the **Use the Z-Wave JS Supervisor add-on**.
-   - Under **WebSocket URL**, enter the correct address for the community add-on.
-     - To find the URL, in the Z-Wave JS UI add-on, open the **Documentation** tab. The URL is listed in the section "Setting up the Home Assistant Z-Wave JS integration".
+   - Under **WebSocket URL**, enter the URL you copied before.
+     - **Troubleshooting**: If it fails to connect although you pasted the right URL, check if you pasted all the security keys in the **Z-Wave JS UI** configuration.
 
-5. Uninstall the official Z-Wave JS add-on.
+5. Uninstall the official **Z-Wave JS** add-on.
 
 6. Enable the Z-Wave integration.
 
--- a/source/_integrations/zwave_js.markdown
+++ b/source/_integrations/zwave_js.markdown
@@ -988,7 +988,7 @@ Switching does not require renaming your devices.
    - From the menu, select **Reconfigure**, then **Reconfigure current adapter**.
    - Uncheck the **Use the Z-Wave JS Supervisor add-on**.
    - Under **Websocket URL**, enter the correct address for the community add-on.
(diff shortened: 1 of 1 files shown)


source/_integrations/zwave_js.markdown
```diff
@@ -983,18 +983,23 @@ Switching does not require renaming your devices.
 
 4. Configure the Z-Wave JS UI add-on with the added control panel, including setting the location of your Z-Wave device and the network security keys.
 
-5. Add the Z-Wave integration again (even though it is still installed), and uncheck the "Use the Z-Wave JS Supervisor add-on". Enter the correct address for the community add-on in the URL field in the next step.
+5. Add the Z-Wave integration again (even though it is still installed).
```
MartinHjelmare: Maybe we should remove step 1 and update the text here to not say to add the Z-Wave integration again? The reconfigure flow will now unload the config entry and stop the add-on during the move from the official add-on to Z-Wave JS UI.

source/_integrations/zwave_js.markdown
```diff
@@ -975,24 +975,22 @@ You can switch from the official **Z-Wave JS** add-on to the **Z-Wave JS UI** ad
 
 Switching does not require renaming your devices.
 
-1. Disable the Z-Wave integration. **Do not remove the Z-Wave integration or you will lose all device and entity naming.** This will automatically stop the official Z-Wave JS add-on.
+1. Note your network security keys from the official add-on.
 
-2. Note your network security keys from the official add-on.
+2. Install and start the Z-Wave JS UI add-on.
```
MartinHjelmare: You shouldn't start the new add-on until reconfiguration is done since that will cause a USB device access conflict.

I'd install the new add-on and configure it with the network keys. Note the URL of the add-on to connect to, as this will be used in the reconfigure flow. Then reconfigure the integration to use the new add-on URL. Then start the new add-on.

MartinHjelmare: Maybe add a note that the add-on will normally take a while to start up after start is pressed and the integration will retry the connection to the server in the add-on if the first attempt fails.

MartinHjelmare: Actually we may have a connection check in the flow already, so we need to start the add-on when coming to the form to enter the URL. That will ensure the connection succeeds before the flow is finished.

c0ffeeca7: how would I add the security keys in the next step if the add-on wasn't started?

source/_integrations/zwave_js.markdown
```diff
@@ -967,34 +967,42 @@ Zwavejs2Mqtt was renamed Z-Wave JS UI in September 2022. They are synonymous wit
...
-To switch between the official Z-Wave JS add-on and the Z-Wave JS UI add-on, follow these steps:
+You can switch from the official **Z-Wave JS** add-on to the **Z-Wave JS UI** add-on. However, you cannot run them both at the same time. Only one of the add-ons can be active at the same time.
 
 Switching does not require renaming your devices.
 
-1. Disable the Z-Wave integration. **Do not remove the Z-Wave integration or you will lose all device and entity naming.** This will automatically stop the official Z-Wave JS add-on.
+1. Note your network security keys from the official add-on.
 
-2. Note your network security keys from the official add-on.
+2. Install and start the **Z-Wave JS UI** add-on.
```
MartinHjelmare, line 980: We can't start the add-on until we come to the URL form in the reconfigure flow.

source/_integrations/zwave_js.markdown
c0ffeeca7: I'm running Core 2025.7.2, Z-Wave JS 0.18, Z-Wave JS UI 4.7. I needed to start the JS UI add on in order to be able to access the configuration settings. It was not an issue for me that the add-on was started.

source/_integrations/zwave_js.markdown
c0ffeeca7: but then how do you change the configuration in the Z-Wave JS Web UI?
I had to start it in order to get here and enter the keys
image

source/_integrations/zwave_js.markdown
```diff
@@ -967,34 +967,42 @@ Zwavejs2Mqtt was renamed Z-Wave JS UI in September 2022. They are synonymous wit
...
-4. Configure the Z-Wave JS UI add-on with the added control panel, including setting the location of your Z-Wave device and the network security keys.
+4. Reconfigure the adapter.
+   - Select the **Z-Wave** integration and select the three-dot {% icon "mdi:dots-vertical" %} menu.
+   - From the menu, select **Reconfigure**, then **Reconfigure current adapter**.
+   - Uncheck the **Use the Z-Wave JS Supervisor add-on**.
+   - Under **WebSocket URL**, enter the URL you copied before.
+     - **Troubleshooting**: If it fails to connect although you pasted the right URL, check if you pasted all the security keys in the **Z-Wave JS UI** configuration.
 
-5. Add the Z-Wave integration again (even though it is still installed), and uncheck the "Use the Z-Wave JS Supervisor add-on". Enter the correct address for the community add-on in the URL field in the next step.
+5. Uninstall the official **Z-Wave JS** add-on.
```
c0ffeeca7: when uninstalling, should I delete all the related data? By default, this setting is disabled.

source/_integrations/zwave_js.markdown
MartinHjelmare: Ok, right. That add-on won't access the device immediately when the add-on is started the first time but after configuring the device and applying that in the Z-Wave JS UI GUI.

So it's ok to start the add-on, but we need to start the reconfigure flow, reaching the URL step, before applying the device settings and starting the driver.

Side note: With the complexity of these steps, we should consider having an explicit choice for reconfiguring from the official add-on to the Z-Wave JS UI add-on in the reconfigure flow, that walks the user through the needed steps.

source/_integrations/zwave_js.markdown
MartinHjelmare: There's no right or wrong answer. If the user wants to switch back it's nice to have the data still there, but otherwise it only takes up space that the user may want to free up.

source/_integrations/zwave_js.markdown
c0ffeeca7: I don't understand what you mean by "but we need to start the reconfigure flow, reaching the URL step, before applying the device settings and starting the driver."
the order described here worked for me.

MartinHjelmare: Having two different processes accessing the same serial (USB) device isn't supported and may lead to issues and weird behavior. That's why we disabled the existing config entry before and why we should start the reconfigure flow before letting the driver run by Z-Wave JS UI access the device.

Starting the reconfigure flow and deselecting add-on to get to the manual URL form will stop the official add-on.

c0ffeeca7: I'm not trying to argue with you, I just don't quite understand what the order is. I reordered the steps now. Is it correct this way? Part of the reconfiguration steps are done twice, but I didn't see how else to do it. Maybe I am still missing something?

MartinHjelmare: You shouldn't reconfigure twice. That flow should still be open. The Z-Wave JS UI can be accessed in a separate tab.

If this is too complicated at the moment, we can keep the instruction to disable the config entry. Although we're moving away from that way of allowing reconfiguration via creating a new config entry which updates an existing config entry.

MindFreeze (approved):

c0ffeeca7:
Thanks!

PR #40217, opened 2025-07-30
Revert "Add info about select entity for heating circuit to Tado docs"

joostlek:
Reverts home-assistant/home-assistant.io#39797

We're in contact with Tado and about API usage. With that in mind, it would be wise to revert this PR until we know the conclusion of these talks.


c0ffeeca7 (approved):
Thank you, @joostlek :+1:

PR #41076, opened 2025-09-29
Revert "Add ConBee thread enable link"

frenck:
Reverts home-assistant/home-assistant.io#41038

We had some discussions on how this might confuse people with this method being supported or not. We don't know... we do know, thread causes tons of issues for people accross the board.

We want to limit this part to the setups we have verified and know actually work well. Hence the revert of this pull request for now.

../Frenck


c0ffeeca7 (approved):
Thank you, @frenck :+1:

PR #41576, opened 2025-11-02
Replace unsupported recommendation

deni:
## Proposed change

Hello,

This change removes the recommendation to use A2 SD cards when installing Home Assistant on Raspberry Pi. The recommendation was introduced with 92301d0 and does not seem to be correct or supported by anything. I wish to note that I delayed my own installation of Home Assistant after having purchased an A1 card for this project and, upon reading the documentation, believing that it would be unideal. I am not happy about this delay and the extra time that I have needed to spend. I can only imagine that other users, like me, take recommendations in the official documentation seriously why I believe that introducing something so specific without being able to back it up is counter-productive. I encourage contributions to open source projects in general, including to documentation, but also ask, which I think is reasonable, that the contributor consider if the change might have any implications. I hope that this is something the Home Assistant community can get behind.

Kind regards,
Denis Smajlović

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/installation/raspberrypi.markdown
+++ b/source/installation/raspberrypi.markdown
@@ -16,7 +16,7 @@ We will need a few things to get started with installing Home Assistant.
 
 - [Raspberry Pi 5](https://amzn.to/3UH6TcD) or [Raspberry Pi 4](https://amzn.to/2S0Gcl1) with [power supply](https://amzn.to/2ReZ2Vq) (make sure to choose a model with at least 2&nbsp;GB of RAM).
 - [Micro SD Card](https://amzn.to/2X0Z2di).
-  - Ideally get one that is [Application Class 2](https://www.sdcard.org/developers/overview/application/index.html). Check for the label **A2** on the card. Application Class 2 cards perform better especially on small read and write operations and are better suited to host applications.
+  - A quality card from a reputable manufacturer is recommended.
   - Make sure to use a card that provides at least 32&nbsp;GB.
 - SD Card reader. This is already part of most laptops, but you can purchase a [standalone USB adapter](https://amzn.to/4lkMfsu) if you don't have one. The brand doesn't matter, just pick the cheapest.
 - [Ethernet cable](https://amzn.com/dp/B00N2VISLW). Required for installation. After installation, Home Assistant can work with Wi-Fi, but an Ethernet connection is more reliable and highly recommended.


agners:
> This change removes the recommendation to use A2 SD cards when installing Home Assistant on Raspberry Pi. The recommendation was introduced with [92301d0](https://github.com/home-assistant/home-assistant.io/commit/92301d052e63810eebcaf9ff9188f7ec77e289f3) and does not seem to be correct or supported by anything.

Raspberry Pi them self sell such SD cards: https://www.raspberrypi.com/products/sd-cards/:

> Optimise data transfer speeds on your Raspberry Pi computer with an official Raspberry Pi SD Card. Rigorously tested to ensure optimal performance on Raspberry Pi computers, these Class A2 microSD cards help ensure you get the smoothest user experience from your device.

A2 stands for application class. They are meant for use cases where random I/O is important, like hosting applications (A stands for applications). It makes very much sense to *recommend* such cards. That does not mean other cards don't work.  The wording clearly says "Ideally"... This should indicate that other SD cards work too, they just might not be ideal. YMMV.

If SD card manufacturer produce SD cards which are well suited for applications, but don't undergo the official testing to be spec'ed as a A2 card, that is up to them. But it makes sense to generally advice A2 cards, as this is a well-known standard which ensures good performance for our use case.

frenck:
I agree with the above response. Going to close this PR for that reason.

../Frenck

deni:
Hello @agners and @frenck,

I see different claims here. One being that the A2 cards can endure more writes, the other that they better support random I/O. I do see that it does say so on the Raspberry Pi store but nothing in documentation or with data to support these claims. In fact, I see posts online [saying the exact opposite](https://www.jeffgeerling.com/blog/2019/a2-class-microsd-cards-offer-no-better-performance-raspberry-pi).

Have you found anything other than marketing material indicating that these cards are actually better?

Kind regards,
Denis Smajlović

agners:
Hi @deni 

> In fact, I see posts online [saying the exact opposite](https://www.jeffgeerling.com/blog/2019/a2-class-microsd-cards-offer-no-better-performance-raspberry-pi).

Yeah the article is a bit dated though. Likely the reason back then was the SDHCI driver not having command queuing support.

> Have you found anything other than marketing material indicating that these cards are actually better?

Yes, e.g. this more recent test on Raspberry Pi 5: https://forums.raspberrypi.com/viewtopic.php?t=367459#p2203671.

pantherale0:
> In fact, I see posts online [saying the exact opposite](https://www.jeffgeerling.com/blog/2019/a2-class-microsd-cards-offer-no-better-performance-raspberry-pi).

Surely if the SD card does not perform to its rated class, then that is a defective SD card and should be returned? I think IOPS is more important to an application like Home Assistant with "high volume" of small and random reads and writes over pure bandwidth, and A2 correctly rates for that usage accordingly.

deni:
First, I need to ask for a minute of silence as I now know more about Command Queuing than I ever wanted or needed.

In summary, this is how I see this. Please let me know if you disagree with any of the points.

- Command Queuing on Raspberry Pi [only exited testing in January 2025](https://forums.raspberrypi.com/viewtopic.php?p=2284988#p2284988).
- The original commit with the recommendation (92301d0) was made in 2019. I assume that it is undisputed that the recommendation was not correct when it was made.
- Command Queuing is only supported [for certain whitelisted SD cards](https://github.com/raspberrypi/linux/pull/6591) and **only** on Raspberry Pi 5

My conclusion is that, at the very least, the recommendation does not make sense in its current form. If kept, it should be specified that it is RPi5-specific and that the card needs to be one of the specific whitelisted ones for it to make a difference.

That said, it is my impression that non-A2 cards do random operations _fine_. So even if having a setup that can support the new features (better-performing at random operations) I do question if it will actually make a noticeable performance difference in Home Assistant. Users with this need have probably already switched away from running the OS from the SD card anyway. To me the central question is if anyone has had a problem with their Home Assistant setup that was solved by using an A2 SD card? I believe that it is unlikely.

PR #41930, opened 2025-11-23
2025.12: Beta release notes

frenck:


source/_posts/2025-12-03-release-202512.markdown
```diff
@@ -0,0 +1,492 @@
...
+  - [Lorem ipsum](#lorem-ipsum-1)
+- [Need help? Join the community](#need-help-join-the-community)
+- [Backward-incompatible changes](#backward-incompatible-changes)
+- [All changes](#all-changes)
+
+_A huge thank you to all the contributors who made this release possible! And a special shout-out to [@frenck], [@frenck], and [@frenck] who helped write the release notes this release. Also, [@frenck], [@frenck], [@frenck], and [@frenck] for putting effort into tweaking its contents. Thanks to them, these release notes are in great shape. ❤️_
+
+[@frenck]: https://github.com/frenck
+
+## Lorem ipsum
```
TimoPtr:
```markdown
🆕 Add to

Starting with Home Assistant Android app version 2025.11.x, a new Add to option is available directly from the More info dialog for supported entities.

This new feature makes it much easier to take advantage of your device’s native capabilities — no more deep-diving into app settings! With just a few taps, you can now:

- Add widgets for quick control of entities right from your home screen. Find more about widgets on the [companion documentation](https://companion.home-assistant.io/docs/integrations/android-widgets)
- Set entities as Android Auto favorites, making them quickly accessible while on your car

The frontend now exposes the Add to menu entry whenever the companion app signals support for it. Selecting the action opens a native dialog that lists available options depending on the entity type.

This is the first step in integrating native mobile features directly into the frontend experience. Future releases will expand this with support for creating shortcuts, tiles, and watch favorites, directly from the same interface.
```

There are some content in the Frontend PR, but it's displaying some features not yet available. I can make a new video if we want a video.

For reference 
- Android PR https://github.com/home-assistant/android/pull/5622
- Frontend PR https://github.com/home-assistant/frontend/pull/26346

laupalombi:
I was trying to commit the changes myself but I don't have push permissions - its OK I can just send the changes here too.

Text changes only - no images yet

```
 ## Picking a default dashboard is now at your system level

Default dashboard setting is now something that can be set easily for any dashboard you choose, has instant effect and puts this Dashboard on top of the list, replacing the current default dashboard. And it won’t need to be configurable by every single person (or the same one entering a new browser), but will be done for everyone in the same Home Assistant installation.

This doesn’t mean someone can’t have a different default dashboard - we added a new setting in your User preferences where you can decide to override this system default and have your own custom default dashboard.

## Order Areas & Floors so they make sense for _you_

When working in our built-in dashboard experiences (Home, Lights, Security, etc) one of the main pain points we kept seeing is the strict ordering of your Areas (alphabetically) and Floors (numerically by level). The reason is, this didn’t make any sense in a real home setup, where your Guest bathroom is never more prominent than your Master Bedroom, and Attic is rarely more relevant than the main floor.

Now you can go to System - Areas and manually drag and drop any area or the whole floor altogether. The change will instantly be made for all the built-in dashboards showing an area/floor representation.

## Experimental mode is gone

With the launch of Labs, it makes no sense to keep existing experiments like Area and Home as Experimental in the dashboard creation list. Home can now be found in the Dashboard list (still not visible by default) and Areas has found its evolution in Home and we no longer want to keep experimenting with it for the time being.

## Home is getting better and better

We added a new sidebar to gather all the accesses we believe are suited For You - more to come, like Home Assistant reminders for admins - and a nicer area/floor distribution that uses space better. At a more internal level, this page is now a proper built-in dashboard and shows in the built-in dashbaord list.

**Sidebar can also be added to your manual dashboards via YAML

**There is a chance your current favorites might disappear in this release and you need to re-add them, due to the migration of this dashboard from strategy to built-in. 
```

source/_posts/2025-12-03-release-202512.markdown
frenck: Thanks!

PR #42839, opened 2026-01-01
[mcp_server] Add a note about the current restrictions for OAuth authentication

conallob:
Add a note callout to point people experiencing issues configuriing OAuth with custom hostnames to https://github.com/home-assistant/core/issues/153820

## Proposed change

`mcp_server` now supports OAuth authentication, in addition to long lived tokens. However,
numerous people are reporting issues completing the OAuth setup flow on https://github.com/home-assistant/core/issues/153820 . On the bug discussion, at least one known and undocumented caveat of the OAuth auth flow is that `ui.nabu.casa` hostnames are required.

## Type of change

## Additional information

See https://github.com/home-assistant/core/issues/153820#issuecomment-3704233027 for a concise summary of the current limitations this note is raising awareness

## Checklist

- [ x The documentation follows the Home Assistant documentation [standards].


source/_integrations/mcp_server.markdown
```diff
@@ -71,6 +71,13 @@ client to provide an authentication token.
 
 #### OAuth
 
+{% note %}
```
allenporter, line 74: I'm not sure this is totally accurate, and it may be depending on something more nuanced. 
(This integration definitely works with OAuth for other domains and certainly does not require nabu casa for OAuth to work)

I believe you that you had an issue and using a nabu casa domain helped you.  However, I think this is more about limitations on the specific MCP Client.

Maybe we can discuss more in the issue or here....

allenporter (changes requested):

source/_integrations/mcp_server.markdown
stgarrity: fwiw I had the same issue with my custom domain (the symptom for me was a redirect to `.../authorize` instead of `.../auth/authorize`, banged my head against it for a while until claude found the issue, i noticed the comment, tried it and it worked great using my ui.nabu.casa URL. This is with Claude Desktop as the MCP client. I actually asked Copilot CLI to write this same PR and then it noticed this one already existed :)

conallob: @allenporter My MCP Clients are Claude and Claude Code, so I don't think it's either of them.


I want both. I want the docs to save people wasting time and I want to try and describe the problem. If we're not 100% sure how to describe the problem, I'm fine with a warning and a pointer to the bug

allenporter: Sounds like a bug in the client that claude desktop is no longer doing Authorization Server Discovery properly
https://modelcontextprotocol.io/docs/tutorials/security/authorization as the home assistant endpoint `.well-known/oauth-authorization-server` has the current url

stgarrity: that might be, but the proposed change alerts users to the potential issue, rather than letting them waste a bunch of time finding it themselves. Wouldn't it be better to note this in the docs with a link to the discussion than not? Happy to propose an update that makes it more specific (eg. to Claude) and clear that it's not a Home Assistant issue, but it seems useful to note

allenporter: There is a misunderstanding.  I would like the docs to actually be accurate describing the situation. I am supportive of updating the document with an accurate statement.

Saying that this integration requires nabu casa is not accurate, right? I gave this feedback last week and am waiting for it to be updated with an accurate statement.

conallob (commented):
I'm not particularly familiar with OAuth internals, but at a cursory reading, it appears to do parallel validation to the https layer (e.g the redirect_url and other settings).

allenporter (changes requested):
Let me simplify what you need to do since I think it's getting lost: This is not correct "As of Jan 1st 2026, OAuth authentication only works for Home Assistant Cloud". Remove it or reword it. For example, update this to say that X or Y clients only work with nabu casa domain.

Background: My understanding is both of the clients (Claude Code and Claude) are using the same infrastructure to connect to remote client and are not following the MCP OAuth spec properly, making assumptions about fixed URL patterns that are not correct.    OAuth authentication for this MCP server does not require Nabu Casa.

conallob:
Abandoning this PR

PR #44043, opened 2026-03-11
Add back-to-top button functionality and styles

dan-long-dev:
## Proposed change

Add a floating back-to-top button on the default template, if a 'table of contents' is present. Typically, table of contents is used on integration pages and blog posts where the content and scroll depth is quite long. 

As a user progresses down the page, the back-to-top button progress indicator updates and is clickable to jump users to the top of the page. The button appears after ~100px of scrolling, and hides when the footer enter the screen. 

The button is hidden on mobile due to too much overlap of content and illegibility (but have another plan on how to introduce a nicer UX for that on mobile - coming in another PR).

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_includes/javascripts/scripts.html
+++ b/source/_includes/javascripts/scripts.html
@@ -82,11 +82,19 @@
 
         const footerTop = footer.offsetTop;
         const windowHeight = window.innerHeight;
-        const documentHeight = document.documentElement.scrollHeight;
 
         const scrollStart = 0;
         const scrollEnd = footerTop - windowHeight;
-        const scrollProgress = Math.min(Math.max((scrollY - scrollStart) / (scrollEnd - scrollStart), 0), 1);
+        let scrollProgress;
+
+        if (scrollEnd <= scrollStart) {
+          scrollProgress = 1;
+        } else {
+          scrollProgress = Math.min(
+            Math.max((scrollY - scrollStart) / (scrollEnd - scrollStart), 0),
+            1
+          );
+        }
 
         const footerOffset = -50;
         const footerIsVisible = scrollY + windowHeight >= footerTop - footerOffset;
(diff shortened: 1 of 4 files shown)


source/_includes/javascripts/scripts.html
```diff
@@ -64,3 +64,64 @@
...
+
+{% if page.toc %}
+<script type="text/javascript">
+  {
+    const scrollButton = document.getElementById("back-to-top-button");
+    const depthCircle = document.getElementById("depth");
+    const footer = document.querySelector("footer");
+
+    if (scrollButton && depthCircle && footer) {
+      const circumference = 106.1258544921875;
```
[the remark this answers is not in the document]
dan-long-dev: I don't want the overhead of calculating this every time when I can set it once. We'll update the value if we change the asset.

source/_includes/javascripts/scripts.html
```diff
@@ -64,3 +64,62 @@
...
```
mrdarrengriffin: I do agree with copilot here tbh. Since we already have `scrollButton` defined, could we move this path length to some attribute inside the svg (maybe data-circumference) so that it ties it to the SVG?

mrdarrengriffin (dismissed):
Happy to approve in it's current state, but moving the circumference would be nice just in case we update the SVG down the road

source/_includes/javascripts/scripts.html
dan-long-dev: Yea OK - I've added a `data-circumference` attr to the progress indicator to tie the value to the SVG.

dan-long-dev: Using getTotalLength() instead - it's only called once and our path is small so the overhead of using the function is minimal in this use case.

mrdarrengriffin (approved):

PR #44460, opened 2026-04-02
Correct remap() edges parameter default and document 'none' option

### LINKED ISSUE #44394: remap() `edges` parameter documentation incorrect
### Feedback

The documentation on Templating page for the `remap()` `edges=` parameter says `clamp` is the default, but the default is actually `none` which is not documented on the page at all. `edges='none'` performs no clamping (out-of-range input values result in out-of-range output values).

The documentation on Templating page documents the `edges=` parameter as follows:

    - `edges='clamp'` (the default) will clamp the output to the min/max output range.
    - `edges='wrap'` will wrap the input value around the input range before remapping.
    - `edges='mirror'` will bounce the input value back and forth within the input range before remapping.

Documentation from source code:

        The edges parameter controls how out-of-bounds input values are handled:
        - "none": No special handling; values outside the input range are extrapolated into the output range.
        - "clamp": Values outside the input range are clamped to the nearest boundary.
        - "wrap": Values outside the input range are wrapped around cyclically.
        - "mirror": Values outside the input range are mirrored back into the range.

### URL

https://www.home-assistant.io/docs/configuration/templating/

### Version

2026.3.4

### Additional information

_No response_

tmchow:
The `remap()` template function documentation states `edges='clamp'` is the default, but the actual default in HA core is `edges='none'` ([source](https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/template/extensions/math.py)). The `none` option was also missing from the documented options.

Updated the edges parameter documentation to list all four options (`none`, `clamp`, `wrap`, `mirror`) and corrected the default indicator.

Fixes #44394

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/source/_docs/configuration/templating.markdown
+++ b/source/_docs/configuration/templating.markdown
@@ -1156,7 +1156,8 @@ The numeric functions and filters raise an error if the input is not a valid num
 - `remap(v, in_min, in_max, out_min, out_max, *, [steps], [edges])` remaps a value `v` from the range `in_min`..`in_max` to the range `out_min`..`out_max`.
   If any of the values `v`, `in_min`, `in_max`, `out_min`, `out_max` cannot be converted to numbers, an error is raised. Can be used as a filter.
   - You can optionally set the `edges` parameter to control how out-of-bounds input values are handled:
-    - `edges='clamp'` (the default) will clamp the output to the min/max output range.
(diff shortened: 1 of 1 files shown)


silamon (approved):

tmchow:
Appreciate the review on this one too.

PR #44538, opened 2026-04-06
Fix broken min/max examples in common template patterns

frenck:
## Proposed change

Fixes two incorrect examples in the "Finding the highest or lowest value" section on the Common patterns page, reported by community member Taras on the forum.

Bugs fixed:

- Warmest room example used max(attribute='state') which compares states as strings. This means "9" would beat "23" alphabetically, and "unknown" could be selected as the "warmest". Replaced with a namespace + loop approach that converts to float for correct numeric comparison while keeping the entity reference so the room name appears in the output.
- Both examples used rejectattr('state', 'in', ['unknown', 'unavailable']). Replaced with selectattr('entity_id', 'has_value') which is shorter, reads as intent, and is correct by definition (checks both states internally).

**Swapped the two examples so they build on each other:**

- Lowest battery (first, simpler): filter chain approach (map → min). You get the value but not which sensor it came from.
- Warmest room (second, builds on it): introduces namespace + loop when you need both the value AND the entity name. Opens with "The previous example gives you the value but not which sensor it came from."

Small change in the rejectattr, I've replaced it with an simpler has_value using the selectattr

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/templating/patterns.markdown
+++ b/source/_docs/templating/patterns.markdown
@@ -65,35 +65,47 @@ When you have several sensors of the same kind, you often want to pick out the e
 A [Group helper](/integrations/group/) can pick the minimum, maximum, or mean from a group of sensors and expose it as a regular sensor. No template needed if you only need the value.
 {% endtip %}
 
-**The warmest room:**
+**The lowest battery:**
+
+{% example %}
+template: |
+  {{
+    states.sensor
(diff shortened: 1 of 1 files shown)


c0ffeeca7 (approved):
Thank you, @frenck 👍

PR #44569, opened 2026-04-07
Align voice_control language support section with Voice Preview Edition

rayenchanchah01:
This update refreshes the “Check supported languages” section in the voice control page so it matches the current Voice Preview Edition content and stays in sync over time.


c0ffeeca7 (changes requested):
Thank you for contributing, @rayenchanchah01 👍 
Before this PR can be processed: 
- Please restore the original PR template and fill in the required information.
- In the description, outline the goal of this PR.
- Remove any changes that are outside the scope of this PR, such as the changes to the Gemfile, etc.

Image

PR #44686, opened 2026-04-11
Warn about DST pitfalls in history stats and template docs

### LINKED ISSUE #44211: local time date math can be wrong due to DST. utcnow() is a workaround.
### Feedback

There is an example in the history stats documentation that says:
...(now() + timedelta(hours=8))...
This is a bad example as it generates unexpected results during a DST transition. 
A simple workaround that generates the correct expected result is:
(utcnow() + timedelta(hours=8))
There is no good reason that all the examples should not be updated from using now() to utcnow() as it eliminates this issue,
To help you understand this issue, lets look at a simple example, assuming we are still in March 2026 when DST began:
```
{% set t1 = now().replace(day=8, hour=3, minute=1) %}
{% set t2 = (t1 - timedelta(hours=1)) %}
{% set ts1 = t1 | as_timestamp %}
{% set ts2 = t2 | as_timestamp %}
{{ t1 }}
{{ t2 }}
{{ ts1 }}
{{ ts2 }}
{{ ts1 == ts2 }}
```
```
2026-03-08 03:01:00.359657-07:00
2026-03-08 02:01:00.359657-08:00
1772964060.359657
1772964060.359657
True
```
So, in this example we see that using now() - timedelta(hours=1)) which is simply taking the current local time and subtracting 1 hour, we get the "wrong" time which is NOT one hour earlier.
But, if we instead were to use utcnow() - timedelta(hours=1))  then the time would be correct. 
So, my proposal is to update the documentation to always use utcnow() instead of now()

### URL

https://www.home-assistant.io/integrations/history_stats/

### Version

2026.3.2

### Additional information

Here is some more details including screen shots of this issue. 
https://github.com/home-assistant/core/issues/165115

frenck:
## Proposed change

Adds a daylight saving time warning to the History Stats integration page and the templating `Dates and times` reference, and rewrites the fragile `Next 4 pm` example so it no longer relies on the brittle `(now() + timedelta(hours=8)).replace(hour=16, ...)` pattern.

On DST-transition days, replacing the hour field of a timezone-aware local datetime can land in a skipped or repeated hour and silently produce the wrong absolute timestamp. The previous `Next 4 pm` example was exactly this pattern.

## Type of change

- [ ] Spelling, grammar or other readability improvements
- [x] Adjusted missing or incorrect information in the current documentation
- [ ] Added documentation for a new feature that has not been released yet, OR improves existing documentation
- [ ] Removed stale documentation

## Additional information

- Closes home-assistant/home-assistant.io#44211


source/_integrations/history_stats.markdown
```diff
@@ -242,12 +246,17 @@ Here, last Monday is today at 00:00, minus the current weekday (the weekday is 0
     end: "{{ today_at('00:00').replace(day=1) }}"
 ```
 
-**Next 4 pm**: 24 hours, from the last 4 pm till the next 4 pm. If it hasn't been 4 pm today, that would be 4 pm yesterday until 4 pm today. If it is already past 4 pm today, it will be 4 pm today until 4 pm tomorrow. When changing the start time, adjust the hour in `today_at()` accordingly.
+**Next 4 pm**: 24 hours, from the last 4 pm until the next 4 pm. If it is not yet 4 pm today, the period runs from 4 pm yesterday to 4 pm today. If it is already past 4 pm today, the period runs from 4 pm today to 4 pm tomorrow. When changing the start time, adjust the hour in `today_at()` accordingly.
```
silamon:
```suggestion
**Next 4 pm**: 24 hours, from the last 4 pm until the next 4 pm. If it is not yet 4 pm today, the period runs from 4 pm yesterday to 4 pm today. If it is at or after 4 pm today, the period runs from 4 pm today to 4 pm tomorrow. When changing the start time, adjust the hour in `today_at()` accordingly.
```

silamon (approved):

PR #44705, opened 2026-04-12
Clarify which Honeywell account to use during Lyric OAuth login

### LINKED ISSUE #44084: Not an easy process!
### Feedback

1. The documented steps do not make clear that the "new App" you need to create is on the Honeywell site, not in Home Assistant.

2. The "callback url" I entered got magically changed to my LAN IP address somehow.  But the error message did say "incorrect callback", so that I was able to correct easily.

3. The app api key and secret key cannot be seen or edited after they are entered, and if the integration hasn't been successfully created yet, there is no way to remove them. I didn't realize that I had 2 Honeywell apps and it was using the wrong one.

4. Biggest issue: when prompted to log in to the Honeywell api site in the final step of setup, login was repeatedly rejected. It turns out that the regular account login for the Honeywell thermostat and the Developer account where the app is set up use two totally separate logins. One naturally assuems that the Devekoper login is required at this step to access the api with the app key your just created and copied from the Developer site, but no! It's the regular Honeywell login that's needed. This is a huge stumbling block that NEEDS TO BE CLARIFIED.

### URL

https://www.home-assistant.io/integrations/lyric/

### Version

2026.3.1

### Additional information

_No response_

frenck:
## Proposed change

Adds an important note to the Honeywell Lyric integration page clarifying that the OAuth sign-in step requires your regular Resideo/Honeywell Home account, not the developer account. These are separate accounts even when they share the same email address, and the docs previously gave no indication of which one to use.

Also clarifies that the Client ID and Client Secret come from the app created on the Honeywell developer site, not from within Home Assistant.

## Type of change

- [ ] Spelling, grammar or other readability improvements
- [x] Adjusted missing or incorrect information in the current documentation
- [ ] Added documentation for a new integration I'm adding to Home Assistant
- [ ] Added documentation for a new feature I'm adding to Home Assistant
- [ ] Removed stale or deprecated documentation

## Additional information

- This PR fixes or closes issue: fixes #44084

## Checklist

### DIFF
--- a/source/_integrations/lyric.markdown
+++ b/source/_integrations/lyric.markdown
@@ -42,11 +42,15 @@ Internal examples: `http://192.168.0.2:8123/auth/external/callback`, `http://hom
 
 {% enddetails %}
 
-You can then add the integration in the frontend via the steps below.
+You can then add the integration in Home Assistant.
 
 {% include integrations/config_flow.md %}
 
-The integration configuration will ask for the **Client ID** and **Client Secret**, which correspond to the **Consumer** values in the app you created above. See [Application Credentials](/integrations/application_credentials) for more details.
+The integration setup will ask for the **Client ID** and **Client Secret**. These correspond to the **Consumer Key** and **Consumer Secret** values from the app you created on the Honeywell developer site. See [Application Credentials](/integrations/application_credentials) for more details.
+
+{% important %}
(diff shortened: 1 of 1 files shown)


timmo001 (approved):
Nice clarification! Hopefully solves some confusion 👍

PR #45552, opened 2026-05-24
Document Assist VAD timing settings

lextiz:
Documents the Assist VAD timing settings.

Core: home-assistant/core#172084


lextiz:
done

PR #45808, opened 2026-06-08
Document limitations of secrets in automations

karwosts:
## Proposed change

Add notes about using secrets in automations and scripts. I don't think this is mentioned currently anywhere in the documentation. 

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_docs/configuration/secrets.markdown
+++ b/source/_docs/configuration/secrets.markdown
@@ -76,5 +76,7 @@ automation ui: !include automation.yaml
 # These automations may contain secrets, and will be read-only in the UI
 automation secret: !include automation-secret.yaml
 ```
-{% caution %} Secrets used in automations will expose their secret value to admins when viewed in the frontend interface, in the automation source viewer and in the trace viewer. {% endcaution %}
+{% caution %}
+Secrets used in automations will expose their secret value to administrators when viewed in the UI, such as in the YAML source viewer and the trace viewer.
+{% endcaution %}
 
--- a/source/_docs/configuration/secrets.markdown
+++ b/source/_docs/configuration/secrets.markdown
@@ -71,10 +71,10 @@ Example `configuration.yaml`:
 
 ```yaml
 # The main automations editable in the UI
-automation ui: !include automation.yaml
+automation ui: !include automations.yaml
 
 # These automations may contain secrets, and will be read-only in the UI
-automation secret: !include automation-secret.yaml
+automation secret: !include automations-secret.yaml
 ```
 {% caution %}
 Secrets used in automations will expose their secret value to administrators when viewed in the UI, such as in the YAML source viewer and the trace viewer.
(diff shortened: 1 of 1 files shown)


source/_docs/configuration/secrets.markdown
```diff
@@ -60,3 +60,23 @@ logger: debug
...
+```yaml
+# The main automations editable in the UI
+automation ui: !include automations.yaml
+
+# These automations may contain secrets, and will be read-only in the UI
+automation secret: !include automations-secret.yaml
+```
+{% caution %}
+Secrets used in automations will expose their secret value to administrators when viewed in the UI, such as in the YAML source viewer and the trace viewer.
+{% endcaution %}
```
c0ffeeca7, line 81:
```suggestion
{% important %}
Secrets used in automations will expose their secret value to administrators when viewed in the UI, such as in the YAML source viewer and the trace viewer.
{% endimportant %}
```
According to ANSI Z535.6 (a standard on how to document hazards in technical documentation), "caution" is a signal word used to identify a hazard that may or will cause injury but not death. For this reason, I would not use "caution" in this context. A note or a tip would work, or, if you want to highlight it a bit more, "important". 

https://en.wikipedia.org/wiki/ANSI_Z535

c0ffeeca7 (changes requested):
tiny tweak

source/_docs/configuration/secrets.markdown
karwosts: Hmm, this rule feels somewhat inconsistently applied. I want to warn users that their secrets may be exposed, so I think we need to have some kind of alert that implies a negative connotation.

According to these copilot instructions, maybe this should be a warning, but a warning is also the class that it says implies physical harm, not caution.

Copilot instructions says:

https://github.com/home-assistant/home-assistant.io/blob/9ef0acd7f0cd9c39063ec16f80b9763aaf5df6f0/.github/copilot-instructions.md?plain=1#L349-L359


Our other current uses of caution include:
 - your password might be exposed
 - your automation may break
 - "This is an advanced feature"
 - your secret code may be sent over the network
 - not recommended to run a serial proxy over wifi
 - a feature is not available before a certain version. 
 
 None of these necessarily imply physical injury either. 
 
 Can we review a bit here and figure out if our alert categories and copilot instructions are accurate?
 
 To use 'important' here I feel doesn't fit the posted guidelines, as I think this is more of a security issue.
 
 If after reading this you still disagree, feel free to merge your suggestion, as I will defer to your judgement.

c0ffeeca7: Good point. Thank you for pointing that out. Your PR follows the guidelines. I disagree with that guideline. I don't think it's correct, but maybe I'm missing something.

So let me do the following: 
Merge your PR as is, check the guidelines, and then, if the guidelines should change, we can standardize the hazard statements throughout the docs.

c0ffeeca7 (approved):
Thank you, @karwosts 👍

PR #46355, opened 2026-06-21
Update Nest GCP setup steps and refresh screenshots

### LINKED ISSUE #38367: Workflow for Nest and GCP outdated
### Feedback

Hi

I am new to both github and HA so I hope the below makes sense! 

I have just run through the workflow to add my Nest thermostat to Home Assistant. The documentation looks good but slightly outdated.  Not sure if this is due to GCP updates changing the process slightly.  

1.) Adding the pub/sub Publisher permissions isn't possible when adding a new principal. When assigning the pub/sub role the Publisher role is not available to choose from. I chose both "Editor" and "Admin" instead and neither worked.  You have to assign your principal the pubsub.Publisher permissions using IAM instead which involves a few more steps.  If this isn't done you cannot create a topic and allow the pub/sub Topic to become enabled.

2.) At the end of the process when GCP is configured and you are linking the account in HA the output in the screenshots are different / not displayed now.  I guess again that this is due to changes within Google.

Regards
Eddie

### URL

https://www.home-assistant.io/integrations/nest/

### Version

2025.4.0

### Additional information

_No response_

1waterrj:
## Proposed change

Updates the Google Nest integration documentation to match the current Google Cloud Platform and account-linking flow. Addresses the two problems reported in #38367:

**1. Pub/Sub topic permissions (Cloud Console):**
- Clarified the navigation to grant the SDM publisher access: open the topic's **Permissions** tab → **Add principal**.
- Corrected the button label (**Create** → **Save**) and noted the **Pub/Sub Publisher** role can be found by browsing under **Pub/Sub** or via the search box.
- Added a tip to grant the role on the topic (not the project **IAM & Admin** page) plus a `gcloud` alternative.
- Refreshed the create-topic, add-principal, and role-selection screenshots.

Note: I could **not** reproduce the report's "Publisher role isn't available when adding a principal" behavior as of June 2026 — the role is still selectable. So the permission section was corrected for accuracy rather than rewritten around that bug.

**2. Account-linking screenshots:**
- Google replaced the old two-step grant flow (and the separate "Confirm your choices" screen) with a **single consent screen**, so the steps and screenshots no longer matched.
- Refreshed `oauth_approve`, `oauth_app_verification`, `oauth_grant1`, `config_flow_with_topic`, and `config_flow_subscription`.
- Updated step 5 to use **Advanced → Go to home-assistant.io (unsafe)**, merged the grant/confirm steps into one **Continue** step, renumbered, and removed the now-unused `oauth_grant2` and `oauth_confirm` images.
- Fixed several incorrect image alt-texts.

All new screenshots have personal data (project IDs, account email, home/device names) redacted.

## Type of change

## Additional information

- This PR fixes or closes issue: fixes #38367

## Checklist

- [x] I have verified that the preview generated by Netlify looks correct.


source/_integrations/nest.markdown
```diff
@@ -215,15 +215,26 @@ If you previously set up events, then your Device Access Project may have alread
...
-7. In **New Principals** enter `sdm-publisher@googlegroups.com`
+7. In **New principals**, enter `sdm-publisher@googlegroups.com`.
 
-8. In **Select a Role** under **Pub/Sub** select **Pub/Sub Publisher** and **Create**.
+8. In the **Assign roles** section, select the **Role** dropdown. Under **Pub/Sub**, select **Pub/Sub Publisher** (or type `Pub/Sub Publisher` in the search box to find it). Select **Save**.
 
-  ![Screenshot of OAuth confirmation](/images/integrations/nest/cloud_pubsub_add_principal_role.png)
+  ![Screenshot of selecting the Pub/Sub Publisher role](/images/integrations/nest/cloud_pubsub_add_principal_role.png)
+
+  {% tip %}
```
allenporter, line 228: Is this here because the other way no longer works or just as an alternative? If just an alternative, I don't think we should introduce `gcloud` here in these docs and folks that want to do this entirely themselves can just follow https://developers.google.com/nest/device-access/subscribe-to-events (linked above?)

allenporter (changes requested):
Thank you for making improvements to the documentation.

Can the screenshots please be in light mode so we're not switching back and forth between dark mode and light mode?

source/_integrations/nest.markdown
1waterrj: Just an alternative — the UI path works fine (the **Pub/Sub Publisher** role is still selectable when adding the principal), so I've removed the `gcloud` block as you suggested. Anyone wanting the CLI route can follow the linked Google doc. I kept a one-line note to grant the role on the topic's **Permissions** tab rather than the project **IAM & Admin** page, since that's the step the original issue tripped on — happy to drop that too if you'd prefer.

1waterrj:
Thanks for the review! Addressed both points:

- **Light mode:** recaptured all six dark screenshots (the three Pub/Sub Console shots, the consent screen, and the two config-flow dialogs) in light mode, so the page no longer switches themes.
- **gcloud:** removed the CLI block (it was only an alternative; the UI path works), keeping a short note to grant the role on the topic's Permissions tab rather than the project IAM page.

Also folded in the Copilot suggestions (wording/alt-text). Ready for another look.

1waterrj:
Updated!

allenporter (commented):
This seems good though a lot of the images are updated that don't really seem like significant changes. They just look different for you because of your devices or setup. Do you agree?

1waterrj:
> This seems good though a lot of the images are updated that don't really seem like significant changes. They just look different for you because of your devices or setup. Do you agree?

Agreed but I believe it created a more consistent setup guide.

allenporter (approved):
OK i really do appreciate this update and don't want to nit pick, but also there are other things changed in the screenshots e.g. the project name redaction has changed:

The original redaction is really intentionally still showing the prefix `projects/` for example because of the way the old SDM topics were set up were owned by sdm-prod and we want to make sure people are now using their own topics.

Do you think we'd be able to do a version that keeps some of that information? I could also agree we just merge this then fix it in a follow up.

1waterrj:
> OK i really do appreciate this update and don't want to nit pick, but also there are other things changed in the screenshots e.g. the project name redaction has changed:
> 
> The original redaction is really intentionally still showing the prefix `projects/` for example because of the way the old SDM topics were set up were owned by sdm-prod and we want to make sure people are now using their own topics.
> 
> Do you think we'd be able to do a version that keeps some of that information? I could also agree we just merge this then fix it in a follow up.

Reviews like this are what keep HA great! I'll get something out ASAP. Really appreciate the input.

PR #46540, opened 2026-06-29
Update Anova codeowners

rhys-saldanha:
Adds `@rhys-saldanha` as codeowner for the Anova integration, to match the update in home-assistant/core#175146.


rhys-saldanha:
> It seems that this PR is targeted against an incorrect branch since it has a parent PR on one of our codebases. Documentation that needs to be updated for an upcoming release should target the `next` branch. Please change the target branch of this PR to `next` and rebase if needed.

Updated to `next`

edit: wow that was a bad idea, PR recreated

PR #46818, opened 2026-07-13
Correct Flux color temperature accepted range to 40000 K

TowyTowy:
## Proposed change

The Flux integration docs state that the accepted color temperature range is "between 1000 and 4000 kelvin", but the code validates all three color-temperature options (`start_colortemp`, `sunset_colortemp`, `stop_colortemp`) with `vol.Range(min=1000, max=40000)` in [`homeassistant/components/flux/switch.py`](https://github.com/home-assistant/core/blob/dev/homeassistant/components/flux/switch.py). A user following the docs would wrongly believe a value like 6500 K is out of range.

This updates the documented upper bound to 40000 K to match the code. (The `4000` values elsewhere on the page are the *default* `start_colortemp`/`stop_colortemp`, which are unaffected.)

## Type of change

- [x] Bugfix (non-breaking change which fixes an issue)

## Additional information

- Documentation only; corrects the stated accepted range to match `vol.Range(min=1000, max=40000)` in the integration code.

## Checklist

- [x] This change is fixing an incorrectly documented value (verified against the integration code).

Disclosure: this fix was prepared with AI assistance (Claude) and verified by me against the integration source.

### DIFF
--- a/source/_integrations/flux.markdown
+++ b/source/_integrations/flux.markdown
@@ -23,7 +23,7 @@ During the day (in between `start_time` and `sunset_time`), it will fade the lig
 
 The value of `sunset_time` is automatically calculated based on the location specified in your [Home Assistant configuration](/docs/configuration/basic).
 
-The color temperature is specified in kelvin, and accepted values are between 1000 and 4000 kelvin. Lower values will seem more red, while higher will look more white.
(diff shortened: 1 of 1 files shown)


c0ffeeca7 (approved):
Thank you, @TowyTowy 👍 

For future PRs, please make sure to use the original PR template (see PULL_REQUEST_TEMPLATE.md)

PR #46928, opened 2026-07-19
Nest: fix Pub/Sub screenshot redaction to preserve the projects/ topic path

1waterrj:
Follow-up to #46355.

As @allenporter noted in review, the refreshed screenshots redacted the entire Pub/Sub path, which hid the intentional `projects/` prefix and `/topics/<name>` structure (the original redaction deliberately kept those visible so users configure a topic under **their own** project rather than the old shared SDM project).

This re-captures the three topic-path screenshots so only the project ID is masked, keeping the path structure readable:

- `cloud_pubsub_create_topic.png` — `Topic name: projects/▇▇▇/topics/home-assistant-nest`
- `config_flow_with_topic.png` — `projects/▇▇▇/topics/home-assistant-nest`
- `config_flow_subscription.png` — `projects/▇▇▇/topics/home-assistant-nest` and `.../subscriptions/home-assistant-nest-sub`

All in light mode, consistent with the surrounding screenshots.


1waterrj:
Thanks for merging #46355! This is the redaction follow-up you flagged — I re-shot the three topic-path screenshots so only the project ID is masked and the `projects/.../topics/<name>` (and `.../subscriptions/<name>`) structure stays visible, per your note about keeping people pointed at their own topic rather than the old shared SDM one. Light mode, consistent with the rest.

allenporter (approved):
Thank you!

PR #46938, opened 2026-07-20
Document UniFi option to ignore private (randomized) MAC addresses

tomwilkie:
## Proposed change

Documents the new opt-in UniFi Network option added in home-assistant/core#174503, which stops the integration creating entities for Wi-Fi clients that connect with a private (randomized) MAC address.

- Adds the option to the **Configuration options** list.
- Rewrites the presence detection note about MAC address randomization. It previously said presence detection "is not compatible with Client MAC Address Randomization" and only offered "turn the feature off on every device". It now explains why tracking breaks and covers both ways to handle it, including the new option.

## Type of change

- [x] Document existing feature (new feature in an existing integration)

## Additional information

- Related core PR: https://github.com/home-assistant/core/pull/174503
- Related discussion: https://github.com/orgs/home-assistant/discussions/4118

## Checklist

- [x] This PR targets the `next` branch, since it documents a feature that has not been released yet.

### DIFF
--- a/source/_integrations/unifi.markdown
+++ b/source/_integrations/unifi.markdown
@@ -115,6 +115,8 @@ Time in seconds from last seen until considered away:
   description: "Number of seconds since last seen before a client is considered away. Defaults to `300` seconds."
 Disable UniFi Network wired bug logic:
   description: "Disable the workaround for a UniFi Network bug that sometimes reports wired clients as wireless."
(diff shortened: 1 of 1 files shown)


Kane610 (approved):
Thanks!

tomwilkie:
Thank you!

PR #47207, opened 2026-08-02
Fix UniFi Access blueprint triggers to ignore unavailable/unknown states

imhotep:
## Summary
- Both UniFi Access blueprints (`unifi_access_doorbell_notification.yaml` and `unifi_access_door_access_notification.yaml`) used a generic `state` trigger on the event entity, which fires on *any* state change — including transitions to/from `unavailable`/`unknown` (e.g. on a Home Assistant restart or a WebSocket reconnect to the UniFi Access controller).
- The doorbell blueprint had no condition to filter these out, so it could send a false "someone's at the door" notification without an actual ring event.
- Switched both to the `event.received` trigger, which only fires for the specified event type(s) and ignores `unavailable`/`unknown` transitions. This also let us drop the extra `condition: state` block from the door access blueprint, since event type filtering now happens at the trigger.
- Bumped `min_version` from `2026.4.0` to `2026.7.0`, since purpose-specific triggers like `event.received` graduated from Labs to the default automation editor in the 2026.7 release.

## Test plan
- [ ] Validated blueprint YAML syntax
- [ ] Imported both blueprints into a test Home Assistant instance with UniFi Access and confirmed the doorbell/access triggers fire only on real events, not on restarts/reconnects

### DIFF
--- a/source/blueprints/integrations/unifi_access_door_access_notification.yaml
+++ b/source/blueprints/integrations/unifi_access_door_access_notification.yaml
@@ -7,7 +7,7 @@ blueprint:
   domain: automation
   author: RaHehl
   homeassistant:
-    min_version: 2026.4.0
+    min_version: 2026.7.0
 
   input:
     access_entity:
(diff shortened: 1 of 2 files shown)


RaHehl (approved):

fdmarcin:
@imhotep Hi! Nobody has reviewed this PR because it targets `next` while also not having a "parent" PR attached, so it fell through the cracks.

If this isn't related to anything specific added with the upcoming release, could you rebase this branch and retarget the PR against `current`? Or if there's a related PR in Core, please add it to the description.

Thanks!

imhotep:
Rebased onto `current` and retargeted — this isn't tied to a specific upcoming Core release, and `event.received` is already available in the stable 2026.7+ releases, so `current` is the right target. Thanks for the heads up!

Laufmaschine (approved):
Thank you, @imhotep!

PR #47283, opened 2026-08-07
Fix code owner check in the task creation restriction

balloob:
## Proposed change

The `Restrict task creation` workflow authorizes the author of a new **Task** issue in two steps: organization membership first, and integration code ownership as a fallback. The second step has two problems.

**The CODEOWNERS file was fetched from a branch that does not exist.** The lookup used `ref: 'dev'`, but this repository has `current` and `next`. Every request returned a 404, which the surrounding `catch` swallowed into a `console.error`, so the code owner path never authorized anyone and organization membership was effectively the only way to pass. A code owner opening a Task issue had it closed automatically, with nothing but a log line to explain why.

This now reads from `next` rather than `dev`. Reading from `next` also covers code owners of integrations that have not been released yet, since their documentation lands there first.

**The author was matched with a substring search.** The check was `codeownersContent.includes('@' + issueAuthor)`, which passes for any username that happens to be a prefix of a listed code owner. For example, a `@shred` account would match the code owner entry `@shred86`. The handles are now parsed out of the file and compared in full, case-insensitively, since GitHub usernames are not case-sensitive. Team entries such as `@home-assistant/core` are skipped, because their members already pass the organization check above.

Verified the parser against the current CODEOWNERS file: it extracts 716 user handles, matches every real code owner including on differing case, and rejects prefixes, team names, and unknown users.

## Type of change

None of the above apply: this changes repository automation in `.github/workflows`, not documentation content.

## Additional information

One behavior is left unchanged and may be worth a follow-up: a failure to fetch CODEOWNERS is still caught and falls through to closing the issue, so a temporary API problem closes a legitimate contributor's Task issue with only a log line as a record. Fixing the branch reference makes that far less likely, but the silent fail-closed path is still there.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_015jLcLJXh97STLm8MkUmKCA)_

### DIFF
--- a/.github/workflows/restrict-task-creation.yml
+++ b/.github/workflows/restrict-task-creation.yml
@@ -55,8 +55,6 @@ jobs:
                 }
               }
 
-              // Compare complete handles, so a username that is only a prefix of an
-              // owner does not pass. GitHub usernames are case-insensitive.
               if (owners.has(issueAuthor.toLowerCase())) {
                 console.log(`✅ ${issueAuthor} is an integration code owner`);
                 return; // Authorized
--- a/.github/workflows/restrict-task-creation.yml
+++ b/.github/workflows/restrict-task-creation.yml
@@ -31,20 +31,33 @@ jobs:
 
             // If not an org member, check if they're a codeowner
             try {
-              // Fetch CODEOWNERS file from the repository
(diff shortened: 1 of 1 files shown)


Laufmaschine (approved):
Thank you, @balloob!

PR #47304, opened 2026-08-08
Rewrite the ZhongHong page for the config entry setup

crhan:
## Proposed change

Rewrite the ZhongHong page for the config entry setup added in home-assistant/core#178506.

The page currently documents only the YAML climate platform, which that PR replaces with a config flow. The three settings are unchanged, so they are now described as the fields of the setup dialog, and the note about YAML says what happens to an existing configuration: it is imported on the first start after the upgrade, and the `zhong_hong` platform can then be removed from the `climate:` block.

While rewriting, the page also picks up the sections the documentation standard asks for, all of which it was missing: prerequisites, supported functionality, known limitations, troubleshooting, and removal.

Two of the limitations are worth calling out, because both cost me time on my own installation and neither was written down anywhere:

- The gateway accepts a single TCP connection at a time. Anything else left connected to it — a phone app, a second Home Assistant — keeps setup from succeeding, and the resulting error says nothing about the real cause.
- The protocol addresses five fan speeds but offers no way to ask an indoor unit which of them it has. A unit with three simply ignores the other two, so all five are offered on every entity.

## Type of change

## Additional information

- Link to parent pull request in the codebase: home-assistant/core#178506


Laufmaschine (approved):
Thanks, @crhan.

PR #47387, opened 2026-08-13
Optimize build performance with domain index generator and static data file

mrdarrengriffin:
## Proposed change

Builds of the website have been getting slower as content grows — slow enough to sometimes crash devcontainers. This PR profiles the build, fixes the hotspots, and cuts a clean build from **~10.5 minutes to ~2.3 minutes (4.6× faster)**, peak memory from **5.0 GB to 0.67 GB**, and generated output from **2.8 GB to 814 MB**. Every visitor also downloads ~517 KB less HTML per page view.

Measured with `jekyll build --profile` on identical hardware:

| | Before | After |
|---|---|---|
| Clean build time | 627 s | ~135 s |
| Peak memory | 5.0 GB | 0.67 GB |
| Output size | 2.8 GB | 814 MB |

### What was slow, and what changed

1. **Sidebar domain lookups scanned all integrations, per domain, per page (54% of build time).** The action/trigger/condition sidebars resolve each domain label with `site.integrations | where: 'ha_domain', ... | first` — a linear scan over 1,500+ integration documents, repeated for all ~280 domains on each of the 985 action pages (~420 million lookups). A new generator (`plugins/domain_index.rb`) now computes the domain→label/icon index once per build, and the sidebars, meta sections, and index pages read from it. It still honors the optional `_data/<kind>_domains.yml` files if they are added later.

2. **~517 KB of JSON was inlined into every page (73% of output size).** The action/trigger/condition/template-function lookup data used by the code-block tooltips was embedded in all 4,079 pages. A new generator (`plugins/doc_data_file.rb`) writes it once to a content-hashed static file (`/static/doc-data-<hash>.js`) that browsers cache across page views.

3. **One page took 19 seconds by itself.** `lg_tv_rs232.markdown` was the only page using `{% link %}`, which iterates every file in the site per call. It now uses plain documentation paths like every other page.

4. **Smaller wins:** the `cache_buster` filter now hashes each asset once per build instead of once per page (~37,000 redundant file reads removed), and the table of contents is rendered once per page instead of twice.

## Type of change

- [ ] Spelling, grammar or other readability improvements (`current` branch).
- [ ] Adjusted missing or incorrect information in the current documentation (`current` branch).
- [ ] Added documentation for a new integration I'm adding to Home Assistant (`next` branch).
  - [ ] I've opened up a PR to add logos and icons in [Brands repository](https://github.com/home-assistant/brands).
- [ ] Added documentation for a new feature I'm adding to Home Assistant (`next` branch).
[...]

PR #47392, opened 2026-08-13
Serve doc tooltip data as a static file instead of inlining it into every page

mrdarrengriffin:
## Proposed change

Split out of #47387 (2 of 5). This is the big output-size and memory win.

The action, trigger, condition, and template function lookup data used by the code-block tooltips (`window.__actions` and friends) was inlined into every generated page — about 517 KB each. Across 4,079 pages that was **~2 GB of the 2.8 GB build output**, it drove the build's peak memory toward 5 GB (the cause of devcontainer crashes), and every visitor re-downloaded the data on each page view.

A new generator (`plugins/doc_data_file.rb`) writes the data once to a content-hashed static file (`/static/doc-data-<hash>.js`) that browsers can cache forever while still picking up new data immediately after a deploy. The file loads with a deferred script tag placed before `prism-template-links.js` and `prism-doc-links.js`, so per the HTML spec's deferred-script ordering it executes before its consumers, exactly like the old inline block.

## Type of change

None of the listed types apply: this is a build performance change with no intended change to page behavior.

## Additional information

- Split from #47387, which has the full investigation and combined measurements.
- Regression tested as part of #47387:
  - The external data file on the deploy preview is **byte-identical** (529,338 bytes) to the JSON the live site inlines.
  - Both variants were loaded in headless Chromium: data dictionaries populated (985 actions / 192 triggers / 149 conditions / 202 template functions), linked code-block tokens, parameter tooltips, and glossary tooltips all have identical counts, with zero JavaScript errors — confirming the deferred file loads before the scripts that consume it.
- Trade-off for reviewers: one extra HTTP request on first page view, in exchange for ~517 KB less HTML on every page view after that.

## Checklist

https://claude.ai/code/session_015zeLdNaEP3Bq8akPj4b76d

---
_Generated by [Claude Code](https://claude.ai/code/session_015zeLdNaEP3Bq8akPj4b76d)_

### DIFF
--- a/plugins/doc_data_file.rb
+++ b/plugins/doc_data_file.rb
@@ -1,0 +1,45 @@
+require 'digest/md5'
+
+# Write the generated template function, action, trigger, and condition
+# lookup data to a single static JavaScript file instead of inlining it
+# into every page.
+#
+# The four JSON blobs together are roughly 500 KB. Inlined through
+# scripts.html they were repeated in every generated page, which added
+# gigabytes to the build output and made every visitor download the data
+# again on each page view. As a static file it is generated once and
+# cached by the browser.
+#
+# The file name carries a content hash, so browsers can cache it forever
+# and still pick up new data immediately after a deploy. scripts.html
+# reads the URL from site.data['doc_data_js_path'].
+#
+# Runs at priority :lowest so it executes after the :low generators that
+# produce the JSON (doc_collections_data.rb and template_functions_data.rb).
+module Jekyll
+  class DocDataFileGenerator < Generator
+    safe true
+    priority :lowest
+
+    def generate(site)
+      content = <<~JS
+        window.__templateFunctions = #{site.data['template_functions_json']};
+        window.__actions = #{site.data['actions_json']};
+        window.__triggers = #{site.data['triggers_json']};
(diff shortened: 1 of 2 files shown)


plugins/doc_data_file.rb
```diff
@@ -0,0 +1,45 @@
+require 'digest/md5'
+
+# Write the generated template function, action, trigger, and condition
+# lookup data to a single static JavaScript file instead of inlining it
+# into every page.
+#
+# The four JSON blobs together are roughly 500 KB. Inlined through
```
balloob: We don't need to include in the comments what we did before. Just what we're doing now.

plugins/doc_data_file.rb
```diff
@@ -0,0 +1,45 @@
...
+# scripts.html they were repeated in every generated page, which added
+# gigabytes to the build output and made every visitor download the data
+# again on each page view. As a static file it is generated once and
+# cached by the browser.
+#
```
balloob:
```suggestion
```

balloob (approved):
This is a good improvement. I think that we can push it further in a future PR by only loading this file when the data is actually needed.

PR #47538, opened 2026-08-20
Rain Bird: correct the 2.x limitation, clarify the controller password

ayanich:
## Proposed change

Two corrections to the Rain Bird integration documentation.

**Known limitations.** The note saying the Rain Bird 2.0 app and firmware are not compatible is out of date. Controllers updated by the 2.x app serve their local API over HTTPS rather than HTTP, and Home Assistant 2026.3.1 or later tries both, so they are supported and still controlled locally with no cloud requirement.

**Password.** The field now identifies the credential as the controller password, set when the WiFi controller was first configured and distinct from the Rain Bird account password. It is 4-8 letters or numbers; the 2.0 app presents it as a 6-digit PIN. Adds links to the Rain Bird articles covering what it protects, how to change it, and what to do when it is forgotten.

## Type of change

## Additional information

### DIFF
--- a/source/_integrations/rainbird.markdown
+++ b/source/_integrations/rainbird.markdown
@@ -50,8 +50,15 @@ irrigations schedules on a calendar.
 Host:
   description: "The IP address of your Rain Bird device. You can find the IP address under the device in the Rain Bird app under **Controller Settings** > **Network Info**."
 Password:
-  description: "The password used to authenticate the Rain Bird device."
+  description: "The **controller password**, which you set when the Rain Bird WiFi controller was first configured. This is not the password for your Rain Bird account. It is 4-8 letters or numbers; if you set it up with the Rain Bird 2.0 app, it is the 6-digit PIN."
 {% endconfiguration_basic %}
+
+The controller password is specific to the controller and protects its network
+and sharing settings. Rain Bird documents it here:
+
+- [WiFi controller password and what it protects](https://wifi.rainbird.com/articles/wifi-controller-password-and-what-it-protects/)
+- [How do I change my controller password?](https://wifi.rainbird.com/articles/how-do-i-change-my-controller-password/)
+- [Forgotten controller passwords](https://wifi.rainbird.com/articles/forgotten-controller-passwords/)
 
 ## Configuration options
 
@@ -107,9 +114,10 @@ The Rain Bird integration provides the following entities.
 
 ## Known Limitations
 
-The new Rain Bird 2.0 App and Firmware is not compatible with Home Assistant.
-The upgrade process will migrate devices to require use of the new Rain Bird
-IQ4 cloud, and Home Assistant will not be able to access the device.
+Controllers updated by the Rain Bird 2.x app and newer firmware serve their local
+API over HTTPS instead of HTTP. Home Assistant 2026.3.1 or later tries both, so
+these controllers are supported and are still controlled locally, with no cloud
+requirement. On earlier versions, setup fails with a connection error.
 
 The Rain Bird LNK WiFi can only receive one incoming request at a time. It may
 not be possible for Home Assistant to send commands to the device while you


ayanich:
Retargeted to `next` and rebased as requested.

One note for whoever reviews: the bot flagged this because I had filled in "Link to parent pull request in the codebase". That was my mistake - home-assistant/core#163915 is not a parent PR for an upcoming release, it shipped in **2026.3.1** back in March. I've cleared that field, since this is a correction to already-published documentation rather than docs for an unreleased change.

So `current` may actually be the more appropriate target if you'd like the correction live sooner - happy to move it back, or to leave it here for the next release. Either is fine by me; I didn't want to silently re-flip the base after the bot asked.

source/_integrations/rainbird.markdown
```diff
@@ -50,7 +50,7 @@ irrigations schedules on a calendar.
 Host:
   description: "The IP address of your Rain Bird device. You can find the IP address under the device in the Rain Bird app under **Controller Settings** > **Network Info**."
 Password:
-  description: "The password used to authenticate the Rain Bird device."
+  description: "The password used to authenticate the Rain Bird device. This is the 6-digit PIN that was set when the controller was first configured, not the password for your Rain Bird app account. It is not displayed anywhere in the Rain Bird app."
```
allenporter: It looks like Rainbird may call this the "controller password" so we can use that terminology. 

Would it be helpful if we link to any of this documentation for users that don't know it?

https://wifi.rainbird.com/articles/wifi-controller-password-and-what-it-protects/
https://wifi.rainbird.com/articles/how-do-i-change-my-controller-password/
https://wifi.rainbird.com/articles/forgotten-controller-passwords/


Also, users may be using the 1.0 or 2.0 app so it can help to clarify: "When configured with the Rainbird 2.0 App, this is the 6-digit PIN" since according to https://wifi.rainbird.com/articles/controller-password-no-special-characters/ it may be 4-8 letters or numbers.

allenporter (changes requested):

ayanich:
Thanks — you were right on both counts, and one of them was an error on my part.

**Controller password.** Switched to Rain Bird's own terminology throughout. I had written "the 6-digit PIN" unconditionally, which was wrong: I generalised it from a single report in home-assistant/core#162671. Rain Bird's own article says the password is ["restricted to letters and numbers, 4-8 digits in length"](https://wifi.rainbird.com/articles/controller-password-no-special-characters/), so the PIN is how the 2.0 app presents it rather than the rule. The field now reads:

> The **controller password**, which you set when the Rain Bird WiFi controller was first configured. This is not the password for your Rain Bird account. It is 4-8 letters or numbers; if you set it up with the Rain Bird 2.0 app, it is the 6-digit PIN.

**Links.** Added all three, after the `{% raw %}{% configuration_basic %}{% endraw %}` block rather than inside the description string, so they render as a readable list.

**Grammar/version wording.** Also fixed per the Copilot comment — "and later probe both" → "or later tries both", matching the "or later" phrasing used elsewhere in the docs.

Rebased onto `next` while I was in there.

On the branch question from earlier: happy either way. core#163915 shipped in 2026.3.1 back in March, so this is a correction to already-published docs rather than docs for an unreleased change — `current` would get the fix live sooner, but I left it on `next` since that is what the bot asked for.

allenporter (changes requested):
Looks good.  I agree with you that this belongs on `current`. The reason the bot suggested next is because it references a specific PR, and typically we want those to go into next since it is assuming the PR is new.

allenporter:
Remove the references to the PRs to get the bot to stop doing that. It's not really needed anyway. You can simplify the description as it looks more like a conversation history than a simple description of the PR.

allenporter (approved):

ayanich:
Done - description simplified and the PR references removed.

PR #47746, opened 2026-08-30
Correct the Lyngdorf network requirements, add per-model features and blueprints

fishloa:
## Proposed change

The same-network requirement was wrong. The serial lookup is a unicast SSDP
M-SEARCH to the device rather than multicast, so a device on another subnet
works fine when added by IP address — the old wording told those users they were
unsupported. Only automatic discovery needs the same subnet.

The prerequisites now name the ports that are fixed, TCP 84 for the control
protocol and UDP 1900 for the SSDP request, and explain that the UPnP
description is served from a port the device assigns itself, so a firewall rule
cannot rely on a particular number. Measured against an MP-60, which serves it
from a high device-assigned port. This also matters in troubleshooting, where
"add the device manually using its IP address" is offered as the fix for a
discovery problem and only works if UPnP is reachable — that step now says so.

The two brand lists under supported devices are replaced by one table of which
entities each model gets. The models differ more than the old list suggested:
the P-series has Zone B, a remote and lip sync but no streaming or trims, while
the TDAI-series has streaming and tone trims but none of the rest, and the
TDAI-2170 gets only a main zone media player and the two RoomPerfect selects.
Every value comes from the library's own per-model capability table.

Also adds TDAI-2210, which the library has always supported but the page never
listed, links every model, and replaces the claim that only the MP-60 has been
tested — the TDAI-1120 and TDAI-3400 both have owners running it, and two of the
library fixes came from them.

The inline examples are replaced by three blueprints under
`source/blueprints/integrations/`, linked with import badges:

- Switches voicing while watching a source
- Applies lip sync delay per source
- Runs actions when surround audio starts

The remote script example is dropped; `remote.send_command` is already covered
under supported functionality.

## Type of change

## Additional information

Targeted at `next` rather than `current` because the sections being corrected
only exist there: the platform documentation for 2026.9 landed on `next` in
#47477, and the corrected paragraphs are part of it.

- Link to parent pull request in the codebase: https://github.com/home-assistant/home-assistant.io/pull/47477

## Checklist

### DIFF
--- a/source/_integrations/lyngdorf.markdown
+++ b/source/_integrations/lyngdorf.markdown
@@ -32,24 +32,22 @@ The **Lyngdorf** {% term integration %} allows you to control [Lyngdorf] and [St
 
 ## Supported devices
 
-Every model gets a main zone media player and the RoomPerfect position and
-voicing selects. The rest depends on what the model has:
+Every model gets a main zone media player and the RoomPerfect position and voicing selects. The rest depends on what the model has:
 
 | Model | Zone B | Now playing and transport | Remote | Lip sync | Bass and treble trims | Channel trims | Input sensors |
 | ----- | ------ | ------------------------- | ------ | -------- | --------------------- | ------------- | ------------- |
-| [MP-40](https://lyngdorf.steinwaylyngdorf.com/lyngdorf-mp-40/) | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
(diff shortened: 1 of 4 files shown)


fishloa:
The sections this corrects only exist on `next`. The feature table describes the number, remote, select, and sensor entities added for 2026.9 in #47477, which `current` does not document — it is still media_player-only. Retargeting would not apply cleanly.

silamon (approved):

PR #47805, opened 2026-09-02
Update groupby function examples...

jack3308:
Fixing example outputs provided in the docs that do not function as described.

## Proposed change

The current example for the groupby function in the docs utilizes an attribute that doesn't exist for the domain in the example.

Light entities don't have their area's stored in an attribute accessible by groupby - the example provided still technically works for what it's trying to communicate, it's just all of the lights will show up as the default instead of only some of them. 

A better fix would be to use an attribute that likely exists for some (but not all) of them (like brightness or similar...) - but this works for now.

## Type of change

## Additional information

## Checklist


fdmarcin:
@jack3308 Hi, thanks for the fix! Would you like to edit this to use another attribute that shows the grouping, like you said in the description?

source/_template_functions/groupby.markdown
```diff
@@ -93,9 +93,7 @@ template: |
   {% endfor %}
 type: string
 output: |
-  Kitchen: light.kitchen_ceiling, light.kitchen_counter
-  Living room: light.living_room_main
-  Unknown: light.unnamed_bulb
+  Unknown: light.unnamed_bulb, light.kitchen_ceiling, light.kitchen_counter, light.living_room_main
```
klaasnicolaas, line 96: Would it make sense to use `attributes.device_class` on sensors for this example instead? That gives us a real optional attribute and demonstrates the `default` parameter more clearly, since some sensors can have a device class while others may not.

For example:

```jinja
{% for device_class, entities in expand("sensor.home_sensors")
  | groupby("attributes.device_class", default="Other") %}
  {{ device_class }}: {{ entities | map(attribute="entity_id") | join(", ") }}
{% endfor %}
```

With an example output such as:

```text
humidity: sensor.bathroom_humidity
temperature: sensor.kitchen_temperature, sensor.living_room_temperature
Other: sensor.daily_energy
```

That way the example shows both normal grouping and the fallback behavior, rather than every entity ending up in the default group.

klaasnicolaas (commented):
I think we can make this example more useful by switching to a real optional attribute rather than having every entity fall into the default group.

source/_template_functions/groupby.markdown
Laufmaschine: @jack3308 could you please check this suggestion?

jack3308:
Agreed! Just haven't had time - I'll test out the brightness attribute when I get home tonight and make the update once I confirm it's valid.

source/_template_functions/groupby.markdown
jack3308: Didn't see this before my previous reply - that looks valid to me so long as `device_class` is stored as an attribute! Just on my phone atm though - so will look properly later this evening

jack3308: I fear I've messed up this PR - this is more or less my first contribution to a project through an actual pull request. I'm so sorry for mucking things up (if I have). 

That being said - the template above looks like a great example and seems to work really well - only edit I would make is to change `expand("sensor.home_sensors")` to `states.sensor` as it's not super clear that `sensor.home_sensors` is a group or template that contains multiple sensors and with the change someone can copy the template into their template editor and see it working.

fdmarcin: @jack3308 No worries and thanks for contributing to Home Assistant! If it will be easier, feel free to close this PR and start over.

PR #47898, opened 2026-09-04
Document CoolMasterNet demand sensor and HVAC action

jameskoziol:
> **Status: on hold with home-assistant/core#181291.** Testing on my own bridge (Mitsubishi Electric M-NET line) shows the demand flag this documents is never set on that line type, so the parent PR stays a draft until CoolAutomation confirms where Demand State is supported.

## Proposed change

Documents the two things home-assistant/core#181291 adds to the CoolMasterNet integration:

- The **Demand** binary sensor. The bridge reports a per-unit demand flag, which is true while an indoor unit is actually calling for heating or cooling rather than merely switched on. I run six units behind a CoolMasterNet and this is the only runtime signal the hardware offers, so it is worth spelling out that a unit which has reached its target temperature reads as off.
- The climate entity's HVAC action, which is derived from that same flag.

Both cases where the action is unknown are documented, because they will otherwise look like a fault: bridges that answer with the older status format do not report demand at all, and a unit in the bridge's automatic mode whose current temperature equals its target gives nothing to infer a direction from.

I have kept this to what that core PR delivers. The integration's other entities (clean filter, error code, reset filter) are already shipped and undocumented, and adding them here would be a change to existing documentation and so belongs on the `current` branch in its own PR. I am happy to open that separately if you would like it.

## Type of change

## Additional information

- Link to parent pull request in the codebase: home-assistant/core#181291

No front matter change is needed: `ha_platforms` already lists `binary_sensor`, because the integration already ships a binary sensor.

Disclosure: I used Claude Code as an editor while drafting this page. I have read and understood it, the content is mine, and I checked it against the behaviour of the code in the parent pull request. `textlint` and `remark` both pass on the edited page.

## Checklist

### DIFF
--- a/source/_integrations/coolmaster.markdown
+++ b/source/_integrations/coolmaster.markdown
@@ -20,3 +20,16 @@ ha_integration_type: hub
 The **CoolMasterNet** {% term integration %} lets you control HVAC through [CoolMasterNet](https://coolautomation.com/products/coolmasternet/).
 
 {% include integrations/config_flow.md %}
+
+## Supported functionality
+
+### Climate
+
+The current activity of each unit is shown as the climate entity's HVAC action. A unit that is calling for heating or cooling shows **heating**, **cooling**, or **drying** to match its mode, a unit in fan only mode shows **fan**, a unit that is on but has already reached its target temperature shows **idle**, and a unit that is switched off shows **off**.
+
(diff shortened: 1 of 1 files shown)


jameskoziol:
Closing along with home-assistant/core#181291. I'll open a separate docs PR for the fan speed change when #181289 is ready.

Issue #46973, opened 2026-07-21
Template: create a separate UI section for each platform


Laufmaschine:
### Description

In Template integration documentation: https://www.home-assistant.io/integrations/template/:
- Refer which template types can be created from the UI
- Create a UI section for each platform

PR #48029, opened 2026-09-09
Template: Add procedure for creating a template helper from the UI and UI options (Part I)

Laufmaschine:
## Proposed change

- Add section describing how to create a template helper from the user interface.
- Refer that only template helpers of state-based entities can be created from the user interface 
- Add section **Options in the UI** to each platform
- Add titles **Options in YAML** and **YAML examples** to the respective content of each existing platform

The changes were made to the platforms:

- Alarm control panel
- Binary sensor
- Button
- Cover
- Device Tracker
- Event
- Fan
- Image
- Light
- Lock
- Number

The others will be handled in a different PR, to make this one shorter.

Partially fixes but not closing: https://github.com/home-assistant/home-assistant.io/issues/46973

## Type of change

## Additional information

## Checklist

### DIFF
--- a/source/_integrations/template.markdown
+++ b/source/_integrations/template.markdown
@@ -96,6 +96,17 @@ Configuration using our user interface provides a more limited subset of options
 If you need more specific features for your use case, the manual [YAML-configuration section](#yaml-configuration) of this integration might provide them.
 {% endnote %}
 
+### Creating a template helper from the user interface
+
+You can create state-based template entities from the user interface. To create a helper of a state-based template entity:
+
+1. Follow the manual configuration steps described in [Configuration](#configuration).
+2. Pick the entity platform of the template helper you want to create from the list.
+3. Fill in the **Name** and other fields shown. Each field is described in the **Options in the UI** section of the corresponding platform section further down this page.
+4. Select **Submit** to save the helper.
+
+After creation, a template helper's options can be changed at any time in **Settings > Devices & services > Helpers**, by selecting the helper and then **Settings** {% icon "mdi:cog-outline" %}.
+
 ## YAML configuration
 
 Entities are defined in your YAML configuration files under the `template:` key. You can define multiple configuration blocks as a list. Each block defines sensor/binary sensor/number/select entities and can contain optional update triggers.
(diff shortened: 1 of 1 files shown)


c0ffeeca7 (dismissed):
Thank you, @Laufmaschine 👍

Laufmaschine:
Hi @Petro31! Could you please check the proposed changes and share your feedback? Thanks!

Petro31:
~~Please do not remove the yaml examples.  I added them because users specifically requested them.  If you plan to add UI sections, by all means, go ahead.  However yaml for template entities is still the better choice because we do not have all options in the UI.~~

Ok, I see what's happening.  The code compare made it look like the yaml examples were removed.

I'm good with this format.

Laufmaschine:
No, they won't be removed nor changed. They will just be placed under the subsection **YAML examples** in each platform section so they will be identified and easily found.

Petro31:
Yes, see my edit.

I like the new format, it works out.  My only concern is that the section titles can't be linked.  E.g. you can't link the alarm control panel ui options or the yaml options.

I do think the titles of those sections are a bit too verbose as well.  Maybe ask AI if it makes more sense to use `UI OPtions` instead of `Options in the UI`.  Same goes for yaml.  I'm not sure which is better.

Change availability template to just Availability.  I never noticed that, I'll update the actual UI in a separate PR.  PR Here https://github.com/home-assistant/core/pull/181740

Laufmaschine:
What do you mean with linking the alarm control panel UI options or the YAML options?

Regarding the titles, I used the same ones that are used in triggers, conditions and actions pages for consistency.

Thank you for linking the PR. I will use the new name **Availability**. 👍

Petro31:
> What do you mean with linking the alarm control panel UI options or the YAML options?

Users create links from the documentation, each platform section is going to have the same headings, so users won't be able to point to the alarm control panel yaml options for example.  Because the title will just be "yaml options" and the markdown link will just bring them to the first hit.

Laufmaschine:
In that case I´m afraid users will need to use the parent link of the section, with the platform name, and then refer the intended "options" subsection.

source/_integrations/template.markdown
```diff
@@ -520,18 +585,20 @@ The template button platform allows you to create button entities with scripts t
...
+### Options in the UI
+
+{% options_ui %}
+Name:
+  description: The name of the button.
+Actions on press:
+  description: The action or actions run when the button is pressed.
+Device class:
+  description: The [device class](/integrations/button/#device-class) used to pick the button's wording and icon.
+Device:
```
klaasnicolaas, line 597: Could you also document **Availability** here? The Template config flow exposes `availability` for Button under the shared **Additional options** section, so the UI options list is currently incomplete. The same applies to the Number section below.

source/_integrations/template.markdown
```diff
@@ -655,6 +731,41 @@ cover:
 
 {% endconfiguration %}
 
+```yaml
```
klaasnicolaas, line 734: Please move the `### YAML examples` heading above this state-based example so both Cover examples are grouped under the same section.

klaasnicolaas (changes requested):
Hi there,

Thanks for working on this. I found a few remaining documentation mismatches with the current Template helper UI:

- The Button UI options are missing **Availability**. Core exposes this under the shared **Additional options** section for Template helpers.
- The Number UI options are also missing **Availability** for the same reason.
- The Cover section currently places the state-based YAML example before the **YAML examples** heading, while the trigger-based example is below it. Please move the heading so both examples are grouped consistently.

There is also still an outdated wording issue in the helper creation steps: “Follow the manual configuration steps described in Configuration” points to the UI config flow rather than a manual configuration path. Please adjust that wording for clarity.

./Klaas

source/_integrations/template.markdown
Laufmaschine: Good catch, thanks! Done.

source/_integrations/template.markdown
Laufmaschine: Thank you! 🤓 Done!

source/_integrations/template.markdown
```diff
@@ -96,6 +96,17 @@ Configuration using our user interface provides a more limited subset of options
 If you need more specific features for your use case, the manual [YAML-configuration section](#yaml-configuration) of this integration might provide them.
 {% endnote %}
 
+### Creating a template helper from the user interface
+
+You can create state-based template entities from the user interface. To create a helper of a state-based template entity:
+
+1. Follow the manual configuration steps described in [Configuration](#configuration).
```
klaasnicolaas: This still refers to “manual configuration” and links to `#configuration`, but this procedure is describing the UI flow and that anchor does not exist on this page.

Could we make this step explicitly refer to the UI configuration flow instead? For example, guide the user from **Helpers** to creating a **Template** helper before selecting the template entity type.

klaasnicolaas (changes requested):
Hi there,

There’s one remaining blocking point inline. Once that is addressed, this should be ready from my side.

./Klaas

source/_integrations/template.markdown
Laufmaschine: Updated the procedure to explicitly refer the UI steps instead. Thank you for the review!

klaasnicolaas (approved):

PR #48077, opened 2026-09-11
Warn about and/or not returning booleans in template condition docs

mcdave:
## Proposed change

`condition: template` (and the shorthand `conditions: "{{ ... }}"`) only passes when the rendered result is the case-insensitive string `true`. `and`/`or` in Python/Jinja don't coerce to a boolean, they return whichever operand determined the result. So `{{ x is defined and x }}` renders as the value of `x` itself (for example, a non-empty list) when `x` is truthy, not `true`, and the condition silently fails even though `x` is genuinely defined and non-empty.

I hit this for real in a script that only worked when a `color_rgb` field was passed: the condition guarding the color branch used `color_rgb is defined and color_rgb`, which never rendered `true` for a non-empty list, so the color was silently never applied.

This PR adds an `{% important %}` box to the Template condition section with a minimal reproducible example (using the `{% example %}` tag to show the actual rendered output) and the fix, right after the existing basic example.

## Type of change

## Additional information

- I checked for an existing issue/PR about this first. The closest one I found, [#143892](https://github.com/home-assistant/core/issues/143892), is about a different (and already resolved as "working as designed") native-typing quirk with lowercase `true`/`false`, not this `and`/`or` operand-passthrough pitfall.
- I checked the current `homeassistant/helpers/condition.py` source to confirm `condition: template` renders with `parse_result=False` and compares the lowercased string against `"true"`, and verified locally that `{{ true and [255, 0, 0] }}` renders as `[255, 0, 0]`, not `True`/`true`, and that `{{ x is defined and x | length > 0 }}` renders `True`/`False` correctly.

## Checklist

- [x] The documentation follows the Home Assistant documentation [standards](https://developers.home-assistant.io/docs/documenting/standards).

### DIFF
--- a/source/_docs/scripts/conditions.markdown
+++ b/source/_docs/scripts/conditions.markdown
@@ -668,15 +668,17 @@ Within an automation, template conditions also have access to the `trigger` vari
 Be careful when combining `and`/`or` with a value that isn't already a boolean, such as a list, string, or number. In Python and Jinja, `and`/`or` don't always evaluate to a boolean — they return whichever operand determined the result. For example:
 
 ```yaml
-condition: template
-value_template: "{{ some_list is defined and some_list }}"
+conditions:
+  - condition: template
+    value_template: "{{ some_list is defined and some_list }}"
 ```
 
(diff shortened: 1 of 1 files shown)


c0ffeeca7 (approved):
Thank you, @mcdave 👍

PR #48125, opened 2026-09-14
Fix devcontainer build and pin the Ruby base image

c0ffeeca7:
The devcontainer build failed at "rvm fix-permissions" with:

    bash: line 1: /usr/local/rvm/bin/rvm: No such file or directory

The base image no longer ships RVM. Upstream devcontainers/images bumped the Ruby image build recipe to 4.x, which upgraded the ruby feature from v1 to v2. That feature installs Ruby with ruby-build instead of RVM, so GEM_HOME moved from /usr/local/rvm/gems/default to /usr/local/bundle and /usr/local/rvm stopped existing. Comparing the image config env shows where it changed:

    3.0.9-3.4-trixie  built 2026-07-30  ruby feature v1  GEM_HOME=/usr/local/rvm/gems/default
    4.0.1-3.4-trixie  built 2026-08-25  ruby feature v2  GEM_HOME=/usr/local/bundle

Because we tracked the floating "3.4" tag, the build started failing without any change on our side.

Remove the obsolete fix-permissions step. It has no replacement: there are no RVM gem directories left to fix, and bundler installs into vendor/bundle inside the workspace via BUNDLE_PATH, owned by the vscode user.

Pin the base image to a full tag so an upstream rebuild cannot change the Ruby install layout again. "3.4" pinned only the Ruby version and left the Debian release, the install method and the version manager free to move. The tag now names the build recipe version as well, which is the part that carried this change, and the Dockerfile documents how to bump it. Drop the VARIANT indirection so the image is named in one place instead of two, and drop its stale default of 2.

Also switch to the current devcontainers/ruby path. The former vscode/devcontainers/ruby path is deprecated and both resolve to the same digest.

Verified the image builds and that bundle install (66 gems, including native extensions) and npm install both succeed as the vscode user, and that bundle exec jekyll reports 4.4.1.

## Type of change

## Additional information

## Checklist


Laufmaschine (approved):
Thank you, @c0ffeeca7!

c0ffeeca7:
Thanks!

PR #48138, opened 2026-09-14
OpenDisplay: Review docs for style

fdmarcin:
## Proposed change

Edit [OpenDisplay doc](https://www.home-assistant.io/integrations/opendisplay/) and [related action](https://www.home-assistant.io/actions/opendisplay.upload_image/) to match current style.

## Type of change

## Additional information

## Checklist


source/_integrations/opendisplay.markdown
```diff
@@ -21,151 +21,171 @@ ha_integration_type: device
...
-- **Battery Voltage**: (Only if the device has a battery configured) Shows the current voltage of the attached battery
+OpenDisplay devices provide the following diagnostic sensors, which are disabled by default:
+
+- **Temperature**: The device chip temperature.
+- **Battery voltage**: The voltage of the attached battery. Available when the device uses battery or solar power.
+- **Battery**: The battery charge as a percentage. Available when the device uses battery or solar power.
 
 ### Button events
 
-OpenDisplay Flex devices with configured physical inputs show up as {% term event %} {% term entities %} in Home Assistant. One {% term event %} {% term entity %} is created for each physical button.
```
fdmarcin, line 71: ℹ️ I removed mentions of Flex. LLM reading the Core code said:

> OpenDisplay Flex is a firmware board-description format, not a distinct device type that Home Assistant can currently identify.
> In the current Core implementation, every OpenDisplay device is treated as Flex because the underlying library cannot distinguish the formats yet.

fdmarcin:
@g4bri3lDev FYA

source/_integrations/opendisplay.markdown
```diff
@@ -21,151 +21,171 @@ ha_integration_type: device
...
+OpenDisplay devices with configured physical inputs provide {% term event %} entities. Home Assistant creates one event entity for each physical button.
```
klaasnicolaas, line 74: Core still explicitly distinguishes Flex from non-Flex devices. Only Flex devices load the `sensor` and `event` platforms; non-Flex devices only get the connectivity binary sensor.

Could we keep the Flex qualifier here? The same applies to the **Sensors** section above, otherwise the documentation suggests these entities are available on every OpenDisplay device.

source/_integrations/opendisplay.markdown
```diff
@@ -21,151 +21,171 @@ ha_integration_type: device
...
 
+### Binary sensors
+
+- **Connectivity**: Reports whether Home Assistant has recently received a Bluetooth Low Energy advertisement from the device.
+
 ### Sensors
 
-- **Temperature**: Chip temperature
-- **Battery Voltage**: (Only if the device has a battery configured) Shows the current voltage of the attached battery
+OpenDisplay devices provide the following diagnostic sensors, which are disabled by default:
```
klaasnicolaas, line 66: Not all of these sensors are disabled by default. Core disables **Temperature** and **Battery voltage** by default, but the **Battery** percentage sensor does not set `entity_registry_enabled_default=False` and is enabled by default.

Could we make the disabled-by-default status explicit per sensor instead of applying it to the whole list?

klaasnicolaas (changes requested):
I’ve left two blocking points inline. Once those and the existing action-label comments are addressed, this should be ready from my side.

source/_integrations/opendisplay.markdown
fdmarcin: yep, my bad

source/_integrations/opendisplay.markdown
```diff
@@ -21,151 +21,171 @@ ha_integration_type: device
...
 
-## Examples
+### Automation: Show a daily image every morning
 
-### Uploading an image
+Show an image from your media folder on the display each morning.
 
-{% details "Upload an image from local media" %}
+- **Trigger**: Time: 07:00:00
+- **Action**: OpenDisplay: Upload image
```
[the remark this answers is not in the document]
fdmarcin: This should stay: https://developers.home-assistant.io/docs/documenting/integration-docs-examples/#referring-to-automation-triggers-conditions-and-actions-in-ui-steps

source/_integrations/opendisplay.markdown
fdmarcin: Thanks, added it back + some explanation and link.

It's still not perfect and the reader might be like "so, do I have a Flex device? (like I thought), but it's an improvement.

klaasnicolaas (approved):

